PR #3, opened 2018-07-09
add support for field mode in TextIOToBigQuery.java

mchalek:
This commit adds support for the `mode` field in bigquery schemas, so that the GCS text to bigquery template can use repeated and/or required fields.  Existing behavior is unchanged, because the mode is only set on the Bigquery schema if it is present in the input schema.


azurezyq:
This change has already merged into our internal codebase and will be sync'ed to this repo in a few days.

Public dataflow templates are already updated and this change is part of that, including:
gs://dataflow-templates/2018-07-11-00/
gs://dataflow-templates/latest/
as well as the UI.

Issue #8, opened 2018-10-01
Record type not supported in TextIOToBigQuery template


cmangani:
Hi guys,

The biq query schema file that is used in this template cannot have a RECORD type defined..else get error below . Looking at the code.. does not look like code accommodates nested/recursive build up of table schema in the withSchema() block of code to deal with RECORD schema definition... I would be happy to code it up if you like....

Error: Field xxxx is type RECORD but has no schema.

PR #11, opened 2018-10-15
Added support for RECORD field type in schema definition

cmangani:
Added support for the RECORD type in BQ Schema definition to the TextIOToBigQuery template.
#8


cmangani:
I signed it!
________________________________

PR #12, opened 2018-10-17
Removed Example Usage

the-thappy:
The provided example usage was misleading / incorrect because, when using the example usage to compile the file, it threw an error, 
`[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.6.0:java (default-cli) on project google-cloud-teleport-java: An exception occured while executing the Java class. Class interface com.google.cloud.teleport.templates.PubsubToText$Options missing a property named 'topic'. -> [Help 1]`
even if `--topic` or `--inputTopic` argument is passed. 
The correct steps has been properly documented here [Dataflow Documentation](https://cloud.google.com/dataflow/docs/templates/overview) in the creating and executing templates section. To reiterate, compile without passing arguments at compile time, and pass the arguments at run time instead when executing the dataflow job.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
@@ -42,26 +42,6 @@ import org.apache.beam.sdk.transforms.windowing.Window;
  * This pipeline ingests incoming data from a Cloud Pub/Sub topic and
  * outputs the raw data into windowed files at the specified output
  * directory.
- *
- * <p> Example Usage:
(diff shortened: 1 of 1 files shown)


ryanmcdowell:
Instead of removing the documentation completely, perhaps updating it to be more comprehensive like the PubSubToBigQuery documentation would be better.

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/master/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java

PR #21, opened 2018-12-26
Update PubSubToBigQuery to use subscription input instead of topic.

wallaceicy06:
The topic input makes the workflow non-deterministic because there is no guarantee of when the workflow will create the subscription for the provided topic. Thus, messages sent to the topic could be dropped. Requiring an existing subscription guarantees that any unread messages published to the topic will be read eventually.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java
@@ -64,7 +64,7 @@ import org.slf4j.LoggerFactory;
  * <p><b>Pipeline Requirements</b>
  *
  * <ul>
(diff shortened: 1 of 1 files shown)


jeffkpayne:
Would be great to see this merged in and released.  I would also recommend allowing both use cases (specifying topic or specifying subscription), but the use case of specifying an existing subscription is definitely the more useful one.  Would be great not to have to maintain our own version of this ;)

sabhyankar:
Note that the Dataflow UI now has a PubsubToBigQuery pipeline that supports reading from a subscription (Cloud Pub/Sub subscription to BigQuery)

sabhyankar:
Closing as the PubsubToBigQuery pipeline now has subscription support.

PR #28, opened 2019-04-03
Update TableRow conversion to support dates

priyankr2411:
The TableRow object in BigQuery requires the dates columns to be formatted before inserting into BigQuery. With the data set with date fields, the template is breaking because Jdbc convertor is not taking date fields into account. This change is to modify the ResultSetToTableRow method to format the date fields in the format acceptable to BigQuery.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
@@ -17,8 +17,12 @@
 package com.google.cloud.teleport.templates.common;
 
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.api.client.util.DateTime;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
+import java.sql.Types;
+import java.text.SimpleDateFormat;
+
 import org.apache.beam.sdk.io.jdbc.JdbcIO;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -101,9 +105,17 @@ public TableRow mapRow(ResultSet resultSet) throws Exception {
       TableRow outputTableRow = new TableRow();
 
       for (int i = 1; i <= metaData.getColumnCount(); i++) {
-        outputTableRow.set(metaData.getColumnName(i), resultSet.getObject(i));
+         int fieldType = metaData.getColumnType(i);
+         if(fieldType == Types.DATE) {
```
sabhyankar: Can we make this a switch block?

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -101,9 +105,17 @@ public TableRow mapRow(ResultSet resultSet) throws Exception {
...
+             SimpleDateFormat bqFormat = new SimpleDateFormat("yyyy-MM-dd");
+             outputTableRow.set(metaData.getColumnName(i), bqFormat.format(resultSet.getDate(i)));
+         } else if(fieldType == Types.TIMESTAMP) {
```
sabhyankar: Can we add handling for TIMESTAMP_WITH_TIMEZONE as well in that case?

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -17,8 +17,12 @@
 package com.google.cloud.teleport.templates.common;
 
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.api.client.util.DateTime;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
+import java.sql.Types;
+import java.text.SimpleDateFormat;
+
```
sabhyankar: Remove this blank line.

PR #31, opened 2019-04-16
add support for bytes, attributes via JSON

david-kiesel-thd:
add support for bytes, attributes via JSON

- PubsubToText.java
- TextToPubsub.java

add to .gitignore

- tmp/
- common IntelliJ files (.idea/, *.iml)

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
@@ -20,12 +20,15 @@ package com.google.cloud.teleport.templates;
 
 import com.google.cloud.teleport.io.WindowedFilenamePolicy;
 import com.google.cloud.teleport.util.DurationUtils;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
 import org.apache.beam.sdk.io.FileBasedSink;
 import org.apache.beam.sdk.io.TextIO;
 import org.apache.beam.sdk.io.fs.ResourceId;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubIO;
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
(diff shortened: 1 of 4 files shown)


david-kiesel-thd:
I signed it!

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
```diff
@@ -140,37 +223,104 @@ public static PipelineResult run(Options options) {
...
-        .apply(
-            "Write File(s)",
-            TextIO.write()
-                .withWindowedWrites()
-                .withNumShards(options.getNumShards())
-                .to(
-                    new WindowedFilenamePolicy(
+    String format = options.getFormat();
+
+    switch (format) {
```
sabhyankar: Note that the value of the format field is determined at graph creation time. i.e. when this template is compiled and stored on GCS. The "runtime" value of this field is ignored and so this template would essentially either follow the "string" path or the "json" path depending on the value passed to --format at compile time. This would mean that this pipeline would work as a normal dataflow job but not as a template (or you need two different templates compiled with different values for --format)

In Dataflow templates, the job graph does not change after it has been compiled. Take a look at the [PubsubToBigQuery template](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/master/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java) as an example. It uses a boolean flag called --useSubscription. This is used to create two different pipelines in the Dataflow UI ("Cloud Pub/Sub subscription to BigQuery" and "Cloud Pub/Sub topic to BigQuery")

azurezyq: I'm second to @sabhyankar. Because of the limitation of the templates feature, this has to be a separate template in our dropdown list. We're working on extending the feature but it may need some time. It'll fit into "several weeks" time frame.

Unfortunately before that you'll have to generate variants of templates on your side.

david-kiesel-thd:
Thanks for following up.

Let me see if I understand.

I'll restrict the discussion to TextToPubsub for the sake of simplicity.

As my fork stands now, the `format` variable set by the --format option can only be considered at compile time (mvn compile ...).  So, if you use the --templateLocation option to create a template (instead of executing immediately), that particular template will be effectively hard-coded for whatever version of `format` was set when the template was created.  So, to have both `format` options available through a template, I'd have to use `mvn compile` twice.  Say, once to create a TextToPubsub template (with the original string and no attributes functionality).  And once to create a TextToPubsubJson template (with the new bytes and attributes functionality).  And this situation is analogous to how the --useSubscription option is handled in PubsubToBigQuery.

Is everything above correct?

Is there a reason not to proceed that way?

Alternatively, I am assuming that I could change my fork so the --format option uses `ValueProvider<String>` instead of `String`.  In that case, you would only need to use `mvn compile` one time to create a single template.  And you would set the `format` value as part of the parameters when you execute the template with `gcloud dataflow jobs run ...`.

Is everything above correct?

Is there anything else I am missing about the change needed to set `format` either way using a single template?

Now that (I think) I understand this, it does sound like a better solution to set `format` at `gcloud dataflow jobs run` time instead of at `mvn compile` time.

Why, in the case of PubsubToBigQuery, did someone choose for --useSubscription to be set at `mvn compile` time?

Are there any drawbacks to having just one TextToPubsub template with `format` settable at `gcloud dataflow jobs run` time?

Would it make a merge of my fork more palatable if I changed the fork to make `format` a parameter at `gcloud dataflow jobs run` time?

Any other comments about the suitability of the fork for merging?

sabhyankar:
> Thanks for following up.
> 
> Let me see if I understand.
> 
> I'll restrict the discussion to TextToPubsub for the sake of simplicity.
> 
> As my fork stands now, the `format` variable set by the --format option can only be considered at compile time (mvn compile ...). So, if you use the --templateLocation option to create a template (instead of executing immediately), that particular template will be effectively hard-coded for whatever version of `format` was set when the template was created. So, to have both `format` options available through a template, I'd have to use `mvn compile` twice. Say, once to create a TextToPubsub template (with the original string and no attributes functionality). And once to create a TextToPubsubJson template (with the new bytes and attributes functionality). And this situation is analogous to how the --useSubscription option is handled in PubsubToBigQuery.
> 
> Is everything above correct?
> 
> Is there a reason not to proceed that way?
> 
> Alternatively, I am assuming that I could change my fork so the --format option uses `ValueProvider<String>` instead of `String`. In that case, you would only need to use `mvn compile` one time to create a single template. And you would set the `format` value as part of the parameters when you execute the template with `gcloud dataflow jobs run ...`.
> 
> Is everything above correct?
> 
> Is there anything else I am missing about the change needed to set `format` either way using a single template?
> 
> Now that (I think) I understand this, it does sound like a better solution to set `format` at `gcloud dataflow jobs run` time instead of at `mvn compile` time.
> 
> Why, in the case of PubsubToBigQuery, did someone choose for --useSubscription to be set at `mvn compile` time?
> 
> Are there any drawbacks to having just one TextToPubsub template with `format` settable at `gcloud dataflow jobs run` time?
> 
> Would it make a merge of my fork more palatable if I changed the fork to make `format` a parameter at `gcloud dataflow jobs run` time?
> 
> Any other comments about the suitability of the fork for merging?

Yes, your understanding is correct in both the questions above. The key thing to keep in mind is that in a Dataflow template, the execution graph is set and cannot be changed at runtime. 
So you cannot do something like this:
Transform-A -> (if booleanFlag Transform-B else Transform-C) -> Transform-D

We had to take the approach of using two separate templates for PubsubToBigQuery because we couldnt 'remove' the template where useSubscription=false. It is currently being used by several live pipelines. Eventually we would want this to be just one template using subscription.

IMO, we have a few different options here:
1. Create a 'new-transform' that would accept a ValueProvider<String> (for the format) and then use that transform instead. In this case the job graph doesnt change. However, you might take different action in the new-transform based on the runtime value for format.

2. If the need is immediate, create a custom template that you can use for your projects ([See creating templates](https://cloud.google.com/dataflow/docs/guides/templates/creating-templates))

3. We will be publishing a new Template feature that will allow you to change the execution graph at runtime and will thus remove this limitation. I don't have a specific timeline for it that we can communicate. However, if this is something you want to explore then we can connect offline and see how this can be done.

PR #41, opened 2019-06-07
Use values provided by users for WindowDuration and NumShards

barata:
Previous version ignores users parameters and always takes the defaults (1 shard and 5m window duration).

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubsubToAvro.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubsubToAvro.java
@@ -123,9 +123,9 @@ public class PubsubToAvro {
 
     @Description("The maximum number of output shards produced when writing.")
     @Default.Integer(1)
-    Integer getNumShards();
+    ValueProvider<Integer> getNumShards();
 
-    void setNumShards(Integer value);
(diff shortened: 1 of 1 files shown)


barata:
I signed it!

src/main/java/com/google/cloud/teleport/templates/PubsubToAvro.java
```diff
@@ -123,9 +123,9 @@
 
     @Description("The maximum number of output shards produced when writing.")
     @Default.Integer(1)
-    Integer getNumShards();
+    ValueProvider<Integer> getNumShards();
```
sabhyankar: This will not work unfortunately as AvroIO.Write.withNumShards(int) does not support a ValueProvider as an input. The same goes for some of the other parameters below.

barata:
I see it now. I was trying to fix an existing bug when you create DF jobs using GCP Console "Export from Pubsub to GCS" feature. I have an on-going internal ticket with Google that I will follow up on. Thanks for your comments anyway.

dtcrout:
Is there any status update on this issue? I'm encountering a similar issue with `windowDuration` for `PubsubToText` template. The default value of `windowDuration` is taken and the parameter value is ignored.

PR #68, opened 2019-11-03
Support read of repeated string fields in BigQuery

larsbkrogvig:
I want to read a repeated string field from BigQuery, but there is no check for field mode in the `BigQueryConverters` class.

This adds such a check for `STRING` fields only. This should be made more general, though,

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/BigQueryConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/BigQueryConverters.java
@@ -27,6 +27,7 @@ import com.google.datastore.v1.Key;
 import com.google.datastore.v1.Key.PathElement;
 import com.google.datastore.v1.PartitionId;
 import com.google.datastore.v1.Value;
+import com.google.datastore.v1.ArrayValue;
 import com.google.protobuf.NullValue;
 import com.google.protobuf.util.Timestamps;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


larsbkrogvig:
@googlebot I signed it!

src/main/java/com/google/cloud/teleport/templates/common/BigQueryConverters.java
```diff
@@ -384,10 +385,34 @@ public static Value columnToValue(TableFieldSchema column, Object columnValue)
...
+                  excludeArrayFromIndexes = true;
+                }
+                arrayValueBuilder.addValues(Value.newBuilder().setStringValue(arrayStringValue).build());
+              }
+              valueBuilder.setArrayValue(arrayValueBuilder.build());
+              valueBuilder.setExcludeFromIndexes(excludeArrayFromIndexes);
+              break;
+
+          }
           break;
```
sabhyankar: If I am reading this correctly, we are only handling REQUIRED and REPEATED fields for STRING types? Why not extend this to the other supported data types (INT64, FLOAT etc)?

src/main/java/com/google/cloud/teleport/templates/common/BigQueryConverters.java
```diff
@@ -384,10 +385,34 @@ public static Value columnToValue(TableFieldSchema column, Object columnValue)
           // Datastore string properties greater than 1500 bytes will not be indexed in order to
           // respect the limit imposed on the maximum size of index-able string properties. See
           // https://cloud.google.com/datastore/docs/concepts/limits
-          String strValue = columnValue.toString();
-          valueBuilder.setStringValue(strValue);
-          boolean excludeFromIndexes = strValue.getBytes().length > MAX_STRING_SIZE_BYTES;
-          valueBuilder.setExcludeFromIndexes(excludeFromIndexes);
+          switch (column.getMode()) {
```
sabhyankar: Lets add unit test coverage as well.

PR #71, opened 2019-11-09
Fix sample command to execute PubSubToSplunk template

rarsan:
- Minor README update to specify UDF support of PubSubToSplunk template
- Fix example gcloud cli command in inline comments of template source code
- No code changes. All tests and checks pass


rarsan:
cc @sabhyankar

PR #75, opened 2019-11-22
Added Pubsub to MongoDB flex template

pritukam:


v2/pubsub-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToMongoDB.java
```diff
@@ -0,0 +1,494 @@
...
+  /**
+   * DoFn that will parse the given string elements as Bson Documents.
+   */
+  public static class ParseAsDocumentsFn extends DoFn<String, Document> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ParseAsDocumentsFn.class);
+
+    @ProcessElement
+    public void processElement(ProcessContext context) {
+      context.output(Document.parse(context.element()));
```
sabhyankar: Will this parsing always be successful? If not, we should put this in a try-catch block and use separate tags to catch these parsing failures (with metric-counters).

v2/pubsub-to-mongodb/src/test/com/google/cloud/teleport/v2/templates/PubSubToMongoDBTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
```
sabhyankar: Missing javadoc.

v2/pubsub-to-mongodb/src/test/com/google/cloud/teleport/v2/templates/PubSubToMongoDBTest.java
```diff
@@ -0,0 +1,335 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.teleport.v2.templates;
+
+import com.google.cloud.teleport.v2.values.FailsafeElement;
```
sabhyankar: Can you reformat/reorder the imports?

v2/pubsub-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToMongoDB.java
pritukam: The function will always get a valid json. Valid json will always get parsed as Bson documents

sabhyankar: Yes in the context of this pipeline. However, what happens when I attempt to use the public static class and pass a String (which is not a valid json)?

PR #77, opened 2019-12-05
Initial Commit of ParquetToBigtable

YadnikiPawar:


YadnikiPawar:
@sabhyankar The change suggested for handling the condition based on the Runners instead of Object's instance type is done. The PR is ready for review.

pom.xml
```diff
@@ -251,12 +251,6 @@
       <version>${truth.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.google.truth</groupId>
-      <artifactId>truth</artifactId>
-      <version>${truth.version}</version>
-      <scope>test</scope>
-    </dependency>
```
sabhyankar: Why are we removing this dependency?

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.KV;
+
+/**
+ * Dataflow pipeline that imports data from Parquet files in GCS to a Cloud Bigtable table. The
+ * Cloud Bigtable table must be created before running the pipeline and must have a compatible table
+ * schema. For example, if {@link BigtableCell} from the Parquet files has a 'family' of "f1", the
+ * Bigtable table should have a column family of "f1".
+ */
```
sabhyankar: Can you make this similar to the other pipeline's javadoc with instructions to compile and execute?

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
```diff
@@ -0,0 +1,156 @@
...
+   */
+  public static void main(String[] args) {
+    Options options = PipelineOptionsFactory.fromArgs(args).withValidation().as(Options.class);
+
+    PipelineResult result = run(options);
+
+    // Wait for pipeline to finish only if it is not constructing a template.
+    if (options.as(DataflowPipelineOptions.class).getTemplateLocation() == null) {
+      result.waitUntilFinish();
+    }
```
sabhyankar: I dont think we need this block.

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
```diff
@@ -0,0 +1,156 @@
...
+
+    return pipeline.run();
+  }
+
+  static class ParquetToBigtableFn
+          extends DoFn<GenericRecord, KV<ByteString, Iterable<Mutation>>> {
+
+    @ProcessElement
+    public void processElement(ProcessContext ctx) {
+      String runner = ctx.getPipelineOptions().getRunner().getCanonicalName();
```
sabhyankar: Instead of getting the canonical name, you can just get the Class and check whether it is assignable.

So:
Class runner = ctx.getPipelineOptions().getRunner();

And in your if statement:
if (runner.isAssignableFrom((DirectRunner.class)) {..} else if (runner.isAssignableFrom(DataflowRunner.class)) {..}

src/test/java/com/google/cloud/teleport/bigtable/ParquetToBigtableTest.java
```diff
@@ -0,0 +1,106 @@
...
+ * Tests for ParquetToBigtable.
+ */
+@RunWith(JUnit4.class)
+public class ParquetToBigtableTest {
+
+  @Rule
+  public final transient TestPipeline pipeline = TestPipeline.create();
+
+  @Test
+  public void applyParquettoBigtableFn() throws Exception {
```
sabhyankar: Add a javadoc

pom.xml
YadnikiPawar: @sabhyankar  It's a duplicate dependency. It was giving warnings will running the code.

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
YadnikiPawar: @sabhyankar How do we handle if the Runner is not Direct or Dataflow?

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
YadnikiPawar: @sabhyankar Its similar to the one in AvroToBigtable. And none of the other Legacy templates have compile and execute details.

pom.xml
sabhyankar: Dont remote anything that we didnt add in this PR :)

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
sabhyankar: Example:
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/master/src/main/java/com/google/cloud/teleport/templates/BulkCompressor.java

src/main/java/com/google/cloud/teleport/bigtable/ParquetToBigtable.java
sabhyankar: Just handle the separate condition for DirectRunner and assume that all the other runners would work similarly. So:
if (runner.isAssignableFrom((DirectRunner.class)) {..} else {..}

PR #80, opened 2020-01-06
Text Import pipeline to import BYTES datatype

adkhare:
Currently, Spanner to Text exports export BYTES data as base64 string. However, imports error out.

This PR will essentially enable text to spanner import dataflow job to import BYTES columns such that it will decode base64 string into byte array and import as BYTES datatype in spanner


adkhare:
@googlebot I fixed it.

adkhare:
@sabhyankar tagging you here since you seem to be the latest contributor who has merge access.

FWIW, this is in response to the support case that I created with google regards to the dataflow template unable to import BYTES datatype in CSV

sabhyankar:
LGTM

Issue #85, opened 2020-02-11
Specify project id for spanner to avro export


jlemaes:
I'd like to run the "Cloud Spanner to Cloud Storage Avro" template in a different project than the project my spanner instance lives. 
Currently I cannot specify the project in the template: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/9a95f05b149cc99a6688c674596d8f4b0dcd5c35/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java#L80-L84
It is however possible in the spanner to text export: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/9a95f05b149cc99a6688c674596d8f4b0dcd5c35/src/main/java/com/google/cloud/teleport/templates/SpannerToText.java#L91-L95

Is this something that can be added?

sabhyankar:
@melburnerodrigues - Can you take a look at this?

PR #86, opened 2020-02-11
Allow to specify the spanner projectid in spanner avro import/export 

jlemaes:
Fix for #85


jlemaes:
@googlebot I signed it!

PR #96, opened 2020-04-13
Change Write Disposition from WRITE_TRUNCATE to WRITE_APPEND

ramnathnayakpr:
Setting a default template to delete all existing data in a BQ table not only limits its use, but is also a bit dangerous (as I just found out, all my existing data was gone when I started using Dataflow). Setting it to Append instead makes the template more useful. Most people who end up using this template probably want to schedule Dataflow to incrementally new data as it arrives. The trucation at the moment is a dangerous side effect, as someone running this template would have no clue they are going to lose all existing data by using this template.


ramnathnayakpr:
@googlebot I signed it!

iconde:
By changing the WRITE_DISPOSITION option, this code is not coherent with what is explained in the Google documentation:
https://cloud.google.com/dataflow/docs/guides/templates/provided-batch#gcstexttobigquery
image

It could be a nice feature to add write_disposition as a template parameter.

PR #116, opened 2020-05-10
Fixed the incorrect use of == for string comparison

leoiii12:


pabloem:
LGTM. Thanks @leoiii12

PR #118, opened 2020-05-12
Use tempLocation correctly for PubsubToText

haakonmo:
Currently `Cloud_PubSub_to_GCS_Text` template doesn't use the `tempLocation` for storing temporary files, but rather the `outputDirectory`.
https://cloud.google.com/dataflow/docs/guides/templates/provided-streaming#pubsub-to-cloud-storage-text

Problem:
If the GCS bucket that is used for `outputDirectory` is configured with `Bucket Lock`, the job is not able to delete the temporary files, hence `tempLocation` should be used instead.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
@@ -19,6 +19,8 @@
 package com.google.cloud.teleport.templates;
 
 import com.google.cloud.teleport.io.WindowedFilenamePolicy;
+import com.google.cloud.teleport.util.DualInputNestedValueProvider;
+import com.google.cloud.teleport.util.DualInputNestedValueProvider.TranslatorInput;
 import com.google.cloud.teleport.util.DurationUtils;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
@@ -57,7 +59,8 @@ import org.apache.beam.sdk.transforms.windowing.Window;
  --windowDuration=2m \
  --numShards=1 \
  --inputTopic=projects/${PROJECT_ID}/topics/windowed-files \
- --outputDirectory=gs://${PROJECT_ID}/temp/ \
+ --userTempLocation=gs://${PROJECT_ID}/tmp/ \
+ --outputDirectory=gs://${PROJECT_ID}/output/ \
  --outputFilenamePrefix=windowed-file \
  --outputFilenameSuffix=.txt"
  * </pre>
@@ -80,6 +83,10 @@ public class PubsubToText {
     @Required
     ValueProvider<String> getOutputDirectory();
     void setOutputDirectory(ValueProvider<String> value);
+
+    @Description("The directory to output temporary files to. Must end with a slash.")
+    ValueProvider<String> getUserTempLocation();
+    void setUserTempLocation(ValueProvider<String> value);
 
     @Description("The filename prefix of the files to write to.")
     @Default.String("output")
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
```diff
@@ -165,10 +162,9 @@ public static PipelineResult run(Options options) {
                         options.getOutputFilenamePrefix(),
                         options.getOutputShardTemplate(),
                         options.getOutputFilenameSuffix()))
-                .withTempDirectory(NestedValueProvider.of(
-                    options.getOutputDirectory(),
-                    (SerializableFunction<String, ResourceId>) input ->
-                        FileBasedSink.convertToFileResourceIfPossible(input))));
+                .withTempDirectory(
+                    FileBasedSink.convertToFileResourceIfPossible(options.getTempLocation())
+                        .getCurrentDirectory()));
```
sabhyankar, line 167: Without the NestedValueProvider, the tempLocation specified at the time this template is built and released (compile time) will be used at runtime regardless of the used provided value of tempLocation. This would not work since the users of the public template in Dataflow do not have access to write to the tempLocation specified at compile time.

IMO, there are two options here:
1. Create a custom template with an additional option (--userTempLocation perhaps?) that is a ValueProvider and would be used instead of the OutputDirectory. This would be non backward compatible change for us
2. Take a look at [DualInputNestedValueProvider](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/645836749a2960347ec7285f112f0b19a477fb3a/src/main/java/com/google/cloud/teleport/util/DualInputNestedValueProvider.java#L29) and how it was used in other templates. IMO, this can be used to provide a backward compatible optional flag (if userTempLocation is provided then use that OR else use outputDirectory).


1.

haakonmo: Thank you for feedback. I went for option 2. Please see updated code.

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
```diff
@@ -166,11 +173,40 @@ public static PipelineResult run(Options options) {
...
     // Execute the pipeline and return the result.
     return pipeline.run();
   }
+
+  /**
+   * Utility method for using optional parameter userTempLocation as TempDirectory.
+   * This is useful when output bucket is locked and temporary data cannot be deleted.
+   *
+   * @param userTempLocation user provided temp location
+   * @param outputLocation default temp location, which should OutputDirectory for backward compatibility
```
sabhyankar, line 191: Maybe change this to:
"user provided outputDirectory to be used as the default temp location"

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
```diff
@@ -166,11 +173,40 @@ public static PipelineResult run(Options options) {
...
+          @Override
+          public String apply(TranslatorInput<String, String> input) {
+            String userProvidedTempLocation = input.getX();
+            String outputLocation = input.getY();
+            if (userProvidedTempLocation == null) {
+              return outputLocation;
+            }
+            return userProvidedTempLocation;
+          }
+        });
```
sabhyankar, line 210: Can this be changed into a simple lambda where you return something like:
(input.getX() != null) ? input.getX() : input.getY();

haakonmo: 👍

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
haakonmo: 👍

haakonmo:
@sabhyankar do you know why the API returns a parameterViolation when using the userTempLocation added in this PR?
```
type.googleapis.com/google.dataflow.v1beta3.InvalidTemplateParameters
parameterViolations:
  - description: Unrecognized parameter
    parameter: userTempLocation
```
Is there something that need to be changed outside this repo?

sabhyankar:
@haakonmo - Looks like this was not added to the parameters file that is used to validate the pipeline. Let me make this change internally. Thank you for pointing that out!

haakonmo:
> @haakonmo - Looks like this was not added to the parameters file that is used to validate the pipeline. Let me make this change internally. Thank you for pointing that out!

Nice! When (ish) can we expect it to work?

PR #123, opened 2020-05-15
Add --tokenKMSEncryptionKey option for PubSubToSplunkTemplate.

### LINKED ISSUE #122: Security concern: Pub/Sub to Splunk HEC requires a token as a parameter in plain text
## What's the issue?

The pub/sub to splunk beta dataflow recently announced in [this blog post](https://cloud.google.com/blog/products/data-analytics/connect-to-splunk-with-a-dataflow-template) ([docs here](https://cloud.google.com/dataflow/docs/guides/templates/provided-streaming#cloudpubsubtosplunk)) has a bit of a security gap in that it requires a plain text `token` as input. Code for getting that token [is here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/master/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java#L249). I know the team considers security a top priority so I'm hoping there's a known stopgap solution. Without some way to hide this param, anyone with viewer access across the service, project, or org would likely have visibility to this token - not ideal.

## Alternative solutions

Some potential workarounds and expected behavior:

1. Dataflow job integration directly with KMS or secrets manager - much in the way cloudbuild can decrypt a secret as an initial step defined in the build template, it would be slick if dataflow had a similar capability, pointing to pre-existing secrets that are automatically injected into the run context at job start. e.g. the SA running a dataflow job has a decrypter role to a KMS key. `sensitiveParameters.foo` specifies 2 attributes - a KMS key ID and the corresponding ciphertext - upon job launch, decryption happens automatically, with the raw contents of the secret available as environment variable `foo`.
2. Instead of the `token` parameter, the template could instead require a secret manager secret version ID and attempt to fetch that secret as an initial step in the job. [This post](https://medium.com/google-cloud/using-google-cloud-key-management-service-with-dataflow-templates-71924f0f841f) takes a similar approach using KMS. This is slightly different than #1 in that it isn't a feature request to dataflow the product, just a change to the relevant template.

Though I haven't used these myself, I've asked around and heard some transforms provide KMS integration e.g. bigqueryio support a `withKmsKey`. I wonder if there's a pattern that can be reused from that. +@melbrodrigues for input.

melbrodrigues:
Adds an optional parameter to specify a Cloud KMS Encryption Key if an encrypted `token` parameter is passed to the template.

Fixes #122

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
@@ -28,6 +28,7 @@ import com.google.cloud.teleport.templates.common.PubsubConverters.PubsubReadSub
 import com.google.cloud.teleport.templates.common.PubsubConverters.PubsubWriteDeadletterTopicOptions;
 import com.google.cloud.teleport.templates.common.SplunkConverters;
 import com.google.cloud.teleport.templates.common.SplunkConverters.SplunkOptions;
+import com.google.cloud.teleport.util.KMSEncryptedNestedValueProvider;
 import com.google.cloud.teleport.values.FailsafeElement;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -246,7 +247,7 @@ public class PubSubToSplunk {
             .apply(
                 "WriteToSplunk",
                 SplunkIO.writeBuilder()
-                    .withToken(options.getToken())
+                    .withToken(maybeDecrypt(options.getToken(), options.getTokenKMSEncryptionKey()))
                     .withUrl(options.getUrl())
                     .withBatchCount(options.getBatchCount())
                     .withParallelism(options.getParallelism())
@@ -362,5 +363,17 @@ public class PubSubToSplunk {
                     }
                   }));
     }
+  }
+  
+  /**
+   * Utility method to decrypt a Splunk HEC token.
+   *
+   * @param unencryptedToken The Splunk HEC token as a Base64 encoded {@link String} encrypted with a Cloud KMS Key.
+   * @param kmsKey The Cloud KMS Encryption Key to decrypt the Splunk HEC token.
+   * @return Decrypted Splunk HEC token.
+   */
+  private static ValueProvider<String> maybeDecrypt(
+      ValueProvider<String> unencryptedToken, ValueProvider<String> kmsKey) {
+    return new KMSEncryptedNestedValueProvider(unencryptedToken, kmsKey);
   }
 }
--- a/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
@@ -116,6 +116,13 @@ public class SplunkConverters {
     ValueProvider<Boolean> getIncludePubsubMessage();
   
     void setIncludePubsubMessage(ValueProvider<Boolean> includePubsubMessage);
+
+    @Description(
+        "KMS Encryption Key for the token. The Key should be in the format "
+            + "projects/{gcp_project}/locations/{key_region}/keyRings/{key_ring}/cryptoKeys/{kms_key_name}")
+    ValueProvider<String> getTokenKMSEncryptionKey();
+
+    void setTokenKMSEncryptionKey(ValueProvider<String> keyName);
   }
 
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
```diff
@@ -116,6 +116,13 @@ public static FailsafeStringToSplunkEvent failsafeStringToSplunkEvent(
     ValueProvider<Boolean> getIncludePubsubMessage();
   
     void setIncludePubsubMessage(ValueProvider<Boolean> includePubsubMessage);
+
+    @Description(
+        "KMS Encryption Key should be in the format "
+            + "projects/{gcp_project}/locations/{key_region}/keyRings/{key_ring}/cryptoKeys/{kms_key_name}")
+    ValueProvider<String> getKMSEncryptionKey();
```
jaketf, line 123: can this parameter be named `--tokenKMSEncryptionKey`?
reasoning: to disambiguate that this is not used by to decrypt pubsub data which is handled fluently by beam https://cloud.google.com/dataflow/docs/guides/customer-managed-encryption-keys#pubsub

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
```diff
@@ -116,6 +116,13 @@ public static FailsafeStringToSplunkEvent failsafeStringToSplunkEvent(
     ValueProvider<Boolean> getIncludePubsubMessage();
   
     void setIncludePubsubMessage(ValueProvider<Boolean> includePubsubMessage);
+
+    @Description(
+        "KMS Encryption Key should be in the format "
```
jaketf, line 121: specify that this is a kms encryption key for the token.

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: Done

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: Done

brandonjbjelland:
This is great! Thanks for the quick work here @melbrodrigues

src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
```diff
@@ -363,4 +364,9 @@ public void processElement(ProcessContext context) {
                   }));
     }
   }
+
```
sabhyankar, line 367: Can you add a JavaDoc here to clarify what we are doing with this static method?

sabhyankar:
@rarsan FYI

src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
melbrodrigues: Done

src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
```diff
@@ -363,4 +364,16 @@ public void processElement(ProcessContext context) {
...
+  /**
+   * Utility method to decrypt a Splunk HEC token.
+   *
+   * @param unencryptedToken The Splunk HEC token as a Base64 encoded {@link String} encrypted with a Cloud KMS Key.
+   * @param kmsKey The Cloud KMS Encryption Key to decrypt the Splunk HEC token.
+   * @return Decrypted Splunk HEC token.
+   */
+  private static ValueProvider<String> maybeDecrypt(
+      ValueProvider<String> unencryptedToken, ValueProvider<String> kmsKey) {
+    return new KMSEncryptedNestedValueProvider(unencryptedToken, kmsKey);
```
rarsan, line 377: thx @melbrodrigues for this update.
Shouldn't this parameter be named `encryptedToken`?

sabhyankar: Good catch @rarsan - I think the value may or may-not-be encrypted tbh since this is backwards compatible. The user could choose to send the hec token without any encryption.

Since this is a private method, we can correct it post merge.

PR #128, opened 2020-05-22
Fix batched messages sent to deadletter topic for PubSubToSplunk template

melbrodrigues:
Fixes concatenated messages sent to the deadletter topic if the `batchCount` is set to higher than 1.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
+++ b/src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
@@ -24,13 +24,14 @@ import com.google.api.client.http.HttpResponseException;
 import com.google.auto.value.AutoValue;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.security.KeyManagementException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.util.List;
-import java.util.Optional;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.metrics.Metrics;
@@ -81,6 +82,9 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
   private Integer batchCount;
   private Boolean disableValidation;
   private HttpEventPublisher publisher;
+  
+  private static final Gson GSON =
+      new GsonBuilder().setFieldNamingStrategy(f -> f.getName().toLowerCase()).create();
 
   public static Builder newBuilder() {
     return new AutoValue_SplunkEventWriter.Builder();
@@ -207,16 +211,15 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
       try {
         // Important to close this response to avoid connection leak.
         response = publisher.execute(events);
-        Optional<SplunkWriteError> maybeError = checkResponse(response);
 
-        if (maybeError.isPresent()) {
-          receiver.output(maybeError.get());
+        if (!response.isSuccessStatusCode()) {
+          flushWriteFailures(
+              events, response.getStatusMessage(), response.getStatusCode(), receiver);
           logWriteFailures(countState);
 
         } else {
           LOG.info("Successfully wrote {} events", countState.read());
           SUCCESS_WRITES.inc(countState.read());
-
         }
 
       } catch (HttpResponseException e) {
@@ -225,24 +228,13 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
             e.getStatusCode(), e.getContent(), e.getStatusMessage());
         logWriteFailures(countState);
 
-        String payload = publisher.getStringPayload(events);
-        SplunkWriteError error = SplunkWriteError.newBuilder()
-            .withStatusCode(e.getStatusCode())
-            .withStatusMessage(e.getStatusMessage())
-            .withPayload(payload)
-            .build();
-        receiver.output(error);
+        flushWriteFailures(events, e.getStatusMessage(), e.getStatusCode(), receiver);
 
       } catch (IOException ioe) {
         LOG.error("Error writing to Splunk: {}", ioe.getMessage());
         logWriteFailures(countState);
 
-        String payload = publisher.getStringPayload(events);
-        SplunkWriteError error = SplunkWriteError.newBuilder()
-            .withStatusMessage(ioe.getMessage())
-            .withPayload(payload)
-            .build();
-        receiver.output(error);
+        flushWriteFailures(events, ioe.getMessage(), null, receiver);
 
       } finally {
         // States are cleared regardless of write success or failure since we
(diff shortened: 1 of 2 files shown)


melbrodrigues:
Sure @sabhyankar we can take a stab at adding those attributes later. 
I'll modify this PR to just send failed messages separately to the deadletter topic for now.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -225,24 +230,32 @@ private void flush(
...
+          String payload = GSON.toJson(event);
+          SplunkWriteError error =
+              SplunkWriteError.newBuilder()
+                  .withStatusCode(e.getStatusCode())
+                  .withStatusMessage(e.getStatusMessage())
+                  .withPayload(payload)
+                  .build();
+          
+          receiver.output(error);
+        }
```
sabhyankar, line 243: Lets see what we can do here to avoid duplicating this block twice in both the exception cases.

melbrodrigues: Shifted these blocks to a new method `flushWriteFailures`.

melbrodrigues:
> I think you are missing 1 of the cases where SplunkWriteError is written:
> 
> * When checkResponse returns an optional with an error.

checkResponse() adds `response.parseAsString()` to the SplunkWriteError payload. Is that a string of concatenated SplunkEvents?

> Also once you make this change, you might be able to mark the getStringPayload method in HttpEventPublisher as private.

This method's currently marked as protected since it's being used in HttpEventPublisherTest. I can add a VisibleForTesting annotation to the method.

sabhyankar:
> > I think you are missing 1 of the cases where SplunkWriteError is written:
> > 
> > * When checkResponse returns an optional with an error.
> 
> checkResponse() adds `response.parseAsString()` to the SplunkWriteError payload. Is that a string of concatenated SplunkEvents?
> 
> > Also once you make this change, you might be able to mark the getStringPayload method in HttpEventPublisher as private.
> 
> This method's currently marked as protected since it's being used in HttpEventPublisherTest. I can add a VisibleForTesting annotation to the method.

Can you make these package private in that case? You can then add the annotation.

melbrodrigues:
@sabhyankar Fixed the case I missed earlier.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -264,16 +266,35 @@ private void logWriteFailures(@StateId(COUNT_STATE_NAME) ValueState<Long> countS
...
    * A helper method to check the {@link HttpResponse} object and wrap the response into an {@link
    * Optional} {@link SplunkWriteError} in case of an unsuccessful response.
    *
    * @param response {@link HttpResponse} object
-   * @return {@link Optional} {@link SplunkWriteError}
+   * @return {@link Optional} {@link SplunkWriteError.Builder}
    */
-  private Optional<SplunkWriteError> checkResponse(HttpResponse response)
-      throws IOException {
+  private Optional<SplunkWriteError.Builder> checkResponse(HttpResponse response) {
```
sabhyankar, line 297: I prefer to not return a builder here. What else can we return to make this cleaner? We basically need the statusCode and statusMessage from the response if it is unsuccessful.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -264,16 +266,35 @@ private void logWriteFailures(@StateId(COUNT_STATE_NAME) ValueState<Long> countS
...
+  
+  /**
+   * Utility method to flush failed write events.
+   *
+   * @param builder {@link SplunkWriteError.Builder} object
+   * @param events List of {@link SplunkEvent}s to un-batch
+   * @param receiver Receiver to write {@link SplunkWriteError}s to
+   */
+  private void flushWriteFailures(
+      SplunkWriteError.Builder builder,
```
sabhyankar, line 278: We basically just need the statusCode and statusMessage. Lets see if we can find a way to just pass those (either individually or wrapped inside of a custom object). We shouldnt need to pass a builder. SplunkWriteError can be built inside this method.

melbrodrigues: I'll change this to accept the statusCode and statusMessage along with the list of SplunkEvents and the output reciever. We'd then be able to build and output SplunkWriteErrors within the method.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: We can check the HttpResponse success status code inside the existing `flush` method itself and call flushWriteFailures to build and output failed events based on the success status code. We won't need the `checkResponse` method then.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -266,29 +258,32 @@ private void logWriteFailures(@StateId(COUNT_STATE_NAME) ValueState<Long> countS
...
+      Integer statusCode,
+      OutputReceiver<SplunkWriteError> receiver) {
 
-      if (response.parseAsString() != null) {
-        builder.withPayload(response.parseAsString());
-      }
-      SplunkWriteError error = builder.build();
-      return Optional.of(error);
+    SplunkWriteError.Builder builder = SplunkWriteError.newBuilder();
+    builder.withStatusMessage(statusMessage);
```
sabhyankar, line 275: Maybe add an if (statusMessage != null) check here similar to what you are doing for statusCode?

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -266,29 +258,32 @@ private void logWriteFailures(@StateId(COUNT_STATE_NAME) ValueState<Long> countS
...
+
+    if (statusCode != null) {
+      builder.withStatusCode(statusCode);
+    }
+
+    for (SplunkEvent event : events) {
```
sabhyankar, line 281: You are assuming `events` would never be null :) If so, better add a checkNotNull() at the beginning of this method as a guardrail?

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -266,29 +258,32 @@ private void logWriteFailures(@StateId(COUNT_STATE_NAME) ValueState<Long> countS
...
-  private Optional<SplunkWriteError> checkResponse(HttpResponse response)
-      throws IOException {
-    if (!response.isSuccessStatusCode()) {
-      SplunkWriteError.Builder builder = SplunkWriteError.newBuilder();
-      builder.withStatusCode(response.getStatusCode());
-
-      if (response.getStatusMessage() != null) {
-        builder.withStatusMessage(response.getStatusMessage());
-      }
+  private void flushWriteFailures(
```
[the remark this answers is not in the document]
melbrodrigues: Changed to static.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Done.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Good catch! I'll add a checkNotNull() at the beginning of this method.

PR #133, opened 2020-05-28
Updated merge statement to avoid duplicates

schalkva:
Changed merges to avoid duplicate PK rows in the replica table caused by rapid inserts and updates on the same PK

### DIFF
--- a/v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/MergeStatementBuildingFn.java
+++ b/v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/MergeStatementBuildingFn.java
@@ -143,8 +143,8 @@ public class MergeStatementBuildingFn
       "ON %s ",
       "WHEN MATCHED AND changelog.operation = \"DELETE\" ",
       "THEN DELETE ",
-      "WHEN MATCHED THEN %s ",
-      "WHEN NOT MATCHED BY TARGET AND changelog.operation != \"DELETE\" THEN %s");
+      "WHEN MATCHED AND changelog.operation = \"UPDATE\" THEN %s ",
+      "WHEN NOT MATCHED BY TARGET AND changelog.operation = \"INSERT\" THEN %s");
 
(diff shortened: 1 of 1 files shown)


v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/MergeStatementBuildingFn.java
```diff
@@ -1,241 +1,241 @@
...
+      "CREATE TABLE IF NOT EXISTS `%s.%s.%s` ");
+
+  static final String MERGE_REPLICA_WITH_CHANGELOG_TABLES = String.join("",
+      "MERGE `%s.%s.%s` AS replica ",
+      "USING (%s) AS changelog ",
+      "ON %s ",
+      "WHEN MATCHED AND changelog.operation = \"DELETE\" ",
+      "THEN DELETE ",
+      "WHEN MATCHED AND changelog.operation = \"UPDATE\" THEN %s ",
+      "WHEN NOT MATCHED BY TARGET AND changelog.operation = \"INSERT\" THEN %s");
```
pabloem, line 147: In this case, we can have a INSERT then UPDATE of the same primary key before a merge occurs. When this happens, we'll have no match on the target, but `operation == INSERT => FALSE` - and the action we actually want to do is to insert.

schalkva:
@pabloem, I'm putting this on hold for the moment. 
If I understand the usage of the timestampMS field correctly, does it map to the Debezium payloady ts_ms field?

pabloem:
that's correct @schalkva

schalkva:
@pabloem, so what I'm seeing is that the different operations collected within the same event batch gets added with the same timestamp. If a primary key has multiple events in the same batch you end up with duplicates. I can workaround this by changing the column used in the max function for the specific case I'm working on but I'm trying to figure out a solution that can be applied to more situations.  I'm not too familiar with the Debezium connector but I did notice GTID's can be included in the payload. If the payload merely contains the UUID and transaction ID, then max ts_ms and transaction ID on the PK could potentially be used to workaround the duplicates. However if it's the UUID with the transaction set it was contained within then it might not be as useful.

PR #136, opened 2020-06-02
Fix for missing table whitelist in Dataflow CDC Example

nbali:


v2/cdc-parent/cdc-embedded-connector/src/main/java/com/google/cloud/dataflow/cdc/connector/DebeziumMysqlToPubSubDataSender.java
```diff
@@ -109,6 +109,7 @@ public DebeziumMysqlToPubSubDataSender(
         .with("database.password", this.mysqlUserPassword)
         .with("database.server.name", mysqlDatabaseInstanceName)
         .with("decimal.handling.mode", "string")
+        .with(MySqlConnectorConfig.TABLE_WHITELIST, dbzWhitelistedTables)
```
pabloem, line 112: Can you add the argument conditionally? So that if no tables are whitelisted, all tables will be allowed[1]

[1] https://debezium.io/documentation/reference/1.1/connectors/mysql.html

nbali: Done.

PR #143, opened 2020-06-13
Add option for write disposition in TextIOToBigQuery

gustavoatt:
Make write disposition configurable in the `TextIOToBigQuery` DataFlow template. This should allow users to define whether tables are truncated or appended when there is data already on the output table.

@sabhyankar

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/TextIOToBigQuery.java
+++ b/src/main/java/com/google/cloud/teleport/templates/TextIOToBigQuery.java
@@ -30,6 +30,7 @@ import org.apache.beam.sdk.io.TextIO;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.CreateDisposition;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.WriteDisposition;
+import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.options.Validation;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/teleport/templates/TextIOToBigQuery.java
```diff
@@ -147,7 +154,7 @@ public TableSchema apply(String jsonPath) {
                         }))
                 .to(options.getOutputTable())
                 .withCreateDisposition(CreateDisposition.CREATE_IF_NEEDED)
-                .withWriteDisposition(WriteDisposition.WRITE_APPEND)
+                .withWriteDisposition(options.getWriteDisposition())
```
sabhyankar, line 157: Since the WriteDisposition method for BigQueryIO does not support a [ValueProvider](https://cloud.google.com/dataflow/docs/guides/templates/creating-templates#runtime-parameters-and-the-valueprovider-interface), the value provided at template creation time is the value used at runtime as well. I.e Even with this patch, the WriteDisposition will always be write_append. 

You could test this by creating a template and then attempting to _run the template_ with different options for this Enum.

gustavoatt: Ohh that makes sense. 

So the only way to allow passing a runtime parameter to determine write disposition would be to add a method in `BigQueryIO` that takes a `ValueProvider<WriteDisposition>` correct?

sabhyankar: That is one possible option. The other option (preferred) would be to convert this pipeline to a Flex template (see [file-format-conversion](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/master/v2/file-format-conversion) as an example). 

I would be happy to review and get that PR merged if you are interested in opening one.

gustavoatt: Sounds good, I'll work on that PR. It does sound like Flex templates are the way to go in this case 👍

sabhyankar: sgtm! Feel free to open a separate PR once you are ready.

PR #145, opened 2020-06-18
added pubsub-to-cloudstorage template based on subscription with compression

nickcarenza:
This is an extension to the PubsubToText template that uses a named subscription as input instead of a topic and supports compression on write.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubSubToCloudStorageCompressed.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubSubToCloudStorageCompressed.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright (C) 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/teleport/templates/PubSubToCloudStorageCompressed.java
```diff
@@ -0,0 +1,222 @@
...
+            options.getWindowDuration() + " Window",
+            Window.into(FixedWindows.of(DurationUtils.parseDuration(options.getWindowDuration()))))
+
+        // Apply windowed file writes. Use a NestedValueProvider because the filename
+        // policy requires a resourceId generated from the input value at runtime.
+        .apply(
+            "Write File(s)",
+            TextIO.write()
+                .withWindowedWrites()
+                .withCompression(options.getCompression())
```
sabhyankar, line 182: Since compression does not have a ValueProvider, the pipeline would use the static value available at _compile time_. This would not be something that the user can change at _runtime_. i.e For example, if the template were compiled with a compression value of GZIP, then the template would always use GZIP as the compression regardless of what is specified at runtime.

nickcarenza: Oooh I see now. Thank you I will update compression to use the a provider.

nickcarenza:
> Thank you for the PR! Given that these template would closely resemble the existing Pub/Sub to Text template with the addition of --subscription and --compression support, I would suggest creating a [Flex template](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/master/v2) that would be **backward compatible** and provide these two additional options.
> 
> We could then deprecate and eventually remove the older template.
> 
> I would be happy to review and help merge a PR if you were interested in opening one for converting this to a Flex template.

Thank you. I will resubmit this as a Flex template as you suggested.

sabhyankar:
Thanks! Closing this based on the previous comment.

PR #146, opened 2020-06-22
To allow whitelistedtables as regex

gkstechie:
Debezium allows whitelisted tables as patterns/regex so this small piece of code/change set will allow only if tables matched with whitelisted patterns/regex.


pabloem:
Have you tested this? I think this may not work because we're also passing the whitelisted tables to debezium. Just confirming - have you tested this? You can skip passing them to debezium in the table.whitelist property, and then this should work.

gkstechie:
@pabloem  yes I've tested that. Debezium supports whitelisted tables as patterns by default in the same parameter.

PR #159, opened 2020-08-13
Add support for custom DateTime patterns for WindowedFilenamePolicy class

### LINKED ISSUE #155: Cloud_PubSub_to_GCS_Text replaces `mm` in path
`recommendation` becomes `reco00endation`, `reco05endation`, ... when using the following template:

`gs://dataflow-templates/2020-07-13-00_RC00/Cloud_PubSub_to_GCS_Text`

melbrodrigues:
Allows WindowedFilenamePolicy class to use a custom DateTime pattern for the output directory.

Resolves #155

### DIFF
--- a/src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
+++ b/src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
@@ -16,6 +16,10 @@
 
 package com.google.cloud.teleport.io;
 
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
 import org.apache.beam.sdk.io.DefaultFilenamePolicy;
 import org.apache.beam.sdk.io.FileBasedSink.FilenamePolicy;
 import org.apache.beam.sdk.io.FileBasedSink.OutputFileHints;
@@ -27,6 +31,7 @@ import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
 import org.apache.beam.sdk.transforms.windowing.IntervalWindow;
 import org.apache.beam.sdk.transforms.windowing.PaneInfo;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.MoreObjects;
 import org.joda.time.DateTime;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -137,22 +156,37 @@ public ResourceId unwindowedFilename(
...
     if (window instanceof IntervalWindow) {
       IntervalWindow intervalWindow = (IntervalWindow) window;
       DateTime time = intervalWindow.end().toDateTime();
       String outputPath = outputDirectory.toString();
-      outputPath = outputPath.replace("YYYY", YEAR.print(time));
-      outputPath = outputPath.replace("MM", MONTH.print(time));
-      outputPath = outputPath.replace("DD", DAY.print(time));
-      outputPath = outputPath.replace("HH", HOUR.print(time));
-      outputPath = outputPath.replace("mm", MINUTE.print(time));
+      String customDateTimePattern = pattern.get();
```
sabhyankar, line 171: Can this never be null?

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -137,22 +156,37 @@ public ResourceId unwindowedFilename(
...
+    
+      if (customDateTimePattern.equals("")) {
```
sabhyankar, line 173: You would get an NPI if customDatetimePattern is a null.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -137,22 +156,37 @@ public ResourceId unwindowedFilename(
...
+        outputPath = outputPath.replace("YYYY", YEAR.print(time));
+        outputPath = outputPath.replace("MM", MONTH.print(time));
+        outputPath = outputPath.replace("DD", DAY.print(time));
+        outputPath = outputPath.replace("HH", HOUR.print(time));
+        outputPath = outputPath.replace("mm", MINUTE.print(time));
```
sabhyankar, line 178: I think what we need is a way for the user to override what the pattern would be for each of the 5 components (year, month, day, hour and minute).

For example, a user may specify that the replacement pattern for minute should be `mx` instead of the default `mm`.

melbrodrigues:
> Take a look at the review comments. My suggestion would be to look at a builder/fluent pattern so as to not make the constructor look unweildy.
> 
> You could look at an example of the pattern [here](https://github.com/apache/beam/blob/2def4cc32188e8ea1128e72bce57e99c398a6bd2/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkIO.java#L179-L247).

@sabhyankar Okay I'll change this to use a builder pattern.

Do I need to add a second builder to accept `String` values as well or just one that accepts `ValueProvider<String>` would suffice? The current constructor that accepts `String` values is not being used anywhere.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: What should the input format for this option be? A comma separated String of the overridden values? Something like `mm=mx,HH=hx`...

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: I was setting an empty String as the default value for customDateTimePattern. Added a check now to avoid the NPE.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
sabhyankar: The fluent builder pattern can have methods that allow us to override the patterns for each of these. So something like this:

```
WindowedFilenamePolicy policy = WindowedFilenamePolicy.of(<existing_input_params>);
...
// Override minute pattern from default "mm" to "xx"
policy.setMinutePattern("xx");

// Override day pattern from default "DD" to "ZZ"
policy.setDayPattern("ZZ");

```

melbrodrigues: So does this mean each template that uses the `WindowedFilenamePolicy` class would have to have a separate input parameter for each Date component? Like:
```
WindowedFilenamePolicy.newBuilder()
                      .withOutputDirectory(options.getOutputDirectory())
                      .withOutputFilenamePrefix(options.getOutputFilenamePrefix())
                      .withShardTemplate(options.getOutputShardTemplate())
                      .withSuffix(options.getOutputFilenameSuffix())
                      .withYearPattern(options.getYearPattern())
                      .withMonthPattern(options.getMonthPattern()) // and so on
                      .build();
```  
\
Currently, I've changed the class to use the Builder pattern as follows: 
(assuming we're taking all the patterns to be overridden in one input parameter `--customDateTimePattern` from the user)

```
WindowedFilenamePolicy.newBuilder()
                      .withOutputDirectory(options.getOutputDirectory())
                      .withOutputFilenamePrefix(options.getOutputFilenamePrefix())
                      .withShardTemplate(options.getOutputShardTemplate())
                      .withSuffix(options.getOutputFilenameSuffix())
                      .withPattern(options.getCustomDateTimePattern())
                      .build();
```

What should the format for `options.getCustomDateTimePattern()` be?

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -38,60 +42,273 @@
...
-  /** The prefix of the file to output. */
-  private final ValueProvider<String> outputFilenamePrefix;
-  /** The filename suffix. */
-  private final ValueProvider<String> suffix;
-  /** The shard template used during file formatting. */
-  private final ValueProvider<String> shardTemplate;
-
+  
+  /** The default shard template used during file formatting. */
+  private static final String DEFAULT_SHARD_TEMPLATE = "W-P-SS-of-NN";
```
sabhyankar, line 51: Why do we need a default template given that we didnt have one earlier?

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -38,60 +42,273 @@
...
+  
+  /** The default filename suffix */
+  private static final String DEFAULT_SUFFIX = "";
+  
+  /** The default filename prefix */
+  private static final String DEFAULT_PREFIX = "output";
```
sabhyankar, line 57: Same here - Why would this be needed now?

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -38,60 +42,273 @@
...
+  
+  public static WindowedFilenamePolicy newBuilder() {
+    return new AutoValue_WindowedFilenamePolicy.Builder().build();
+  }
```
sabhyankar, line 61: A newBuilder() should return a Builder and not the actual annotated class.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -38,60 +42,273 @@
...
+  abstract ValueProvider<String> outputFilenamePrefix();
+  
+  @Nullable
+  abstract ValueProvider<String> shardTemplate();
+  
+  @Nullable
+  abstract ValueProvider<String> suffix();
+  
+  @Nullable
+  abstract ValueProvider<String> year();
```
sabhyankar, line 76: I prefer we call this (and below) yearPattern() so as to make it less ambiguous.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -38,60 +42,273 @@
...
+  @Nullable
+  abstract ValueProvider<String> day();
+  
+  @Nullable
+  abstract ValueProvider<String> hour();
+  
+  @Nullable
+  abstract ValueProvider<String> minute();
+  
+  abstract Builder toBuilder();
```
sabhyankar, line 90: Why do we have both a toBuilder() and a newBuilder()?

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -137,22 +378,70 @@ public ResourceId unwindowedFilename(
...
+      String dayPattern = "DD";
+      String hourPattern = "HH";
+      String minutePattern = "mm";
+  
+      // Use a custom pattern, if provided.
+      if (year() != null) {
+        if (year().get() != null) {
+          yearPattern = year().get();
+        }
+      }
```
sabhyankar, line 406: Do you think it would be useful to log this at a debug level perhaps? At most we have a log every minute? See how this is done using MoreObjects.firstNotNull

https://github.com/apache/beam/blob/aa345dbeb9f6bdb36f5acf0fb8958777253acd51/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkEventWriter.java#L119-L129

sabhyankar (dismissed):

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: Ah, I got these from the default values set in the Pubsub to Text template. I'll take them out.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: I'll change this to `writeWindowedFiles()` since we're not returning a Builder.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: `toBuilder()` would be called internally after a `withX()` method is called.

Changing `newBuilder()` to `writeWindowedFiles()`

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: Yup, I'll add a few log statements for this

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: Done

melbrodrigues:
Hi @sabhyankar , Would be great to get your feedback on this.

sabhyankar:
Hi @melbrodrigues - I will take a look at this.

an2x:
I'm not sure how this change resolves the original issue #155. The problem was that, for example, "mm" in "recommendation" in the directory name was getting replaced with minutes.

Now, with this change, if I want to have the directory as something like `gs://bucket/recommendations/<minute>` what do I do?
If I set `minutePattern` to `m` or `mm`, it's still going to replace all the letters m in "recommendations". 
If I set `minutePattern` to something like `%mm` and the directory to `gs://bucket/recommendations/%mm`, it still won't work properly: the result will be `gs://bucket/recommendations/%12` (`%` won't be replaced).

I suggest implementing a slightly different fix:

1. Introduce just one new parameter, let's call it `directoryDateTimePattern`, example value that the user would set: `YYYY/MM/DD/HH-mm`.
2. In the directory name, replace `${dateTime}` placeholder with the datetime formatted using `directoryDateTimePattern`. If the placeholder doesn't exist, fallback to the old logic (replace `YYYY`, `MM`, `DD`, etc. separately), for backwards compatibility.
3. The directory specified by the user would look like `gs://bucket/recommendations/${dateTime}`.

I think this would be better because this approach is more flexible (what if I don't want minutes in my directory name at all?) and there is no need to set 5 different parameters (and if you do, you may run into many other issues, e.g. what if `minutePattern` is overridden but `hourPattern` isn't - is this a misconfiguration?).

melbrodrigues:
I think the intention was to give users the ability to override default patterns and avoid unintentional changes to their file paths.

For the `gs://bucket-name/recommendations/` example, the user need only pass one additional parameter `--withMinutePattern` (which can be set to say `mmmm`) to override the default pattern `mm`.

And if the user wants to include patterns in the directory name they can still do so with specific overrides. For example, an override of `--withMinutePattern=mmmm` would result in the path `gs://bucket-name/recommendations/mmmm` being transformed as `gs://bucket-name/recommendations/<minute>`.

Note that the user does not need to pass all parameter overrides and the overrides must be a valid pattern as documented in [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
```diff
@@ -137,22 +370,69 @@ public ResourceId unwindowedFilename(
...
+        final DateTimeFormatter minuteFormatter = DateTimeFormat.forPattern(minutePattern);
+    
+        outputPath = outputPath.replace(yearPattern, yearFormatter.print(time));
+        outputPath = outputPath.replace(monthPattern, monthFormatter.print(time));
+        outputPath = outputPath.replace(dayPattern, dayFormatter.print(time));
+        outputPath = outputPath.replace(hourPattern, hourFormatter.print(time));
+        outputPath = outputPath.replace(minutePattern, minuteFormatter.print(time));
+      } catch (IllegalArgumentException e) {
+        LOG.error("Invalid DateTime pattern: {}", e.getMessage());
+        throw new RuntimeException(e);
```
an2x, line 433:
```suggestion
        throw new RuntimeException("Invalid DateTime pattern", e);
```
Usually logging and re-throwing the same exception is considered an anti-pattern. Any strong reason to do it here?

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
```diff
@@ -119,6 +119,31 @@
...
+  
+    @Description("The custom hour pattern to use for the output directory.")
+    ValueProvider<String> getHourPattern();
+  
+    void setHourPattern(ValueProvider<String> hourPattern);
+  
+    @Description("The custom minute pattern to use for the output directory.")
+    ValueProvider<String> getMinutePattern();
+  
+    void setMinutePattern(ValueProvider<String> minutePattern);
```
an2x, line 146: I suggest moving all these options (+ the old output* options like `outputShardTemplate`) to a separate class (e.g. `WindowedFilenamePolicyOptions`), similar to how `StreamingOptions` are extended here, to avoid duplication in `PubsubToAvro` and `PubsubToText` templates. I'd put it in ` com.google.cloud.teleport.options` package.

src/test/java/com/google/cloud/teleport/io/WindowedFilenamePolicyTest.java
```diff
@@ -214,18 +218,126 @@ public void testWithDynamicDirectory() throws IOException {
...
+    WindowedFilenamePolicy policy =
+        WindowedFilenamePolicy.writeWindowedFiles()
+            .withOutputDirectory(StaticValueProvider.of(outputDirectory.toString()))
+            .withOutputFilenamePrefix(StaticValueProvider.of("output"))
+            .withShardTemplate(StaticValueProvider.of("-SSS-of-NNN"))
+            .withSuffix(StaticValueProvider.of(null))
+            .withMinutePattern(StaticValueProvider.of(invalidMinutePattern));
+    policy.windowedFilename(1, 1, window, paneInfo, new TestOutputFileHints());
+  }
 }
```
an2x, line 343: Please add a test to verify that when a **custom** pattern is set the **default** pattern is not being replaced.
E.g. if the directory is something like ".../recommendations/mmmm" and the `minutePattern` is `mmmm`, then the resulting directory is ".../recommendations/0047" and not ".../reco47endations/0047".

an2x (changes requested):

an2x:
> And if the user wants to include patterns in the directory name they can still do so with specific overrides. For example, an override of `--minutePattern=mmmm` would result in the path `gs://bucket-name/recommendations/mmmm` being transformed as `gs://bucket-name/recommendations/<minute>`.

OK, got it, thanks. I think this is good enough. I left a few other comments, should be the last review round :)

src/main/java/com/google/cloud/teleport/templates/PubsubToText.java
melbrodrigues: Makes sense. Will move these to a new `WindowedFilenamePolicyOptions` interface

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: Not really. I saw it at a few other places in the repo back then. I'll can take out the log statements.

src/test/java/com/google/cloud/teleport/io/WindowedFilenamePolicyTest.java
melbrodrigues: Ah yes, I had added a test for this after our discussion yesterday. Will push it along with the other changes.

src/main/java/com/google/cloud/teleport/io/WindowedFilenamePolicy.java
melbrodrigues: Should I remove the try-catch block as well? The catch block was just to log the error message and throw a new RuntimeException. If I remove the log statement all we're doing is catching an IllegalArgumentException and throwing a RuntimeException.

an2x: I'd keep it because IllegalArgumentException doesn't make it clear for the user what pattern is incorrect (e.g. if an empty pattern is supplied it just says "Invalid pattern specification"). Here the RuntimeException at least says that the error was caused by a DateTime pattern.

I'd go even further and include even more details in the error message, e.g. what pattern is incorrect (add a helper method to avoid try-catch around each of the 5 patterns). But it's up to you if you want to make such a change.

melbrodrigues: I was wondering if it's a good idea to throw a RuntimeException... Since this resolution happens after we read from Pub/Sub, users may lose some data. Maybe we should use the default values if the pattern provided is invalid? WDYT?

an2x: I'd still throw the Exception. If a user set a pattern override, this means the default pattern doesn't work for them for some reason. If the override was set incorrectly, simply ignoring this and falling back to the fault pattern may cause unwanted side effects (e.g. output written to a wrong folder, potentially overwriting existing data). IMO it's better to throw an exception and not process anything in this case.

As for the data loss, it shouldn't happen. Dataflow guarantees "exactly once" processing semantics ([some info about it](https://cloud.google.com/dataflow/docs/concepts/streaming-with-cloud-pubsub)). In case of PubSub, messages read from a topic are not going to be ACKed if there is a failure in the middle of the pipeline and PubSub is going to re-deliver them on the next processing attempt.

melbrodrigues: Perfect. I'll leave the RuntimeException there then

an2x (approved):

src/main/java/com/google/cloud/teleport/options/WindowedFilenamePolicyOptions.java
```diff
@@ -0,0 +1,64 @@
+package com.google.cloud.teleport.options;
```
zhoufek, line 1: This file seems to be the cause of the `import/copybara` failure preventing merging. You'll need to add a license header similar to the other files. You could copy/paste from another file, though please update the year to 2021.

zhoufek (changes requested):

src/main/java/com/google/cloud/teleport/options/WindowedFilenamePolicyOptions.java
melburnerodrigues: Thanks, I'll add that. 

`mvn clean install` seems to complain about a missing _package-info_ file for the new _options_ package, so I'll add that as well.

melbrodrigues: Thanks, I'll add that.

`mvn clean install` seems to complain about a missing _package-info_ file for the new _options_ package, so I'll add that as well.

zhoufek (approved):

PR #160, opened 2020-08-17
(WIP) add support for throttling bulk writes

aananthram:


aananthram:
@googlebot I signed it!

src/main/java/com/google/cloud/teleport/bigtable/AvroToBigtable.java
```diff
@@ -107,10 +115,24 @@ public static void main(String[] args) {
   public static PipelineResult run(Options options) {
     Pipeline pipeline = Pipeline.create(options);
 
+    BulkOptions.Builder bulkOptionsBuilder = BulkOptions.newBuilder();
+    if (options.getEnableBulkMutationThrottling() != null && options.getEnableBulkMutationThrottling().get()) {
+        bulkOptionsBuilder.enableBulkMutationThrottling();
+    }
+
```
sabhyankar, line 122: Hi @aananthram - This part runs at graph creation time and not at runtime so whether we `enableBulkMutationThrottling` will be determined at the template creation and is not going at runtime.

sabhyankar: Looking at our SDK, it seems the projectid and instanceid should not be provided[1]

"WARNING: instanceId and projectId should not be provided here and should be provided over withProjectId(org.apache.beam.sdk.options.ValueProvider<java.lang.String>) and withInstanceId(org.apache.beam.sdk.options.ValueProvider<java.lang.String>)."

[1] https://beam.apache.org/releases/javadoc/2.23.0/org/apache/beam/sdk/io/gcp/bigtable/BigtableIO.Write.html#withBigtableOptionsConfigurator-org.apache.beam.sdk.transforms.SerializableFunction-

aananthram: oh, i see. so that allows them to be provided at runtime. is there a way to enable bulk mutation at runtime then?

PR #163, opened 2020-08-27
cdc pipeline band-aids

dzoellner:
This isn't a real fix to the underlying issue, but a band-aid to allow the tool to continue functioning until a real fix is implemented.


dzoellner:
Ooops meant to make this PR to my fork

pabloem:
thanks for showing this fix. Yes, we need to add something like this to the pipeline. I'll work on that.

PR #166, opened 2020-08-31
Support map attributes for StreamingDataGenerator

ricardolui:
Added support to generate dynamic attributes for PubSub messages.

### DIFF
--- a/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
+++ b/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
@@ -190,7 +190,8 @@ public class StreamingDataGenerator {
 
     /**
      * The {@link MessageGeneratorFn} class generates {@link PubsubMessage} objects from a supplied
-     * schema and populating the message with fake data.
+     * schema and populating the message with fake data. Also supports supplying schema for PubSub
+     * message attributes.
      *
      * <p>See <a href="https://github.com/vincentrussell/json-data-generator">json-data-generator</a>
      * for instructions on how to construct the schema file.
--- a/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
+++ b/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
@@ -19,6 +19,14 @@ package com.google.cloud.teleport.v2.templates;
 import com.github.vincentrussell.json.datagenerator.JsonDataGenerator;
 import com.github.vincentrussell.json.datagenerator.JsonDataGeneratorException;
 import com.github.vincentrussell.json.datagenerator.impl.JsonDataGeneratorImpl;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.channels.WritableByteChannel;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TimeZone;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
 import org.apache.beam.sdk.io.FileSystems;
@@ -38,15 +46,6 @@ import org.joda.time.Duration;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.nio.channels.Channels;
-import java.nio.channels.ReadableByteChannel;
-import java.nio.channels.WritableByteChannel;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.TimeZone;
 
 
 /**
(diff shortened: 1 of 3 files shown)


sabhyankar:
@prathapreddy123 - Can you review this PR?

prathapreddy123:
@ricardolui -  Adding attribute support is great.   However our goal is to support all the following scenarios:

- Attributes with fields that are not part of event schema (similar to the current PR) 
- Include certain fields that are part of event schema as attributes (values should remain same) 
- Mix of both as well

v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
```diff
@@ -97,144 +100,180 @@
...
+                    try (ReadableByteChannel readerChannel = FileSystems.open(metadataAttr.resourceId())) {
+                        try (WritableByteChannel writerChannel = Channels.newChannel(byteArrayOutputStream)) {
+                            ByteStreams.copy(readerChannel, writerChannel);
+                        }
+                    }
+                    attributeSchema = byteArrayOutputStream.toString();
+                }
+            }
+            //Set Timezone to UTC to generate datetime with UTC
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
prathapreddy123, line 248: Elaborate on why is this required?  By default DataGenerator generates UTC timestamp values

v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
```diff
@@ -45,103 +34,167 @@
...
+        //
+        PAssert.that(results)
+                .satisfies(
+                        input -> {
+                            PubsubMessage message = input.iterator().next();
+                            assertThat(message, is(notNullValue()));
+                            assertThat(message.getPayload(), is(notNullValue()));
+                            assertThat(message.getAttributeMap(), is(notNullValue()));
+                            assertThat(message.getAttribute("attribute_test"), is(notNullValue()));
+                            assertThat(Integer.parseInt(message.getAttribute("attribute_test")), is(greaterThan(5)));
```
prathapreddy123, line 169: Datagenerator uses random.nextInt(upperbound) + lowerbound to generate random integer values between min and max
https://docs.oracle.com/javase/8/docs/api/java/util/Random.html
https://github.com/vincentrussell/json-data-generator/blob/master/src/main/java/com/github/vincentrussell/json/datagenerator/functions/impl/FunctionUtils.java.
Should compare with greaterThanOrEqualTo

ricardolui:
Hi @prathapreddy123 , Have made the changes to include all cases.
Thanks for the comments. 
Implementation relies on passing {"payload": {}, "attributes": {}}

v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
ricardolui: Fixed. I implemented this because I ran a simple test outside of beam/dataflow.

v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
ricardolui: Fixed and changed implementation.

prathapreddy123:
> Hi @prathapreddy123 , Have made the changes to include all cases.
> Thanks for the comments.
> Implementation relies on passing {"payload": {}, "attributes": {}}

To support backward compatibility we cannot change the event schema model (i.e payload and attributes).   Below design outlines the approach to handle earlier described scenarios:

  Add two additional **optional** pipeline options
   a) attrbuteFieldNames -> comma separated list of field names that already exists in event schema but also needs to be included as attributes along with payload.  attributeValues and values inside payload should match.
   b) attrbuteSchemaLocation ->  Json schema of fields to be added as attributes that are not part of event schema

 For each generated event:
- If attrbuteFieldNames is provided parse the payload for corresponding fields and add them as attributes
- If attrbuteSchemaLocation is provided generate attribute values matching schema using JsonDataGenerator 

 We are working internally on below enhancements and should be available in Q4:

-  Attributes support 
-  Avro encoded messages

ricardolui:
@prathapreddy123 , the implementation **IS** backward compatible. 
If you don't specify the 2 fields (payload and attributes) the JSON, it will behave just like it was before, whilst covering the other scenarios.
I implemented in a way that it would take advantage of the updated features of the JSON data generator library with methods such as put and get value.
For Avro encoding on payload, would probably be easy to implement.

kenzik:
@prathapreddy123 - Is there an update on attribute support? It's a feature that I require when streaming GCS text files' rows to PubSub. Without the attributes (eg. filename, custom meta data, etc.) we have no context available in the message payload, aside from the subscription name.

You noted that work was being done and expected in Q4 2020. The current templates do not appear to support attributes yet. I'm looking to manage some expectations before moving on to alternate implementations.

prathapreddy123:
@kenzik - Yes the attribute support has been added. You can check https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/main/v2/streaming-data-generator#example-schema-file-to-generate-pubsub-message-with-attributes for example

PR #168, opened 2020-09-08
Options to Exclude Big Query Replica

andreasholmesberge:
In this PR I allow the option to execute the dataflow with fewer options and bypassing the automatic replica dataset part.

Note: I myself then generate my own custom replica with a query that is better optimised with clustering to avoid high costs for large tables.

Example
mvn exec:java -pl cdc-change-applier -Dexec.args="--runner=DataflowRunner \
              --inputSubscriptions=${PUBSUB_SUBSCRIPTION} \
              --changeLogDataset=${CHANGELOG_BQ_DATASET} \
              --project=${GCP_PROJECT} \
              --useSingleTopic=true"

### DIFF
--- a/v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/CdcToBigQueryChangeApplierPipeline.java
+++ b/v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/CdcToBigQueryChangeApplierPipeline.java
@@ -93,11 +93,12 @@ public class CdcToBigQueryChangeApplierPipeline {
     String getChangeLogDataset();
     void setChangeLogDataset(String dataset);
 
-    @Description("The BigQuery dataset where the Replica tables are to be kept.")
+    @Description("The BigQuery dataset where the Replica tables are to be kept. If empty replica operation is skipped.")
     String getReplicaDataset();
     void setReplicaDataset(String dataset);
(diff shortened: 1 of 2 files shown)


andreasholmesberge:
@googlebot I signed it!

v2/cdc-parent/cdc-change-applier/src/main/java/com/google/cloud/dataflow/cdc/applier/CdcToBigQueryChangeApplierPipeline.java
```diff
@@ -93,11 +93,12 @@
...
     void setChangeLogDataset(String dataset);
 
-    @Description("The BigQuery dataset where the Replica tables are to be kept.")
+    @Description("The BigQuery dataset where the Replica tables are to be kept. If empty replica operation is skipped.")
     String getReplicaDataset();
     void setReplicaDataset(String dataset);
 
-    @Description("How often the pipeline will issue updates to the BigQuery replica table.")
+    @Description("How often the pipeline will issue updates to the BigQuery replica table. Default Hourly")
+    @Default.Integer(1440)
```
pabloem, line 101: Can we avoid changing the default for now? I'd rather we don't change this behavior without consulting with other potential users.

pabloem (commented):
LGTM. Can we just avoid a default merge periodicity so users will specify it explicitly?

PR #178, opened 2020-10-13
Update README.md

larchiu:
Using the existing  imag spec causes issues - this is the correct one that works.


sabhyankar (approved):
LGTM
Thanks @larchiu

PR #181, opened 2020-10-23
Replacing Nashorn JS Engine with GraalVM js implementation.

prodriguezdefino:
Given the deprecation of Nashorn as the JS Engine for Java, and the future potential changes on the JDK requirements (from JDK8 to JDK11+) for this project, this PR adds the needed maven dependencies to automatically switch the JS Engine in use from the default one to the GraalVM implementation (truffle API). 

In the future, if this change sticks, we could completely remove the dependencies to the ScriptEngine API and replace them with GraalVM Polyglot Context's ones.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
@@ -13,7 +13,6 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-
 package com.google.cloud.teleport.templates.common;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -134,8 +143,7 @@ public Invocable getInvocable() throws ScriptException, IOException {
      */
     @Nullable
     private static Invocable newInvocable(Collection<String> scripts) throws ScriptException {
-      ScriptEngineManager manager = new ScriptEngineManager();
-      ScriptEngine engine = manager.getEngineByName("JavaScript");
+      ScriptEngine engine = new ScriptEngineManager().getEngineByName("JavaScript");
```
prodriguezdefino, line 146: We could enforce GraalVM by requesting by name `graal.js`.

sabhyankar:
@prathapreddy123 - Can you help review this to see if we should consider this. Note that we are still running JDK 8 for the time being.

@prodriguezdefino - Can you cleanup the formatting diff as its making this patch looking bigger than it really is?

prathapreddy123:
@prodriguezdefino - As sameer pointed can you rebase to latest version and update code to reflect the actual diff.

Issue #108, opened 2020-04-30
Improve handling of NUMERIC/DECIMAL, Time and Timestamp in CDC solution


pabloem:
Currently, these types are converted like so:

- `NUMERIC`/`DECIMAL` are converted to `STRING` in BigQuery
- `Time` is converted into milliseconds from midnight
- `Timestamp` is converted into milliseconds since epoch

damienhollis:
@pabloem any idea when these improvements will be made, we would like to use this solution but these limitations are holding us back?

pabloem:
oops - only NUMERIC is handled as of right now. Time and Timestamp are still remaining.

PR #185, opened 2020-11-03
Added support for debezium decimal.handling.mode=precise

damienhollis:
This pull request adds support for kafka Decimal fields, which means decimals will be stored as NUMERIC data type in BigQuery.

This partially resolves https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/108


pabloem:
thanks! Approving.

Issue #188, opened 2020-11-10
BigtableToAvro - Option to only retrieve latest version of a column


simenl:
**Scenario**
Dataflow jobs are occasionally failing due to large bigtable rows (above 256MiB).
The cause of the large bigtable rows are older versions that has not been garbage collected, yet.

**Possible solution**
Add option to only retrieve the latest versions from a row in the BigtableToAvro template.

<details>
<summary>Error message</summary>
<p>

Caused by: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Error while reading table '...' : Read returned 347MiB from row '...' which exceeds the limit of 256MiB. Make sure you are setting an appropriate request filter to retrieve only recent versions and only the columns you want. If columns are accumulating more versions than you need to read, you can also create a garbage collection policy: https://cloud.google.com/bigtable/docs/configuring-garbage-collection#versions io.grpc.Status.asRuntimeException(Status.java:521)

</p>
</details>

PR #189, opened 2020-11-10
BigtableToAvro: Option to limit the number of cells per column. #188

simenl:
This is a suggested solution to https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/188

### DIFF
--- a/src/main/java/com/google/cloud/teleport/bigtable/BigtableToAvro.java
+++ b/src/main/java/com/google/cloud/teleport/bigtable/BigtableToAvro.java
@@ -20,6 +20,7 @@ import com.google.bigtable.v2.Cell;
 import com.google.bigtable.v2.Column;
 import com.google.bigtable.v2.Family;
 import com.google.bigtable.v2.Row;
+import com.google.bigtable.v2.RowFilter;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/teleport/bigtable/BigtableToAvro.java
```diff
@@ -106,6 +113,11 @@ public static PipelineResult run(Options options) {
             .withInstanceId(options.getBigtableInstanceId())
             .withTableId(options.getBigtableTableId());
 
+    if (options.getCellsPerColumn().isAccessible()) {
```
sabhyankar, line 116: This would not actually work in a template as the conditional statement is evaluated at compile time (when the template is created) and not at runtime. As a result, the template will either be with or without this option depending on the value of --cellsPerColumn at compile time.

You should be able to verify it in your project.

simenl: Thanks! That makes sense.

It seems like this feature would be non-trivial to add as part of the template.

PR #192, opened 2020-11-12
Add support for Timestamps

damienhollis:
This PR adds support for mapping DATETIME columns in source DB to TIMESTAMP columns in BigQuery. 

Note: it does change the default behaviour of the template, which used to map DATETIME to INT64. Although, I think this was probably unusable by a lot of people because the value might have been millis, micros or nanos and there was no way to tell within BigQuery.

This addresses the Timestamp support in https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/108


damienhollis:
@pabloem is there any chance of getting this reviewed and into a release soon. I'm just trying to decide whether we need to use our own patched version of the dataflow template or there will be an official version soon?

pabloem:
so sorry that I missed this. LGTM.

prathapreddy123:
@damienhollis - Can you please rebase to latest master

damienhollis:
@prathapreddy123 I've rebased it off master

pabloem:
ah I'm seeing a couple internal tests failing due to this. Taking a look...

PR #202, opened 2020-12-21
Only consider changelog entries that were added within the last window

damienhollis:
The PR keeps track of the oldest timestamp of the changelog messages processed within a window, which allows us to only apply those changelog messages to the replica table. When the changelog is partitioned or clustered by timestampMs, this makes the query to find the changelog entries to apply more cost efficient.

I'm not an expert with windows, so I wasn't 100% sure I needed two windows (which I've done) or whether there was a better way. I also wasn't sure whether this might cause any issues that I haven't thought of.


damienhollis:
This PR was based on changes that have not been accepted yet, so I'll create a new PR.

PR #203, opened 2020-12-21
Only apply change log records within the current window to replica ta…

damienhollis:
The PR keeps track of the oldest timestamp of the changelog messages processed within a window, which allows us to only apply those changelog messages to the replica table. When the changelog is partitioned or clustered by timestampMs, this makes the query to find the changelog entries to apply more cost efficient.

I'm not an expert with windows, so I wasn't 100% sure I needed two windows (which I've done) or whether there was a better way. I also wasn't sure whether this might cause any issues that I haven't thought of.


damienhollis:
I had to revert this change locally because I got a strange error - I don't recall the details. 

I was thinking I might leave this open for now for your feedback on whether this is a good idea or not in the long run and if it is, get some feedback on the approach.

PR #207, opened 2020-12-27
use compiled regex

jyane:
Hi, thank you for your work as always, I'm skamimori@.

I'm not sure why we do this only for STRING.
If I understand BNF [1] on our public doc correctly, BYTES could have length attr as well.
In addition, the BNF explains the length attr can be hex.
It seems that our current implementation doesn't expect the hex value.

I don't think it improves so much, but parseSpannerDataType would
be called O(columns) times.

Thanks,

[1] https://cloud.google.com/spanner/docs/data-definition-language#scalars


jyane:
Friendly ping.

PR #221, opened 2021-02-21
Add gradle as part of dependency

omarismail94:
Let's have fun and add Gradle as an automation tool!


omarismail94:
Just like `mvn test`,  running `./gradlew test` or `build` throws out errors, but compiling works, and uploading a template seems to work!

omarismail94:
R: @sabhyankar

PR #227, opened 2021-03-04
Fix malformed URL exception for Pub/Sub to Splunk template

melbrodrigues:
This PR fixes the `java.net.MalformedURLException` thrown when the HEC URL does not include the protocol.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
+++ b/src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
@@ -22,8 +22,11 @@ import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Precondi
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.HttpResponseException;
 import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
+import com.google.common.net.InetAddresses;
+import com.google.common.net.InternetDomainName;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.io.IOException;
@@ -32,6 +35,8 @@ import java.security.KeyManagementException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.metrics.Metrics;
@@ -70,6 +75,12 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
   private static final String BUFFER_STATE_NAME = "buffer";
   private static final String COUNT_STATE_NAME = "count";
   private static final String TIME_ID_NAME = "expiry";
+  
+  @VisibleForTesting
+  protected static final String INVALID_URL_FORMAT_MESSAGE =
+      "Invalid url format. Url format should match PROTOCOL://HOST[:PORT], where PORT is optional. "
+          + "Supported Protocols are http and https. eg: http://hostname:8088";
+
   @StateId(BUFFER_STATE_NAME)
   private final StateSpec<BagState<SplunkEvent>> buffer = StateSpecs.bag();
 
@@ -106,6 +117,7 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
   public void setup() {
 
     checkArgument(url().isAccessible(), "url is required for writing events.");
+    checkArgument(isValidUrlFormat(url().get()), INVALID_URL_FORMAT_MESSAGE);
     checkArgument(token().isAccessible(), "Access token is required for writing events.");
 
     // Either user supplied or default batchCount.
@@ -291,6 +303,22 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
     }
   }
 
+  /**
+   * Checks whether the HEC URL matches the format PROTOCOL://HOST[:PORT].
+   *
+   * @param url for HEC event collector
+   * @return true if the URL is valid
+   */
+  private static boolean isValidUrlFormat(String url) {
+    Pattern pattern = Pattern.compile("^http(s?)://([^:]+)(:[0-9]+)?$");
+    Matcher matcher = pattern.matcher(url);
+    if (matcher.find()) {
+      String host = matcher.group(2);
+      return InetAddresses.isInetAddress(host) || InternetDomainName.isValid(host);
+    }
+    return false;
+  }
+
   @AutoValue.Builder
   abstract static class Builder {
 
@@ -317,6 +345,9 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
      */
     public Builder withUrl(ValueProvider<String> url) {
       checkArgument(url != null, "withURL(url) called with null input.");
+      if (url.isAccessible()) {
+        checkArgument(isValidUrlFormat(url.get()), INVALID_URL_FORMAT_MESSAGE);
+      }
       return setUrl(url);
     }
 
@@ -328,6 +359,7 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
      */
     public Builder withUrl(String url) {
       checkArgument(url != null, "withURL(url) called with null input.");
+      checkArgument(isValidUrlFormat(url), INVALID_URL_FORMAT_MESSAGE);
       return setUrl(ValueProvider.StaticValueProvider.of(url));
     }
 
@@ -384,5 +416,4 @@ public abstract class SplunkEventWriter extends DoFn<KV<Integer, SplunkEvent>, S
       return autoBuild();
     }
   }
-
 }
--- a/src/test/java/com/google/cloud/teleport/splunk/SplunkEventWriterTest.java
+++ b/src/test/java/com/google/cloud/teleport/splunk/SplunkEventWriterTest.java
@@ -80,7 +80,53 @@ public class SplunkEventWriterTest {
   }
 
   /**
-   * /** Test building {@link SplunkEventWriter} with missing token.
+   * Test building {@link SplunkEventWriter} with missing URL protocol.
+   */
+  @Test
+  public void eventWriterMissingURLProtocol() {
+
+    Exception thrown =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> SplunkEventWriter.newBuilder().withUrl("test-url").build());
+
+    assertThat(thrown).hasMessageThat().contains(SplunkEventWriter.INVALID_URL_FORMAT_MESSAGE);
+  }
+
+  /**
+   * Test building {@link SplunkEventWriter} with an invalid URL.
+   */
+  @Test
+  public void eventWriterInvalidURL() {
+
+    Exception thrown =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> SplunkEventWriter.newBuilder().withUrl("http://1.2.3").build());
+
+    assertThat(thrown).hasMessageThat().contains(SplunkEventWriter.INVALID_URL_FORMAT_MESSAGE);
+  }
+
+  /**
+   * Test building {@link SplunkEventWriter} with the 'services/collector/event' path appended to
+   * the URL.
+   */
+  @Test
+  public void eventWriterFullEndpoint() {
+
+    Exception thrown =
+        assertThrows(
+            IllegalArgumentException.class,
+            () ->
+                SplunkEventWriter.newBuilder()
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -317,6 +318,9 @@ abstract Builder setDisableCertificateValidation(
      */
     public Builder withUrl(ValueProvider<String> url) {
       checkArgument(url != null, "withURL(url) called with null input.");
+      if (url.isAccessible()) {
+        checkArgument(url.get().startsWith("http"), "input url has no protocol.");
```
prathapreddy123, line 322: This check doesn't fully address the isssue.  If input is provides as ```httpbadurl``` then input passes the check but still throws a MalformedURLException.

Instead can you verify the simple pattern like ```"^http(s?):\\/\\/.+$"```  to ensure Url matches either http://domain or https://domain

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -106,6 +106,7 @@ public static Builder newBuilder() {
   public void setup() {
 
     checkArgument(url().isAccessible(), "url is required for writing events.");
+    checkArgument(url().get().startsWith("http"), "url has no protocol.");
```
prathapreddy123, line 109: This check is not required since the format is being validated while building instance

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -328,6 +332,7 @@ public Builder withUrl(ValueProvider<String> url) {
      */
     public Builder withUrl(String url) {
       checkArgument(url != null, "withURL(url) called with null input.");
+      checkArgument(url.startsWith("http"), "input url has no protocol.");
```
prathapreddy123, line 335: Same as earlier comment.  Instead of doing twice move implementation to method say isValidUrlFormat(String url) returning boolean value.  
```
checkArgument(isValidUrlFormat(url),"Invalid url format. Url format should match either http://domain or https://domain");
```

prathapreddy123 (changes requested):

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: The other checks won't be evaluated if the pipeline is run from a template

melbrodrigues:
@rarsan FYI

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Done

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Added method `isValidUrlFormat(String url)` to validate URLs

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -290,6 +295,18 @@ private static void flushWriteFailures(
...
   }
+  
+  /**
+   * Checks whether the HEC URL matches the format 'http://domain' or 'https://domain'.
+   *
+   * @param url for HEC event collector
+   * @return true if the URL is valid
+   */
+  private static boolean isValidUrlFormat(String url) {
+    Pattern pattern = Pattern.compile("^http(s?)://.+$");
```
rarsan, line 306: Good to have this new function `isValidUrlFormat`.
However, incorrect URLs like https://example or https://1.2.3 will be marked as valid here.

I recommend restricting the regex to ensure correct URL, including either a FQDN or IP address.

rarsan (commented):
Looks good. One requested change for URL validation implementation, along with additional unit tests (e.g. http://example, http://1.2.3, etc.).

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
sabhyankar: Why not use a validator library instead of attempting to rewrite the regex and parsing logic? Something like Apache Commons Validator? There are a number of permutations in verifying the validity of a URL and the validator library covers most of those.

melbrodrigues: Restricted the regex to allow URLs that match the format [PROTOCOL]://[HOST]:[PORT]. This would also take care of the [issue](https://stackoverflow.com/questions/64866113/pub-sub-to-splunk-dataflow-template-the-requested-url-was-not-found-on-this-se) where `services/collector/event` is appended to the URL.

Also added the `InetAddresses` and `InternetDomainName` utilities from Guava to validate the host

src/test/java/com/google/cloud/teleport/splunk/SplunkEventWriterTest.java
```diff
@@ -78,17 +78,68 @@ public void eventWriterMissingURL() {
...
+  public void eventWriterMissingURLProtocol() {
+    
+    Exception thrown =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> SplunkEventWriter.newBuilder().withUrl("test-url").build());
 
+    assertThat(thrown)
+        .hasMessageThat()
+        .contains("Invalid url format. Url format should match [PROTOCOL]://[HOST]:[PORT]");
```
prathapreddy123, line 95: Can refer to Constant to ensure that message is aligned as expected

prathapreddy123 (changes requested):
Thanks for changes.  Included few minor comments.   Post these changes rebase to latest master and run `mvn test` to ensure tests and style checks are successful

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -317,6 +340,11 @@ abstract Builder setDisableCertificateValidation(
      */
     public Builder withUrl(ValueProvider<String> url) {
       checkArgument(url != null, "withURL(url) called with null input.");
+      if (url.isAccessible()) {
+        checkArgument(
+            isValidUrlFormat(url.get()),
+            "Invalid url format. Url format should match [PROTOCOL]://[HOST]:[PORT]");
```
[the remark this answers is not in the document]
melbrodrigues: Removed [] brackets and defined an `INVALID_URL_FORMAT_MESSAGE` constant with the additional error details.

melbrodrigues:
Thanks for the feedback @prathapreddy123 . I rebased against master and ran `mvn verify`. All tests and style checks passed successfully.

prathapreddy123:
> Thanks for the feedback @prathapreddy123 . I rebased against master and ran `mvn verify`. All tests and style checks passed successfully.

Thanks @melbrodrigues

prathapreddy123 (approved):

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
rarsan: The port is optional. I think this will invalidate any URL without a port.
Also a path in the URL is completely valid: it's possible there's load balancer at the destination performing some path-based routing before it goes to HEC.

prathapreddy123: Template already internally appends path fragment(services/collector/event) to user provided URL. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/788f6b7af095d3455d8c8f5da18c320baeeaab71/src/main/java/com/google/cloud/teleport/splunk/HttpEventPublisher.java#L75 

It will bit tricky to handle all scenarios. For e.g: even using apache commons urlvalidator is also not supporting all our required scenarios (e.g: not failing when path fragment is included)

```
 boolean isValidUrlFormat(String url) {
        String[] schemes = {"http", "https"};
        UrlValidator urlValidator =   new UrlValidator(schemes,  UrlValidator.ALLOW_LOCAL_URLS + UrlValidator.NO_FRAGMENTS);
        return urlValidator.isValid(url);
    }

 for(String url : new String[]{"http://testurl"
                ,"https://example.splunk.com:1234"
                ,"https://10.209.127.123:8088",
                "https://example.splunk.com:1234/services/collector/event"}) {
            System.out.printf("Url: %s - isValid: %s\n", url, isValidUrlFormat(url));
        }
```

We would like the users to stick to a specific format as per the documentation https://cloud.google.com/dataflow/docs/guides/templates/provided-streaming#cloudpubsubtosplunk.  
https://docs.splunk.com/Documentation/Splunk/8.1.2/Data/UsetheHTTPEventCollector (<protocol>://<host>:<port>)

If we need to make port optional then the regex can be updated to ```"^http(s?)://([^\\:]+)(:[0-9]+)?$"``` with enough testing

melbrodrigues: Updated the regex to make the port optional. Also added a check to invalidate the URL if the host has `services/collector/event` appended to it.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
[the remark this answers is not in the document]
melbrodrigues: Added angular brackets around PORT since it would be optional now. So the error message looks like `Invalid url format. Url format should match PROTOCOL://HOST[:PORT], where PORT is optional. Supported Protocols are http and https. eg: http://hostname:8088`

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -291,6 +303,23 @@ private static void flushWriteFailures(
...
+   * @param url for HEC event collector
+   * @return true if the URL is valid
+   */
+  private static boolean isValidUrlFormat(String url) {
+    Pattern pattern = Pattern.compile("^http(s?)://([^:]+)(:[0-9]+)?$");
+    Matcher matcher = pattern.matcher(url);
+    if (matcher.find()) {
+      String host = matcher.group(2);
+      return (InetAddresses.isInetAddress(host) || InternetDomainName.isValid(host))
+          && !host.endsWith(HttpEventPublisher.HEC_URL_PATH);
```
prathapreddy123, line 318: This check is redundant.  Regex allows 2 types of urls: 
a) urls with port number (i.e protocol://hostname:port)
b) urls  without port (i.e protocol://hostname).  

It automatically returns false for any url that has path fragment  (ie protocol://hostname[:port]/) where trailing slash & anything beyond that is considered as path fragment. For e.g in url  ```https://nonporttrailingslash.splunk.com/ ``` the host will be nonporttrailingslash.splunk.com/  and InetAddresses.isInetAddress(host) returns false due to trailing slash.  Hence  ```https://nonport.splunk.com/services/collector/event``` will return false even without host.endsWith(HttpEventPublisher.HEC_URL_PATH) explicit check.

prathapreddy123 (changes requested):
Thanks.  Need a minor change and good to move forward

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Oh, I missed that! Thanks for the detailed explanation :) Fixed it now

prathapreddy123 (approved):
Thanks. LGTM

prathapreddy123:
@rarsan , @sabhyankar - Can you please verify and let me know if you have any concerns. If not I can go ahead with merge process

rarsan:
LGTM. Thx everyone.
Has this been integration tested with a Splunk HEC instance?

prathapreddy123:
> LGTM. Thx everyone.
> Has this been integration tested with a Splunk HEC instance?

Yes will be verified through Integration test

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -106,6 +117,7 @@ public static Builder newBuilder() {
   public void setup() {
 
     checkArgument(url().isAccessible(), "url is required for writing events.");
+    checkArgument(isValidUrlFormat(url().get()), INVALID_URL_FORMAT_MESSAGE);
```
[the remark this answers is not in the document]
sabhyankar: ReadMessages is a composite transform where messages are read via the built-in PubsubIO [1]. So the messages will still be read by PubsubIO and would not necessarily "fail-fast". I.e the pipeline will be running and this would fail at runtime as we try to publish messages to Splunk.


[1] https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/604b79630b5dbfa61b10e2e0da7dfff7fb35a96f/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java#L326-L328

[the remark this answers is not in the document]
melbrodrigues: Got it. Thanks a ton :)

PR #234, opened 2021-03-25
Log UDF errors

melbrodrigues:
-  Add optional UDF error logging support for `FailsafeJavascriptUdf`
-  Enable UDF error logging for Pub/Sub to Splunk template

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
@@ -52,10 +52,14 @@ import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.sdk.values.TupleTag;
 import org.apache.beam.sdk.values.TupleTagList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** A Text UDF Transform Function. Note that this class's implementation is not threadsafe */
 @AutoValue
 public abstract class JavascriptTextTransformer {
+  
+  private static final Logger LOG = LoggerFactory.getLogger(JavascriptTextTransformer.class);
 
   /** Necessary CLI options for running UDF function. */
   public interface JavascriptTextTransformerOptions extends PipelineOptions {
@@ -154,7 +158,7 @@ public abstract class JavascriptTextTransformer {
     public String invoke(String data) throws ScriptException, IOException, NoSuchMethodException {
       Invocable invocable = getInvocable();
       if (invocable == null) {
-        throw new RuntimeException("No udf was loaded");
+        throw new RuntimeException("No UDF was loaded");
       }
 
       Object result = getInvocable().invokeFunction(functionName(), data);
@@ -270,6 +274,8 @@ public abstract class JavascriptTextTransformer {
     public abstract @Nullable ValueProvider<String> fileSystemPath();
 
     public abstract @Nullable ValueProvider<String> functionName();
+    
+    public abstract @Nullable ValueProvider<Boolean> loggingEnabled();
 
     public abstract TupleTag<FailsafeElement<T, String>> successTag();
 
@@ -285,6 +291,8 @@ public abstract class JavascriptTextTransformer {
       public abstract Builder<T> setFileSystemPath(@Nullable ValueProvider<String> fileSystemPath);
 
       public abstract Builder<T> setFunctionName(@Nullable ValueProvider<String> functionName);
+      
+      public abstract Builder<T> setLoggingEnabled(@Nullable ValueProvider<Boolean> loggingEnabled);
 
       public abstract Builder<T> setSuccessTag(TupleTag<FailsafeElement<T, String>> successTag);
 
@@ -300,12 +308,17 @@ public abstract class JavascriptTextTransformer {
           ParDo.of(
                   new DoFn<FailsafeElement<T, String>, FailsafeElement<T, String>>() {
                     private JavascriptRuntime javascriptRuntime;
+                    private boolean loggingEnabled;
 
                     @Setup
                     public void setup() {
                       if (fileSystemPath() != null && functionName() != null) {
                         javascriptRuntime =
                             getJavascriptRuntime(fileSystemPath().get(), functionName().get());
+                      }
+
+                      if (loggingEnabled() != null && loggingEnabled().isAccessible()) {
+                        loggingEnabled = loggingEnabled().get();
                       }
                     }
 
@@ -323,6 +336,21 @@ public abstract class JavascriptTextTransformer {
                           context.output(
                               FailsafeElement.of(element.getOriginalPayload(), payloadStr));
                         }
+                        
+                      } catch (ScriptException | IOException | NoSuchMethodException e) {
+                        if (loggingEnabled) {
+                          LOG.warn(
+                              "Exception occurred while applying UDF '{}' from file path '{}' due to '{}'",
+                              functionName().get(),
+                              fileSystemPath().get(),
+                              e.getMessage());
+                        }
+                        context.output(
+                            failureTag(),
+                            FailsafeElement.of(element)
+                                .setErrorMessage(e.getMessage())
+                                .setStacktrace(Throwables.getStackTraceAsString(e)));
+                        
                       } catch (Exception e) {
                         context.output(
                             failureTag(),
--- a/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java
@@ -227,6 +227,7 @@ public class PubSubToSplunk {
                 FailsafeJavascriptUdf.<String>newBuilder()
                     .setFileSystemPath(options.getJavascriptTextTransformGcsPath())
                     .setFunctionName(options.getJavascriptTextTransformFunctionName())
+                    .setLoggingEnabled(ValueProvider.StaticValueProvider.of(true))
                     .setSuccessTag(UDF_OUT)
                     .setFailureTag(UDF_DEADLETTER_OUT)
                     .build());


melbrodrigues:
@rarsan Can you review this please?

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -154,7 +158,8 @@ private static Invocable newInvocable(Collection<String> scripts) throws ScriptE
     public String invoke(String data) throws ScriptException, IOException, NoSuchMethodException {
       Invocable invocable = getInvocable();
       if (invocable == null) {
-        throw new RuntimeException("No udf was loaded");
+        LOG.error("No UDF was loaded");
```
rarsan, line 161: The log statement you added in the catch block below should be sufficient, no?

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -166,6 +171,7 @@ public String invoke(String data) throws ScriptException, IOException, NoSuchMet
 
       } else {
         String className = result.getClass().getName();
+        LOG.error("UDF Function did not return a String. Instead got: " + className);
```
rarsan, line 174: Same as above. New log statement in the catch block should cover this.

rarsan (changes requested):
Thanks you for improving the logging!
Might be able to reduce changes to just L333 in the catch block.

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
melbrodrigues: Added it to log errors for other pipelines that use the `TransformTextViaJavascript` PTransform, which does not have a catch block. ([#L257](https://github.com/melbrodrigues/DataflowTemplates/blob/log-udf-errors/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java#L257))

Pub/Sub to Splunk uses the `FailsafeJavascriptUdf` PTransform for which the catch block would be sufficient.

prathapreddy123:
All exceptions gets redirected to error output (i.e failureTag) containing original payload along with error message. 
Logging per record is an anti pattern and is discouraged.

melbrodrigues:
@prathapreddy123 Does this mean that logging is not required for all UDF errors or that logging is not required only for the `FailsafeJavascriptUdf` PTransform since it outputs failures to a deadletter? The original issue in buganizer states that all UDF errors for Pub/Sub to Splunk (which uses `FailsafeJavascriptUdf`) should be logged. Small excerpt from buganizer -  

> UDF errors break entire pipeline, yet don't surface in job details/logs page.

>Logging these would accelerate process for end-user to identify & fix user errors when using UDF functions, such as:
>UDF function not returning a string
>UDF function has a JavaScript syntax error

\
Also, a few other templates like _Pub/Sub to Datastore_ and _Datastore to Text_ use the `TransformTextViaJavascript` PTransform which does not output failures to a deadletter.

prathapreddy123:
> @prathapreddy123 Does this mean that logging is not required for all UDF errors or that logging is not required only for the `FailsafeJavascriptUdf` PTransform since it outputs failures to a deadletter? The original issue in buganizer states that all UDF errors for Pub/Sub to Splunk (which uses `FailsafeJavascriptUdf`) should be logged. Small excerpt from buganizer -
> 
> > UDF errors break entire pipeline, yet don't surface in job details/logs page.
> 
> > Logging these would accelerate process for end-user to identify & fix user errors when using UDF functions, such as:
> > UDF function not returning a string
> > UDF function has a JavaScript syntax error
> 
> 
> Also, a few other templates like _Pub/Sub to Datastore_ and _Datastore to Text_ use the `TransformTextViaJavascript` PTransform which does not output failures to a deadletter.

TransformTextViaJavascript propogates up all the exceptions to the caller which in turns shows up in worker logs at pipeline level.   But In case of FailsafeJavascriptUdf any errors that occur while processing UDF are redirected through a separate PCollection back to caller which normally writes to Deadletter Channel.  Whenever Pipeline is capable of writing errors to deadletter channel, Pipeline owners needs to monitor deadletter channel as well to verify if there are any failures.

melbrodrigues:
I think the intention was to log errors in addition to pushing them to a deadletter. @rarsan your thoughts on this?

rarsan:
Yes. We need to log these errors in worker logs, regardless of whether messages are dropped in DL topic or not.

prathapreddy123:
Since [JavascriptTextTransformer](src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java) is common module across templates the changes will have impact across the board. If SplunkIO prefers logging errors in addition to deadletter queue, you can log the errors inside the template by adding additional Sink.

Be cautious about additional costs and throttling issues with respect to cloud logging while logging per element.   Monitoring log error count will help to act immediately during such cases.

mhite:
Take this question for what it is worth (which may not be much given my lack of experience programming Dataflow), but is it possible to implement a rate-limited logger? IE. Emit a message like "log rate exceeded; throttling."

melbrodrigues:
Dataflow logs a message if the throttling limit is reached - https://cloud.google.com/dataflow/docs/guides/logging#LogLimits

melbrodrigues:
@rarsan @prathapreddy123 Added setLogUdfErrors() to `FailsafeJavascriptUdf`. Defaults to false, so this should not affect other templates

prathapreddy123 (changes requested):

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -324,6 +337,9 @@ public void processElement(ProcessContext context) {
                               FailsafeElement.of(element.getOriginalPayload(), payloadStr));
                         }
                       } catch (Exception e) {
+                        if (logUdfErrors) {
+                          LOG.warn("Could not apply UDF. Reason: {}", e.getMessage());
```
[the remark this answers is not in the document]
melbrodrigues: Updated the error message

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -270,6 +274,8 @@ public void processElement(ProcessContext c)
     public abstract @Nullable ValueProvider<String> fileSystemPath();
 
     public abstract @Nullable ValueProvider<String> functionName();
+    
+    public abstract @Nullable ValueProvider<Boolean> logUdfErrors();
```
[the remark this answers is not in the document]
melbrodrigues: Changed to **loggingEnabled**

src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java
```diff
@@ -337,8 +337,12 @@ public void processElement(ProcessContext context) {
                               FailsafeElement.of(element.getOriginalPayload(), payloadStr));
                         }
                       } catch (Exception e) {
-                        if (logUdfErrors) {
-                          LOG.warn("Could not apply UDF. Reason: {}", e.getMessage());
+                        if (loggingEnabled) {
```
prathapreddy123, line 340: You still need to have a check on javascriptRuntime as mentioned in earlier comment since FailsafeJavascriptUdf transform executes as an identity transform even during the cases when JSUDF name and FilePath are null.  In such instances if any exception occurs, referring functionName().get throws NPE without proper guard.

melbrodrigues: The `javascriptRuntime` remains null if the `fileSystemPath` and `functionName` are null [1]. And there is a check on the  `javascriptRuntime` before the UDF is invoked [2], so it should not enter the catch block. In that case it will simply output the payload string as is without any transformations.

Do we still need the null check or am I missing something here?

[1] https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/d114a4293596ddec0a87e6aead5d0c581c4a1731/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java#L315-L318

[2] https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/d114a4293596ddec0a87e6aead5d0c581c4a1731/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java#L331-L333

prathapreddy123: Good callout.  Since the below code always execute independent of JS function any exception occurred while executing the code will still be caught by the same catch block.  In such cases if JS function is not available then catch block log messages inside the catch block throws NPE while referring functionName/fileSystemPath.  To avoid such cases safeguard  (i.e javascriptRuntime != null) is included in the condition.  
```
  if (!Strings.isNullOrEmpty(payloadStr)) {                          
        context.output( FailsafeElement.of(element.getOriginalPayload(), payloadStr));                        
   }
```

melbrodrigues: I'm a bit confused as to why that block might throw an exception... 

`Strings.isNullOrEmpty()` and `FailsafeElement.of()` are static method calls, and `element` will never be null so `element.getOriginalPayload()` will not result in an NPE (if it does then `element.getPayload()` at [L328](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/d114a4293596ddec0a87e6aead5d0c581c4a1731/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java#L328) should also be in the try block).

Can you point out the possible points of failure in the block of code you've mentioned?

prathapreddy123: Runtime exceptions can occur anywhere including the code inside library methods (e.g:  context.output ) due to either system faults, interaction with other services such as runner or external I/O etc. All these exceptions lands in closest catch all block.  In such cases if JS related variables are null will lead to another exception inside catch block. With additional code inside try block as part of defensive programming adding check would be helpful.

I guess the cleanest & preferred way is to explicitly declare a checked exception for JS UDF exceptions (either when given file/function is missing or in case of error while executing JS Code) and log inside that catch block leaving the default catch block for any other runtime exceptions.

prathapreddy123 (approved):
LGTM

prathapreddy123:
@melbrodrigues  -  can you rebase with master

melbrodrigues:
@prathapreddy123 Rebased and squashed all commits

PR #235, opened 2021-04-01
Support inputSubscripton for pubsub-to-text/avro

n-oden:
Allow using a pre-existing PubSub subscription for the
Cloud_PubSub_to_GCS_Text and Cloud_PubSub_to_GCS_Avro templates

Signed-off-by: Nathan J. Mehl <n@oden.io>


prathapreddy123:
In Classic templates , for the pipeline options to be available during run time the options should be exposed as ValueProviders and can be accessed only inside PTransforms / DoFn's.  if else condition construct is available only during template build time.  

For e.g:  In [PubSubToBigQuery](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/master/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java) despite of using if else approach  the code is compiled as two different templates (i.e Pub/Sub Topic to BQ, Pub/Sub Subscription to BQ).

However this is no longer the case with Flex templates (i.e all pipelines under v2).

PR #230, opened 2021-03-16
Fix incompatible PubsubMessage for Pub/Sub to Splunk template

melbrodrigues:
- Publish Pub/Sub message body as a JSON object instead of a stringified JSON
- Nest Pub/Sub message body under `data` field instead of `message` field

PR #241, opened 2021-04-16
Fix parsing timestamp nested under Pub/Sub message 'data' field

melbrodrigues:
This PR modifies the Pub/Sub to Splunk template to add support for parsing `timestamp` field nested under the `data` field for Pub/Sub messages, when `includePubsubMessage=true`.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
@@ -60,6 +60,8 @@ public class SplunkConverters {
   private static final String HEC_SOURCE_TYPE_KEY = "sourcetype";
 
   private static final String TIMESTAMP_KEY = "timestamp";
+  
+  private static final String DATA = "data";
 
   /**
    * Returns a {@link FailsafeStringToSplunkEvent} {@link PTransform} that consumes {@link
@@ -183,6 +185,9 @@ public class SplunkConverters {
                           String parsedTimestamp;
                           if (metadataAvailable) {
                             parsedTimestamp = metadata.optString(HEC_TIME_KEY);
+                          } else if (json.optJSONObject(DATA) != null
+                              && json.getJSONObject(DATA).has(TIMESTAMP_KEY)) {
+                            parsedTimestamp = json.getJSONObject(DATA).getString(TIMESTAMP_KEY);
                           } else {
                             parsedTimestamp = json.optString(TIMESTAMP_KEY);
                           }
--- a/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
@@ -61,7 +61,8 @@ public class SplunkConverters {
 
   private static final String TIMESTAMP_KEY = "timestamp";
   
-  private static final String DATA = "data";
+  protected static final String PUBSUB_MESSAGE_ATTRIBUTE_FIELD = "attributes";
+  protected static final String PUBSUB_MESSAGE_DATA_FIELD = "data";
 
   /**
    * Returns a {@link FailsafeStringToSplunkEvent} {@link PTransform} that consumes {@link
@@ -146,7 +147,7 @@ public class SplunkConverters {
       this.splunkEventOutputTag = splunkEventOutputTag;
       this.splunkDeadletterTag = splunkDeadletterTag;
     }
-
+  
     @Override
     public PCollectionTuple expand(PCollection<FailsafeElement<String, String>> input) {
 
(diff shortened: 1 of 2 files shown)


melbrodrigues:
Marking this as a draft PR till PR #230 is merged, since this expects the Pub/Sub payload to be nested under `data` instead of `message`

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
```diff
@@ -60,6 +60,8 @@
   private static final String HEC_SOURCE_TYPE_KEY = "sourcetype";
 
   private static final String TIMESTAMP_KEY = "timestamp";
+  
+  private static final String DATA = "data";
```
rarsan, line 64: Possible to re-use same constant `PUBSUB_MESSAGE_DATA_FIELD`?
https://github.com/melbrodrigues/DataflowTemplates/blob/incompatible-ps-msg/src/main/java/com/google/cloud/teleport/templates/PubSubToSplunk.java#L160

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
```diff
@@ -183,6 +185,9 @@ public void processElement(ProcessContext context) {
                           String parsedTimestamp;
                           if (metadataAvailable) {
                             parsedTimestamp = metadata.optString(HEC_TIME_KEY);
+                          } else if (json.optJSONObject(DATA) != null
+                              && json.getJSONObject(DATA).has(TIMESTAMP_KEY)) {
+                            parsedTimestamp = json.getJSONObject(DATA).getString(TIMESTAMP_KEY);
```
rarsan, line 190: I think we want to make this condition restricted to `includePubsubMessage=true` since this fix is specific to that use case. Otherwise, this can lead to unexpected behavior where we're applying this logic to arbitrary payload that happens to have data.timestamp nested field (that is without PubsubMessage wrapper i.e. `includePubsubMessage=false`).

rarsan (changes requested):
Thx for the PR bugfix @melbrodrigues . Some minor changes requested.
Keep in mind this automatic timestamp parsing is special treatment only for Cloud Logging ([LogEntry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry)) incoming payload and its `timestamp` field in RFC3339 format, so we should only look for top-level `timestamp` field in PubSub message. Given that common use case of exporting cloud logs, this is meant for convenience to avoid requiring a UDF just to set _metadata.timestamp to Splunk HEC.

rarsan:
cc @mhite @pauld-splunk

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
mhite: That sounds like a good idea to me, too, @rarsan.

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: yes, this is just a placeholder till the other PR is merged

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: This can be done but we'd be tying up `SplunkConverters` tightly with PubSub. 

The class is placed in a [common directory](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/master/src/main/java/com/google/cloud/teleport/templates/common) so other templates can make use of it in the future. Might be best to leave that parameter out if later on we decide to add a new template that has a different source with Splunk as the sink.

mhite: @melbrodrigues - I see what you mean. Any ideas on loosening that coupling while keeping with the general spirit of the request?

melbrodrigues: Discussed this with @rarsan yesterday and we feel it's better to add an additional check to see if the payload contains the `logName` field nested under `data`. (`logName` field is present for all Cloud Log events)

This should reduce chances of the template parsing the timestamp for arbitrary payloads.

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: Renamed to `PUBSUB_MESSAGE_DATA_FIELD` per our last discusstion.

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
melbrodrigues: As discussed earlier, we won't be checking if `data.logName` field exists before parsing the timestamp to keep the behaviour consistent with `includePubsubMessage=false`.

Additionally, I've added a method `isPubsubMessage()` that checks for the `data` and `attributes` fields before parsing the timestamp.
```java
private boolean isPubsubMessage(JSONObject json) {
  return json.has(PUBSUB_MESSAGE_DATA_FIELD) && json.has(PUBSUB_MESSAGE_ATTRIBUTE_FIELD);
}
```

mhite:
Thanks, I tested `includePubsubMessage=false` and `includePubsubMessage=true` with this PR after you rebased and force-pushed this again. Looks good as far as I can tell. TImestamps match in both cases.

prathapreddy123 (approved):
LGTM.  Can be merged once https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/242 is fixed by early next week

src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
rarsan: Nice. LGTM.

PR #250, opened 2021-05-07
Update README.md

omarismail94:
I tried launching a job without specifying this field from the UI, and it failed. Looking at the logs, I saw this error:
java.lang.IllegalArgumentException: Missing required value --topic for PUBSUB sink type


omarismail94:
Ignore this.

PR #253, opened 2021-05-16
Fix JdbcToBigQuery template to handle DATETIME, TIMESTAMP, DATE datatypes

ayush-poddar:
Current JdbcToBigQuery template fails if any of the columns in SQL database is of type `DATETIME`, `TIMESTAMP`, `DATE`. 

The issue I believe is due to:
- BigQuery not supporting loads in milliseconds for `DATETIME` and `TIMESTAMP` but rather supports UNIX TIME which is in seconds for loads from GCS [2].
- `DATE` fails due to JDBC drivers give result in milliseconds but BigQuery uses canonical format for DATE [1]

Current workaround used is converting all these data types to their canonical format [1]:

- TIMESTAMP: `yyyy-MM-dd hh:mm:ss.SSSSSSXXX`
- DATETIME:    `yyyy-MM-dd hh:mm:ss.SSSSSS`
- TIME:             `yyyy-MM-dd`

[1] https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types
[2] https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#details_of_loading_json_data

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
@@ -19,6 +19,10 @@ package com.google.cloud.teleport.templates.common;
 import com.google.api.services.bigquery.model.TableRow;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
+import java.text.SimpleDateFormat;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.TemporalAccessor;
+
 import org.apache.beam.sdk.io.jdbc.JdbcIO;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
@@ -99,6 +103,11 @@ public class JdbcConverters {
    */
   private static class ResultSetToTableRow implements JdbcIO.RowMapper<TableRow> {
 
+    static SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd");
+    static DateTimeFormatter datetimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd hh:mm:ss.SSSSSS");
+    static SimpleDateFormat timestampFormatter = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss.SSSSSSXXX");
+
+
     @Override
     public TableRow mapRow(ResultSet resultSet) throws Exception {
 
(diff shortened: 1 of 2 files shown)


ayush-poddar:
+R @prathapreddy123

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -99,6 +103,11 @@
    */
   private static class ResultSetToTableRow implements JdbcIO.RowMapper<TableRow> {
 
+    static SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd"); // date format
```
prathapreddy123, line 106: Better variable names would be dateFormatter, datetimeFormatter, timestampFormatter

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -107,7 +116,31 @@ public TableRow mapRow(ResultSet resultSet) throws Exception {
...
+         * TIMESTAMP: EPOCH MILLISECONDS -> yyyy-MM-dd hh:mm:ss.SSSSSSXXX
+         *
+         * MySQL drivers have ColumnTypeName in all caps and postgres in small case
+         */
+        switch (metaData.getColumnTypeName(i).toLowerCase()) {
+          case "date":
+            outputTableRow.set(metaData.getColumnName(i), df.format(resultSet.getObject(i)));
+            break;
+          case "datetime":
+            outputTableRow.set(metaData.getColumnName(i), dtf.format(Timestamp.valueOf((LocalDateTime)resultSet.getObject(i))));
```
prathapreddy123, line 136: Is there any specific reason why converting to Timestamp.valueOf which also requires LocalDateTime conversion.

prathapreddy123 (commented):
Thanks for contribution.  Include unit tests to validate all possible combinations (i.e value being null, columns with different date related types and a non date type)

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
ayush-poddar: done

src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
ayush-poddar: done

ayush-poddar:
@prathapreddy123 I have added the test cases, please let me know if there are any other cases to be tested for?

prathapreddy123 (commented):
LGTM. Can you rebase current branch with master.

ayush-poddar:
@prathapreddy123 done, can you verify

PR #254, opened 2021-05-19
Refactor maven modules

Cherepushko:
Created Elasticsearch Maven module that combine old PubSub-To-Elasticsearch, BigQuery-To-Elasticsearch and Csv-To-Elasticsearch modules. Added main class to run application specified in options. Added options to pass Elastic endpoint credentials. Moved some inner classes into different .java files. Created initial hierarchy for options interfaces.
Tested on Dataflow (case PubSub to Elasticsearch). Other submodules wasn't tested.


Cherepushko:
@googlebot I signed it!

PR #259, opened 2021-06-01
add PubSubToPubSub idattribute options to remove message duplication

case-k-git:
use idattribute method to remove message duplication. Dataflow  remove message duplication automatically based on created pub/sub message id at least once policy. but message gonna duplicate if publisher send same message several time.
by using idattribute. we can remove message duplication based on idattributekey(message id ) set by publisher

https://beam.apache.org/releases/javadoc/2.27.0/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.Read.html


bvolpato:
Hi @case-k-git -- thanks for your contribution here. Unfortunately we ran behind on reviews, and the structure of the project changed a bit. Is there still interest in pursuing this template? We can work together on a plan to resolve conflicts + review the code.

PR #260, opened 2021-06-01
add PubSubToBigQuery dynamic destinations tables

case-k-git:
add DynamicDestinations options to determine table based on pub/sub message value . using this option reduce dataflow instance cost .

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java
+++ b/src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java
@@ -19,6 +19,8 @@ package com.google.cloud.teleport.templates;
 import static com.google.cloud.teleport.templates.TextToBigQueryStreaming.wrapBigQueryInsertError;
 
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.api.services.bigquery.model.TableSchema;
+import com.google.api.services.bigquery.model.TableReference;
 import com.google.cloud.teleport.coders.FailsafeElementCoder;
 import com.google.cloud.teleport.templates.common.BigQueryConverters.FailsafeJsonToTableRow;
 import com.google.cloud.teleport.templates.common.ErrorConverters;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/teleport/templates/PubSubToBigQuery.java
```diff
@@ -191,6 +196,43 @@
     ValueProvider<String> getOutputDeadletterTable();
 
     void setOutputDeadletterTable(ValueProvider<String> value);
+
+    @Description(
+        "This determines whether the template use dynamic destinationsto to write BigQuery table based on message value")
+    @Default.Boolean(false)
+    Boolean getDynamicDestinations();
```
an2x, line 203: I'm afraid this won't work.
With [classic templates](https://cloud.google.com/dataflow/docs/guides/templates/creating-templates), you can't change the pipeline graph in runtime (after the template creation) based on the parameters supplied via `ValueProvider`. As a "workaround" you can declare this parameter as a regular option (not a `ValueProvider`) but then this parameter must be set at the **template creation time**.

The problem is that this template already has another parameter `useSubscription`, which is set at creation time too (and hence we generate 2 separate templates: "Pub/Sub Subscription to BigQuery" and "Pub/Sub Topic to BigQuery").
Now, with this new parameter, we'll have to generate 4 template (for all possible combinations of `useSubscription` and `dynamicDestinations`), which is not sustainable and already too much IMO.

One option I see is re-implementing the whole template as a [Flex template](https://cloud.google.com/dataflow/docs/guides/templates/using-flex-templates), which allows changing the pipeline graph after the template creation.
There is no easy way to replace an existing classic template with a Flex template as there may be users depending on the old template, but I guess we can deprecate the classic template and slowly move everyone to Flex if it gets implemented.

Precha24:
Good, thanks

Precha24 (approved):

damccorm:
Looks like this PR is stale and needs significant revisions so I'm going to close it; if you're interested in taking this forwards, feel free to reopen.

PR #267, opened 2021-07-02
Improve datastream to bigquery

diegolnasc:
Currently, the template allows the output of the Datastream to another project, however, there are some bugs along the way that prevent it from working.

Minor: added a parameter to ignore fields, fix typo and improved pipeline.

```
Also, I believe it is ideal to make explicit/recommend a machine with a considerable number of cpu and memory. Personally, I would indicate n1-highcpu-8 or more.

In some tests, I faced several problems (out of memory) with the machine (n1-highcpu-4) that was in the readme.
```


diegolnasc:
@googlebot I signed it!

diegolnasc:
@dhercher

zhoufek:
Hi, sorry for the delay in getting this reviewed. I'm going to try to do so today.

v2/datastream-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToBigQuery.java
```diff
@@ -31,8 +31,9 @@
 import com.google.cloud.teleport.v2.transforms.UDFTextTransformer.InputUDFToTableRow;
 import com.google.cloud.teleport.v2.values.FailsafeElement;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
+
```
zhoufek, line 36: Please remove this blank line. It is causing a checkstyle violation.

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/mappers/MergeInfoMapper.java
```diff
@@ -68,8 +69,10 @@
 
   public MergeInfoMapper(
       DataStreamClient dataStreamClient,
+      String projectId,
```
zhoufek, line 72: Could you please reformat the parameters so that only one appears per line? (I know it isn't your fault, but the inconsistency is becoming more apparent, and the one-per-line is better aligned with how we'll enforce formatting in the future.)

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/mappers/MergeInfoMapper.java
```diff
@@ -152,12 +155,14 @@ public DataStreamPkCache getPkCache() {
                         allPkFields,
                         allSortFields,
                         METADATA_DELETED,
-                        String.format("%s.%s", // Staging Table
+                        String.format("%s.%s.%s", // Staging Table
+                                projectId,
```
zhoufek, line 159: Please make sure that the arguments are aligned with each other. `projectId` in both cases is further indented than the other arguments.

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/merge/DataStreamBigQueryMerger.java
```diff
@@ -51,9 +52,11 @@
 
   public DataStreamBigQueryMerger(
       GcpOptions options,
+      String projectId,
```
zhoufek, line 55: Similar to the other constructor, could you please make sure there's only one parameter per line for consistency?

v2/datastream-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToBigQuery.java
```diff
@@ -422,6 +419,10 @@ public void process(
     return pipeline.run();
   }
 
+  private static Set<String> getFieldsToIgnore(String fields) {
+    return Arrays.stream(fields.trim().split("\\s*,\\s*")).collect(Collectors.toSet());
```
zhoufek, line 423: Could you switch this to use [Guava Splitter](https://guava.dev/releases/23.0/api/docs/com/google/common/base/Splitter.html#on-java.util.regex.Pattern-)? I'm pretty sure you can use it without any changes to the POM.

zhoufek (changes requested):
Overall, looks good. Sorry for being such a stickler for formatting, but it's something we're trying to get under control. Unfortunately, we don't have any formatting plugins set up (yet!).

diegolnasc:
> Overall, looks good. Sorry for being such a stickler for formatting, but it's something we're trying to get under control. Unfortunately, we don't have any formatting plugins set up (yet!).

Thanks for the review.
All suggestions make sense and were made as requested.

zhoufek (approved):

zhoufek:
LGTM - Just need to make sure it doesn't break any of our internal tests before I merge.

v2/datastream-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToBigQuery.java
```diff
@@ -30,9 +30,10 @@
 import com.google.cloud.teleport.v2.transforms.UDFTextTransformer.InputUDFOptions;
 import com.google.cloud.teleport.v2.transforms.UDFTextTransformer.InputUDFToTableRow;
 import com.google.cloud.teleport.v2.values.FailsafeElement;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.regex.Pattern;
+import com.google.common.base.Splitter;
```
zhoufek, line 36: (Sorry, missed this!) Imports need to be ordered alphabetically. Otherwise, it will cause a checkstyle violation. So the `Splitter` import should come between `import com.google.cloud.teleport.v2.values.FailsafeElement;` and `import java.util.HashSet;`.

For future reference, you can run checkstyle with `mvn checkstyle:checkstyle` from the module with the change or from v2/ with `mvn -pl m0,m1,...,mN checkstyle:checkstyle`, so for this PR, you could run 

`mvn -pl common,datastream-to-bigquery checkstyle:checkstyle`

zhoufek (changes requested):

v2/datastream-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToBigQuery.java
diegolnasc: Sure!

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/mappers/DataStreamMapper.java
```diff
@@ -72,7 +72,7 @@ public TableId getTableId(TableRow input) {
         BigQueryConverters.formatStringTemplate(tableNameTemplate, input)
         .replaceAll("\\$", "_");
 
-    return TableId.of(datasetName, tableName);
+    return TableId.of(getProjectId(), datasetName, tableName);
```
zhoufek, line 75: Unfortunately, this line seems to be causing some failures to internal tests:

```
java.lang.NullPointerException
com.google.common.base.Preconditions.checkNotNull(Preconditions.java:877)
com.google.cloud.bigquery.TableId.of(TableId.java:82)
com.google.cloud.teleport.v2.cdc.mappers.DataStreamMapper.getTableId(DataStreamMapper.java:75)
com.google.cloud.teleport.v2.cdc.mappers.DataStreamMapper.getTableId(DataStreamMapper.java:35)
com.google.cloud.teleport.v2.cdc.mappers.BigQueryMapper$1.apply(BigQueryMapper.java:214)
org.apache.beam.sdk.transforms.Contextful.lambda$fn$36334a93$1(Contextful.java:112)
org.apache.beam.sdk.transforms.MapElements$1.processElement(MapElements.java:143)
```

Since `outputProjectId` is not a required parameter, I think the factory method should be chosen based on whether or not `getProjectId()` returns null.

zhoufek (changes requested):

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/mappers/DataStreamMapper.java
diegolnasc: Mmm.
This doesn't feel right. Does DataStreamMapper receive a null parameter in the constructor?
Also, the description of the _outputProject_ is as follows:

`The BigQuery Project ID. Default is the project for the Dataflow job.`

diegolnasc: Can we add the snippet below to _validateOptions_ method?

```
    String outputProjectId = options.getOutputProjectId();

    if (StringUtils.isBlank(outputProjectId)) {
      options.setOutputProjectId(options.as(GcpOptions.class).getProject());
    }
```

That way, I think the journey is more coherent, considering even the current description.

zhoufek: Getting the value from GcpOptions does seem to be a better option, but I'd rather set the project id value in `DataStreamMapper` because:

1. Overwriting pipeline options in production code should be avoided.
2. We won't have to keep duplicating the same logic across templates.

That should be enough to fulfill the documented behavior of defaulting to the current GCP project.

diegolnasc: I agree.

zhoufek (approved):

zhoufek:
Everything looks good, and internal tests are succeeding. Thanks!

PR #274, opened 2021-07-22
Add support for Secret Manager

melbrodrigues:
- Add `SecretManagerValueProvider` to fetch secrets stored in Secret Manager
- Add `--tokenSecretId` parameter to include Secret Manager support for Splunk template
- Add `--tokenSource` parameter to determine Splunk token source


src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java
```diff
@@ -132,6 +132,18 @@ public static FailsafeStringToSplunkEvent failsafeStringToSplunkEvent(
...
     void setTokenKMSEncryptionKey(ValueProvider<String> keyName);
+
+    @Description(
+        "Secret Manager Secret ID for the token. Should be in the format "
+            + "projects/{project}/secrets/{secret}/versions/{secret_version}")
+    ValueProvider<String> getTokenSecretId();
+
+    void setTokenSecretId(ValueProvider<String> secretId);
+    
+    @Description("Source of the token. One of PLAINTEXT, KMS or SECRET_MANAGER")
```
an2x, line 143:
```suggestion
    @Description("Source of the token. One of PLAINTEXT, KMS or SECRET_MANAGER.")
```

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -159,6 +185,66 @@ public void setup() {
...
+        return new KMSEncryptedNestedValueProvider(token(), tokenKmsEncryptionKey());
+
+      case PLAINTEXT:
+        checkNotNull(token().get(), "token is required for writing events");
+
+        LOG.warn("Using plaintext token");
+        return token();
+
+      default:
+        throw new RuntimeException("tokenSource must be one of PLAINTEXT, KMS or SECRET_MANAGER");
```
an2x, line 244:
```suggestion
        throw new RuntimeException("tokenSource must be one of PLAINTEXT, KMS or SECRET_MANAGER, but found: " + tokenSource);
```

src/main/java/com/google/cloud/teleport/splunk/SplunkIO.java
```diff
@@ -157,9 +184,82 @@ public Builder withToken(ValueProvider<String> token) {
...
+       * Same as {@link Builder#withTokenSource(ValueProvider)} but without {@link ValueProvider}.
+       *
+       * @param tokenSource Source of the token
+       * @return {@link Builder}
+       */
+      public Builder withTokenSource(String tokenSource) {
+        checkArgument(tokenSource != null, "withTokenSource(tokenSource) called with null input");
+        
+        if (!EnumUtils.isValidEnum(SplunkTokenSource.class, tokenSource)) {
+          LOG.warn("tokenSource is invalid. Defaulting to {}", UNKNOWN_TOKEN_SOURCE);
```
an2x, line 257:
```suggestion
          LOG.warn("tokenSource {} is invalid. Defaulting to {}", tokenSource, UNKNOWN_TOKEN_SOURCE);
```

src/main/java/com/google/cloud/teleport/splunk/SplunkIO.java
```diff
@@ -157,9 +184,82 @@ public Builder withToken(ValueProvider<String> token) {
...
+       * Source of the token. One of PLAINTEXT, KMS or SECRET_MANAGER.
+       *
+       * @param tokenSource Source of the token
+       * @return {@link Builder}
+       */
+      public Builder withTokenSource(ValueProvider<String> tokenSource) {
+        checkArgument(tokenSource != null, "withTokenSource(tokenSource) called with null input");
+        
+        if (tokenSource.isAccessible() && !EnumUtils.isValidEnum(SplunkTokenSource.class, tokenSource.get())) {
+          LOG.warn("tokenSource is invalid. Defaulting to {}", UNKNOWN_TOKEN_SOURCE);
```
an2x, line 240:
```suggestion
          LOG.warn("tokenSource {} is invalid. Defaulting to {}", tokenSource, UNKNOWN_TOKEN_SOURCE);
```

src/main/java/com/google/cloud/teleport/splunk/SplunkIO.java
```diff
@@ -238,7 +338,16 @@ public Builder withDisableCertificateValidation(Boolean disableCertificateValida
 
       public Write build() {
         checkNotNull(url(), "HEC url is required.");
-        checkNotNull(token(), "Authorization token is required.");
+        
+        if (tokenSecretId() == null && tokenKmsEncryptionKey() == null && token() == null) {
```
an2x, line 342: This validation logic is spread across at least 3 places:
`SplunkIO.Builder#build()`
`SplunkEventWriter.Builder#build()`
`SplunkEventWriter#resolveTokenSource()` and `#getToken()`

And in the first 2 location it's incomplete anyway (e.g. checking that at least 1 param is provided is not enough, depending on the `tokenSource` we may still have another required param missing and fail later).

Can we move this validation logic to one place?

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -159,6 +185,66 @@ public void setup() {
...
   }
 
+  /**
+   * Utility method that attempts to determine the token source based on user provided parameters.
+   *
+   * @return {@link SplunkTokenSource}
+   */
+  private SplunkTokenSource resolveTokenSource() {
+
+    if (tokenSecretId() != null && tokenSecretId().get() != null) {
```
an2x, line 195: I would exclude the `SECRET_MANAGER` branch from `resolveTokenSource`. The idea is that this method should support only the already existing token sources (i.e. PLAINTEXT and KMS), for reverse compatibility.
If someone wants to use the new `SECRET_MANAGER` source, they will **have** to set `tokenSource` as well (to avoid potential misconfiguration).

an2x (changes requested):

src/main/java/com/google/cloud/teleport/splunk/SplunkIO.java
melbrodrigues: The null checks in the build methods were added so that the template would fail fast in case of missing params. The ones in `#resolveTokenSource()` may or may not be executed depending on whether the user passes the `tokenSource`, and the ones in `#getToken()` are final guard rails.

I can remove the checks in the build methods but the checks in `#resolveTokenSource()` and `#getToken()` would be required. Would that be okay?

Or should I create separate null check methods for each of the three cases?

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: Makes sense. I'll take that branch out.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -386,9 +510,82 @@ public Builder withToken(ValueProvider<String> token) {
...
+     * @param tokenSource Source of the token
+     * @return {@link Builder}
+     */
+    public Builder withTokenSource(String tokenSource) {
+      checkArgument(tokenSource != null, "withTokenSource(tokenSource) called with null input");
+  
+      if (!EnumUtils.isValidEnum(SplunkTokenSource.class, tokenSource)) {
+        LOG.warn("tokenSource is invalid. Defaulting to {}", UNKNOWN_TOKEN_SOURCE);
+        return setTokenSource(ValueProvider.StaticValueProvider.of(UNKNOWN_TOKEN_SOURCE));
+      }
```
an2x, line 585: Remove this block? The same validation will be performed in `withTokenSource(ValueProvider<String> tokenSource)` method, and `StaticValueProvider` always returns `isAccessible() == true`.

Same for other `with*` methods.

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
```diff
@@ -386,9 +510,82 @@ public Builder withToken(ValueProvider<String> token) {
...
+    /**
+     * Source of the token. One of PLAINTEXT, KMS or SECRET_MANAGER.
+     *
+     * @param tokenSource Source of the token
+     * @return {@link Builder}
+     */
+    public Builder withTokenSource(ValueProvider<String> tokenSource) {
+      checkArgument(tokenSource != null, "withTokenSource(tokenSource) called with null input");
+  
+      if (tokenSource.isAccessible() && !EnumUtils.isValidEnum(SplunkTokenSource.class, tokenSource.get())) {
```
an2x, line 565: `with*` methods and some validation logic (e.g. checking for `isValidEnum`) are still duplicated in `SplunkIO` and `SplunkEventWriter`.
I'm OK with submitting this as is, but my recommendation would be to move `resolveTokenSource` and `getToken` to `SplunkIO`, resolve the token there, and pass an already resolved token to `SplunkEventWriter`. Then you won't need to duplicate `withTokenSource`, `withTokenSecretId`, `withTokenKmsEncryptionKey` and their validation logic here. This is also better from the "fail fast" perspective.

an2x (changes requested):

src/main/java/com/google/cloud/teleport/splunk/SplunkEventWriter.java
melbrodrigues: We may not be able to build a template for the pipeline if try to resolve it in `SplunkIO`. The code inside `SplunkIO`'s `expand()` method is evaluated while building the template and since the token params won't be present at build time, it'll fail with an exception.

Maybe we should go ahead with a new `TokenNestedValueProvider` class that accepts the token params and outputs the final resolved token like you suggested offline? Might need a little more refactoring from the current approach but I agree it would be cleaner.

src/test/java/com/google/cloud/teleport/util/TokenNestedValueProviderTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    assertEquals(SplunkTokenSource.PLAINTEXT, tokenSource);
+  }
+
+  /**
+   * Test that {@link TokenNestedValueProvider#resolveTokenSource()} fails when a Secret Manager
+   * Secret ID is passed without a token source.
+   */
+  @Test
+  public void testResolveTokenSource_secretManager() {
```
an2x, line 77: Can you please add one more positive test case for SECRET_MANAGER, i.e. when the `tokenSource`and `tokenSecretId` are both provided, it shouldn't fail and should return `SplunkTokenSource.SECRET_MANAGER`.

Otherwise all looks good.

an2x (changes requested):

src/test/java/com/google/cloud/teleport/util/TokenNestedValueProviderTest.java
melbrodrigues: Couldn't add a test case for the exact scenario you mentioned since we decided `resolveTokenSource()` will only resolve a missing source for KMS and PLAINTEXT in [this comment](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/274#discussion_r677814129), but I added another test case that checks whether an instance of `SecretManagerValueProvider` is returned if both `tokenSource` and `tokenSecretId` are provided.

Also added additional similar test cases for the KMS and PLAINTEXT scenarios as well.

an2x (approved):

zhoufek (changes requested):

pom.xml
```diff
@@ -593,6 +594,21 @@
       <artifactId>proto</artifactId>
       <version>${tensorflow.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-secretmanager</artifactId>
```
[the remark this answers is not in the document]
melbrodrigues: Added `libraries-bom` v20.0.0 for secret-manager.

.
Had to explicitly add `proto-google-cloud-secretmanager-v1` to the dependency list because `maven-dependency-plugin` complains about the dependency being used but not declared and causes `mvn clean install` to fail.
```
[WARNING] Used undeclared dependencies found:
[WARNING]    com.google.api.grpc:proto-google-cloud-secretmanager-v1:jar:1.4.2:compile
```
I've removed it from the dependency list now and set `maven-dependency-plugin` to ignore the warning.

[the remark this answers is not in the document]
zhoufek: Could you please follow the convention of declaring the versions in the properties list? So something like:

```
<cloud-libraries-bom.version>20.0.0</cloud-libraries-bom.version>
...
      <version>${cloud-libraries-bom.version}</version>
```

If nothing else, it's more consistent with how we manage versions for other dependencies.

Also, could you remove the comment about this being for secret manager? While that is the purpose for now, `libraries-bom` covers a lot more of Google Cloud's artifacts. It'll come in handy in other cases sooner or later.

zhoufek (changes requested):

pom.xml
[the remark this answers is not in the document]
melbrodrigues: Sure thing. 

Added a property for the `libraries-bom` version and removed the comment. Please check again.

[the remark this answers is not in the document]
zhoufek: Thanks!

zhoufek (approved):

zhoufek:
Looks like there's a ton of merge conflicts on the attempted import. I think the last commit should be removed. Rather than merging main into this branch, do a rebase and force-push.

It also looks like `mvn spotless:apply` needs to be run.

noahmercado:
+1 to needing this

bvolpato:
Thanks for the contributions! I see that contributions are part of the codebase, so I will go ahead and close this PR.

PR #276, opened 2021-07-26
Relax the schema attribute pattern to match other white spaces

otourzan:
Current pattern doesn't allow any other white spaces like space before
the payload, for example the sample pattern in the README also doesn't
match.

### DIFF
--- a/v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
+++ b/v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
@@ -161,12 +161,12 @@ public class StreamingDataGeneratorTest {
 
   /** Tests generation of fake Json data message with attributes. */
   @Test
-  public void testJsonMessageGenerator_WithAttributes_returnsFakeMessageContaingAttributes()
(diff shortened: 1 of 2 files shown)


v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
```diff
@@ -161,12 +161,12 @@ public void testMessageGenerator_returnsFakeMessage() throws IOException {
 
   /** Tests generation of fake Json data message with attributes. */
   @Test
-  public void testJsonMessageGenerator_WithAttributes_returnsFakeMessageContaingAttributes()
```
prathapreddy123, line 164: Instead of modifying the existing test can you add another test

prathapreddy123 (changes requested):

v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorTest.java
otourzan: Done.

prathapreddy123 (approved):
LGTM

PR #285, opened 2021-08-31
fixed docs, made dataset and namespace optional in JSON config

Cherepushko:
Update README for Elasticsearch templates with content related to out of the box dashboards, getting started, better text for parameters, etc.


v2/googlecloud-to-elasticsearch/src/main/java/com/google/cloud/teleport/v2/elasticsearch/templates/PubSubToElasticsearch.java
```diff
@@ -160,7 +160,7 @@ public static PipelineResult run(PubSubToElasticsearchOptions options) {
         .apply(
             "WriteTransformFailuresToBigQuery",
             ErrorConverters.WritePubsubMessageErrors.newBuilder()
-                .setErrorRecordsTable(options.getErrorOutputTable())
+                .setErrorRecordsTable(options.getErrorOutputTable().trim())
```
prathapreddy123, line 163: Any reason to specify trim() only on this option as extra spaces might be possible for any options. If there is no specific reason, can remove trim() to maintain uniformity with other options.

prathapreddy123 (approved):

PR #288, opened 2021-09-03
Added ApiKey support, removed username and password options

Cherepushko:
Added ApiKey support.
Username and password options were removed.

### DIFF
--- a/v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/options/ElasticsearchWriteOptions.java
+++ b/v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/options/ElasticsearchWriteOptions.java
@@ -28,7 +28,8 @@ public interface ElasticsearchWriteOptions extends PipelineOptions {
 
   void setConnectionUrl(String connectionUrl);
 
-  @Description("Specify ApiKey")
+  @Description("API key for access without requiring basic authentication. " +
+          "Refer  https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html#security-api-create-api-key-request")
   @Validation.Required
   String getApiKey();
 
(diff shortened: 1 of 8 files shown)


Cherepushko:
@prathapreddy123 @suyograo
ElasticsearchIO from Beam 2.31 is not compatible because it requires the following options that we're not using:
idFn indexFn routingFn typeFn docVersionFn isDeleteFn
So in this pull request I've used previous version and added apikey feature manually.

suyograo:
> ElasticsearchIO from Beam 2.31 is not compatible because it requires the following options that we're not using:
idFn indexFn routingFn typeFn docVersionFn isDeleteFn

I am not sure why 2.31 made these required. They are functions that is executed to extract the id, index, type, etc. We already pass these directly by setting them. Very weird. 

I am ok with adding this feature on top of 2.29 until we get to 2.33 (hopefully soon)

v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/options/ElasticsearchWriteOptions.java
```diff
@@ -28,17 +28,11 @@
 
   void setConnectionUrl(String connectionUrl);
 
-  @Description("Username for Elasticsearch endpoint")
+  @Description("Specify ApiKey")
```
prathapreddy123, line 31: Better description would be:
```
API key for access without requiring basic authentication. Refer  https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html#security-api-create-api-key-request
```

prathapreddy123 (changes requested):

prathapreddy123 (approved):
LGTM. @suyograo - Can you please review from elastic end

suyograo:
LGTM from Elastic side. @prathapreddy123

prathapreddy123 (approved):

PR #292, opened 2021-09-20
feat: BigQuery to Bigtable template

billyjacobson:
This is a fairly simplified template to get data from BigQuery to Bigtable. It treats all values as strings and creates a rowkey from one of the unique columns in the query.


src/main/java/com/google/cloud/teleport/templates/BigQueryToBigtable.java
```diff
@@ -0,0 +1,64 @@
+/*
```
zhoufek, line 1: Is there a reason for making this a Classic Template rather than a Flex Template?

Note that, from the root directory, Classic Templates are defined under src/, which Flex Templates are defined under v2/.

src/main/java/com/google/cloud/teleport/templates/common/BigQueryConverters.java
```diff
@@ -174,6 +179,93 @@ public void processElement(ProcessContext context) {
     }
   }
 
+  /** Reads data from BigQuery and converts it to Bigtable mutation. */
+  @AutoValue
+  public abstract static class BigQueryToMutation
+      extends PTransform<PBegin, PCollection<Mutation>> {
+
+    abstract ValueProvider<String> query();
```
zhoufek, line 187: Related to my question about Classic vs. Flex, if this were a Flex Template, you wouldn't need to wrap these in a `ValueProvider`. The `ValueProvider` is a work-around in Classic Templates, since the graph was pre-serialized and static, but Flex Templates construct the graph during template launching, so options can be passed in just like any other Beam pipeline.

src/test/java/com/google/cloud/teleport/spanner/ExportRelatedTablesCheckTest.java
```diff
@@ -124,38 +124,105 @@ private void createAndPopulate(Ddl ddl, int numBatches) throws Exception {
   /* Validates behavior of exporting full db without selecting any tables */
```
zhoufek, line 124: Are you able to revert this file? I'm guessing you ran spotless (thanks!), but we generally prefer the old way of formatting. I must have missed guarding these tests against spotless formatting.

src/test/java/com/google/cloud/teleport/templates/common/BigQueryConvertersTest.java
```diff
@@ -157,6 +171,41 @@
   private String dateTimeFieldDesc = "Full publication date";
   private String dateTimeFieldValue = "2013-08-19 23:28:20.000567";
 
+  /** Tests that {@link BigQueryConverters.AvroToMutation} creates a Mutation. */
+  @Test
+  public void testAvroToMutation() {
+    // Create test data
```
zhoufek, line 177: Could you replace these comments with `// Arrange`, `// Act`, and `// Assert`? That seems to be what the test comments are mapping to, but the Arrange/Act/Assert convention is a bit more consistent with other tests.

zhoufek (changes requested):

README.md
```diff
@@ -15,6 +15,7 @@ their functionality.
 
 ## Template Pipelines
 
+* [BigQuery to Bigtable](v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java)
```
billyjacobson, line 18: Wasn't sure if this should be in here since the other v2 ones aren't in here.

v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
```diff
@@ -0,0 +1,118 @@
...
+ * Dataflow template which reads BigQuery data and writes it to Bigtable. The source data can be
+ * either a BigQuery table or an SQL query.
+ */
+public class BigQueryToBigtable {
+
+  /**
+   * The {@link BigQueryToBigtableOptions} class provides the custom execution options passed by the
+   * executor at the command-line.
+   */
+  public interface BigQueryToBigtableOptions extends PipelineOptions {
```
billyjacobson, line 39: Do I need the JSON file that specifies the parameters for this or does this repo take care of that?

billyjacobson:
@zhoufek Hey Zachary, I think I am ready for another review. I tried deploying and running it but got a bit confused on how to properly get the query to work as input since it can have a lot of special characters that need escaping.

README.md
zhoufek: It's ok to be here. We're probably going to completely rewrite the README once we start setting up proper documentation.

zhoufek:
> I tried deploying and running it but got a bit confused on how to properly get the query to work as input since it can have a lot of special characters that need escaping.

What query have you tried?

v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryConverters.java
```diff
@@ -88,6 +93,49 @@
...
+    public Mutation apply(SchemaAndRecord record) {
+      GenericRecord row = record.getRecord();
+      String rowkey = row.get(rowkey()).toString();
+      Put put = new Put(Bytes.toBytes(rowkey));
+
+      List<TableFieldSchema> columns = record.getTableSchema().getFields();
+      for (TableFieldSchema column : columns) {
+        String columnName = column.getName();
+        String columnValue = row.get(columnName).toString();
+        // TODO: handle other types and column families
```
zhoufek, line 129: Could you please add your GitHub username here? It'll make it easier to track and get context.

zhoufek (changes requested):

billyjacobson:
> > I tried deploying and running it but got a bit confused on how to properly get the query to work as input since it can have a lot of special characters that need escaping.
> 
> What query have you tried?

I was trying something like this to take advantage of one of the public datasets

```SELECT CONCAT(SenderCompID,'#', OrderID) as rowkey, \* FROM \`bigquery-public-data.cymbal_investments.trade_capture_report\` LIMIT 100```

**Comment by zhoufek** [comment]
> I was trying something like this to take advantage of one of the public datasets
> 
> `` SELECT CONCAT(SenderCompID,'#', OrderID) as rowkey, \* FROM \`bigquery-public-data.cymbal_investments.trade_capture_report\` LIMIT 100 ``

I don't think * and \` need to be escaped.

You can probably leave off the \` characters around the dataset as well. At least that's what I normally do.

**Comment by billyjacobson** [comment]
> > I was trying something like this to take advantage of one of the public datasets
> > `` SELECT CONCAT(SenderCompID,'#', OrderID) as rowkey, \* FROM \`bigquery-public-data.cymbal_investments.trade_capture_report\` LIMIT 100 ``
> 
> I don't think * and ` need to be escaped.
> 
> You can probably leave off the ` characters around the dataset as well. At least that's what I normally do.

I figured out the issue. I needed to specify a [different delimiter ](https://cloud.google.com/sdk/gcloud/reference/topic/escaping) but was doing it wrong. This is now working as expected and ready for review

File: v2/bigquery-to-bigtable/README.md [hunk 8]
```diff
@@ -0,0 +1,162 @@
...
+```sh
+export JOB_NAME="${TEMPLATE_MODULE}-`date +%Y%m%d-%H%M%S-%N`"
+gcloud beta dataflow flex-template run ${JOB_NAME} \
+        --project=${PROJECT} --region=${REGION} \
+        --template-file-gcs-location=${TEMPLATE_IMAGE_SPEC} \
+        --parameters ^~^readQuery="${READ_QUERY}" \
+        --parameters readIdColumn=${READ_ID_COLUMN},bigtableWriteProjectId=${BIGTABLE_WRITE_PROJECT_ID},bigtableWriteInstanceId=${BIGTABLE_WRITE_INSTANCE_ID},bigtableWriteTableId=${BIGTABLE_WRITE_TABLE_ID},bigtableWriteColumnFamily=${BIGTABLE_WRITE_COLUMN_FAMILY}
+```
+
+Note: The `^~^` prefix on readQuery indicates `~` as a delimiter for the query. If your
```
zhoufek, line 152: I think this should be reworded to make it clear that '~' is used to make sure that commas aren't treated as delimiters. Something like: "The `^~^` prefix on `readQuery` is used to make `~` a delimiter instead of commas. This allows commas to be used in the query."

I think that makes it clearer why it is needed. It'll also help explain why something other than `~` should be used if it's in the query or indicate cases where the custom delimiter is unnecessary (e.g. `SELECT * FROM foo.bar`).

Maybe also add a link to the custom delimiter page so that people can get more info on this if they aren't familiar with this feature of `gcloud`.

zhoufek (changes requested):

v2/bigquery-to-bigtable/README.md
billyjacobson: great, updated

zhoufek:
Looks good! Thanks for adding this!

PR #304, opened 2021-10-25
Add debug info to Googlecloud-to-elasticsearch

Cherepushko:
Added optional parameter **verboseDebugMessages** that enables additional output when error occurs. It shows full response when there's an issue writing to Elasticsearch. Changes made by this PR will be discarded after migration to Apache Beam 2.33.


bvolpato:
I believe we don't need this anymore? We've moved way past 2.33.0.

Sorry for taking so long to review this.

PR #308, opened 2021-11-04
Feat/logging 6223 define sane defaults

jsubirat:


jsubirat:
Sorry, created by accident

PR #312, opened 2021-11-10
Add PubsubToNewRelic template

jsubirat:
## Description
The following PR adds a new Dataflow template which takes log records from a PubSub subscription and forwards them to New Relic using the HTTP Logs API.

## Tests
- Unitary tests have been provided for relevant classes.
- Several full end-to-end integration tests (only mocking PubSub and using MockServer to capture the emitted HTTP requests) can be found in `NewRelicPipelineTest`.

## IMPORTANT NOTE
It seems that when we ran `mvn spotless:apply`, as suggested in the `README.md`, other non-related files were also formatted (only applies to white spaces/lines). These files are:

- `src/main/java/com/google/cloud/teleport/spanner/SpannerRecordConverter.java`
- `src/main/java/com/google/cloud/teleport/templates/SpannerToText.java`
- `src/main/java/com/google/cloud/teleport/util/GCSUtils.java`
- `src/test/java/com/google/cloud/teleport/spanner/ExportRelatedTablesCheckTest.java`

Signed-off-by: Josep Subirats Castell <jsubirats@newrelic.com>


jsubirat:
@googlebot I signed it!

jsubirat:
@googlebot I signed it!

jsubirat:
Hey folks, any chance to get a review on this? Thanks a lot in advance!

src/main/java/com/google/cloud/teleport/newrelic/dtos/coders/NewRelicLogApiSendErrorCoder.java
```diff
@@ -0,0 +1,76 @@
...
+      NullableCoder.of(STRING_UTF_8_CODER);
+  private static final NullableCoder<Integer> INTEGER_NULLABLE_CODER =
+      NullableCoder.of(BigEndianIntegerCoder.of());
+
+  public static NewRelicLogApiSendErrorCoder getInstance() {
+    return SINGLETON;
+  }
+
+  @Override
+  public void encode(NewRelicLogApiSendError value, OutputStream out) throws IOException {
```
an2x, line 50: Coders have to be designed carefully because ideally they have to be backwards-compatible, e.g. to support in-place streaming job updates. This means that coder v2 should be able to read data produced by coder v1 at the minimum.

With this implementation, it will be hard to add a 4th new field in the future, for example.

I suggest adding something like a version number as the first field in the encoded data to make future changes easier, unless you're absolutely sure this won't change in the future or users won't need to update their jobs.

Something like:

```
  public void encode(NewRelicLogApiSendError value, OutputStream out) throws IOException {
    VAR_INT_CODER.encode(2); // encoding version
    INTEGER_NULLABLE_CODER.encode(value.getStatusCode(), out);
    STRING_NULLABLE_CODER.encode(value.getStatusMessage(), out);
    STRING_NULLABLE_CODER.encode(value.getPayload(), out);
    STRING_NULLABLE_CODER.encode(value.newFieldAddedInVersion2(), out);
  }

  public NewRelicLogApiSendError decode(InputStream in) throws IOException {
    final version = VAR_INT_CODER.decode(in);
    if (version >= 1) {
        statusCode = INTEGER_NULLABLE_CODER.decode(in);
        statusMessage = STRING_NULLABLE_CODER.decode(in);
        payload = STRING_NULLABLE_CODER.decode(in);
        newFieldAddedInVersion2 = <some default for older versions>;
    }
    if (version >= 2) {
        newFieldAddedInVersion2 = STRING_NULLABLE_CODER.decode(in);
    }

    ...
  }
```

an2x, line 50: Coders have to be designed carefully because ideally they have to be backwards-compatible, e.g. to support in-place streaming job updates. This means that coder v2 should be able to read data produced by coder v1 at the minimum.

With this implementation, it will be hard to add a 4th new field in the future, for example.

I suggest adding something like a version number as the first field in the encoded data to make future changes easier.
Something like:

```
  public void encode(NewRelicLogApiSendError value, OutputStream out) throws IOException {
    INTEGER_CODER.encode(2); // encoding version
    INTEGER_NULLABLE_CODER.encode(value.getStatusCode(), out);
    STRING_NULLABLE_CODER.encode(value.getStatusMessage(), out);
    STRING_NULLABLE_CODER.encode(value.getPayload(), out);
    STRING_NULLABLE_CODER.encode(value.newFieldAddedInVersion2(), out);
  }

  public NewRelicLogApiSendError decode(InputStream in) throws IOException {
    final version = INTEGER_CODER.decode(in);
    if (version >= 1) {
        statusCode = INTEGER_NULLABLE_CODER.decode(in);
        statusMessage = STRING_NULLABLE_CODER.decode(in);
        payload = STRING_NULLABLE_CODER.decode(in);
        newFieldAddedInVersion2 = <some default for older versions>;
    }
    if (version >= 2) {
        newFieldAddedInVersion2 = STRING_NULLABLE_CODER.decode(in);
    }

    ...
  }
```

src/main/java/com/google/cloud/teleport/newrelic/dtos/coders/NewRelicLogApiSendErrorCoder.java
```diff
@@ -0,0 +1,76 @@
...
+    INTEGER_NULLABLE_CODER.encode(value.getStatusCode(), out);
```
an2x, line 51: If status code is a small number, consider using VarIntCoder instead of BigEndianIntegerCoder to save a few bytes.

an2x, line 51: If status code is a small number, consider using VarIntCoder instead of BigEndianIntegerCoder to save a few bytes.

src/main/java/com/google/cloud/teleport/newrelic/dtos/coders/NewRelicLogApiSendErrorCoder.java
```diff
@@ -0,0 +1,76 @@
...
+    return new NewRelicLogApiSendError(payload, statusMessage, statusCode);
+  }
+
+  @Override
+  public TypeDescriptor<NewRelicLogApiSendError> getEncodedTypeDescriptor() {
+    return TYPE_DESCRIPTOR;
+  }
+
+  @Override
+  public void verifyDeterministic() throws Coder.NonDeterministicException {
```
an2x, line 71: Unless I'm missing something, this coder is deterministic.
"Deterministic" means that given 2 NewRelicLogApiSendError objects that are equal, the coder will produce the same encoded data for them, which seems to be true.

Typically, verifyDeterministic() methods look like this and simply check if the other underlying coders are deterministic:

```
  @Override
  public void verifyDeterministic() throws NonDeterministicException {
    VAR_INT_CODER.verifyDeterministic();
    INTEGER_NULLABLE_CODER.verifyDeterministic();
    STRING_NULLABLE_CODER.verifyDeterministic();
  }
```

an2x, line 71: Unless I'm missing something, this coder is deterministic.
"Deterministic" means that given 2 NewRelicLogApiSendError objects that are equal, the coder will produce the same encoded data for them, which seems to be true.

Typically, verifyDeterministic() methods look like this and simply check if the other underlying coders are deterministic:

```
  @Override
  public void verifyDeterministic() throws NonDeterministicException {
    VAR_INT_CODER.verifyDeterministic();
    INTEGER_NULLABLE_CODER.verifyDeterministic();
    STRING_NULLABLE_CODER.verifyDeterministic();
  }
```

src/main/java/com/google/cloud/teleport/newrelic/dtos/coders/NewRelicLogRecordCoder.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.beam.sdk.coders.NullableCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.values.TypeDescriptor;
+
+/**
+ * A {@link org.apache.beam.sdk.coders.Coder} for {@link NewRelicLogRecord} objects. It allows
+ * serializing and deserializing {@link NewRelicLogRecord} objects, which can then be transmitted
+ * through a Beam pipeline using PCollection/PCollectionTuple objects.
+ */
+public class NewRelicLogRecordCoder extends AtomicCoder<NewRelicLogRecord> {
```
an2x, line 33: See comments for NewRelicLogApiSendErrorCoder. The same applies for this one too.

an2x, line 33: See comments for NewRelicLogApiSendErrorCoder. The same applies for this one too.

an2x (changes requested):
Is there any specific reason this is implemented as a Classic template? Have you considered [Flex templates](https://cloud.google.com/dataflow/docs/guides/templates/using-flex-templates)?

Classic is fine, but in general we recommend to implement new templates as Flex templates because they are more flexible, e.g. it's impossible to construct the pipeline dynamically depending on a template parameter value with Classic templates, you may run into such issues in the future when the template evolves, and it will be hard to switch from Classic to Flex once there are customers already using the template. Just FYI.

bvolpato:
Hi @jsubirat -- thanks for your contribution here. Unfortunately we ran behind on reviews / merges, and the structure of the project changed a bit. Is there still interest in pursuing this template? We can work together on a plan to resolve conflicts + review the code.

damccorm:
Looks like this PR is stale and needs significant revisions. Additionally, at this point we're generally not adding new templates without first getting approval - see https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/code-contributions.md#adding-new-templates

I'm going to close this PR, if you're still interested in contributing it please file an issue in the public tracker as described in https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/code-contributions.md#adding-new-templates

PR #321, opened 2022-01-05
Run Spotless on PRs

zhoufek:
Spotless violations frequently slip through. This should prevent that.

The check will only run on the relevant template "family" (i.e. Classic or Flex) to hopefully speed things up. We can make it more granular later if Flex Templates start taking too long, but the check normally completes in a matter of seconds at the current time.


zhoufek:
Closing this one and running it from the equivalent PR322 so that the check is picked up. (It's picked up here but only after I opened the other PR.)

PR #332, opened 2022-01-14
Add YAML Templates for Github Issues

WeiwenXu21:
Added two YAML templates for `bug` and `feature request` issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-template.yml
+++ b/.github/ISSUE_TEMPLATE/bug-template.yml
@@ -5,70 +5,15 @@ labels: ["bug", "p2", "needs triage"]
 body:
   - type: markdown
     attributes:
-      value: |
-        Thank you so much for taking the time to fill out this bug report! Please note that changing the issue label from p2 to p0/p1 will *NOT* speed up triaging. **Any issue that is currently affecting your production should be directed to Google Cloud customer support.** If you cannot find the template in the dropdown menu below, we probably no longer support the template.
-  - type: dropdown
+      value: >
+        Thank you so much for taking the time to fill out this bug report!
+        
+        **Please note: any issue that is currently affecting your production should be directed to Google Cloud customer support.**
+  - type: input
     id: template
     attributes:
-      label: Related Template
-      options:
-        - Pubsub-to-BigQuery (classic, streaming)
-        - PubsubAvro-to-BigQuery (flex, steaming)
-        - PubsubProto-to-BigQuery (flex, streaming)
-        - Pubsub-to-Pubsub (classic, streaming)
-        - Pubsub-to-Splunk (classic, streaming)
-        - Pubsub-to-Avro (classic, streaming)
-        - Pubsub-to-Text (classic, streaming)
-        - Pubsub-to-MongoDB (flex, streaming)
-        - Pubsub-to-Elasticsearch (flex, streaming)
-        - Datastream-to-Spanner (flex, streaming)
-        - Text-to-BigQuery-stream (classic, streaming)
-        - Text-to-Pubsub-stream (classic, streaming)
-        - DLPText-to-BigQuery-stream (classic, streaming)
-        - CDC (flex, streaming)
-        - Kafka-to-BigQuery (flex, streaming)
-        - Datastream-to-BigQuery (flex, streaming)
-        - Datastream-to-Postgres (flex, streaming)
-        - JDBC-to-Pubsub (flex, streaming)
-        - Pubsub-to-JDBC (flex, streaming) 
-        - BigQuery-to-TFRecord (classic, batch)
-        - BigQuery-to-Parquet (flex, batch)
-        - BigQuery-to-Elasticsearch (flex, batch)
-        - Bigtable-to-Avro (classic, batch)
-        - Bigtable-to-Parquet (classic, batch)
-        - Datastore-to-Text (classic, batch)
-        - Spanner-to-Avro/Spanner-Export (classic, batch)
-        - Spanner-to-Text (classic, batch)
-        - Avro-to-Bigtable (classic, batch)
-        - Avro-to-Spanner/Spanner-Import (classic, batch)
-        - Parquet-to-Bigtable (classic, batch)
-        - TextIO-to-BigQuery (classic, batch)
-        - Text-to-Datastore (classic, batch)
-        - Text-to-Pubsub-batch (classic, batch)
-        - Text-to-Spanner/Spanner-TextImport (classic, batch)
-        - GCS-to-Elasticsearch (flex, batch)
-        - JDBC-to-BigQuery (classic, batch)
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-template.yml
```diff
@@ -0,0 +1,88 @@
...
+      label: Beam Version
+      description: What version of beam are you using?
+      options:
+        - 2.35.0
+        - 2.34.0
+        - 2.33.0
+        - 2.32.0
+        - 2.31.0
+        - 2.30.0
+        - Older than 2.30.0
```
zhoufek, line 71: Maybe also a "Newer than 2.35.0" in case we forget to update it with a new Beam release.

.github/ISSUE_TEMPLATE/bug-template.yml
```diff
@@ -0,0 +1,88 @@
...
+    validations:
+      required: true
+  - type: textarea
```
zhoufek, line 74: Maybe put this above setting the version.

I think the two most important parts, and the two we're likely to always require, are the affected template and the user's concerns. Logs and Beam version seem more like additional info that can give us context, like if someone is using a really old Beam version.

.github/ISSUE_TEMPLATE/feature-request-template.yml
```diff
@@ -0,0 +1,65 @@
+name: Feature Request
+description: File a feature request issue
+title: "[Feature Request]: <Please Replace This with a Summary of Your Request>"
+labels: ["addition", "p2", "needs triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you so much for taking the time to fill out this feature request report! Please note that changing the issue label from p2 to p0/p1 will *NOT* speed up triaging. **Any issue that is currently affecting your production should be directed to Google Cloud customer support.** If you cannot find the template in the dropdown menu below, we probably no longer support the template.
```
zhoufek, line 9: I think the following is good enough:

```
Thank you so much for taking the time to fill out this feature request!
```

The rest seems more relevant to bug reports.

zhoufek (changes requested):

.github/ISSUE_TEMPLATE/bug-template.yml
```diff
@@ -0,0 +1,88 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you so much for taking the time to fill out this bug report! Please note that changing the issue label from p2 to p0/p1 will *NOT* speed up triaging. **Any issue that is currently affecting your production should be directed to Google Cloud customer support.** If you cannot find the template in the dropdown menu below, we probably no longer support the template.
+  - type: dropdown
+    id: template
+    attributes:
+      label: Related Template
+      options:
```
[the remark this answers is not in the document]
WeiwenXu21: No worries! I forgot to add a comment on this part because I didn't quite like it either since it requires constant maintenance (adding/removing to the list). I chose this over `input` mainly to avoid customers trying to submit an issue related to unsupported templates. Submitting a bug issue only to be told that it's not supported any more a few days later sounds disappointing to me lol

.github/ISSUE_TEMPLATE/bug-template.yml
```diff
@@ -0,0 +1,88 @@
+name: Bug
+description: File a bug issue
+title: "[Bug]: <Please Replace with a Summary of the Bug> "
+labels: ["bug", "p2", "needs triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you so much for taking the time to fill out this bug report! Please note that changing the issue label from p2 to p0/p1 will *NOT* speed up triaging. **Any issue that is currently affecting your production should be directed to Google Cloud customer support.** If you cannot find the template in the dropdown menu below, we probably no longer support the template.
```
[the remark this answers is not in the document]
WeiwenXu21: Oh I didn't know external contributors don't have the access to change the issue label. I'll fix that!

.github/ISSUE_TEMPLATE/bug-template.yml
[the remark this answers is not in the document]
an2x: +1 to make this an input.

1. I don't like the idea of keeping another list up-to-date manually. I'd prefer to read a free-form text to figure out the template instead.
2. Sometimes there are bug that are not specific to a single template, an input will allow you to enter something like "All Spanner-related templates".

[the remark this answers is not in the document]
zhoufek: > Submitting a bug issue only to be told that it's not supported any more a few days later sounds frustrating to me lol

Maybe we could have a label we apply to issues to indicate that it won't be addressed officially by our team, but someone outside the team could address the issue themselves. 

Alternatively, we might want to consider what it means to have no support. That's at least starting to sound like dead code, but that's a whole other discussion.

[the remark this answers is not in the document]
WeiwenXu21: SGTM! I changed this part into input type.

.github/ISSUE_TEMPLATE/bug-template.yml
```diff
@@ -0,0 +1,48 @@
...
+    attributes:
+      value: >
+        Thank you so much for taking the time to fill out this bug report!
+        
+        **Please note: any issue that is currently affecting your production should be directed to Google Cloud customer support.**
+  - type: input
+    id: template
+    attributes:
+      label: Related Template(s)
+      description: "Which template(s) this bug is affecting?"
```
zhoufek, line 16: Could you change "this bug is" to "is this bug"? The latter reads more naturally as a sentence.

.github/ISSUE_TEMPLATE/feature-request-template.yml
```diff
@@ -0,0 +1,22 @@
+name: Feature Request
+description: File a feature request issue
+title: "[Feature Request]: <Please Replace This with a Summary of Your Request>"
+labels: ["addition", "p2", "needs triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you so much for taking the time to fill out this feature request report!
```
zhoufek, line 9: optional: I think we can remove "report". Might just be a matter of opinion, but I tend to think of "bug report" and "feature request", not "feature request report".

zhoufek (changes requested):
Thanks for making the changes! Overall, this looks good. I just have a couple minor nits about wording.

.github/ISSUE_TEMPLATE/bug-template.yml
WeiwenXu21: My bad! Fixed.

.github/ISSUE_TEMPLATE/feature-request-template.yml
WeiwenXu21: Makes sense. Done.

PR #345, opened 2022-02-03
Add precondition check for running checkstyle

zhoufek:
TODO: Update when ready


zhoufek:
Seems I misunderstood checkstyle. From the [limitations](https://checkstyle.org/writingchecks.html#Limitations): "To get valid violations, code have to be compilable, in other case you can get not easy to understand parse errors."

Furthermore, it seems that using `.class` files rather than raw source code is preferred for many checks. I guess that explains all the time to compile.

I'll repurpose the valuable stuff into another PR. Otherwise, checkstyle will just run in a build check.

PR #352, opened 2022-02-14
Add build and unit test checks to Java PRs

zhoufek:
Add an action to make sure everything in a PR builds properly. This will also run checkstyle after the build is complete. Due to some existing issues with our dependencies, dependency analysis is turned off for now.

Other improvements:
* Cleanup of the different Go files.
* The Action is now filtered to just Java changes. Some filtering is still done within the workflow to account for some edge cases where a workflow may trigger without any relevant changes to one of the affected modules.
* We now stream both stdout and stderr when running a command. (Note: The implementation guarantees that two lines won't be printed at the same time, but it's maybe possible the lines will be printed slightly out-of-order. I'm not sure how practically possible that is, though.)

### DIFF
--- a/cicd/internal/flags/common-flags_test.go
+++ b/cicd/internal/flags/common-flags_test.go
@@ -74,6 +74,11 @@ func TestChangedFilesRegexes(t *testing.T) {
 			regexes:  []string{"\\d"},
 			expected: []string{"file1", "file2"},
 		},
+		{
+			files:    "foo.c,bar.cc",
+			regexes:  []string{"\\.c$"},
+			expected: []string{"foo.c"},
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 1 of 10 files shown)


zhoufek:
Made a simple Java change in a now-removed commit just to validate that the changes work. Here's the link to the workflow run: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/1841638087

zhoufek:
Added running unit tests, which required setting up a common `clean install` workflow. I also gave spotless its own cache entry.

I think that this mostly gets the PR checks up to where we want them (minus integration tests, but that's a separate workflow). We still need to get Maven to actually build dependencies that aren't under the same POM, though.

cicd/internal/flags/common-flags_test.go
```diff
@@ -45,10 +45,50 @@ func TestChangedFiles(t *testing.T) {
...
 	}
 }
+
+func TestChangedFilesRegexes(t *testing.T) {
+	tests := []struct {
+		files    string
+		regexes  []string
+		expected []string
+	}{
+		{
```
an2x, line 62: nit: I'd add a test case with an end of line character, which is very typical, e.g. "\.java$" regexp should match Foo.java, but not Bar.javaa.

an2x (approved):

PR #354, opened 2022-02-15
Take more control of Java dependency cache

zhoufek:
A drawback of the `actions/setup-java` is that its caching uses `actions/cache`, which doesn't allow forcing an overwrite of the cached data. For us, this means that a short-running workflow, like checking formatting, can easily claim the cache entry. This effectively makes the cache useless for longer-running workflows. The only real workaround is to change the key, which is what is done here.

Since `actions/setup-java` is intended for simple uses of the cache, this switches to using `actions/cache` directly. It'll attempt to refresh the cache daily, but we can manually claim a future date in case something goes wrong. Worst case scenario is one of those short workflows runs and keeps the cache constant, but since it would pull from the previous cache, it's better than what we have now.

### DIFF
--- a/.github/actions/setup-java-env/action.yml
+++ b/.github/actions/setup-java-env/action.yml
@@ -17,33 +17,48 @@
 # without the key. Otherwise, today's date and yetserday's date (both in UTC) will be
 # tried before going to the first match without any date.
 
-name: 'Setup Java Cache'
-description: 'Sets up the Java dependency cache.'
+name: 'Setup Java Environment'
+description: 'Sets up the full Java environment.'
 
 inputs:
-  key:
+  cache-key:
     type: string
     description: 'Key to use for the cache entry'
     required: false
     default: ''
+  java-version:
+    type: string
+    description: 'The version of Java to install'
+    required: false
+    default: '8'
+outputs:
+  cache-hit:
+    description: 'Whether or not there was a cache hit'
+    value: ${{ steps.setup-cache.outputs.cache-hit }}
 
 runs:
   using: 'composite'
   steps:
+    - name: Setup Java
+      uses: actions/setup-java@a12e082d834968c1847f782019214fadd20719f6
+      with:
+        distribution: 'zulu'
+        java-version: ${{ inputs.java-version }}
     - name: Prepare Key
       shell: bash
       env:
-        KEY: ${{ inputs.key }}
+        KEY: ${{ inputs.cache-key }}
       run: |
-        if [[ "$KEY" == "" ]]; then
+        if [[ "$KEY" == "" || "$KEY" == "''" ]]; then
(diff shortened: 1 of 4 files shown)


zhoufek:
Somewhat interesting: This PR serves as a good example of the problem. A run happened for `20220216`, and the short-running Spotless check claimed the cache entry ahead of the workflow meant to prepare the cache, preventing the more useful value from claiming that entry:

* [Spotless run](https://github.com/GoogleCloudPlatform/DataflowTemplates/runs/5217234224?check_suite_focus=true)
* [Prepare cache run](https://github.com/GoogleCloudPlatform/DataflowTemplates/runs/5217234240?check_suite_focus=true)

The message under `Post Setup Java` for the prepare cache run notes another job claimed the key.

zhoufek:
@damccorm

Thanks for helping with this!

.github/actions/setup-java-env/action.yml
```diff
@@ -0,0 +1,68 @@
...
+  steps:
+    - name: Setup Java
+      uses: actions/setup-java@a12e082d834968c1847f782019214fadd20719f6
+      with:
+        distribution: 'zulu'
+        java-version: ${{ inputs.java-version }}
+    - name: Prepare Key
+      shell: bash
+      env:
+        KEY: ${{ inputs.key }}
```
damccorm, line 50: Should this be inputs.cache-key?

.github/actions/setup-java-env/action.yml
```diff
@@ -0,0 +1,68 @@
...
+      run: |
+        if [[ "$KEY" == "" ]]; then
+          echo "TODAY=$(date -u +%Y%m%d)" >> $GITHUB_ENV
+          echo "YESTERDAY=$(date -u --date='1 day ago' +%Y%m%d)" >> $GITHUB_ENV
+        else
+          echo "TODAY=${{ inputs.key }}" >> $GITHUB_ENV
+          echo "YESTERDAY=${{ inputs.key }}" >> $GITHUB_ENV
```
damccorm, line 57: Nit: prefer `$KEY` over `${{ inputs.key }}` for consistency since we're already using it above (or just use `${{ inputs.key }}`)

damccorm (commented):
I think this generally looks good to me, left a couple notes

.github/actions/setup-java-env/action.yml
zhoufek: Yes. Looks like a few things in this file became stale.

.github/actions/setup-java-env/action.yml
zhoufek: Made it more consistent. I prefer `$KEY` here, since I think it is a bit cleaner.

damccorm (approved):

PR #358, opened 2022-02-17
Stop caching the Templates with their dependencies

zhoufek:
Tries to optimize cache use by removing the artifacts under `com/google/cloud/teleport/`. Since we rebuild these artifacts every time, it makes no sense to cache them. It also takes up a lot of space, so this will help a lot with keeping the cache size under control.


zhoufek:
Going to wait on PR #352 to properly add cleanup-env to the Java PR workflow.

pranavbhandari24 (approved):
LGTM

PR #359, opened 2022-02-28
Add GoogleAdsToBigQuery template

sjvanrossum:
This template enables users to write custom queries in the Google Ads Query Language and execute them against the Google Ads reporting API.
The results are transformed such that every selected nested field in the GoogleAdsRow is pulled up as a top-level field.
The table schema is inferred using the GoogleAdsRow protobuf descriptor, but users are advised to pre-create tables (optionally partitioned) such that e.g. date fields (which are encoded as strings in the protobuf) are converted to BigQuery DATE fields.


kerrydc:
R: @WeiwenXu21

v2/pom.xml
```diff
@@ -658,6 +658,7 @@
         <module>datastream-to-sql</module>
         <module>elasticsearch-common</module>
         <module>file-format-conversion</module>
+        <module>googleads-to-bigquery</module>
```
WeiwenXu21, line 661: For scalability reasons, we are keeping new google product to google product templates under `googlecloud-to-googlecloud` module so please move this template to `googlecloud-to-googlecloud` module as well.

WeiwenXu21 (changes requested):

v2/googleads-to-bigquery/src/main/java/com/google/cloud/teleport/v2/io/GoogleAdsIO.java
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  public static ReadAll readAll() {
+    return new AutoValue_GoogleAdsIO_ReadAll.Builder()
+        .setGoogleAdsClientFactory(DefaultGoogleAdsClientFactory.getInstance())
+        .setRateLimitPolicyFactory(RateLimitPolicyFactory.withoutRatelimiter())
+        .setInferBeamSchema(false)
+        .build();
+  }
+
```
pabloem, line 65: Since this IO may eventually be added to Beam for others to use, I would recommend adding extensive javador in the file header and for the `read` and `readAll` functions.

bvolpato:
Hi @sjvanrossum -- thanks for your contribution here. Unfortunately we ran behind on reviews, and the structure of the project changed a bit. Is there still interest in pursuing this template? We can work together on a plan to resolve conflicts + review the code.

sjvanrossum:
There's still interest in pursuing this, but we don't have all the thumbs ups we need to move forward with publishing this as of yet. I'll update the PR if I have any news on that.

sjvanrossum:
Hi @bvolpato, I've updated the template to use the GoogleAdsIO connector which shipped with Beam 2.50.0.
I've added the new template annotations and generated a README to accompany the template.
Tests validate the transformation from the API's reporting proto to a flattened record for BigQuery.
The rate limiting policy is simple (uses a Guava RateLimiter), but configurable to limit the per worker QPS.
Setting up integration tests would require setting up a Google Ads test account for this project, so I've left out the e2e test for now until we can get that in place.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/utils/GoogleAdsRateLimitPolicyFactory.java
```diff
@@ -0,0 +1,62 @@
...
+              @Override
+              public void onSuccess(String developerToken, String customerId, Message request) {}
+
+              @Override
+              public void onError(
+                  String developerToken,
+                  String customerId,
+                  Message request,
+                  GoogleAdsError error) {}
+            });
```
bvolpato, line 60: Shouldn't we at least log here?

bvolpato (approved):

PR #360, opened 2022-02-28
JdbcToBigQuery template: Fix java.lang.ClassCastException: class java.sql.Timestamp

callidustaurus:
### Context
This PR is an extension to this https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/253.
It turned out that MySQL is not returning a `java.time.LocalDateTime` even if the column-type received with `java.sql.ResultSet.getMetaData().getColumnTypeName()` is "datetime", but instead it returns `java.sql.Timestamp`.
That leads to `java.lang.ClassCastException: class java.sql.Timestamp cannot be cast to class java.time.temporal.TemporalAccessor` as it tries to cast the result to TemporalAccessor (see [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java#L148)) 

I've tried multiple JDBC drivers to verify this issue, all with the same result:
* net.sourceforge.jtds.jdbc.Driver
* com.mysql.jdbc.Driver
* com.mysql.cj.jdbc.Driver
* com.microsoft.sqlserver.jdbc.SQLServerDriver

All the drivers return a java.sql.Timestamp instance when `resultSet.getObject()` is called.

I've added a simple check to avoid a ClassCastException for this particular case.

### Tests
I extended the existing tests with a related test case. Sadly i wasn't able to execute all the tests as i got connection exception from Cassandra which seems not to be able to connect to localhost somehow (i guess). So i executed test classes individually, i hope thats enough 🤞

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
+++ b/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
@@ -143,6 +143,7 @@ public class JdbcConverters {
             outputTableRow.set(
                 metaData.getColumnName(i), dateFormatter.format(resultSet.getObject(i)));
             break;
+          case "datetime2":
           case "datetime":
             Object item = resultSet.getObject(i);
             String datetime;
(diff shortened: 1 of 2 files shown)


bvolpato:
Hi @callidustaurus -- thanks for your contribution here.

Unfortunately we ran behind on reviews, and the structure of the project changed a bit. Is there still interest in pursuing this change?

We can work together on a plan to resolve conflicts + review the code.

bvolpato:
I believe we have fixed a similar error after a customer faced it, but it just doesn't have all the changes that are contained here.

PR #371, opened 2022-03-27
update & add jdbc template

case-k-git:

PR #372, opened 2022-04-04
fix jdbc to BigQuery datetime template

case-k-git:
dataflow timestamp format is wrong use HH instead of hh

please read this documents
```
   h       clock-hour-of-am-pm (1-12)  number            12
   H       hour-of-day (0-23)          number            0
```

 https://docs.oracle.com/javase/jp/8/docs/api/java/time/format/DateTimeFormatter.html

please look at after this pr because this template can not use for SQL Server
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/371


bvolpato:
Thanks for the contributions! We ran behind on reviews, but checking the code, I see this was already addressed a part of another change. I'll go ahead and close this for now.

PR #383, opened 2022-05-10
Adding GCS-to-Splunk template

andreigurau:
## Description
The following PR adds a new Dataflow template which reads CSV records from a GCS bucket, converts the record to JSON, and forwards them to Splunk. This PR is merely a rough draft of what this template may look like, and will be further refined and developed as time goes on.

## Tests
Unit tests have been written for the GCSToSplunk class, found in the GCSToSplunkTest.java file. The unit tests contain the transforms necessary for the pipeline to execute. The only parts of the pipeline that haven't been added as parts of the unit tests are the `4) Write SplunkEvents to Splunk's HEC end point.` step, which should be covered in Beam's [SplunkIOTest](https://github.com/apache/beam/blob/243128a8fc52798e1b58b0cf1a271d95ee7aa241/sdks/java/io/splunk/src/test/java/org/apache/beam/sdk/io/splunk/SplunkIOTest.java) class, and the writing of error logs as part of step 5, which should be covered in the [ErrorConvertersTest](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/e57f23e793a8ba20e626983b208ebda65abaadb8/v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/ErrorConvertersTest.java) class
 
Any other unit tests in this PR were copied over from the classic templates directory, as part of the integration of Splunk converters/coders to the flex templates directory. 

## IMPORTANT NOTE
A lot of these files have been copied over from the classic templates' directory. These files are:

- `v2/common/src/main/java/com/google/cloud/teleport/v2/coders/SplunkEventCoder.java`
- `v2/common/src/test/java/com/google/cloud/teleport/v2/coders/SplunkEventCoderTest.java`
- `v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java`
- `v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/SplunkConvertersTest.java`
- `v2/common/src/main/java/com/google/cloud/teleport/v2/coders/SplunkWriteErrorCoder.java`
- `v2/common/src/test/java/com/google/cloud/teleport/v2/coders/SplunkWriteErrorCoderTest.java`

However, these files have been modified as to not use or process the `field` attribute, present in the [classic templates' SplunkEvent class](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/src/main/java/com/google/cloud/teleport/splunk/SplunkEvent.java). This is due to the fact that we are using the [SplunkEvent class from the Beam SDK](https://github.com/apache/beam/blob/master/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkEvent.java), which doesn't contain that attribute (done to minimize the amount of copying/pasting existing code. If updates need to be made to Splunk-related classes, they will ideally be done in the Beam-SDK side).

Similarly, a lot of options previously found in the classic templates [SplunkConverters](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/292c0dfdc5d3b1d0e68014ff96bda6651e8347ed/src/main/java/com/google/cloud/teleport/templates/common/SplunkConverters.java) have been removed in the flex template's version of the same class, due to the fact that the [SplunkIO class on Beam](https://github.com/apache/beam/blob/85a122735f84c0ee46ba0fb583d9ff9e05dcf2fc/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkIO.java) does not support these features.


andreigurau:
R: @zhoufek

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+ * <p>NOTE: This is a work in progress, do not attempt to run this pipeline
+ */
+public class GCSToSplunk {
+
+  /** String/String Coder for FailsafeElement. */
+  private static final FailsafeElementCoder<String, String> FAILSAFE_ELEMENT_CODER =
+      FailsafeElementCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of());
+
+  /** The tag for the headers of the CSV if required. */
+  static final TupleTag<String> CSV_HEADERS = new TupleTag<String>() {};
```
zhoufek, line 63: I don't think these are intended to be used anywhere other than the template. If so, please make them private. If they _must_ be package-private for testing purposes, please add the `@VisibleForTesting` annotation. (This is generally discouraged but not always avoidable.)

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+  static final Logger LOG = LoggerFactory.getLogger(GCSToSplunk.class);
+
+  /**
+   * The {@link GCSToSplunkOptions} class provides the custom execution options passed by the
+   * executor at the command-line.
+   */
+  public interface GCSToSplunkOptions extends CsvConverters.CsvPipelineOptions, SplunkOptions {
+
+    @Description("Pattern of where to write errors, ex: gs://mybucket/somepath/errors.txt")
+    ValueProvider<String> getInvalidOutputPath();
```
zhoufek, line 96: Just `String`, not a `ValueProvider`, which is only for Classic Templates.

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+   */
+  public static PipelineResult run(GCSToSplunkOptions options) {
+    Pipeline pipeline = Pipeline.create(options);
+
+    CoderRegistry registry = pipeline.getCoderRegistry();
+    registry.registerCoderForClass(SplunkEvent.class, SplunkEventCoder.of());
+    registry.registerCoderForType(
+        FAILSAFE_ELEMENT_CODER.getEncodedTypeDescriptor(), FAILSAFE_ELEMENT_CODER);
+
+    /*
```
zhoufek, line 131: Just my personal opinion: I think this comment is unnecessary. Labels should be enough to clearly indicate what the step is doing, and if we have branches, so should the variables. If anything weird is done as part of a step, it should be commented closer to that weirdness, and these comments aren't great for templates that don't always produce the same graph or have a high branching factor.

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+     *  3) Convert successfully transformed messages into SplunkEvent objects
+     *  4) Write SplunkEvents to Splunk's HEC end point.
+     *  4a) Wrap write failures into a FailsafeElement.
+     *  5) Collect errors from UDF transform (#3), SplunkEvent transform (#4)
+     *     and writing to Splunk HEC (#5) and place into a GCS folder.
+     */
+
+    PCollectionTuple convertToEventTuple =
+        pipeline
+            // 1) Read CSV file(s) from Cloud Storage using {@link CsvConverters.ReadCsv}.
```
zhoufek, line 144: In line with the comment above, I think commenting the steps like this is unnecessary. We've applied a "ReadCsv" label to this step. Comments should be about what isn't obvious or documenting an API.

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+            .apply(
+                "ReadCsv",
```
zhoufek, line 146: Labels don't need to be one word, so this could be "Read CSV". Having spaces probably makes it more readable when rendered in Dataflow.

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+                    .setDelimiter(options.getDelimiter())
+                    .setHasHeaders(options.getContainsHeaders())
+                    .setInputFileSpec(options.getInputFileSpec())
+                    .setHeaderTag(CSV_HEADERS)
+                    .setLineTag(CSV_LINES)
+                    .setFileEncoding(options.getCsvFileEncoding())
+                    .build())
+
+            // 2) Apply user provided UDF (if any) on the input strings.
+            .apply(
```
zhoufek, line 158: If the user doesn't provide a UDF, then this step will still be part of the graph. That's confusing and wasteful. It's unavoidable with Classic Templates, but for Flex Templates, we should take advantage of the flexibility with the graph when it makes sense, so I'd only add this node when necessary.

An example template already adding UDFs only when necessary is [Pub/Sub Proto to BigQuery](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/pubsub-binary-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/PubsubProtoToBigQuery.java).

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+                "ConvertLine",
+                com.google.cloud.teleport.v2.transforms.CsvConverters.LineToFailsafeJson
```
zhoufek, line 160: Avoid using the package name. It is unnecessary.

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,253 @@
...
+            LogErrors.newBuilder()
+                .setErrorWritePath(options.getInvalidOutputPath().get())
+                .setErrorTag(COMBINED_ERRORS)
+                .build());
+
+    return pipeline.run();
+  }
+}
+
+class SplunkWriteErrorToFailsafeElementDoFn
```
zhoufek, line 216: These classes should be `private static` ones.

v2/gcs-to-splunk/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSplunkTest.java
```diff
@@ -0,0 +1,311 @@
...
+package com.google.cloud.teleport.v2.templates;
+
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.COMBINED_ERRORS;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.CSV_HEADERS;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.CSV_LINES;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.SPLUNK_EVENT_DEADLETTER_OUT;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.SPLUNK_EVENT_OUT;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.UDF_DEADLETTER_OUT;
+import static com.google.cloud.teleport.v2.templates.GCSToSplunk.UDF_OUT;
+import static org.hamcrest.CoreMatchers.equalTo;
```
zhoufek, line 25: Like other tests, please use Truth instead of Hamcrest.

v2/gcs-to-splunk/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSplunkTest.java
```diff
@@ -0,0 +1,311 @@
...
+    // Assert
+    PAssert.that(readCsvOut.get(UDF_OUT))
+        .satisfies(
+            collection -> {
+              FailsafeElement element = collection.iterator().next();
+              assertThat(element.getPayload(), is(equalTo(stringifiedJsonRecord)));
+              return null;
+            });
+
+    // Act
```
zhoufek, line 102: Tests should only have one Arrange, Act, and Assert block each. Please combine the two act and assert blocks into one. Same goes for other tests.

v2/gcs-to-splunk/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSplunkTest.java
```diff
@@ -0,0 +1,311 @@
...
+  private static final String TRANSFORM_FILE_PATH =
+      Resources.getResource("elasticUdf.js").getPath();
+  private static final String NO_HEADER_CSV_FILE_PATH =
+      Resources.getResource("no_header.csv").getPath();
+  private static final String HEADER_CSV_FILE_PATH =
+      Resources.getResource("with_headers.csv").getPath();
+  private static final String JSON_SCHEMA_FILE_PATH =
+      Resources.getResource("testSchema.json").getPath();
+
+  /** Tests the {@link GCSToSplunk} pipeline using a Udf to parse the Csv. */
```
zhoufek, line 70: Like elsewhere, please remove method comments for tests.

v2/gcs-to-splunk/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSplunkTest.java
```diff
@@ -0,0 +1,311 @@
...
+    PCollectionTuple flattenedErrors = getStringifiedFailsafeErrors(wrappedSplunkWriteErrors);
+
+    // Assert
+    PAssert.that(flattenedErrors.get(COMBINED_ERRORS)).containsInAnyOrder(stringifiedError);
+
+    //  Execute pipeline
+    pipeline.run();
+  }
+
+  private PCollectionTuple readFromCsvAndTransform(GCSToSplunkOptions options) {
```
zhoufek, line 251: It looks like this method was copied over from the production code. The pipeline code should be structured in a way to be tested directly. Requiring a copy/paste is error-prone and confusing for future developers. Same goes for the other cases.

v2/gcs-to-splunk/src/test/resources/elasticUdf.js
```diff
@@ -0,0 +1,36 @@
+/*
```
zhoufek, line 1: Please rename this file so that it no longer references Elasticsearch.

v2/gcs-to-splunk/src/test/resources/elasticUdf.js
```diff
@@ -0,0 +1,36 @@
...
+    var split = line.split(",");
+      var obj = new Object();
+      obj.id = split[0];
+      obj.state = split[1];
+      obj.price = parseFloat(split[2]);
+      var jsonString = JSON.stringify(obj);
+      return jsonString;
+
+  } catch (e) {
+    return false;
```
zhoufek, line 33: nit: Personally, I'd just let it error and be handled by the pipeline. It's a lot more natural to work with errors in Java than dynamic return types.

v2/gcs-to-splunk/src/test/resources/elasticUdf.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * This function splits a csv and turns it into a pre-defined JSON.
+ * @param {string} line is a line from TextIO.
+ * @return {JSON} a JSON created after parsing the line.
+ */
+function transform(line) {
+  try {
+    var split = line.split(",");
+      var obj = new Object();
```
zhoufek, line 25: Please use consistent indentation.

zhoufek (changes requested):
Sorry for all the comments. I know most of this was copied over from the Pub/Sub -> Splunk template, but some stuff has changed since, and I'd rather try to converge on consistent dependencies and styles rather than the current practice of everyone doing their own thing.

v2/common/pom.xml
```diff
@@ -113,6 +119,11 @@
             </exclusion>
           </exclusions>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
andreigurau: It's actually not, I had to import this manually. I moved the import to the `googlecloud-to-splunk` directory though if that's where we're going to keep all Splunk related classes

v2/common/src/main/java/com/google/cloud/teleport/v2/coders/SplunkEventCoder.java
```diff
@@ -0,0 +1,188 @@
...
+  private static final StringUtf8Coder STRING_UTF_8_CODER = StringUtf8Coder.of();
+  private static final NullableCoder<String> STRING_NULLABLE_CODER =
+      NullableCoder.of(STRING_UTF_8_CODER);
+  private static final NullableCoder<Long> LONG_NULLABLE_CODER =
+      NullableCoder.of(BigEndianLongCoder.of());
+
+  private static final Gson GSON = new Gson();
+
+  // Version markers must be >= 2.
+  private static final int VERSION_3 = 3;
```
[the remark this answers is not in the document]
andreigurau: To be honest, I'm not sure if my edits are what you were referring to, but I definitely ended up simplifying the code to only use one version (1 in this case), and removed all methods/tests that used another version

v2/gcs-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
andreigurau: I'm not sure how to validate a `PCollectionTuple`'s contents without accessing these `TupleTag`, so I used the `@ VisibleForTesting`annotation instead of marking them private.

v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,291 @@
...
+              .setLineTag(CSV_LINES)
+              .setUdfOutputTag(UDF_OUT)
+              .setUdfDeadletterTag(UDF_ERROR_OUT)
+              .build());
+    }
+
+    // For testing purposes, we need to do this check before creating the PTransform. Otherwise, we
+    // get a NullPointerException due to the PTransform not returning
+    // a value.
+    if (Strings.isNullOrEmpty(options.getJavascriptTextTransformFunctionName())) {
```
zhoufek, line 189: Sorry, missed this the last time. It seems that the `CsvConverters.LineToFailsafeJson` will handle all of the dynamic graph construction, so you can probably just call it with the options.

There is one bug [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/901278b958bfc0b154ca6f0a6ec34d6150ce7a78/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/CsvConverters.java#L295-L305), though, since the check allows empty strings (it shouldn't) and doesn't account for the function name not being set (it should). However, I'd accept that for now, since this isn't production-ready, and fix that in the next PR.

v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,291 @@
...
+            .setUdfFunctionName(options.getJavascriptTextTransformFunctionName())
+            .setJsonSchemaPath(options.getJsonSchemaPath())
+            .setHeaderTag(CSV_HEADERS)
+            .setLineTag(CSV_LINES)
+            .setUdfOutputTag(UDF_OUT)
+            .setUdfDeadletterTag(UDF_ERROR_OUT)
+            .build());
+  }
+
+  static PCollectionTuple convertToSplunkEvent(
```
zhoufek, line 209: I think it makes sense, if there's only one `apply`, to return the `PTransform` rather than a `PCollection` or `PCollectionTuple`. That'll make `run` a bit cleaner and potentially make pipeline construction more fluent. We also really only care about testing the `PTransform`, not the label, so nothing is lost in testing.

This also goes for `writeToSplunk`

v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java
```diff
@@ -0,0 +1,259 @@
...
+                            // have been parsed and populated in the SplunkEvent metadata.
+                            json.remove(METADATA_KEY);
+                            // If the event was not overridden in metadata above, use
+                            // the received JSON as event.
+                            if (event == null) {
+                              builder.withEvent(json.toString());
+                            }
+                          }
+
+                        } catch (IllegalStateException illegalStateException) {
```
zhoufek, line 231: Since we don't care about exceptions, just combine the like:

```
catch (IllegalStateException | JsonSyntaxException e) {
  // comment
}
```

zhoufek (changes requested):
Forgot this last time: Please remove "(rough draft)" from the title. We don't want that appearing in the commit history.

v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java
```diff
@@ -0,0 +1,255 @@
...
+  /** The tag for failed {@link SplunkEvent} conversion. */
+  @VisibleForTesting
+  static final TupleTag<FailsafeElement<String, String>> SPLUNK_EVENT_ERROR_OUT =
+      new TupleTag<FailsafeElement<String, String>>() {};
+
+  /** The tag for all the elements that failed. */
+  @VisibleForTesting static final TupleTag<String> COMBINED_ERRORS = new TupleTag<String>() {};
+
+  /** Logger for class. */
+  static final Logger LOG = LoggerFactory.getLogger(GCSToSplunk.class);
```
zhoufek, line 91: Please make this private.

v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java
```diff
@@ -0,0 +1,266 @@
...
+        TupleTag<FailsafeElement<String, String>> splunkEventErrorTag) {
+      this.splunkEventOutputTag = splunkEventOutputTag;
+      this.splunkEventErrorTag = splunkEventErrorTag;
+    }
+
+    @Override
+    public PCollectionTuple expand(PCollection<FailsafeElement<String, String>> input) {
+
+      return input.apply(
+          "ConvertToSplunkEvent",
```
zhoufek, line 152: nit: Please split this into multiple words.

zhoufek (approved):
Since googlecloud-to-splunk succeeds for build and unit tests, I'm going to ignore the unrelated failures and approve.

PR #392, opened 2022-05-24
Remove hive-to-bigquery template

zhoufek:
We stopped officially supporting this template months ago and have no intention to pick it back up, and it has started breaking our GitHub Actions. Time to just delete it.


zhoufek:
The test failure seems related to a known issue where a Datastream test doesn't work with certain timezones.

an2x (approved):

Issue #387, opened 2022-05-13
Hive Paritioning isn't working, the DD part of YYYY-MM-DD is day of year (133)


brooks-j:
### Related Template(s)

Cloud_PubSub_to_Avro

### What happened?

This is using Google's Dataflow.  

Here are some slices of the relevant terraform code.  This is the template being used.

template_gcs_path       = "gs://dataflow-templates/latest/Cloud_PubSub_to_Avro" 

And here is the date formatting.

outputDirectory   = "gs://${var.gcs_path}/${var.topic_name}/dt=YYYY-MM-DD"

Instead of getting dt=YYYY-MM-DD, I'm seeing this:  dt=2022-05-133

Note the day of year at the end of the date string.  

### Beam Version

2.35.0

pranavbhandari24:
Fixed

PR #395, opened 2022-05-25
Fixing problem with day of the year in WindowedFilenamePolicy

kwoz:
I noticed a bug exactly like #387. It looks like WindowedFilenamePolicy incorrectly uses DD default value which is the day of the year. This should correct it.


pranavbhandari24:
This fix has already been merged.

Issue #397, opened 2022-05-31
[Bug]: ElasticsearchIO does not properly respect getMaxBatchSizeBytes option


ggprod:
### Related Template(s)

BigQueryToElasticsearch, GCSToElasticsearch, PubsubToElasticsearch

### What happened?

when setting a max batch size bytes option that value is not respected and some batches can be larger.  This is particularly evident if setting a batch size close to the maximum allowed by Elasticsearch (when working with 7.17 this appears to be around 4080218931 bytes).. In this case some documents may exceed that size and then cause the job to fail with Elasticsearch errors that the batch size is too large (larger than the maximum 4080218931)

### Beam Version

Newer than 2.35.0

### Relevant log output

```shell
[...]

PR #398, opened 2022-05-31
fix bug where maxBatchSizeBytes may be exceeded

ggprod:
Resolves this isssue: https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/397

### DIFF
--- a/v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/utils/ElasticsearchIO.java
+++ b/v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/utils/ElasticsearchIO.java
@@ -1441,6 +1441,11 @@ public class ElasticsearchIO {
           }
         }
 
+        long newBatchSizeBytes = currentBatchSizeBytes + document.getBytes(StandardCharsets.UTF_8).length;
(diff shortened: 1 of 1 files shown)


v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/utils/ElasticsearchIO.java
```diff
@@ -1441,6 +1441,11 @@ public void processElement(ProcessContext context) throws Exception {
           }
         }
 
+        long newBatchSizeBytes = currentBatchSizeBytes + document.getBytes(StandardCharsets.UTF_8).length;
```
bvolpato, line 1444: Can you save `document.getBytes(StandardCharsets.UTF_8)` to a variable and reuse below (`currentBatchSizeBytes +=`)?

`String.getBytes` creates a new byte array every time it is called.

ggprod: yes, good idea, let me make that update

bvolpato (approved):

bvolpato (approved):

PR #400, opened 2022-06-06
Add README file for GCS-To-Splunk template

andreigurau:


andreigurau:
R: @zhoufek

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
+# GCS To Splunk Dataflow Template
+
+The [GCSToSplunk](../../src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java) pipeline ingests
+data from CSV files stored in GCS, optionally executes a UDF, converts the output to [SplunkEvent](https://github.com/apache/beam/blob/master/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkEvent.java) objects,
```
zhoufek, line 4: I don't think we need to describe converting to a `SplunkEvent`. This is more describing what it does, not how it does it, and the specific objects we use is more "how" rather than "what".

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
...
+* parallelism: Maximum number of parallel requests. Default: 1 (no parallelism).
+* disableCertificateValidation: Disable SSL certificate validation (true/false). Default: false (validation enabled). If true, the certificates are not validated (all certificates are trusted) and  `rootCaCertificatePath` parameter is ignored.
+
+The template has the following optional parameters:
+* delimiter: Delimiting character in CSV file(s). Default: use delimiter found in csvFormat
+* csvFormat: Csv format according to [Apache Commons CSV format](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html). Default is: Default
+  <br><b>NOTE:</b> Must match format names exactly found [here](http://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.Predefined.html#Default)
+* jsonSchemaPath: Path to JSON schema, ex gs://path/to/schema. Default: null.
+* javascriptTextTransformGcsPath: GCS path to Javascript UDF source. UDF will be preferred option for transformation if supplied. Default: null
+* javascriptTextTransformFunctionName: UDF Javascript Function Name. Default: null
```
zhoufek, line 221: What happens if the GCS path or function name is provided but not both?

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
...
+export SPLUNK_URL=<splunk-hec-url>
+export SPLUNK_TOKEN=<splunk-token>
+export BATCH_COUNT=<batch-count>
+export PARALLELISM=1
+
+gcloud config set project ${PROJECT}
+```
+* Build and push image to Google Container Repository
+```sh
+mvn clean package \
```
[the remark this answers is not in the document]
andreigurau: When I wrote this page, I went through all the commands I written, and it all worked for me. What suggests to you that this won't work?

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
+# GCS To Splunk Dataflow Template
+
+The [GCSToSplunk](../../src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java) pipeline ingests
+data from CSV files stored in GCS, optionally executes a UDF, converts the output to [SplunkEvent](https://github.com/apache/beam/blob/master/sdks/java/io/splunk/src/main/java/org/apache/beam/sdk/io/splunk/SplunkEvent.java) objects,
+and writes those records into Splunk's HEC endpoint.
+The template creates the SplunkEvent's payload as a JSON element using one of the following:
+1. Javascript UDF (if provided)
```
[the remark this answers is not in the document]
andreigurau: I kept the order as is (since that's the order where it gets prioritized), but clarified what happens when multiple of such inputs were present

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
...
+The template unit tests can be run using:
+```sh
+mvn test
+```
+
+### Executing Template
+
+The template requires the following parameters:
+* inputFileSpec: Pattern of where data lives, ex: gs://mybucket/somepath/*.csv
+* containsHeaders: Set to true or false if CSV file contains headers.
```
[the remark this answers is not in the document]
andreigurau: `containsHeaders` is a required parameter that defaults to `null` if it's not specified, which will also cause the pipeline to not execute if not specified

v2/googlecloud-to-splunk/docs/GCSToSplunk/gcs-to-splunk-image-spec.json
```diff
@@ -0,0 +1,128 @@
...
+        "isOptional": false,
+        "label": "GCS file pattern to search for CSV files",
+        "name": "inputFileSpec",
+        "paramType": "TEXT",
+        "regexes": [
+          "^gs:\\/\\/[^\\n\\r]+$"
+        ]
+      },
+      {
+        "helpText": "True if CSVs include headers or False if they do not.",
```
zhoufek, line 19: The help text implies "True" or "False" are values, but the regex wouldn't allow those. Please update this to valid values and mention the default. It should also make a note that all the files from `inputFileSpec` must hold to this option.

Note that everything here is what would appear in the UI if someone is going through the Cloud Console. It's arguably more important to be helpful here than in the README.

v2/googlecloud-to-splunk/docs/GCSToSplunk/gcs-to-splunk-image-spec.json
```diff
@@ -0,0 +1,128 @@
...
+        "isOptional": false,
+        "label": "GCS file pattern to search for CSV files",
+        "name": "containsHeaders",
+        "paramType": "TEXT",
+        "regexes": [
+          "^true|^false"
```
zhoufek, line 25: ^(true|false)$

v2/googlecloud-to-splunk/docs/GCSToSplunk/gcs-to-splunk-image-spec.json
```diff
@@ -0,0 +1,128 @@
...
+      },
+      {
+        "helpText": "The delimiter that the CSV uses. Example: ,",
+        "isOptional": true,
+        "label": "CSV delimiter",
+        "name": "delimiter",
+        "paramType": "TEXT"
+      },
+      {
+        "helpText": "Csv format according to [Apache Commons CSV format](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html). Default is: Default.<br><b>NOTE:</b> Must match format names exactly found [here](http://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.Predefined.html#Default)",
```
zhoufek, line 36: Please update this to match the README. I also don't think the Markdown link will render properly, so I'd remove it, but feel free to test this out manually. Same goes for other booleans.

zhoufek (changes requested):

v2/googlecloud-to-splunk/docs/GCSToSplunk/gcs-to-splunk-image-spec.json
andreigurau: The link doesn't render properly but I still think the URL is helpful for the end users, so I just removed the Markdown formatting from the JSON

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,237 @@
...
+* Maven
+* Cloud Storage bucket exists
+* Splunk HEC instance exists
+
+### Building Template
+This is a Flex Template meaning that the pipeline code will be containerized, and the container will be
+run on Dataflow.
+
+#### Building Container Image
+* Set environment variables that will be used in the build process.
```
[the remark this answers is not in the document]
andreigurau: To be honest, I'm still not exactly sure what I'm supposed to be looking, but I separated out the `gcloud config set project` command and bolded the header for each separate block. The thing is that other templates combine the `gcloud` command with the 'set environment variables' code block, which is why I tried to follow what the other templates had.

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,109 @@
+# GCS To Splunk Dataflow Template
+
+The [GCSToSplunk](../../src/main/java/com/google/cloud/teleport/v2/templates/GCSToSplunk.java) pipeline ingests
+data from CSV files stored in GCS, optionally executes a UDF,
+and writes those records into Splunk's HEC endpoint.
+
+The template creates the Splunk payload as a JSON element using one of the following:
+
+1. CSV headers* (default option, used only if none of the below two inputs were provided)
```
zhoufek, line 9: What is the asterisk for?

I also think "default" is enough. There's no need to say that it will be used if the other inputs aren't provided, since that's implied by "default". You can also probably remove the "(if provided)" on the other two for the same reason.

v2/googlecloud-to-splunk/docs/GCSToSplunk/README.md
```diff
@@ -0,0 +1,109 @@
...
+* Splunk HEC instance exists
+
+### Building Template
+
+This is a Flex Template meaning that the pipeline code will be containerized, and the container will be
+run on Dataflow.
+
+#### Building Container Image
+
+__Set environment variables that will be used in the build process.__
```
zhoufek, line 34: Section headers are identified by `#` symbols. It would look something like:

```
#### Set Environment Variables

...

#### Set Cloud Project

...

#### Build and Push Image
```

Also, I don't think setting the Cloud project is necessary, since no `gcloud` commands are run under `Building Template`, so you can leave that out. Whether or not to combine the other two sections is up to you.

zhoufek (changes requested):

zhoufek:
LGTM - Thanks for updating everything!

PR #411, opened 2022-06-23
[do not merge] Validating 2.40.0 release

pabloem:


pabloem:
@oleg-semenov I wonder if the failure here is normal or not

PR #412, opened 2022-06-23
Running all unit tests before throwing failure

pabloem:
This allows Java unit tests to run to completion, and only then fail. This grants us full information about how many tests are passing / failing / skipped.


pabloem:
See example of full unit test suite being run here: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/411

Specifically this flow: https://github.com/GoogleCloudPlatform/DataflowTemplates/runs/7033164469

pranavbhandari24 (approved):

PR #425, opened 2022-07-22
Created Spanner Resource Manager

ike-albert:
Created Spanner Resource Manager interface and implemented default client.

### DIFF
--- a/it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
@@ -1,0 +1,261 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.spanner;
+
+import static com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils.generateDatabaseId;
+import static com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils.generateInstanceId;
+import static com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils.generateNewId;
+
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.InstanceAdminClient;
+import com.google.cloud.spanner.InstanceConfigId;
+import com.google.cloud.spanner.InstanceId;
+import com.google.cloud.spanner.InstanceInfo;
+import com.google.cloud.spanner.KeySet;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.Struct;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import java.util.concurrent.ExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default class for implementation of {@link SpannerResourceManager} interface.
+ *
+ * <p>The class supports one instance, one database, and multiple tables per manager object. The
+ * instance and database are created when the first table is created.
+ *
+ * <p>The instance and database ids are formed using testId. The database id will be {testId}, with
+ * some extra formatting. The instance id will be "{testId}-{ISO8601 time, microsecond precision}",
+ * with additional formatting. Note: If testId is more than 30 characters, a new testId will be
+ * formed for naming: {first 21 chars of long testId} + “-” + {8 char hash of testId}.
+ *
+ * <p>The class is thread-safe.
+ */
+public final class DefaultSpannerResourceManager implements SpannerResourceManager {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultSpannerResourceManager.class);
+  private static final int MAX_BASE_ID_LENGTH = 30;
+
+  private boolean hasInstance = false;
+  private boolean hasDatabase = false;
+
+  private final String projectId;
+  private final String instanceId;
+  private final String databaseId;
+  private final String region;
+
+  private final Spanner spanner;
+  private final InstanceAdminClient instanceAdminClient;
+  private final DatabaseAdminClient databaseAdminClient;
+
+  private DefaultSpannerResourceManager(Builder builder) {
+    this(
+        SpannerOptions.newBuilder().setProjectId(builder.projectId).build().getService(),
+        builder.testId,
+        builder.projectId,
+        builder.region);
+  }
+
+  @VisibleForTesting
+  DefaultSpannerResourceManager(Spanner spanner, String testId, String projectId, String region) {
+    if (testId.length() > MAX_BASE_ID_LENGTH) {
+      testId = generateNewId(testId, MAX_BASE_ID_LENGTH);
+    }
+    this.projectId = projectId;
+    this.instanceId = generateInstanceId(testId);
+    this.databaseId = generateDatabaseId(testId);
+
+    this.region = region;
+    this.spanner = spanner;
+    this.instanceAdminClient = spanner.getInstanceAdminClient();
+    this.databaseAdminClient = spanner.getDatabaseAdminClient();
+  }
+
+  public static Builder builder(String testId, String projectId, String region) {
+    return new Builder(testId, projectId, region);
+  }
+
+  private synchronized void maybeCreateInstance() {
+    checkIsUsable();
+    if (hasInstance) {
+      return;
+    }
+    LOG.info("Creating instance {} in project {}.", instanceId, projectId);
+    InstanceInfo instanceInfo =
+        InstanceInfo.newBuilder(InstanceId.of(projectId, instanceId))
+            .setInstanceConfigId(InstanceConfigId.of(projectId, region))
+            .setDisplayName(instanceId)
+            .setNodeCount(1)
+            .build();
+    try {
+      instanceAdminClient.createInstance(instanceInfo).get();
+      hasInstance = true;
+      LOG.info("Successfully created instance {}.", instanceId);
+    } catch (ExecutionException | InterruptedException | SpannerException e) {
+      cleanupAll();
+      throw new SpannerResourceManagerException("Failed to create instance.", e);
+    }
+  }
+
+  private synchronized void maybeCreateDatabase() {
+    checkIsUsable();
+    if (hasDatabase) {
+      return;
+    }
+    LOG.info("Creating database {} in instance {}.", databaseId, instanceId);
+    try {
+      databaseAdminClient.createDatabase(instanceId, databaseId, ImmutableList.of()).get();
+      hasDatabase = true;
+      LOG.info("Successfully created database {}.", databaseId);
+    } catch (ExecutionException | InterruptedException | SpannerException e) {
+      cleanupAll();
+      throw new SpannerResourceManagerException("Failed to create database.", e);
+    }
+  }
+
+  private void checkIsUsable() throws IllegalStateException {
+    if (spanner.isClosed()) {
+      throw new IllegalStateException("Manager has cleaned up all resources and is unusable.");
+    }
+  }
+
+  private void checkHasInstanceAndDatabase() throws IllegalStateException {
+    if (!hasInstance) {
+      throw new IllegalStateException("There is no instance for manager to perform operation on.");
+    }
+    if (!hasDatabase) {
+      throw new IllegalStateException("There is no database for manager to perform operation on");
+    }
+  }
+
+  /** Creates an instance and database as well if there are none at the time of method call. */
+  @Override
+  public synchronized void createTable(String statement) throws IllegalStateException {
+    checkIsUsable();
+    maybeCreateInstance();
+    maybeCreateDatabase();
+
+    LOG.info("Creating table in database {} using statement '{}'.", databaseId, statement);
+    try {
+      databaseAdminClient
+          .updateDatabaseDdl(
+              instanceId, databaseId, ImmutableList.of(statement), /* operationId= */ null)
+          .get();
+      LOG.info("Successfully created table in database {}.", databaseId);
+    } catch (ExecutionException | InterruptedException | SpannerException e) {
+      throw new SpannerResourceManagerException("Failed to create table.", e);
+    }
+  }
+
+  @Override
+  public synchronized void write(Mutation tableRecord) throws IllegalStateException {
+    write(ImmutableList.of(tableRecord));
+  }
+
+  @Override
+  public synchronized void write(Iterable<Mutation> tableRecords) throws IllegalStateException {
+    checkIsUsable();
+    checkHasInstanceAndDatabase();
+
+    LOG.info("Sending {} mutations to {}.{}", Iterables.size(tableRecords), instanceId, databaseId);
+    try {
+      DatabaseClient databaseClient =
+          spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+      databaseClient.write(tableRecords);
+      LOG.info("Successfully sent mutations to {}.{}", instanceId, databaseId);
+    } catch (SpannerException e) {
+      throw new SpannerResourceManagerException("Failed to write mutations.", e);
+    }
+  }
+
+  @Override
+  public synchronized ImmutableList<Struct> readTableRecords(String tableId, String... columnNames)
+      throws IllegalStateException {
+    return readTableRecords(tableId, ImmutableList.copyOf(columnNames));
+  }
+
+  @Override
+  public synchronized ImmutableList<Struct> readTableRecords(
+      String tableId, Iterable<String> columnNames) throws IllegalStateException {
+    checkIsUsable();
+    checkHasInstanceAndDatabase();
+
+    LOG.info(
+        "Loading columns {} from {}.{}.{}",
+        Iterables.toString(columnNames),
+        instanceId,
+        databaseId,
+        tableId);
+    DatabaseClient databaseClient =
+        spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+
+    try (ResultSet resultSet =
+        databaseClient.singleUse().read(tableId, KeySet.all(), columnNames)) {
+      ImmutableList.Builder<Struct> tableRecordsBuilder = ImmutableList.builder();
+
+      while (resultSet.next()) {
+        tableRecordsBuilder.add(resultSet.getCurrentRowAsStruct());
+      }
+      ImmutableList<Struct> tableRecords = tableRecordsBuilder.build();
+      LOG.info(
+          "Loaded {} records from {}.{}.{}", tableRecords.size(), instanceId, databaseId, tableId);
+      return tableRecords;
+    } catch (SpannerException e) {
+      throw new SpannerResourceManagerException("Error occurred while reading table records.", e);
+    }
+  }
+
+  @Override
+  public synchronized void cleanupAll() {
+    LOG.info("Attempting to cleanup manager.");
+    try {
+      instanceAdminClient.deleteInstance(instanceId);
+      hasInstance = false;
+      hasDatabase = false;
+    } catch (SpannerException e) {
+      throw new SpannerResourceManagerException("Failed to delete instance.", e);
+    } finally {
+      spanner.close();
+    }
+    LOG.info("Manager successfully cleaned up.");
+  }
+
+  /** Builder for {@link DefaultSpannerResourceManager}. */
+  public static final class Builder {
+
+    private final String testId;
+    private final String projectId;
+    private final String region;
+
+    private Builder(String testId, String projectId, String region) {
+      this.testId = testId;
+      this.projectId = projectId;
+      this.region = region;
+    }
+
+    public DefaultSpannerResourceManager build() {
+      return new DefaultSpannerResourceManager(this);
+    }
+  }
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.spanner;
+
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.Struct;
+import com.google.common.collect.ImmutableList;
+
+/** Interface for managing spanner resources in integration tests. */
+public interface SpannerResourceManager {
+
+  /**
+   * Creates a table given a CREATE TABLE DDL statement.
+   *
+   * <p>Note: Implementations may do instance creation and database creation here.
+   *
+   * @param statement The CREATE TABLE DDL statement.
+   * @throws IllegalStateException if method is called after resources have been cleaned up.
+   */
+  void createTable(String statement) throws IllegalStateException;
+
+  /**
+   * Writes a given record into a table. This method requires {@link
+   * SpannerResourceManager#createTable(String)} to be called for the target table beforehand.
+   *
+   * @param tableRecord A mutation object representing the table record.
+   * @throws IllegalStateException if method is called after resources have been cleaned up or if
+   *     the manager object has no instance or database.
+   */
+  void write(Mutation tableRecord) throws IllegalStateException;
+
+  /**
+   * Writes a collection of table records into one or more tables. This method requires {@link
+   * SpannerResourceManager#createTable(String)} to be called for the target table beforehand.
+   *
+   * @param tableRecords A collection of mutation objects representing table records.
+   * @throws IllegalStateException if method is called after resources have been cleaned up or if
+   *     the manager object has no instance or database.
+   */
+  void write(Iterable<Mutation> tableRecords) throws IllegalStateException;
+
+  /**
+   * Reads all the rows in a table. This method requires {@link
+   * SpannerResourceManager#createTable(String)} to be called for the target table beforehand.
+   *
+   * @param tableId The id of the table to read rows from.
+   * @param columnNames The table's column names.
+   * @return A ResultSet object containing all the rows in the table.
+   * @throws IllegalStateException if method is called after resources have been cleaned up or if
+   *     the manager object has no instance or database.
+   */
+  ImmutableList<Struct> readTableRecords(String tableId, String... columnNames)
+      throws IllegalStateException;
+
+  /**
+   * Reads all the rows in a table.This method requires {@link
+   * SpannerResourceManager#createTable(String)} to be called for the target table beforehand.
+   *
+   * @param tableId The id of table to read rows from.
+   * @param columnNames A collection of the table's column names.
+   * @return A ResultSet object containing all the rows in the table.
+   * @throws IllegalStateException if method is called after resources have been cleaned up or if
+   *     the manager object has no instance or database.
+   */
+  ImmutableList<Struct> readTableRecords(String tableId, Iterable<String> columnNames)
+      throws IllegalStateException;
+
+  /**
+   * Deletes all created resources (instance, database, and tables) and cleans up all Spanner
+   * sessions, making the manager object unusable.
+   */
+  void cleanupAll();
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.spanner;
+
+import static com.google.common.hash.Hashing.goodFastHash;
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.base.CharMatcher;
+import com.google.common.hash.HashFunction;
+import com.google.re2j.Pattern;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Random;
+
+/**
+ * Utilities for {@link com.google.cloud.teleport.it.spanner.SpannerResourceManager}
+ * implementations.
+ */
+public final class SpannerResourceManagerUtils {
+  private static final DateTimeFormatter TIME_FORMAT =
+      DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
+  private static final Pattern ILLEGAL_DATABASE_CHARS = Pattern.compile("[\\W-]");
+  private static final Pattern ILLEGAL_INSTANCE_CHARS = Pattern.compile("[\\W_]");
+
+  private SpannerResourceManagerUtils() {}
+
+  /**
+   * Generates a new id string from an existing one.
+   *
+   * @param id The id string to generate a new id from.
+   * @param targetLength The length of the new id to generate. Must be greater than 8.
+   */
+  static String generateNewId(String id, int targetLength) {
+    if (id.length() <= targetLength) {
+      return id;
+    }
+
+    if (targetLength <= 8) {
+      throw new IllegalArgumentException("targetLength must be greater than 8");
+    }
+
+    HashFunction hashFunction = goodFastHash(32);
+    String hash = hashFunction.hashUnencodedChars(id).toString();
+    return id.substring(0, targetLength - hash.length() - 1) + "-" + hash;
+  }
+
+  /**
+   * Generates an instance id from a given string.
+   *
+   * @param baseString The string to generate the id from.
+   * @return The instance id string.
+   */
+  static String generateInstanceId(String baseString) {
+    checkArgument(baseString.length() != 0, "baseString cannot be empty!");
+    String illegalCharsRemoved =
+        ILLEGAL_INSTANCE_CHARS.matcher(baseString.toLowerCase()).replaceAll("-");
+
+    LocalDateTime localDateTime = LocalDateTime.now(ZoneId.of("UTC"));
+
+    // if first char is not a letter, replace with letter, so it doesn't violate spanner's instance
+    // naming rules
+    if (!Character.isLetter(baseString.charAt(0))) {
+      char padding = generatePadding();
+      illegalCharsRemoved = padding + illegalCharsRemoved.substring(1);
+    }
+    return illegalCharsRemoved + "-" + localDateTime.format(TIME_FORMAT);
+  }
+
+  /**
+   * Generates a database id from a given string.
+   *
+   * @param baseString The string to generate the id from.
+   * @return The database id string
+   */
+  static String generateDatabaseId(String baseString) {
+    checkArgument(baseString.length() != 0, "baseString cannot be empty!");
+    String illegalCharsRemoved =
+        ILLEGAL_DATABASE_CHARS.matcher(baseString.toLowerCase()).replaceAll("_");
+
+    // replace hyphen with underscore, so there's no need for backticks
+    String trimmed = CharMatcher.is('_').trimTrailingFrom(illegalCharsRemoved);
+
+    checkArgument(
+        trimmed.length() > 0,
+        "Database id is empty after removing illegal characters and trailing underscores");
+
+    // if first char is not a letter or last char is a digit, replace with letter, so it doesn't
+    // violate spanner's database naming rules
+    char padding = generatePadding();
+
+    if (Character.isDigit(trimmed.charAt(trimmed.length() - 1))) {
+      trimmed = trimmed.substring(0, trimmed.length() - 1) + padding;
+    }
+    if (!Character.isLetter(trimmed.charAt(0))) {
+      trimmed = padding + trimmed.substring(1);
+    }
+    return trimmed;
+  }
+
+  /** Helper method to generate random letter for padding. */
+  private static char generatePadding() {
+    Random random = new Random();
+    return (char) ('a' + random.nextInt(26));
+  }
+}
--- a/it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
+++ b/it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
@@ -1,0 +1,482 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.spanner;
+
+import static com.google.cloud.spanner.Value.int64;
+import static com.google.cloud.spanner.Value.string;
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Database;
+import com.google.cloud.spanner.Instance;
+import com.google.cloud.spanner.InstanceAdminClient;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.Struct;
+import com.google.common.collect.ImmutableList;
+import java.util.concurrent.ExecutionException;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Answers;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** Unit tests for {@link DefaultSpannerResourceManager}. */
+@RunWith(JUnit4.class)
+public final class DefaultSpannerResourceManagerTest {
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+  private Spanner spanner;
+
+  @Mock private Database database;
+  @Mock private Instance instance;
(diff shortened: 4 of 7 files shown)


ike-albert:
R: @zhoufek

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerManager.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  public static Builder builder(String testId, String projectId, String region) {
+    return new Builder(testId, projectId, region);
+  }
+
+  @Override
+  public void createTable(String statement) throws Exception {
+    if (spanner.isClosed()) {
+      throw new Exception("Manager has cleaned up all resources and cannot complete operation.");
```
ike-albert, line 72: Should i make specific exception class in another file, maybe SpannerManagerExceptions, instead of throwing a general exception?

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerManager.java
```diff
@@ -0,0 +1,200 @@
...
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+/** Default class for implementation of {@link SpannerResourceManager} interface. */
```
zhoufek, line 42: Please document when the instance is created. That's an expensive, state-affecting behavior that users may want to know about. It should also document that there's only one database and that all tables are under this one database. Again, that impacts the state.

It's also worth documenting what the instance and database ids are, since it doesn't exactly follow what is passed in the test id.

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.SpannerException;
+
+/** Interface for managing spanner resources in integration tests. */
+public interface SpannerResourceManager {
+
+  /**
+   * Creates a table given a CREATE TABLE DDL statement.
+   *
+   * @param statement The CREATE TABLE DDL statement. Example: "CREATE TABLE Singers (" + "SingerId
```
zhoufek, line 28: I don't think we need an example. That's just part of Spanner.

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
```diff
@@ -0,0 +1,70 @@
...
+   *     INT64 NOT NULL," + " FirstName STRING(1024)," + " LastName STRING(1024)," + " SingerInfo
+   *     BYTES(MAX)" + ") PRIMARY KEY (SingerId)"
+   */
+  void createTable(String statement) throws Exception;
+
+  /**
+   * Inserts a given record into a table.
+   *
+   * @param tableRecord A mutation object representing the table record.
+   * @throws SpannerException if there is an issue inserting the table record.
```
zhoufek, line 38: Once all the exceptions are cleaned up, please make sure that the `throws` comments and method signatures are in sync.

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Reads all the rows in a table.
+   *
+   * @param tableId The id of table to read rows from.
+   * @param columnNames A collection of the table's column names. @@return A ResultSet object
+   *     containing all the rows in the table.
+   */
+  ResultSet readTableRecords(String tableId, Iterable<String> columnNames) throws Exception;
+
+  /** Deletes all created resources (instance, database, and tables). */
```
zhoufek, line 68: It also cleans up all sessions and makes the object unusable.

zhoufek (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerManager.java
ike-albert: Document as comments? How?

zhoufek: Add it to the class comment.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+  private DatabaseClient databaseClient;
+
+  private DefaultSpannerResourceManager(Builder builder) {
+    this(builder.projectId, builder.instanceId, builder.databaseId);
+  }
+
+  @VisibleForTesting
+  DefaultSpannerResourceManager(String projectId, String instanceId, String databaseId) {
+    this.instanceId = InstanceId.of(projectId, instanceId);
+    this.databaseId = DatabaseId.of(projectId, instanceId, databaseId);
```
ike-albert, line 80: Having this here could cause the constructor to fail. If there is an invalid instance id or database id, InstanceId.of() and DatabaseId.of() will fail. The alternative is to make projectId a class property, that way, I can use DatabaseId.of() in createTable(). Should I or do I leave it as is?

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.concurrent.ExecutionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default class for implementation of {@link SpannerResourceManager} interface.
+ *
+ * <p>There are one instance, one database, and multiple tables per manager object. The tables are
```
zhoufek, line 49: Few comments on this paragraph:

1. I would rephrase that it it supports multiple tables, not that it necessarily will have multiple tables.
2. The comment on how instances, databases, and tables are structured is unnecessary. That's standard Spanner details. The reason for mentioning the number is so that users know what it can and can't be used for in a test.
3. Avoid giving implementation details. The instance being created when the object is is relevant, but we don't have to  specifically call out `build()` as the place that it happens, and we definitely shouldn't discuss *how* it's doing that. (Besides, `createInstance` isn't public.) Doc comments are about describing behavior, not implementation.
4. The comment on database creation should be added to the `createTable` method. Specifically, it only does so on the first call to `createTable`, so the state impact of later calls is different.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+ * under the database, and the database is under the instance. Instance creation is handled by
+ * {@link Builder#build()} using {@link DefaultSpannerResourceManager#createInstance(String,
+ * String)}. Database creation is triggered by the {@link
+ * DefaultSpannerResourceManager#createTable(String)} method, if there is no existing database.
+ *
+ * <p>The instance and database ids are formed using testId. The database id string will be
+ * {testId}, with some extra formatting. The instance id will be "{testId}-{ISO8601 time,
+ * microsecond precision}", with additional formatting.
+ *
+ * <p>Note: If testId is more than 30 characters, a new testId will be formed for naming: {first 22
```
zhoufek, line 59: This paragraph can be added to the last one.

You should also make a paragraph saying that this class is thread safe. That's the guarantee we're offering, but we normally assume classes aren't thread safe.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+      LOG.info("Successfully created instance.");
+    } catch (ExecutionException | InterruptedException e) {
+      throw new CreateInstanceException("Failed to create instance.", e);
+    }
+  }
+
+  @Override
+  public synchronized void createTable(String statement)
+      throws CreateDatabaseException, IllegalStateException, UpdateDatabaseException {
+    if (spanner.isClosed()) {
```
zhoufek, line 109: This check is being reused a lot and with an identical error message. It would be better to have a private method that does the check and throws the exception.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerManager.java
```diff
@@ -0,0 +1,200 @@
...
+
+  /** Builder for {@link DefaultSpannerManager}. */
+  public static final class Builder {
+    private static final int MAX_BASE_ID_LENGTH = 30;
+    private static final HashFunction BASE_ID_HASH_FUNCTION = Hashing.goodFastHash(32);
+
+    private final String testId;
+    private final String projectId;
+    private final String region;
+    private InstanceId instanceId;
```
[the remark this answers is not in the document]
zhoufek: This still hasn't been addressed.

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManager.java
```diff
@@ -0,0 +1,79 @@
...
+   * @param columnNames A collection of the table's column names.
+   * @return A ResultSet object containing all the rows in the table.
+   * @throws IllegalStateException if method is called after resources have been cleaned up.
+   */
+  ResultSet readTableRecords(String tableId, Iterable<String> columnNames)
+      throws IllegalStateException;
+
+  /**
+   * Deletes all created resources (instance, database, and tables) and cleans up all Spanner
+   * sessions, making the Spanner and manger objects unusable.
```
zhoufek, line 76: Don't mention the  Spanner object. That's an implementation detail. Users only care if the resource manager is unusable.

zhoufek (changes requested):
Please add the package-info.java file that's causing the build check to fail.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+    this.spanner = SpannerOptions.newBuilder().setProjectId(projectId).build().getService();
+    this.instanceAdminClient = spanner.getInstanceAdminClient();
+    this.databaseAdminClient = spanner.getDatabaseAdminClient();
+  }
+
+  public static Builder builder(String testId, String projectId, String region) {
+    return new Builder(testId, projectId, region);
+  }
+
+  @VisibleForTesting
```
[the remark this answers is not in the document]
ike-albert: It doesn't. I'll take note

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+      DateTimeFormatter timeFormat = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
+      LocalDateTime localDateTime = LocalDateTime.now(ZoneId.of("UTC"));
+      instanceId = baseIdString.replace("_", "-") + "-" + localDateTime.format(timeFormat);
+
+      // replace hyphen with underscore, so there's no need for backticks, then remove trailing
+      // underscores.
+      databaseId = CharMatcher.is('_').trimTrailingFrom(baseIdString.replace("-", "_"));
+
+      DefaultSpannerResourceManager manager = new DefaultSpannerResourceManager(this);
+      manager.createInstance(projectId, region);
```
an2x: I think the instance needs to be created lazily, similar to the database, only when someone calls `createTable` or `readTableRecords` forst time.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  @Override
+  public synchronized ResultSet readTableRecords(String tableId, Iterable<String> columnNames)
+      throws IllegalStateException {
+    if (spanner.isClosed()) {
+      throw new IllegalStateException(
+          "Manager has cleaned up all resources and cannot complete operation.");
+    }
+    return databaseClient.singleUse().read(tableId, KeySet.all(), columnNames);
```
an2x, line 159: Possible problem with the `databaseClient` not being initialized here, because AFAIU it's created only in the `createTable` method. If someone tries to read table records before calling `createTable` this will simply throw an NPE.

If reading a table before creating at least 1 is not supported, at least a better check/error message is needed.

But in fact I don't see why we should have this restriction. I think we can easily have a template that creates the target Spanner tables itself. So an integration test would need to read this data without calling `createTable` to verify the results.

an2x (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,219 @@
...
+      // Make hashing part a function, so I can test?
+      if (baseIdString.length() > MAX_BASE_ID_LENGTH) {
+        String hash = BASE_ID_HASH_FUNCTION.hashUnencodedChars(baseIdString).toString();
+        baseIdString =
+            baseIdString.substring(0, MAX_BASE_ID_LENGTH - hash.length() - 1) + "-" + hash;
+      }
+
+      DateTimeFormatter timeFormat = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
+      LocalDateTime localDateTime = LocalDateTime.now(ZoneId.of("UTC"));
+      instanceId = baseIdString.replace("_", "-") + "-" + localDateTime.format(timeFormat);
```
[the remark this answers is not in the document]
ike-albert: Noted

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,211 @@
...
+  @Override
+  public synchronized ResultSet readTableRecords(String tableId, Iterable<String> columnNames)
+      throws IllegalStateException {
+    checkIsCleanedUp();
+    return databaseClient.singleUse().read(tableId, KeySet.all(), columnNames);
+  }
+
+  @Override
+  public synchronized void cleanupAll() {
+    instanceAdminClient.deleteInstance(instanceId);
```
zhoufek, line 158: It's not really clear from the documentation how this will respond if the instance doesn't exist (i.e. `createInstance` failed or was never called in the first place). We should have a way of tracking if `createInstance` successfully created the instance. Basically, a boolean field that's false by default, gets set to true if `createInstance` succeeds, and is checked here before making a call.

zhoufek (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,211 @@
...
+      // replace hyphen with underscore, so there's no need for backticks, then remove trailing
+      // underscores.
+      databaseId = CharMatcher.is('_').trimTrailingFrom(baseIdString.replace('-', '_'));
+
+      if (spanner == null) {
+        spanner = SpannerOptions.newBuilder().setProjectId(projectId).build().getService();
+      }
+      DefaultSpannerResourceManager manager = new DefaultSpannerResourceManager(this);
+      manager.createInstance(region);
+      manager.createDatabase();
```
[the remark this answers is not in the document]
ike-albert: Noted!

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,244 @@
...
+
+    this.region = region;
+    this.spanner = spanner;
+    this.instanceAdminClient = spanner.getInstanceAdminClient();
+    this.databaseAdminClient = spanner.getDatabaseAdminClient();
+  }
+
+  public static Builder builder(String testId, String projectId, String region) {
+    checkArgument(testId.length() >= 2, "testId length must be at least 2!");
+    checkArgument(Character.isLetter(testId.charAt(0)), "testId must start with a letter!");
```
ike-albert, line 98: @zhoufek 

Should I static import isLetter() here and isDigit() in the Utils class since the names are self-explanatory?

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,244 @@
...
+  public synchronized void insertTableRecord(Mutation tableRecord) throws IllegalStateException {
+    insertTableRecords(ImmutableList.of(tableRecord));
+  }
+
+  @Override
+  public synchronized void insertTableRecords(Iterable<Mutation> tableRecords)
+      throws IllegalStateException {
+    String tableId = Iterables.get(tableRecords, 0).getTable();
+    LOG.info(
+        "Inserting {} records into table {}.{}.{}",
```
zhoufek, line 176: There's actually nothing about this implementation requires users to insert into a single table, and the interface's doc comment states that updating an existing record is possible.

I think we should just rename this to `write` and let users do whatever they intend to do. If we ever want more granular control, we can add that, but it's harder on our end and may not offer much benefit to usability. The log statement would update to something like: "Sending {} mutations to {}.{}" Basically, leave off the table and just make a generic reference to the number of mutations, not a specific type of mutation.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,244 @@
...
+        Iterables.size(tableRecords),
+        instanceId,
+        databaseId,
+        tableId);
+
+    checkIsUsable();
```
zhoufek, line 182: Maybe move these checks before the log? It's less likely to give off the impression the write was attempted.

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
```diff
@@ -0,0 +1,86 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Random;
+
+/**
+ * Utilities for {@link com.google.cloud.teleport.it.spanner.SpannerResourceManager}
+ * implementations.
+ */
+public final class SpannerResourceManagerUtils {
```
zhoufek, line 31: These utilities should be package-private. They only exist to support common resource manager tasks.

zhoufek (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
ike-albert: Okay

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,109 @@
...
+    assertThat(actual).isEqualTo(shortId);
+  }
+
+  @Test
+  public void testGenerateNewIdLongString() {
+    String longId = "long-test-id-string";
+
+    String actual = generateNewId(longId, 13);
+
+    assertThat(actual).isNotEqualTo(longId);
```
an2x, line 46: I think this is not an important check. Do we care if the generate id is equal to the base id or not? No, we don't, as long as it has the correct length and it's a valid Spanner ID.
I'd rather check that the length of the id is <= 13.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,109 @@
...
+import static com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils.generateNewId;
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils}. */
+@RunWith(JUnit4.class)
+public class SpannerResourceManagerUtilsTest {
```
an2x, line 29: What about test cases for other characters present in the input id?
For example, I imagine resource manager users may be passing the Java class name of the test as the testId, which may contain `.` or `$` chars, which are not allowed in Spanner IDs.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,311 @@
...
+    //   arrange
+    String statement =
+        "CREATE TABLE Singers (\n"
+            + "  SingerId   INT64 NOT NULL,\n"
+            + "  FirstName  STRING(1024),\n"
+            + "  LastName   STRING(1024),\n"
+            + ") PRIMARY KEY (SingerId)";
+
+    testManager.createTable(statement);
+    // act & assert
```
an2x, line 138: "act" here is `testManager.createTable`

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,311 @@
...
+            statementCaptor.capture(),
+            any());
+
+    String actualInstanceId = instanceIdCaptor.getValue();
+    String actualDatabaseId = databaseIdCaptor.getValue();
+    Iterable<String> actualStatement = statementCaptor.getValue();
+
+    assertThat(actualInstanceId).startsWith("test-id");
+    assertThat(actualInstanceId).isNotEqualTo("test-id");
+    assertThat(actualDatabaseId).isEqualTo("test_id");
```
an2x, line 152: The same can be verified as follows (a much shorter version):
```
    verify(spanner.getDatabaseAdminClient())
        .updateDatabaseDdl(
            argThat(s -> s.startsWith("test-id") && !s.equals("test-id")),
            eq("test_id"),
            eq(ImmutableList.of(statement)),
            any());
```
All static methods from org.mockito.ArgumentMatchers.

an2x (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,243 @@
...
+      throw new IllegalStateException("Manager has cleaned up all resources and is unusable.");
+    }
+  }
+
+  @Override
+  public synchronized void createTable(String statement) throws IllegalStateException {
+    checkIsUsable();
+    if (!hasInstance) {
+      createInstance();
+    }
```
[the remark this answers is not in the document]
ike-albert: The functions are separated because the database creation may fail while the instance creation succeeds. In that case, the manager will have an instance but no database.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
ike-albert: I have updated the logic for both generateInstanceId and generateDatabaseId to account for those. For other characters not allowed in Spanner IDs, I'll leave it to Spanner to throw an error

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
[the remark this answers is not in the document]
an2x: True, but the resource manager is unusable if any of these fail. If it has an instance but no database, and I call any other method, it won't work. So I think the approach should be "all or nothing". Create the instance AND the database. If both succeed, the resource manager is ready to handle other requests. If any of these 2 fail, throw an exception, cleanup the resources, etc. We can't run the integration test anyway. Right?

I'm OK with keeping this as separate methods, though I would still at least move the if check inside the method. Basically any new method in this resource manager that reads from or writes to the database will have to have this code block copy-pasted. So the smaller it is the better.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+
+    // act & assert
+    assertThrows(SpannerResourceManagerException.class, () -> testManager.createTable(statement));
+  }
+
+  @Test
+  public void testCreateTable_throwsExceptionWhenSpannerCreateDatabaseFails()
+      throws ExecutionException, InterruptedException {
+    // arrange
+    when(spanner.getDatabaseAdminClient().createDatabase(any(), any(), any()).get())
```
zhoufek, line 99: Make sure to explicitly set up successful mocks, such as for instance creation. You can create a helper method to handle this.

This goes for other tests as well.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+            + "  SingerId   INT64 NOT NULL,\n"
+            + "  FirstName  STRING(1024),\n"
+            + "  LastName   STRING(1024),\n"
+            + ") PRIMARY KEY (SingerId)";
+
+    // act
+    testManager.createTable(statement);
+
+    // assert
+    verify(spanner.getDatabaseAdminClient())
```
zhoufek, line 143: Use a captor and then use Truth to validate the captor's results. This:

1. Is more consistent with other tests (see: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/it/src/test/java/com/google/cloud/teleport/it/dataflow/DataflowOperatorTest.java)
2. Will create more consistent error messages to other assertions.
3. Will make this a lot easier to read. `updateDatabaseDdl(instanceCaptor.capture(), ...)` is way more clear than what is here. (Honestly, it took me a bit for me to figure out what these were, especially the one for the instance id.)

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+        .updateDatabaseDdl(
+            argThat(s -> s.startsWith("test-id") && !s.equals("test-id")),
```
zhoufek, line 145: This should be more specific, since what follows "test-id" is important. With Truth, you can provide a regex that also matches the rest of the string.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+            eq("test_id"),
```
zhoufek, line 146: I only know what "test_id" is because I know the implementation. Someone less familiar wouldn't know where this value came from. It's probably best for these base tests to have the test id not have any special characters and pass in the constant for matching. Later we can have tests specifically to make sure characters were replaced and introduce variables for just those tests.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+            .to("Richards")
+            .build();
+
+    // act
+    testManager.write(testMutation);
+
+    // assert
+    verify(spanner.getDatabaseClient(any())).write(writeMutationCaptor.capture());
+    Iterable<Mutation> actualWriteMutation = writeMutationCaptor.getValue();
+    assertThat(actualWriteMutation).isEqualTo(ImmutableList.of(testMutation));
```
zhoufek, line 172: For iterables, it's best to use `containsExactly` or `containsExactlyElementsIn`: https://javadoc.io/doc/com.google.truth/truth/latest/com/google/common/truth/IterableSubject.html

Some iterables won't play nice with `isEqualTo`, so if anything changes about the underlying implementation, this may start failing.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  @Test
+  public void testWriteSingleRecord_throwsExceptionWhenSpannerWriteFails() {
+    // arrange
+    when(spanner.getDatabaseClient(any()).write(any())).thenThrow(SpannerException.class);
+    testManager.createTable("");
+    // spotless:off
```
zhoufek, line 180: Not every mutation follows this, such as the one from the last test method. Please make this the standard (it's the cleanest).

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+import org.mockito.Answers;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** Unit tests for {@link DefaultSpannerResourceManager}. */
+@RunWith(JUnit4.class)
+public class DefaultSpannerResourceManagerTest {
```
zhoufek, line 52: Please mark all test classes as final.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @Test
+  public void testGenerateDatabaseId_replacesDigitLastCharWithLetter() {
+    String testBaseString = "test_database_0";
+
+    String actual = generateDatabaseId(testBaseString);
+
+    assertThat(actual).startsWith("test_database_");
```
zhoufek, line 37: Use `matches` and provide an appropriate regex. This will remove the need for the second assert.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    assertThat(actual).isEqualTo("test_database");
+  }
+
+  @Test
+  public void testGenerateDatabaseId_replacesNonLetterFirstCharWithLetter() {
+    String testBaseString = "0_test_database";
+
+    String actual = generateDatabaseId(testBaseString);
+
+    assertThat(actual.charAt(0))
```
zhoufek, line 77: Similar to above. Use `matches` with an appropriate regex.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    assertThat(actual).startsWith("test-instance");
+  }
+
+  @Test
+  public void testGenerateInstanceId_replacesNonLetterFirstCharWithLetter() {
+    String testBaseString = "0-test-instance";
+
+    String actual = generateInstanceId(testBaseString);
+
+    assertThat(actual.charAt(0))
```
zhoufek, line 135: Why not test the full string?

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    assertThat(actual).startsWith("test-instance");
+  }
+
+  @Test
+  public void testGenerateNewId_inputLengthLongerThanTargetLength() {
+    String longId = "long-test-id-string";
+
+    String actual = generateNewId(longId, 13);
+
+    assertThat(actual.length()).isAtMost(13);
```
zhoufek, line 165: This should also use a regex. We know that the value is the right length, but what if it has invalid characters?

zhoufek (changes requested):
Thanks for adding the thorough tests. Lots of non-test comments are very nit-picky at this point.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,312 @@
...
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+  private Spanner spanner;
+
+  @Mock(serializable = true)
```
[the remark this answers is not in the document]
ike-albert: I thought it did, but I was mistaken

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  @Test
+  public void testWriteSingleRecordShouldWorkWhenSpannerWriteSucceeds()
+      throws ExecutionException, InterruptedException {
+    // arrange
+    prepareCreateInstanceMock();
+    prepareCreateDatabaseMock();
+    prepareUpdateDatabaseMock();
+    testManager.createTable("");
```
ike-albert, line 194: Since prepareCreateInstanceMock() ... testManager.createTable("")  is repeated from here on, can I move it to setup and have methods that need a different behavior(e.g. throw error for createInstance)  redefine it?

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,364 @@
...
+            + "  SingerId   INT64 NOT NULL,\n"
+            + "  FirstName  STRING(1024),\n"
+            + "  LastName   STRING(1024),\n"
+            + ") PRIMARY KEY (SingerId)";
+
+    // act
+    testManager.createTable(statement);
+
+    // assert
+    verify(spanner.getDatabaseAdminClient(), times(2))
```
zhoufek, line 176: Are we really calling this twice? I thought that we only called `updateDatabaseDdl` once when creating the table. What's the second one for?

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,364 @@
...
+            instanceIdCaptor.capture(),
+            databaseIdCaptor.capture(),
+            statementCaptor.capture(),
+            any());
+
+    String actualInstanceId = instanceIdCaptor.getValue();
+    String actualDatabaseId = databaseIdCaptor.getValue();
+    Iterable<String> actualStatement = statementCaptor.getValue();
+
+    assertThat(actualInstanceId).startsWith(TEST_ID);
```
zhoufek, line 187: As mentioned before, could we use a regex to match the instance id? That'll be clearer and more complete than `startsWith` and `isNotEqual`.

https://truth.dev/api/0.41/com/google/common/truth/StringSubject.html#matches-java.lang.String-

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,364 @@
...
+    testManager = new DefaultSpannerResourceManager(spanner, TEST_ID, PROJECT_ID, REGION);
+
+    // act
+    testManager.cleanupAll();
+
+    // assert
+    verify(spanner.getInstanceAdminClient()).deleteInstance(any());
+    verify(spanner).close();
+  }
+}
```
zhoufek, line 364: Sorry, missed this last time. Could we add tests for what happens when we try to use this after `cleanupAll` is called?

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,166 @@
...
+    assertThat(actual).isEqualTo("test_database");
+  }
+
+  @Test
+  public void testGenerateInstanceId_replacesDollarSignWithHyphen() {
+    String testBaseString = "test$instance";
+
+    String actual = generateInstanceId(testBaseString);
+
+    assertThat(actual).matches("test-instance-[0-9]{8}-[0-9]{6}-[0-9]{6}");
```
zhoufek, line 109: nit: `\d` is a cleaner way of matching any number than `[0-9]`. Same goes for other uses.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link com.google.cloud.teleport.it.spanner.SpannerResourceManagerUtils}. */
+@RunWith(JUnit4.class)
+public final class SpannerResourceManagerUtilsTest {
+
+  @Test
+  public void testGenerateDatabaseId_replacesDigitLastCharWithLetter() {
```
zhoufek, line 32: Could you update the test names to be consistent with `DefaultSpannerResourceManagerTest`?

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final DateTimeFormatter TIME_FORMAT =
+      DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
+
+  private SpannerResourceManagerUtils() {}
+
+  /**
+   * Generates a new id string from an existing one.
+   *
+   * @param id The id string to generate a new id from.
+   * @param targetLength The length of the new id to generate. Must be greater than 8.
```
zhoufek, line 43: Sorry, missed this: It doesn't look like the restriction on being greater than eight is ever enforced.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  @Test
+  public void testGenerateNewId_inputLengthLongerThanTargetLength() {
+    String longId = "long-test-id-string";
+
+    String actual = generateNewId(longId, 13);
+
+    assertThat(actual.length()).isAtMost(13);
+    assertThat(actual).matches("(\\w|-)*");
```
zhoufek, line 155: Could we make this regex clearer? Don't we know what the start of the string is as well as what pattern the hash should match?

zhoufek (changes requested):

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * Generates an instance id from a given string.
+   *
+   * @param baseString The string to generate the id from.
+   * @return The instance id string.
+   */
+  static String generateInstanceId(String baseString) {
+    Pattern illegalCharsPattern = Pattern.compile("\\.|\\$|_");
```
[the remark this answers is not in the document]
ike-albert: Noted!

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
ike-albert: The second is when we mock updateDatabaseDdl.get() in prepareUpdateDatabaseMock()

it/src/main/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtils.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Random;
+
+/**
+ * Utilities for {@link com.google.cloud.teleport.it.spanner.SpannerResourceManager}
+ * implementations.
+ */
+public final class SpannerResourceManagerUtils {
+  private static final DateTimeFormatter TIME_FORMAT =
+      DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
+  private static final Pattern ILLEGAL_DATABASE_CHARS = Pattern.compile("[\\W-]");
```
ike-albert, line 36: Can I do this since it takes care of all illegal characters at once or should I leave it as it was and account for only '$', '-' and '.'?

zhoufek: You can leave it as it is. I think this will cover most, if not all, illegal characters test authors may actually use. I don't think the other characters are necessary. Most likely, someone chooses a delimiter between words that isn't allowed. I think '$' and '.' could be signs of other problems with the test.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
zhoufek: Ok. Could you verify the other mocks as well? All of them are changing external state that will impact future calls, so it'll be good to know that the instance and database were created correctly. This seems to only verify that creating a table was done correctly.

I think doing it in this test is enough.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,395 @@
...
+            .to("Richards")
+            .build();
+
+    // act & assert
+    assertThrows(IllegalStateException.class, () -> testManager.createTable(statement));
+    assertThrows(
+        IllegalStateException.class, () -> testManager.readTableRecords("Singers", "SingerId"));
+    assertThrows(IllegalStateException.class, () -> testManager.write(testMutation));
+  }
+
```
zhoufek, line 378: Maybe I missed it, but could you also add tests for when a user tries to read or write before calling `createTable`?

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
```diff
@@ -0,0 +1,395 @@
...
+    when(spanner.isClosed()).thenReturn(true);
+    testManager = new DefaultSpannerResourceManager(spanner, TEST_ID, PROJECT_ID, REGION);
+    testManager.cleanupAll();
+    String statement =
+        "CREATE TABLE Singers (\n"
+            + "  SingerId   INT64 NOT NULL,\n"
+            + "  FirstName  STRING(1024),\n"
+            + "  LastName   STRING(1024),\n"
+            + ") PRIMARY KEY (SingerId)";
+    Mutation testMutation =
```
zhoufek, line 362: Please make this consistent with the other mutations. That'll require disabling spotless similarly to what was done with them.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
zhoufek: Please also remove the check on the length and make it part of the regex. Regexes support an exact count (`{N}`) or a range (`{M,N}`), and since the prefix ("long-") is fixed, we know how long the rest of it should be.

it/src/test/java/com/google/cloud/teleport/it/spanner/SpannerResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  @Test
+  public void testGenerateNewIdShouldReturnNewIdWhenInputLengthIsLongerThanTargetLength() {
+    String longId = "long-test-id-string";
+
+    String actual = generateNewId(longId, 13);
+
+    assertThat(actual.length()).isAtMost(13);
+    assertThat(actual).matches("long-([a-zA-Z]|\\d)*");
```
zhoufek, line 177: This could be simplified a bit: `"long-[a-zA-Z\\d]{8}"` (This assumes the length is always 13.)

You can also do `"long-[a-zA-Z0-9]{8}"` if you'd rather stay consistent with using ranges, since the `\\d` shorthand breaks from that.

zhoufek (changes requested):

an2x (approved):

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,263 @@
...
+
+    LOG.info(
+        "Loading columns {} from {}.{}.{}",
+        Iterables.toString(columnNames),
+        instanceId,
+        databaseId,
+        tableId);
+
+    ImmutableList.Builder<Struct> tableRecordsBuilder = ImmutableList.builder();
+    ImmutableList<Struct> tableRecords;
```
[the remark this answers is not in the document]
ike-albert: Noted!

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -0,0 +1,262 @@
...
+        instanceId,
+        databaseId,
+        tableId);
+    DatabaseClient databaseClient =
+        spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+
+    try (ResultSet resultSet =
+        databaseClient.singleUse().read(tableId, KeySet.all(), columnNames)) {
+      ImmutableList.Builder<Struct> tableRecordsBuilder = ImmutableList.builder();
+      ImmutableList<Struct> tableRecords;
```
zhoufek, line 216: I meant below to have: `ImmutableList<Struct> tableRecords = tableRecordsBuilder.build();`

This variable serves no purpose until then.

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
zhoufek: Could you add it for all the read/write methods, not just one of each? 

I know that these methods call the others, but tests shouldn't be written as if we know exactly how each method is implemented. In theory, each method could have its own unique implementation without calling the other.

zhoufek (changes requested):

it/src/test/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManagerTest.java
zhoufek: There still needs to be one for `readTableRecords(String, Iterable<String>)`. It looks like there's only one for `readTableRecords(String, String...)`. Feel free to just add a similar assertion to the same test method.

zhoufek (changes requested):
Only a minor change now. After you address it, could you also squash the commits. Looking at the commits, you can squash them all down into one that has the same message as the first commit in this PR.

an2x (approved):

PR #427, opened 2022-07-26
Updated static to parallel processing

theshanbhag:
Hi Nick. I have updated my templates to imporve the processing efficiency of the templates. I will add the Regex files on the same PR

### DIFF
--- a/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
+++ b/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
@@ -19,9 +19,6 @@ import com.google.api.services.bigquery.model.TableRow;
 import com.google.api.services.bigquery.model.TableSchema;
 import com.google.cloud.teleport.v2.mongodb.options.MongoDbToBigQueryOptions.BigQueryWriteOptions;
 import com.google.cloud.teleport.v2.mongodb.options.MongoDbToBigQueryOptions.MongoDbOptions;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
 import org.apache.beam.sdk.io.mongodb.MongoDbIO;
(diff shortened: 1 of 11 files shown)


v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
```diff
@@ -77,11 +74,37 @@ public static boolean run(Options options) {
...
-            MapElements.via(
-                new SimpleFunction<Document, TableRow>() {
-                  @Override
-                  public TableRow apply(Document document) {
-                    return MongoDbUtils.generateTableRow(document, userOption);
+            ParDo.of(
+                new DoFn<Document, TableRow>() {
+
+                  @ProcessElement
+                  public void process(ProcessContext c) {
```
an2x, line 81: This code seems to be duplicated in MongoDbToBigQueryCdc. I suggest moving it to the Utils class or something like this, e.g. convert the input Document to Map here, and the input String to Map in MongoDbToBigQueryCdc, and then convert Map->Row via a common utils method.

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
```diff
@@ -77,11 +74,37 @@ public static boolean run(Options options) {
...
+                              case "java.util.Integer":
+                                row.set(key, value);
+                                break;
+                              default:
+                                row.set(key, value.toString());
+                            }
+                          });
+                    } else {
+                      DateTimeFormatter timeFormat =
+                          DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
```
an2x, line 101: This will create a new instance of DateTimeFormatter for each element, resulting in low performance. I believe the DateTimeFormatter instance can be made static final as it's immutable and thread-safe.

an2x (changes requested):

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
theshanbhag: Hi Nick, This method serves different purpose and takes and returns different parameters from that of the MongoDbToBigQueryCdc template.

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
theshanbhag: I have moved the mapping part to utils to a static function and made the DateTimeFormatter as static final.

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbUtils.java
```diff
@@ -76,4 +84,56 @@ public static Document getMongoDbDocument(String uri, String dbName, String coll
...
+    } else {
+      LocalDateTime localdate = LocalDateTime.now(ZoneId.of("UTC"));
+      row.set("id", parsedMap.get("_id").toString())
+          .set("source_data", parsedMap.toString())
+          .set("timestamp", localdate.format(TIMEFORMAT));
+    }
+    return row;
+  }
+
+  public static TableRow getTableSchema(Document document, String userOption) {
```
an2x, line 114: This method seems to be doing the same thing as `getTableSchemaCDC(HashMap<String, Object> parsedMap, String userOption)`, and `Document` implements `Map<String, Object>`, so a document can be passed to the `getTableSchemaCDC` method too.

Maybe remove this method and use the one that accepts a Map? And then rename it from getTableSchemaCDC to getTableSchema or something more generic.

an2x (changes requested):

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbUtils.java
theshanbhag: Updated. Removed the method getTableSchema and renamed getTableSchemaCDC to getTableSchema.

an2x (approved):

an2x (approved):

PR #430, opened 2022-08-01
Add manual workflow to create Releases.

pranavbhandari24:
The Dataflow templates team creates GitHub releases manually every Monday. This GitHub workflow can be used to automate part of this process and make it simpler for the team. 

This workflow allows inputs of release branch name, tag name, branch commit, and any cherry pick commits to include. It creates a release branch with the branch commit and cherry-picks, and tags the latest commit of the release branch. It also makes a release with an empty notes section for the team to fill out. ( A future improvement can be to create the notes automatically).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,75 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      candidateName:
+        description: 'Name of the candidate'
+        type: string
+        required: true
+      branchCommit:
+        description: 'Commit to check out. If it is the most recent commit then leave blank.'
+        type: string
+        required: false
+        default: ''
+      cherrypickCommits:
+        description: 'Comma separated commits to cherry pick'
+        type: string
+        required: false
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get releaser identity
+        run: |
+          git config --global user.name '${{github.actor}}'
+          git config --global user.email '${{github.actor}}@users.noreply.github.com'
+      - name: Declare release branch name and tag name
+        id: variables
+        run: |
+          echo "::set-output name=releaseBranchName::release_${CANDIDATE_NAME,,}"
+          echo "::set-output name=tagName::${CANDIDATE_NAME^^}"
+        env:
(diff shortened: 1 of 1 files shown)


pranavbhandari24:
@an2x I created a GitHub workflow which can be used to reduce the time taken to create the releases on github. Please review.

pranavbhandari24:
Currently, the workflow can only be triggered manually, but we can schedule it each Monday if that works better.

pranavbhandari24:
@zhoufek

zhoufek (dismissed):
Thanks! This should really reduce some of the release toil!

zhoufek:
LGTM - @damccorm just to verify.

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Create release branch
+        run: |
+          if [[ -z "${{ inputs.branchCommit }}" ]]; then
+            git checkout -b $RELEASE_BRANCH_NAME
+          else
+            git checkout -b $RELEASE_BRANCH_NAME $BRANCH_COMMIT
+          fi
```
damccorm, line 52:
```suggestion
          git checkout -b $RELEASE_BRANCH_NAME $BRANCH_COMMIT
```

I think this can be simplified since in the case where we enter the if, $BRANCH_COMMIT is '' anyways

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+  contents: write
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Monday date
+        id: date
+        run: echo "::set-output name=date::$(date -dmonday +'%Y-%m-%d')"
```
damccorm, line 30: Optional - I think this would be simpler if you just inline it where you use it. My rule of thumb is that if you're not reusing it, its probably not worth making it an output vs just inlining. I think there are a few other instances of single use output variables that I'd just inline.

That's mostly personal stylistic/readability related, so take it or leave it.

damccorm (dismissed):
Left some optional nits, overall LGTM though

.github/workflows/release.yml
pranavbhandari24: Made changes accordingly

damccorm (approved):
LGTM

PR #433, opened 2022-08-04
Fix issue with null error messages in BulkDecompressor

zhoufek:
If the `IOException` returns null for `getMessage`, which may happen with `EOFException`, this will now return an empty string rather than letting the null get returned to a `KV` that isn't set up to handle nulls.

Also ran spotless, since there's been a lot of changes to Classic Templates that violate spotless.


pranavbhandari24 (approved):
LGTM!

PR #436, opened 2022-08-11
fix special characters in dataset name

marengaz:
google have a [much more restrictive set of characters that are allowed for dataset names](https://cloud.google.com/bigquery/docs/datasets#:~:text=Letters%20(uppercase%20or%20lowercase)%2C%20numbers%2C%20and%20underscores) than tables. in our case, the source mysql database had a `-` in its name. this is not allowed in a bq dataset. 

this pr replaces invalid characters with an `_`


bvolpato:
I think this looks good (non-binding). Maybe we could just keep the link https://cloud.google.com/bigquery/docs/datasets#dataset-naming in the JavaDoc for future reference?

marengaz:
@bvolpato - no probs, done!

PR #445, opened 2022-08-19
Issue 442 - Create Bigtable Resource Manager

Polber:
Signed-off-by: Jeff Kinard <jkinard@google.com>

### DIFF
--- a/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.bigtable;
+
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import com.google.common.collect.ImmutableList;
+import org.threeten.bp.Duration;
+
+/** Interface for managing bigtable resources in integration tests. */
+public interface BigtableResourceManager {
+
+  /**
+   * Creates a Bigtable instance in which all clusters, nodes and tables will exist.
+   *
+   * @param clusters Collection of BigtableResourceManagerCluster objects to associate with the
+   *     given Bigtable instance.
+   * @throws BigtableResourceManagerException if there is an error creating the instance in
+   *     Bigtable.
+   */
+  void createInstance(Iterable<BigtableResourceManagerCluster> clusters);
+
+  /**
+   * Creates a table within the current instance given a table ID and a collection of column family
+   * names.
+   *
+   * <p>The columns in this table will be automatically garbage collected after one hour.
+   *
+   * <p>Note: Implementations may do instance creation here, if one does not already exist.
+   *
+   * @param tableId The id of the table.
+   * @param columnFamilies A collection of column family names for the table.
+   * @throws BigtableResourceManagerException if there is an error creating the table in Bigtable.
+   */
+  void createTable(String tableId, Iterable<String> columnFamilies);
+
+  /**
+   * Creates a table within the current instance given a table ID and a collection of column family
+   * names.
+   *
+   * <p>The columns in this table will be automatically garbage collected once they reach the age
+   * specified by {@code maxAge}.
+   *
+   * <p>Note: Implementations may do instance creation here, if one does not already exist.
+   *
+   * @param tableId The id of the table.
+   * @param columnFamilies A collection of column family names for the table.
+   * @param maxAge Sets the maximum age the columns can persist before being garbage collected.
+   * @throws BigtableResourceManagerException if there is an error creating the table in Bigtable.
+   */
+  void createTable(String tableId, Iterable<String> columnFamilies, Duration maxAge);
+
+  /**
+   * Writes a given row into a table. This method requires {@link
+   * BigtableResourceManager#createTable(String, Iterable)} to be called for the target table
+   * beforehand.
+   *
+   * @param tableRow A mutation object representing the table row.
+   * @throws BigtableResourceManagerException if method is called after resources have been cleaned
+   *     up, if the manager object has no instance, if the table does not exist or if there is an
+   *     IOException when attempting to retrieve the bigtable data client.
+   */
+  void write(RowMutation tableRow);
+
+  /**
+   * Writes a collection of table rows into one or more tables. This method requires {@link
+   * BigtableResourceManager#createTable(String, Iterable)} to be called for the target table
+   * beforehand.
+   *
+   * @param tableRows A collection of mutation objects representing table rows.
+   * @throws BigtableResourceManagerException if method is called after resources have been cleaned
+   *     up, if the manager object has no instance, if the table does not exist or if there is an
+   *     IOException when attempting to retrieve the bigtable data client.
+   */
+  void write(Iterable<RowMutation> tableRows);
+
+  /**
+   * Reads all the rows in a table. This method requires {@link
+   * BigtableResourceManager#createTable(String, Iterable)} to be called for the target table
+   * beforehand.
+   *
+   * @param tableId The id of table to read rows from.
+   * @return A List object containing all the rows in the table.
+   * @throws BigtableResourceManagerException if method is called after resources have been cleaned
+   *     up, if the manager object has no instance, if the table does not exist or if there is an
+   *     IOException when attempting to retrieve the bigtable data client.
+   */
+  ImmutableList<Row> readTable(String tableId);
+
+  /**
+   * Deletes all created resources (instance and tables) and cleans up all Bigtable clients, making
+   * the manager object unusable.
+   *
+   * @throws BigtableResourceManagerException if there is an error deleting the instance or tables
+   *     in Bigtable.
+   */
+  void cleanupAll();
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerCluster.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerCluster.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.bigtable;
+
+import com.google.auto.value.AutoValue;
+import com.google.cloud.bigtable.admin.v2.models.StorageType;
+
+/**
+ * Class for storing the metadata of a Bigtable cluster object.
+ *
+ * <p>A cluster belongs to a single bigtable instance and represents the service in a given zone. A
+ * cluster can have multiple nodes operating on the data. The cluster also has a storage type of
+ * either SSD or HDD depending on the user's needs.
+ */
+@AutoValue
+public abstract class BigtableResourceManagerCluster {
+
+  public static BigtableResourceManagerCluster create(
+      String clusterId, String zone, int numNodes, StorageType storageType) {
+    return new AutoValue_BigtableResourceManagerCluster(clusterId, zone, numNodes, storageType);
+  }
+
+  /**
+   * Returns the cluster ID of the Bigtable cluster object.
+   *
+   * @return the ID of the Bigtable cluster.
+   */
+  public abstract String clusterId();
+
+  /**
+   * Returns the operating zone of the Bigtable cluster object.
+   *
+   * @return the zone of the Bigtable cluster.
+   */
+  public abstract String zone();
+
+  /**
+   * Returns the number of nodes the Bigtable cluster object should be configured with.
+   *
+   * @return the number of nodes for the Bigtable cluster.
+   */
+  public abstract int numNodes();
+
+  /**
+   * Returns the type of storage the Bigtable cluster object should be configured with (SSD or HDD).
+   *
+   * @return the storage type of the Bigtable cluster.
+   */
+  public abstract StorageType storageType();
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.bigtable;
+
+import static com.google.cloud.teleport.it.common.ResourceManagerUtils.generateResourceId;
+
+import com.google.cloud.bigtable.admin.v2.models.StorageType;
+import com.google.common.collect.ImmutableList;
+import com.google.re2j.Pattern;
+
+/**
+ * Utilities for {@link com.google.cloud.teleport.it.bigtable.BigtableResourceManager}
+ * implementations.
+ */
+public final class BigtableResourceManagerUtils {
+
+  private static final int MAX_CLUSTER_ID_LENGTH = 30;
+  private static final Pattern ILLEGAL_CLUSTER_CHARS = Pattern.compile("[^a-z0-9-]");
+  private static final String REPLACE_CLUSTER_CHAR = "-";
+  public static final int MAX_INSTANCE_ID_LENGTH = 30;
+  private static final Pattern ILLEGAL_INSTANCE_ID_CHARS = Pattern.compile("[^a-z0-9-]");
+  private static final String REPLACE_INSTANCE_ID_CHAR = "-";
+  private static final int MIN_TABLE_ID_LENGTH = 1;
+  private static final int MAX_TABLE_ID_LENGTH = 30;
+  private static final Pattern ILLEGAL_TABLE_CHARS = Pattern.compile("[^a-zA-Z0-9-_.]");
+
+  private BigtableResourceManagerUtils() {}
+
+  /**
+   * Generates a collection that contains a single BigtableResourceManagerCluster object using the
+   * given parameters.
+   *
+   * @param baseString the test id to associate the bigtable cluster to.
+   * @param zone the zone/region that the cluster will be deployed to.
+   * @param numNodes the number of nodes that the cluster will contain.
+   * @param storageType the type of storage to configure the cluster with (SSD or HDD).
+   * @return Collection containing a single BigtableResourceManagerCluster object.
+   */
+  static Iterable<BigtableResourceManagerCluster> generateDefaultClusters(
+      String baseString, String zone, int numNodes, StorageType storageType) {
+
+    String clusterId =
+        generateResourceId(
+            baseString, ILLEGAL_CLUSTER_CHARS, REPLACE_CLUSTER_CHAR, MAX_CLUSTER_ID_LENGTH);
+    BigtableResourceManagerCluster cluster =
+        BigtableResourceManagerCluster.create(clusterId, zone, numNodes, storageType);
+
+    return ImmutableList.of(cluster);
+  }
+
+  /**
+   * Generates an instance id from a given string.
+   *
+   * @param baseString The string to generate the id from.
+   * @return The instance id string.
+   */
+  static String generateInstanceId(String baseString) {
+    return generateResourceId(
+        baseString, ILLEGAL_INSTANCE_ID_CHARS, REPLACE_INSTANCE_ID_CHAR, MAX_INSTANCE_ID_LENGTH);
+  }
+
+  /**
+   * Checks whether the given table ID is valid according to GCP constraints.
+   *
+   * @param idToCheck the table ID to check.
+   * @throws IllegalArgumentException if the table ID is invalid.
+   */
+  static void checkValidTableId(String idToCheck) {
+    if (idToCheck.length() < MIN_TABLE_ID_LENGTH) {
+      throw new IllegalArgumentException("Table ID " + idToCheck + " cannot be empty.");
+    }
+    if (idToCheck.length() > MAX_TABLE_ID_LENGTH) {
+      throw new IllegalArgumentException(
+          "Table ID "
+              + idToCheck
+              + " cannot be longer than "
+              + MAX_TABLE_ID_LENGTH
+              + " characters.");
+    }
+    if (ILLEGAL_TABLE_CHARS.matcher(idToCheck).find()) {
+      throw new IllegalArgumentException(
+          "Table ID "
+              + idToCheck
+              + " is not a valid ID. Only letters, numbers, hyphens, underscores and exclamation points are allowed.");
+    }
+  }
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
@@ -20,6 +20,7 @@ import static com.google.cloud.teleport.it.common.ResourceManagerUtils.generateR
 import com.google.cloud.bigtable.admin.v2.models.StorageType;
 import com.google.common.collect.ImmutableList;
 import com.google.re2j.Pattern;
+import java.time.format.DateTimeFormatter;
 
 /**
  * Utilities for {@link com.google.cloud.teleport.it.bigtable.BigtableResourceManager}
@@ -36,6 +37,8 @@ public final class BigtableResourceManagerUtils {
   private static final int MIN_TABLE_ID_LENGTH = 1;
   private static final int MAX_TABLE_ID_LENGTH = 30;
   private static final Pattern ILLEGAL_TABLE_CHARS = Pattern.compile("[^a-zA-Z0-9-_.]");
+  private static final DateTimeFormatter TIME_FORMAT =
+      DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss-SSSSSS");
 
   private BigtableResourceManagerUtils() {}
 
@@ -54,7 +57,11 @@ public final class BigtableResourceManagerUtils {
 
     String clusterId =
         generateResourceId(
-            baseString, ILLEGAL_CLUSTER_CHARS, REPLACE_CLUSTER_CHAR, MAX_CLUSTER_ID_LENGTH);
+            baseString,
+            ILLEGAL_CLUSTER_CHARS,
+            REPLACE_CLUSTER_CHAR,
+            MAX_CLUSTER_ID_LENGTH,
+            TIME_FORMAT);
     BigtableResourceManagerCluster cluster =
         BigtableResourceManagerCluster.create(clusterId, zone, numNodes, storageType);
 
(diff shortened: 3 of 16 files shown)


it/pom.xml
```diff
@@ -114,6 +115,12 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-inline</artifactId>
```
bvolpato, line 120: Maybe use `${mockito-inline.version}` here?

it/pom.xml
```diff
@@ -211,4 +218,13 @@
...
 
+  <profiles>
+    <profile>
+      <id>apple-silicon</id>
+      <properties>
+        <os.detected.classifier>osx-x86_64</os.detected.classifier>
+      </properties>
+    </profile>
+  </profiles>
+
```
bvolpato, line 229: Not sure if there's any impact for non-M1. What I have done is put the profiles on my own `~/.m2/settings.xml`, so at least we don't need to push anything:
```
<settings>
  <activeProfiles>
    <activeProfile>
      apple-silicon
    </activeProfile>
  </activeProfiles>
  <profiles>
    <profile>
      <id>apple-silicon</id>
      <properties>
        <os.detected.classifier>osx-x86_64</os.detected.classifier>
      </properties>
    </profile>
  </profiles>
</settings>
```

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Creates a table within the current instance given a table ID.
+   *
+   * <p>Note: Implementations may do instance creation here, if one does not already exist.
+   *
+   * @param tableId The id of the table.
+   * @param columnFamilies A collection of column family names for the table.
+   * @throws RuntimeException if there is an error creating the table in Bigtable.
+   */
+  void createTable(String tableId, Iterable<String> columnFamilies) throws IllegalStateException;
```
bvolpato, line 44: Inconsistent `@throws`

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
```diff
@@ -0,0 +1,90 @@
...
+   * Writes a given row into a table. This method requires {@link
+   * BigtableResourceManager#createTable(String, Iterable)} to be called for the target table
+   * beforehand.
+   *
+   * @param tableRow A mutation object representing the table row.
+   * @throws RuntimeException if method is called after resources have been cleaned up, if the
+   *     manager object has no instance, if the table does not exist or if there is an IOException
+   *     when attempting to retrieve the bigtable data client.
+   */
+  void write(RowMutation tableRow) throws IllegalStateException;
```
bvolpato, line 56: Improve throws (same as above)

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  ImmutableList<Row> readTable(String tableId) throws IllegalStateException;
+
+  /**
+   * Deletes all created resources (instance and tables) and cleans up all Bigtable clients, making
+   * the manager object unusable.
+   *
+   * @throws IllegalStateException if there is an error deleting the instance or tables in Bigtable.
+   */
+  void cleanup() throws IllegalStateException;
```
bvolpato, line 89: Just a nitpick, other ResourceManagers have/are being created with `cleanupAll()`

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
```diff
@@ -0,0 +1,55 @@
...
+   * @param testId the test id to associate the bigtable cluster to.
+   * @param zone the zone/region that the cluster will be deployed to.
+   * @param numNodes the number of nodes that the cluster will contain.
+   * @param storageType the type of storage to configure the cluster with (SSD or HDD).
+   * @return Collection containing a single BigtableResourceManagerCluster object.
+   */
+  static Iterable<BigtableResourceManagerCluster> generateDefaultClusters(
+      String testId, String zone, int numNodes, StorageType storageType) {
+    checkArgument(
+        testId.length() > MIN_CLUSTER_ID_LENGTH - 3,
```
bvolpato, line 48: nitpick: Took me a bit to understand why `- 3`. Perhaps extracting `"-c1"` to a constant and use `.length()` can make it cleaner.

it/src/main/java/com/google/cloud/teleport/it/common/ResourceManagerUtils.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.google.common.hash.HashFunction;
+import com.google.re2j.Pattern;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Random;
+
+/** Common utilities for ResourceManager implementations. */
+public class ResourceManagerUtils {
```
bvolpato, line 30: just fyi this will clash with https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/448/, but merging should be quite simple

it/src/test/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+        IllegalArgumentException.class,
+        () -> generateDefaultClusters("a", ZONE, NUM_NODES, STORAGE_TYPE));
+  }
+
+  @Test
+  public void testGenerateDefaultClustersShouldShortenTestIdWhenTooLong() {
+    Iterable<BigtableResourceManagerCluster> cluster =
+        generateDefaultClusters(
+            "really-really-really-long-testing-id", ZONE, NUM_NODES, STORAGE_TYPE);
+    assertEquals(cluster.iterator().next().getClusterId(), "really-really-really-long-t-c1");
```
bvolpato, line 62: It's recommended to use https://github.com/google/truth: `assertThat(x).isEqualTo(z)`;

it/src/test/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManagerTest.java
```diff
@@ -0,0 +1,392 @@
...
+
+    assertThrows(BigtableResourceManagerException.class, () -> testManager.cleanup());
+  }
+
+  @Test
+  public void testCleanupShouldWorkWhenBigtableDoesNotThrowAnyError() {
+    prepareCreateInstanceAdminClientMock();
+    prepareCreateTableAdminClientMock();
+
+    testManager.cleanup();
```
bvolpato, line 390: Can we put some asserts/verify() to make sure it worked as planned?

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
Polber: I was contemplating renaming the spanner RM to use cleanup() along with the BigQuery RM I'm working on, but I went ahead and refactored to cleanupAll() as suggested.

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
Polber: Fixed.

it/src/test/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtilsTest.java
Polber: Fixed.

it/src/test/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManagerTest.java
Polber: There are 3 other tests to make sure the intended resources throw errors when there is a problem, and therefore tests to make sure the 3 necessary close/delete functions are called. This test then subsequently ensures no errors are thrown when the 3 functions work as intended. Special logic could be implemented, but I feel as though that would be redundant in this case.

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Creates a Bigtable instance in which all clusters, nodes and tables will exist.
+   *
+   * @param clusters Collection of BigtableResourceManagerCluster objects to associate with the
+   *     given Bigtable instance.
+   * @throws BigtableResourceManagerException if there is an error creating the instance in
+   *     Bigtable.
+   */
+  void createInstance(Iterable<BigtableResourceManagerCluster> clusters)
+      throws BigtableResourceManagerException;
```
bvolpato, line 34: nitpick: BigtableResourceManagerException is unchecked (RuntimeException), so no need to declare everywhere

it/src/test/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String TEST_ID = "test-id";
+  private static final String ZONE = "us-central1-a";
+  private static final int NUM_NODES = 1;
+  private static final StorageType STORAGE_TYPE = StorageType.SSD;
+
+  @Test
+  public void testGenerateDefaultClustersShouldWorkWhenAllParametersValid() {
+    Iterable<BigtableResourceManagerCluster> cluster =
+        generateDefaultClusters(TEST_ID, ZONE, NUM_NODES, STORAGE_TYPE);
+    assertEquals(cluster.iterator().next().getClusterId(), TEST_ID + "-c1");
```
bvolpato, line 45: nitpick: Truth is preferable.

bvolpato: also, maybe saving the BigtableResourceManagerCluster instance and doing the assertions on it will look simpler, instead of `.iterator().next()` on every line

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param zone the zone/region that the cluster will be deployed to.
+   * @param numNodes the number of nodes that the cluster will contain.
+   * @param storageType the type of storage to configure the cluster with (SSD or HDD).
+   * @return Collection containing a single BigtableResourceManagerCluster object.
+   */
+  static Iterable<BigtableResourceManagerCluster> generateDefaultClusters(
+      String testId, String zone, int numNodes, StorageType storageType) {
+    checkArgument(
+        testId.length() > MIN_CLUSTER_ID_LENGTH - CLUSTER_ID_APPEND.length(),
+        "testId cannot be less than " + MIN_CLUSTER_ID_LENGTH + " characters long.");
```
an2x, line 51: Error message is inconsistent with the check performed. E.g. for a testId length 7 it will print "testId cannot be less than 6 characters long", while it's 7 characters long.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,429 @@
...
+      String testId,
+      String projectId,
+      String instanceId) {
+    this.projectId = projectId;
+    this.instanceId = instanceId;
+    this.testId = testId;
+    this.bigtableResourceManagerClientHandler = bigtableResourceManagerClientHandler;
+
+    // Check that the project and instance ID's conform to GCS standards
+    if (!isValidId(
```
an2x, line 133: Makes sense to make it a separate method `ResourceManagerUtils.checkValidProjectId(String projectId) throws IllegalArgumentException` and move all constants like `MIN_PROJECT_ID_LENGTH` there. And then it will be a 1-liner here (and in other resource managers).

it/src/main/java/com/google/cloud/teleport/it/common/ResourceManagerUtils.java
```diff
@@ -0,0 +1,98 @@
...
+    return id.substring(0, targetLength - hash.length() - 1) + "-" + hash;
+  }
+
+  /**
+   * Generates an instance id from a given string.
+   *
+   * @param baseString The string to generate the id from.
+   * @return The instance id string.
+   */
+  public static String generateInstanceId(
```
an2x, line 65: What's the purpose of this method? Is it supposed to generate Spanner instance IDs or an arbitrary resource ID? Please add this to javadoc. If Spanner IDs, then it should be moved to `SpannerResourceManagerUtils`. If any resource ID, then I'd change its name to `generateResourceId` (but then how is it different from `generateNewId`? please rename and/or add javadoc details), and remove comments like "so it doesn't violate spanner's instance".

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -83,7 +85,7 @@ private DefaultSpannerResourceManager(Builder builder) {
       testId = generateNewId(testId, MAX_BASE_ID_LENGTH);
     }
     this.projectId = projectId;
-    this.instanceId = generateInstanceId(testId);
+    this.instanceId = generateInstanceId(testId, ILLEGAL_INSTANCE_CHARS, MAX_BASE_ID_LENGTH);
```
an2x, line 88: I don't like the fact that the ID generation logic is now spread all over the place: `generateInstanceId` is in `ResourceManagerUtils`, `generateDatabaseId` and `ILLEGAL_DATABASE_CHARS` are in `SpannerResourceManagerUtils`, `ILLEGAL_INSTANCE_CHARS` is here and in `DefaultBigQueryResourceManager`. It makes it hard to answer the question "how do we generate Spanner instance and database ID" for example.

I suggest re-working this and extracting as many generic method is possible.
Maybe something like this?

1) Keep common methods that just generate arbitrary IDs based on the source testId  in `ResourceManagerUtils` (with some options like limit the length to N).
2) All Spanner ID generation logic should be in `SpannerResourceManagerUtils` (move `generateInstanceId` back there, even if it's going to be just a 1-liner that calls `ResourceManagerUtils`). So both `generateInstanceId` and `generateDatabaseId` should be there.
3) All Bigtable ID generation logic should be in `BigtableResourceManagerUtils` (create new class).
4) All constants like  `ILLEGAL_DATABASE/INSTANCE_CHARS` and "rules" (max length, allowed chars, whether to auto-replace some chars, etc.) of how the IDs have to be generated for a specific resource should be only in the `XResourceManagerUtils` class.
5) Consider renaming `ResourceManagerUtils` to `ResourceIdUtils` as I believe this ID generation logic may be used outside of Resource Managers in the future.

an2x (changes requested):

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerUtils.java
Polber: I refactored this method to simply generate the cluster ID using the common generateResourceId() function you suggested below, so this is no longer relevant.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: I agree. I refactored to support this.

it/src/main/java/com/google/cloud/teleport/it/common/ResourceManagerUtils.java
Polber: It was intended to generate Instance ID's for both BigTable and Spanner as they both have a concept of Instances with the same constraints. However, I agree having a `generateResourceId` function is cleaner and more adaptable. I refactored to support this.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
Polber: I agree with points 1-4, however, I'm not sure about point 5. This class is meant to contain common functions used between resource managers, which just so happens to be generating ID's at this point. But I figured there will be other types of common functions contained in this class. Do you think there should be a separate class just for ID generation aside from other common functions?

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -90,6 +91,9 @@ private DefaultSpannerResourceManager(Builder builder) {
     this.spanner = spanner;
     this.instanceAdminClient = spanner.getInstanceAdminClient();
     this.databaseAdminClient = spanner.getDatabaseAdminClient();
+
+    // Check that the project ID conforms to GCS standards
```
pranavbhandari24, line 95: GCP standards?

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
```diff
@@ -90,6 +91,9 @@ private DefaultSpannerResourceManager(Builder builder) {
...
+    checkValidProjectId(projectId);
```
pranavbhandari24, line 96: Maybe check this at the beginning of the method? This way we can fail fast before creating the clients

it/src/test/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManagerTest.java
Polber: Refactored to use the `Mockito.verify` method/check.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
Polber: I had Google Cloud Storage in my head for some reason. Fixed, thanks.

it/src/main/java/com/google/cloud/teleport/it/spanner/DefaultSpannerResourceManager.java
Polber: Fixed in Spanner and Bigtable constructors. Thanks.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,421 @@
...
+    private CredentialsProvider credentialsProvider;
+    private final BigtableInstanceAdminSettings.Builder bigtableInstanceAdminSettings;
+    private final BigtableTableAdminSettings.Builder bigtableTableAdminSettings;
+    private final BigtableDataSettings.Builder bigtableDataSettings;
+
+    private Builder(String testId, String projectId) {
+
+      // Generate and format ID's
+      if (testId.length() > MAX_TEST_ID_LENGTH) {
+        testId = generateNewId(testId, MAX_TEST_ID_LENGTH);
```
an2x, line 365: I don't think we should truncate `testId` itself. The only reason a limit is imposed is because `instaceId` length is limited. But `generateInstanceId()` already takes care of it and truncates `testId` if needed. I'd just remove this whole block that checks `testId` length.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,421 @@
...
+  private final BigtableResourceManagerClientHandler bigtableResourceManagerClientHandler;
+
+  private boolean hasInstance = false;
+
+  @VisibleForTesting
+  DefaultBigtableResourceManager(
+      String testId, String projectId, CredentialsProvider credentialsProvider) throws IOException {
+    this(
+        DefaultBigtableResourceManager.builder(testId, projectId)
+            .setCredentialsProvider(credentialsProvider));
```
an2x, line 77: `credentialsProvider` seems to be unused.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,421 @@
...
+      String projectId,
+      BigtableResourceManagerClientHandler bigtableResourceManagerClientHandler)
+      throws IOException {
+    this(
+        bigtableResourceManagerClientHandler,
+        DefaultBigtableResourceManager.builder(testId, projectId));
+  }
+
+  private DefaultBigtableResourceManager(DefaultBigtableResourceManager.Builder builder)
+      throws IOException {
```
an2x, line 92: I suggest refactoring all these constructors:
1) Keep only this one (that accepts a Builder).
2) Builder should keep an instance of `BigtableResourceManagerClientHandler`, not 3 individual `Bigtable*Settings` objects.
3) Tests can do `Builder#setBigtableResourceManagerClientHandler()` to override the handler with a mock implementation.
4) If no overrides provided, Builder will create the default handler instance in the `build()` method.

This way we'll have only 1 constructor instead of 4, no need for `@VisibleForTesting` annotations, and no constructors like `DefaultBigtableResourceManager(BigtableResourceManagerClientHandler, Builder)`, which actually accepts 2 sets of bigtable settings at the same time (one in the handler argument and one in the builder).

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,421 @@
...
+
+    private Builder(String testId, String projectId) {
+
+      // Generate and format ID's
+      if (testId.length() > MAX_TEST_ID_LENGTH) {
+        testId = generateNewId(testId, MAX_TEST_ID_LENGTH);
+      }
+      this.testId = testId;
+      this.projectId = projectId;
+      this.instanceId = generateInstanceId(testId);
```
an2x, line 369: I think it's better to to generate instance ID and AdminSettings in the resource manager constructor.

an2x (changes requested):

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: I was thinking the same thing, but I left it to reflect the Spanner resource manager. I think the Spanner RM now needs some refactoring to stay consistent. Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: This was another instance of trying to reflect the pattern used in the Spanner RM, however, I agree it is clunky. I refactored as you suggested, and it looks much cleaner.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: I'm not sure how to accomplish this since the Client Settings are built before being wrapped in the BigtableResourceManagerClientHandler class, and the ID's must be set in each of the Settings objects before building.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,390 @@
...
+  private boolean hasInstance = false;
+
+  private DefaultBigtableResourceManager(DefaultBigtableResourceManager.Builder builder)
+      throws IOException {
+    // Check that the project ID conforms to GCP standards
+    checkValidProjectId(builder.projectId);
+
+    this.projectId = builder.projectId;
+    this.instanceId = builder.instanceId;
+    this.testId = builder.testId;
```
an2x, line 77: `testId` seems to be unused.
There is a test that checks its values, but why do we need to check the value of a field that's not used anywhere else? I would remove both this field and the test case.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,390 @@
...
+    public String getTestId() {
+      return testId;
+    }
+
+    public Builder setProjectId(String projectId) {
+      this.projectId = projectId;
+      return this;
+    }
+
+    public String getProjectId() {
```
an2x, line 324: Builders are typically made write-only. I'd remove all getters.
Also consider using `@AutoBuilder`, which generates a lot of boilerplate code for you.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,390 @@
...
+      return projectId;
+    }
+
+    public Builder setInstanceId(String instanceId) {
```
an2x, line 328: I'd remove `setInstanceId`. We don't really need it at the moment (the expectation is that we'll always auto-generate it based on `testId`) and having this setter makes the builder logic too complicated (set testId -> generate a unique instanceId -> override it with setInstanceId -> what's the instanceId value now? not clear for the user of the builder; moreover, calling build() method twice with an overridden instanceId will result in creating to RM instances that will try to write to the same instance). I like the implementation that is guaranteed to auto-generate a unique instanceId on every build() invocation.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
an2x: I'd move `generateInstanceId(testId)` to the `build()` method then, to generate a new id for every new RM instance constructed by the builder.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -0,0 +1,390 @@
...
+            BigtableDataSettings.newBuilder().setProjectId(projectId).setInstanceId(instanceId);
+
+        // add the credentials to the builders, if set.
+        if (this.credentialsProvider != null) {
+          bigtableInstanceAdminSettings.setCredentialsProvider(this.credentialsProvider);
+          bigtableTableAdminSettings.setCredentialsProvider(this.credentialsProvider);
+          bigtableDataSettings.setCredentialsProvider(this.credentialsProvider);
+        }
+
+        this.bigtableResourceManagerClientHandler =
```
an2x, line 379: I'd avoid altering the builder state in the `build()` method (i.e. create a temporary local bigtableResourceManagerClientHandler variable here instead of updating the instance variable), because this is error-prone. For example, if you do this:
```
builder = DefaultBigtableResourceManager.builder(...);
resourceManager1 = builder.build();
resourceManager2 = builder.setCredentialsProvider(...).build();
```
`resourceManager2` actually won't use the credentials provider. But if you re-create `bigtableResourceManagerClientHandler` on every `build()` call it will.

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerCluster.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final String clusterId;
+
+    private final String zone;
+
+    private final int numNodes;
+
+    private final StorageType storageType;
+
+    private Builder(String clusterId, String zone, int numNodes, StorageType storageType) {
```
an2x, line 101: IMO there is no point in having a Builder if all its properties are required (you can just call the main class constructor directly exactly the same way).

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerClientHandler.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /** Builder for {@link BigtableResourceManagerClientHandler}. */
+  public static final class Builder {
+
+    private final BigtableInstanceAdminSettings instanceSettings;
+    private final BigtableTableAdminSettings tableSettings;
+    private final BigtableDataSettings dataSettings;
+
+    private Builder(
```
an2x, line 87: Same: no need for a builder if all its properties are required.

an2x (changes requested):

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerCluster.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManagerClientHandler.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: Fixed.

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
Polber: Fixed.

an2x (dismissed):

pranavbhandari24 (approved):
LGTM!

PR #448, opened 2022-08-22
Pub/Sub Resource Manager + Sample Integration test PubsubToTextIT

bvolpato:
I have created a structure for integration tests using Pub/Sub. Currently, the following operations are available:

- PubsubResourceManager
  - createTopic
  - createSubscription
  - publish

In order to test it, I have created an integration test `PubsubToTextIT`, based on the available integration tests for templates (`StreamingDataGeneratorIT`).

Since Pub/Sub topics are created directly below the project, and there's no concepts of groups, the manager will keep track of all the topics/subscriptions that were created by that instance -- which are then used on the `cleanupAll`.  
If `cleanupAll` is not called, resources will stay in the project, and require external cleanups (in which having the timestamp on the topic/subscription name will be handy).

### DIFF
--- a/it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubPublisherFactory.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubPublisherFactory.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.pubsub.v1.TopicName;
+import java.io.IOException;
+
+/**
+ * Default class for implementation of {@link PubsubPublisherFactory} interface.
+ *
+ * <p>The class provides an interaction with the real Pub/Sub client, with operations related to
+ * creating a Pub/Sub Publisher.
+ */
+class DefaultPubsubPublisherFactory implements PubsubPublisherFactory {
+
+  private final CredentialsProvider credentialsProvider;
+
+  DefaultPubsubPublisherFactory(CredentialsProvider credentialsProvider) {
+    this.credentialsProvider = credentialsProvider;
+  }
+
+  @Override
+  public Publisher createPublisher(TopicName topic) {
+    try {
+      return Publisher.newBuilder(topic).setCredentialsProvider(credentialsProvider).build();
+    } catch (IOException e) {
+      throw new PubsubResourceManagerException("Error creating publisher for topic", e);
+    }
+  }
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
+import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
+import com.google.cloud.pubsub.v1.TopicAdminClient;
+import com.google.cloud.pubsub.v1.TopicAdminSettings;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
+import com.google.pubsub.v1.PushConfig;
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default class for implementation of {@link PubsubResourceManager} interface.
+ *
+ * <p>The class provides an interaction with the real Pub/Sub client, with operations related to
+ * management of topics and subscriptions.
+ */
+public final class DefaultPubsubResourceManager implements PubsubResourceManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultPubsubResourceManager.class);
+
+  private static final int DEFAULT_ACK_DEADLINE_SECONDS = 600;
+  private static final String RESOURCE_NAME_SEPARATOR = "-";
+
+  private final String testId;
+  private final String projectId;
+  private final PubsubPublisherFactory publisherFactory;
+  private final TopicAdminClient topicAdminClient;
+  private final SubscriptionAdminClient subscriptionAdminClient;
+
+  private final Set<TopicName> createdTopics;
+  private final Set<SubscriptionName> createdSubscriptions;
+
+  private DefaultPubsubResourceManager(Builder builder) throws IOException {
+    this(
+        builder.testName,
+        builder.projectId,
+        new DefaultPubsubPublisherFactory(builder.credentialsProvider),
+        TopicAdminClient.create(
+            TopicAdminSettings.newBuilder()
+                .setCredentialsProvider(builder.credentialsProvider)
+                .build()),
+        SubscriptionAdminClient.create(
+            SubscriptionAdminSettings.newBuilder()
+                .setCredentialsProvider(builder.credentialsProvider)
+                .build()));
+  }
+
+  @VisibleForTesting
+  DefaultPubsubResourceManager(
+      String testName,
+      String projectId,
+      PubsubPublisherFactory publisherFactory,
+      TopicAdminClient topicAdminClient,
+      SubscriptionAdminClient subscriptionAdminClient) {
+    this.projectId = projectId;
+    this.testId = PubsubUtils.createTestId(testName);
+    this.publisherFactory = publisherFactory;
+    this.topicAdminClient = topicAdminClient;
+    this.subscriptionAdminClient = subscriptionAdminClient;
+    this.createdTopics = Collections.synchronizedSet(new HashSet<>());
+    this.createdSubscriptions = Collections.synchronizedSet(new HashSet<>());
+  }
+
+  public static Builder builder(String testName, String projectId) {
+    checkArgument(!Strings.isNullOrEmpty(testName), "testName can not be null or empty");
+    checkArgument(!projectId.isEmpty(), "projectId can not be empty");
+    return new Builder(testName, projectId);
+  }
+
+  /**
+   * Create a topic based on the given topic name. However, it adds a prefix based on the test id
+   * and timestamp, to uniquely identify the current execution instance.
+   *
+   * @see PubsubResourceManager#createTopic(String)
+   */
+  @Override
+  public TopicName createTopic(String topicName) {
+    checkArgument(!topicName.isEmpty(), "topicName can not be empty");
+    checkIsUsable();
+
+    LOG.info("Creating topic '{}'...", topicName);
+
+    Topic topic = topicAdminClient.createTopic(getTopicName(topicName));
+    TopicName reference = PubsubUtils.toTopicName(topic);
+    createdTopics.add(reference);
+
+    LOG.info("Topic '{}' was created successfully!", topicName);
+
+    return reference;
+  }
+
+  /**
+   * Create a subscription based on the given name. However, it adds a prefix based on the test id
+   * and timestamp, to uniquely identify the current execution instance.
+   *
+   * @see PubsubResourceManager#createSubscription(TopicName, String)
+   */
+  @Override
+  public SubscriptionName createSubscription(TopicName topicName, String subscriptionName) {
+    checkArgument(!subscriptionName.isEmpty(), "subscriptionName can not be empty");
+    checkIsUsable();
+
+    if (!createdTopics.contains(topicName)) {
+      throw new IllegalArgumentException(
+          "Can not create a subscription for a topic not managed by this instance.");
+    }
+
+    LOG.info("Creating subscription '{}' for topic '{}'", subscriptionName, topicName);
+
+    Subscription subscription =
+        subscriptionAdminClient.createSubscription(
+            getSubscriptionName(subscriptionName),
+            topicName,
+            PushConfig.getDefaultInstance(),
+            DEFAULT_ACK_DEADLINE_SECONDS);
+    SubscriptionName reference = PubsubUtils.toSubscriptionName(subscription);
+    createdSubscriptions.add(reference);
+
+    LOG.info(
+        "Subscription '{}' for topic '{}' was created successfully!", topicName, subscriptionName);
+
+    return reference;
+  }
+
+  @Override
+  public String publish(TopicName topic, Map<String, String> attributes, ByteString data)
+      throws PubsubResourceManagerException {
+    checkIsUsable();
+
+    if (!createdTopics.contains(topic)) {
+      throw new IllegalArgumentException(
+          "Can not publish to a topic not managed by this instance.");
+    }
+
+    LOG.info("Publishing message with {} bytes to topic '{}'", data.size(), topic);
+
+    PubsubMessage pubsubMessage =
+        PubsubMessage.newBuilder().putAllAttributes(attributes).setData(data).build();
+
+    try {
+      String messageId = publisherFactory.createPublisher(topic).publish(pubsubMessage).get();
+      LOG.info("Message published with id '{}'", messageId);
+      return messageId;
+    } catch (Exception e) {
+      throw new PubsubResourceManagerException("Error publishing message to Pubsub", e);
+    }
+  }
+
+  @Override
+  public synchronized void cleanupAll() {
+    // Ignore call if it was cleaned up before
+    if (isNotUsable()) {
+      return;
+    }
+
+    LOG.info("Attempting to cleanup manager.");
+
+    try {
+      for (SubscriptionName subscription : createdSubscriptions) {
+        LOG.info("Deleting subscription '{}'", subscription);
+        subscriptionAdminClient.deleteSubscription(subscription);
+      }
+
+      for (TopicName topic : createdTopics) {
+        LOG.info("Deleting topic '{}'", topic);
+        topicAdminClient.deleteTopic(topic);
+      }
+    } finally {
+      subscriptionAdminClient.close();
+      topicAdminClient.close();
+    }
+
+    LOG.info("Manager successfully cleaned up.");
+  }
+
+  SubscriptionName getSubscriptionName(String subscriptionName) {
+    return SubscriptionName.of(projectId, testId + RESOURCE_NAME_SEPARATOR + subscriptionName);
+  }
+
+  TopicName getTopicName(String topicName) {
+    return TopicName.of(projectId, testId + RESOURCE_NAME_SEPARATOR + topicName);
+  }
+
+  /**
+   * Check if the clients started by this instance are still usable, and throwing {@link
+   * IllegalStateException} otherwise.
+   */
+  private void checkIsUsable() throws IllegalStateException {
+    if (isNotUsable()) {
+      throw new IllegalStateException("Manager has cleaned up resources and is unusable.");
+    }
+  }
+
+  private boolean isNotUsable() {
+    return topicAdminClient.isShutdown() || subscriptionAdminClient.isShutdown();
+  }
+
+  /** Builder for {@link DefaultPubsubResourceManager}. */
+  public static final class Builder {
+
+    private final String projectId;
+    private final String testName;
+    private CredentialsProvider credentialsProvider;
+
+    private Builder(String testName, String projectId) {
+      this.testName = testName;
+      this.projectId = projectId;
+    }
+
+    public Builder credentialsProvider(CredentialsProvider credentialsProvider) {
+      this.credentialsProvider = credentialsProvider;
+      return this;
+    }
+
+    public DefaultPubsubResourceManager build() throws IOException {
+      return new DefaultPubsubResourceManager(this);
+    }
+  }
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubPublisherFactory.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubPublisherFactory.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.pubsub.v1.TopicName;
+
+/** Interface for building Pub/Sub publishers in integration tests. */
+interface PubsubPublisherFactory {
+
+  /** Create a {@link Publisher} instance for the given topic reference. */
+  Publisher createPublisher(TopicName topic);
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubResourceManagerException.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubResourceManagerException.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+/** Custom exception for {@link PubsubResourceManager} implementations. */
+class PubsubResourceManagerException extends RuntimeException {
+
+  PubsubResourceManagerException(String errorMessage, Throwable err) {
+    super(errorMessage, err);
+  }
+}
--- a/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.base.CaseFormat;
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
+import com.google.re2j.Matcher;
+import com.google.re2j.Pattern;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+/** Utilities to make working with Pubsub easier. */
+final class PubsubUtils {
+  private PubsubUtils() {}
+
+  /**
+   * Regular expression to validate/extract a topic name string, following
+   * projects/{project_id}/subscription/{topic_name}.
+   */
+  private static final Pattern TOPICS_PATTERN =
+      Pattern.compile("^projects/(?P<project_id>[^/]+)/topics/(?P<topic_name>[^/]+)$");
+
+  /**
+   * Regular expression to validate/extract a subscription name string, following
+   * projects/{project_id}/subscription/{subscription_name}.
+   */
+  private static final Pattern SUBSCRIPTIONS_PATTERN =
+      Pattern.compile(
+          "^projects/(?P<project_id>[^/]+)/subscriptions/(?P<subscription_name>[^/]+)$");
+
+  /**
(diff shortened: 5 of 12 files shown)


bvolpato:
Spotless errors are caused by different files, maybe we should run a general `spotless:apply` on main?

The unit test errors are also in a different component, but I have addressed them here https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/449

zhoufek (changes requested):
Thanks for adding the resource manager and for adding a couple more integration tests!

Along with the other changes, could you please add a method for reading from Pub/Sub, or are you planning that for a future change? (Delaying it may be preferable to keep the size of this PR small(er).)

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+      }
+    } finally {
+      this.topicAdminClient.close();
+    }
+
+    LOG.info("Manager successfully cleaned up.");
+  }
+
+  @VisibleForTesting
```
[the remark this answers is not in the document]
bvolpato: We have tests both validating functionality, as well as using the method for assertions (see testCleanupSubscriptions). Let me know if you still think this is not proper.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  @Test
+  public void testValidSubscriptionName() {
+    SubscriptionName subscriptionName = testManager.getSubscriptionName("valid-sub");
+    assertThat(subscriptionName).isEqualTo(SubscriptionName.of(PROJECT_ID, "it-test-id-valid-sub"));
+  }
+
+  @Test
+  public void testCreateTopic() {
```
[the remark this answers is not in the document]
bvolpato: Done, looks much organized indeed, thanks!

it/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1,2 @@
+# Required to mock final methods:
+mock-maker-inline
```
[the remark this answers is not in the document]
bvolpato: Unfortunately, a bunch of final methods. Entire `TopicAdminClient` and `SubscriptionAdminClient` are locked down.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,224 @@
...
+        DEFAULT_ACK_DEADLINE_SECONDS);
+  }
+
+  @Override
+  public synchronized void cleanupAll() {
+    LOG.info("Attempting to cleanup manager.");
+
+    // as more resources can exist on the project, we delete only subscriptions for current testId
+    try {
+      ListSubscriptionsPagedResponse subsResponse =
```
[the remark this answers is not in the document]
bvolpato: Replaced with a List of TopicName/SubscriptionName. There's not much details that we have available/would be interested on Topic/Subscription.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubPublisherTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Test
+  public void testBuilderWithInvalidTopicName() {
+    expectedException.expect(IllegalArgumentException.class);
+    expectedException.expectMessage("topicName can not be empty");
+    DefaultPubsubPublisher.builder("");
+  }
+
+  @Test
+  public void testPublishMessage() throws ExecutionException, InterruptedException {
```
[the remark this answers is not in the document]
bvolpato: I like that! Thanks!

v2/common/pom.xml
```diff
@@ -165,6 +165,12 @@
               </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
bvolpato: It was to allow usage of assertThrows on DurationUtilsTest - but I am going to move it to another PR

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/pubsubtotext/PubsubToTextIT.java
```diff
@@ -0,0 +1,243 @@
...
+                });
+
+    // Assert
+    assertThat(result).isEqualTo(Result.CONDITION_MET);
+
+    // Make sure that files contain the job name
+    assertThat(
+        artifacts.get().stream()
+                .anyMatch(artifact -> new String(artifact.contents()).contains(messageString)))
+        .isTrue();
```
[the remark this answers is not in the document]
bvolpato: Valid concern. I am saving all the contents to a single stream and `assertThat(...).contains()` now. Do you think that is enough? Our PubsubMessage's data is only sending `messageString`, so we expect at least that `messageString` is on the files. 

Any ideas?

bvolpato:
> Thanks for adding the resource manager and for adding a couple more integration tests!
> 
> Along with the other changes, could you please add a method for reading from Pub/Sub, or are you planning that for a future change? (Delaying it may be preferable to keep the size of this PR small(er).)

Yes, I think I can do that soon, but I agree with splitting on another PR.
Maybe I could do it while creating an integration test which needs it? (e.g. any XYZToPubsubIT). 

I found doing the ResourceManager+IT together helpful to shape the API for testability, as well as a good way to fully testing it.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,224 @@
...
+            subscriptionAdminClient.deleteSubscription(subscription.getName());
+          }
+        }
+      }
+    } finally {
+      this.subscriptionAdminClient.close();
+    }
+
+    // as more resources can exist on the project, we delete only topics for current testId
+    try {
```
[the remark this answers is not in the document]
bvolpato: We might have more than one on the same ResourceManager instance (e.g. input, output, DLQ).

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,220 @@
...
+  @VisibleForTesting
+  DefaultPubsubResourceManager(
+      @Nullable String testId,
+      String projectId,
+      CredentialsProvider credentialsProvider,
+      @Nullable TopicAdminClient topicAdminClient,
+      @Nullable SubscriptionAdminClient subscriptionAdminClient)
+      throws IOException {
+    this.projectId = projectId;
+    this.testId = (!Strings.isNullOrEmpty(testId) ? testId : "");
```
zhoufek, line 77: An empty id should be an error, and it's probably best to error closer to where the problem arose:

```
public static Builder builder(String projectId, String testId) {
  checkArgument(!Strings.isNullOrEmpty(testId), ...);
  // Other checks
  return new Builder(projectId, testId);
}
```

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,220 @@
...
+    this.credentialsProvider = credentialsProvider;
+    this.topicAdminClient =
+        (topicAdminClient != null
+            ? topicAdminClient
+            : TopicAdminClient.create(
```
zhoufek, line 82: Maybe instead of doing a check here, just create the `TopicAdminClient` and `SubscriptionAdminClient` in the constructor that accepts the builder. That'll make this overall simpler, since we just have to set the resulting client whether it is a real or mock one.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,220 @@
...
+    return new Builder(projectId);
+  }
+
+  @Override
+  public TopicName createTopic(String topicName) {
+    LOG.info("Creating topic '{}'", topicName);
+
+    Topic topic = topicAdminClient.createTopic(getTopicName(topicName));
+    TopicName reference = PubsubUtils.toTopicName(topic);
+    createdTopics.add(reference);
```
zhoufek, line 108: To make sure that this class is thread safe, we should either use a thread-safe collection or make these calls in a `synchronize` block. (I think synchronizing the whole method is a bit much.)

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubResourceManagerException.java
```diff
@@ -0,0 +1,24 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+/** Custom exception for {@link PubsubResourceManager} implementations. */
+public class PubsubResourceManagerException extends RuntimeException {
+
+  public PubsubResourceManagerException(String errorMessage, Throwable err) {
```
zhoufek, line 21: Maybe make this package private. The class should be public, but I don't think we want any other packages creating one of these exceptions.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Convert the instance of {@link Topic} to a {@link TopicName}.
+   *
+   * <p>The idea of this project is to abstract the conversion/parse of the format
+   * <strong>projects/{project_id}/topics/{topic_name}</strong>
+   *
+   * @param topic Base topic.
+   * @return The reference to the topic.
+   */
+  public static TopicName toTopicName(Topic topic) {
```
zhoufek, line 36: This seems like an error-prone approach, or can we guarantee that `Topic` always returns the correct format? This should probably use a regex and capture group. (Note: We use RE2J rather than Java's built-in regex library.)

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
+
+/** Utilities to make working with Pubsub easier. */
+public final class PubsubUtils {
```
zhoufek, line 24: Utilities should be package private.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+  private DefaultPubsubResourceManager testManager;
+
+  @Captor private ArgumentCaptor<TopicName> topicNameCaptor;
+  @Captor private ArgumentCaptor<SubscriptionName> subscriptionNameCaptor;
+  @Captor private ArgumentCaptor<PubsubMessage> pubsubMessageCaptor;
+
+  @Before
+  public void setUp() throws IOException {
+    testManager =
+        spy(
```
zhoufek, line 81: Is there a reason for making this a spy?

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  @Test
+  public void testValidSubscriptionName() {
+    SubscriptionName subscriptionName = testManager.getSubscriptionName("valid-sub");
+    assertThat(subscriptionName).isEqualTo(SubscriptionName.of(PROJECT_ID, "test-id-valid-sub"));
+  }
+
+  @Test
+  public void testCreateTopic() {
```
zhoufek, line 127: Sorry, missed this last time: When naming tests, we generally follow `testXShouldY`. It makes it easier to tell what the expected behavior is without reading the test.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+    // Arrange
```
zhoufek, line 128: nit: You can leave out the comments if it is obvious. The comments are meant for when the blocks are large and require blank lines to keep them readable. But if the only blank lines are to separate the blocks, the comments can be left out.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+    // Act
+    SubscriptionName createSub = testManager.createSubscription("topic-name", "subscription-name");
+
+    // Assert
+    assertThat(createSub).isNotNull();
+    verify(subscriptionAdminClient)
+        .createSubscription(
+            subscriptionNameCaptor.capture(),
+            topicNameCaptor.capture(),
+            any(PushConfig.class),
```
zhoufek, line 158: Just `any()` is good enough. (I'm not sure specifying the class offers any value here.)

Same goes for things like `anyInt()`.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+    testManager.createSubscription("topic1", "topic1-sub1");
+    testManager.createSubscription("topic2", "topic1-sub0");
+    testManager.cleanupAll();
+
+    // Assert
+    verify(subscriptionAdminClient, times(3)).deleteSubscription(subscriptionNameCaptor.capture());
+    assertThat(subscriptionNameCaptor.getAllValues()).hasSize(3);
+    assertThat(subscriptionNameCaptor.getAllValues())
+        .containsExactlyElementsIn(
+            Arrays.asList(subscriptionName1, subscriptionName2, subscriptionName3));
```
zhoufek, line 234: nit: `containsExactly` is a little more readable, since you can just pass in the subscriptions without explicitly creating a collection: https://truth.dev/api/0.41/com/google/common/truth/IterableSubject.html#containsExactly-java.lang.Object...-

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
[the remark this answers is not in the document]
zhoufek: Ack.

it/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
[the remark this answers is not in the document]
zhoufek: That's unfortunate. Most Google clients are much better designed for mocking.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/pubsubtotext/PubsubToTextIT.java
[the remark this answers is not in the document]
zhoufek: I guess I'm wondering if (or why) there'd be anything other than `messageString`. Is this expected behavior, or does it indicate a problem with the pipeline? If the former, this is fine. If the latter, I'm a bit concerned that this would hide an error with the pipeline.

zhoufek (changes requested):

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
bvolpato: Yes, to inject our mocked Publisher on `getPublisher(topic)`:
```
    doReturn(publisher).when(testManager).getPublisher(any());
```

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
bvolpato: Thanks, will use `any()` for non-overloaded methods.
However, `any()` breaks when used in lieu of primitives.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,218 @@
...
+      CredentialsProvider credentialsProvider,
+      TopicAdminClient topicAdminClient,
+      SubscriptionAdminClient subscriptionAdminClient) {
+    this.projectId = projectId;
+    this.testId = testId;
+    this.credentialsProvider = credentialsProvider;
+    this.topicAdminClient = topicAdminClient;
+    this.subscriptionAdminClient = subscriptionAdminClient;
+    this.createdTopics = Collections.synchronizedList(new ArrayList<>());
+    this.createdSubscriptions = Collections.synchronizedList(new ArrayList<>());
```
bvolpato, line 92: relevant: added `Collections.synchronizedList` here

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
bvolpato: Thanks. Added `Collections.synchronizedList` above.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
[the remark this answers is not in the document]
bvolpato: I think I see that you want to simplify, but my initial idea for passing in a name is to add something meaningful to the test logic, and that may in case ITs need to be debugged.

It may be not too clear on PubsubToTextIT, but for example, there are cases that you could potentially have 3 topics (e.g., "input", "output", "dlq"), and they could be created "test-pubsub-to-pubsub-202208260956-(input|output|dlq)" so that the names are somewhat meaningful rather than just using sequences.

What do you think?

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
bvolpato: `PubsubUtils.createTestId` is being used on `PubsubToTextIT` to pass in as a testId. Let me know if you prefer it organized in a different way.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
bvolpato: Makes sense, thanks!

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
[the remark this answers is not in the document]
zhoufek: Maybe rather than receiving the full topic name, just receiving a suffix if tests need to differentiate their purpose. Just make it clear from the name/documentation how it impacts the actual topic name.

Part of the purpose of this framework is to not think about isolation in the tests themselves, and receiving the full topic name goes against that.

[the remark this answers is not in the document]
bvolpato: `testId` always gets prepended on the topic name for such isolation. I'll make this clearer on the docs/parameter names.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/pubsubtotext/PubsubToTextIT.java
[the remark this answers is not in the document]
zhoufek: Yeah, but having unexpected data is also a bug, and since we only pass a single message to Pub/Sub, I would think we should know the exact data and its order (order isn't possible to know on multiple inputs). That said, I'd just prefer trying to be more specific if possible. If that's not feasible, feel free to just resolve this comment.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
zhoufek: It should be that the test has its own id (generally the test method or class name), and then the resource manager will call `createTestId`. This just makes it easier to write tests.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @param prefix a prefix for the topic
+   * @return the prefix plus some way of identifying it separate from other topics with the same
+   *     prefix
+   */
+  public static String createTestId(String prefix) {
+    String convertedPrefix =
+        CaseFormat.UPPER_CAMEL.converterTo(CaseFormat.LOWER_HYPHEN).convert(prefix);
+    String formattedTimestamp =
+        DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
```
zhoufek, line 104: Could we maybe get this down to microsecond precision? That'll decrease the chance of collision. (Second precision is good too, but I think Pub/Sub can handle greater precision, and that'll further reduce the chance of collision.)

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** Unit tests for {@link DefaultPubsubResourceManager}. */
+@RunWith(JUnit4.class)
+public final class DefaultPubsubResourceManagerTest {
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  private static final String TEST_ID = "test-id-";
```
zhoufek, line 57: Sorry, missed this: I don't think it's expected that test ids have a '-' as a suffix. If it is needed to construct a readable string, we should probably be adding the '-' in the resource manager.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
zhoufek: We generally do dependency injection through the constructor so that the class under test is a real object. If there's a bug in `getPublisher`, this would hide that. Is there something about the publisher that prevents us from making it a field?

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,218 @@
...
+  public synchronized void cleanupAll() {
+    LOG.info("Attempting to cleanup manager.");
+
+    try {
+      for (SubscriptionName subscription : createdSubscriptions) {
+        LOG.info("Deleting subscription '{}'", subscription);
+        subscriptionAdminClient.deleteSubscription(subscription);
+      }
+    } finally {
+      subscriptionAdminClient.close();
```
zhoufek, line 172: Sorry, missed this before, but I think this puts the class in an unusable state. We should be checking that the clients haven't been closed and throwing an `IllegalStateException` if they are. (This method could just return, since we're in the expected final state.)

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,218 @@
...
+    checkArgument(!Strings.isNullOrEmpty(testId), "testId can not be null or empty");
+    checkArgument(!projectId.isEmpty(), "projectId can not be empty");
+    return new Builder(testId, projectId);
+  }
+
+  @Override
+  public TopicName createTopic(String topicName) {
+    checkArgument(!topicName.isEmpty(), "topicName can not be empty");
+
+    LOG.info("Creating topic '{}'", topicName);
```
zhoufek, line 105: Could we also log after a successful creation? That'll make it a bit easier to see what's going on the test.

Same goes for subscriptions.

zhoufek (changes requested):

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
bvolpato: Publishers are created for a specific topic/subscription, which are not known ahead of time.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
bvolpato: Thanks. I've looked Spanner up and using `checkIsUsable` now.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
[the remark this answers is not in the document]
bvolpato: It is just the suffix. I've made it clearer on the JavaDoc that the topic/subscription will not be used as-is.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubUtils.java
bvolpato: Agreed, and changed the class to be package private.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/pubsubtotext/PubsubToTextIT.java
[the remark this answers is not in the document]
bvolpato: Got it. I've changed slightly to replace anything that we expect for "", and validate if it yields empty. It seems a bit more convoluted, but I agree it's better / covers the no unexpected data case.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,239 @@
...
+    return createSubscription(getTopicName(topicName), subscriptionName);
+  }
+
+  @Override
+  public SubscriptionName createSubscription(TopicName topicName, String subscriptionName) {
+    checkArgument(!subscriptionName.isEmpty(), "subscriptionName can not be empty");
+
+    LOG.info("Creating subscription '{}' for topic '{}'", subscriptionName, topicName);
+
+    Subscription subscription =
```
zhoufek, line 133: I think it makes sense to verify that the topic is being tracked before adding the subscription, since this should be keeping everything self-contained (i.e. not adding subscriptions to topics not part of this test). We don't need to have a strict mapping between topics and subscriptions right now, but we should make sure that the subscription is being added to some topic that we're aware of.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,239 @@
...
+  public SubscriptionName createSubscription(String topicName, String subscriptionName) {
+    checkArgument(!topicName.isEmpty(), "topicName can not be empty");
+    checkArgument(!subscriptionName.isEmpty(), "subscriptionName can not be empty");
+    checkIsUsable();
+
+    return createSubscription(getTopicName(topicName), subscriptionName);
+  }
+
+  @Override
+  public SubscriptionName createSubscription(TopicName topicName, String subscriptionName) {
```
zhoufek, line 128: I'm not sure if this is supposed to be a helper or alternative public method. (Maybe just a helper for now?)

Either way, the checks are a bit inconsistent. We double-check the subscription name if coming through the other method but don't call `checkIsUsable` here, which seems more critical here than in the other method.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  @VisibleForTesting
+  Publisher getPublisher(TopicName topic) throws IOException {
+    return Publisher.newBuilder(topic).setCredentialsProvider(credentialsProvider).build();
+  }
+
+  @Override
+  public synchronized void cleanupAll() {
+    checkIsUsable();
```
zhoufek, line 174: I don't think we need to fail if it isn't usable, since we're already in the desired state.

it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java
zhoufek: I think then it makes sense to have a package-private `PublisherFactory` class that handles this and inject that. I think it would be better to inject that mock than use a spy.

it/src/test/java/com/google/cloud/teleport/it/pubsub/PubsubUtilsTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Test
+  public void testCreateTopicName() {
+    String name = "create-topic-name";
+    assertThat(createTestId(name)).matches(name + "-\\d{17}");
+  }
+
+  @Test
+  public void testCreateTopicNameWithUppercase() {
+    assertThat(createTestId("testWithUpperCase")).matches("test-with-upper-case" + "-\\d{17}");
```
zhoufek, line 76: nit: I don't think we need the string concatenation here.

zhoufek (changes requested):

it/src/test/java/com/google/cloud/teleport/it/pubsub/PubsubUtilsTest.java
bvolpato: Thanks, yeah it was weird - probably some IDE formatting thing broke it up.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubPublisherFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.pubsub.v1.TopicName;
+import java.io.IOException;
+
+/**
+ * Default class for implementation of {@link PubsubPublisherFactory} interface.
+ *
+ * <p>The class provides an interaction with the real Pub/Sub client, with operations related to
+ * creating a Pub/Sub Publisher.
+ */
+public final class DefaultPubsubPublisherFactory implements PubsubPublisherFactory {
```
zhoufek, line 29: I don't think these should be final. It should just be an implementation detail not used outside of the resource manager.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,251 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultPubsubResourceManager.class);
+
+  private static final int DEFAULT_ACK_DEADLINE_SECONDS = 600;
+  private static final String RESOURCE_NAME_SEPARATOR = "-";
+
+  private final String testId;
+  private final String projectId;
+  private final PubsubPublisherFactory publisherFactory;
+  // private final CredentialsProvider credentialsProvider;
```
zhoufek, line 58: Please remove commented out code.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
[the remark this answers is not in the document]
zhoufek: Ok, getting back to the original ask: Could you remove these `@VisibleForTesting` annotations? It should be possible to use a captor and assert the captured value is correct rather than testing these methods directly.

it/src/main/java/com/google/cloud/teleport/it/pubsub/PubsubPublisherFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.pubsub;
+
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.pubsub.v1.TopicName;
+
+/** Interface for building Pub/Sub publishers in integration tests. */
+public interface PubsubPublisherFactory {
```
zhoufek, line 22: Similar to the class, this should be package-private.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,251 @@
...
+  public String publish(TopicName topic, Map<String, String> attributes, ByteString data)
+      throws PubsubResourceManagerException {
+    checkIsUsable();
+
+    LOG.info("Publishing message with {} bytes to topic '{}'", data.size(), topic);
+
+    PubsubMessage pubsubMessage =
+        PubsubMessage.newBuilder().putAllAttributes(attributes).setData(data).build();
+
+    try {
```
zhoufek, line 167: Forgot to mention this last time: We should verify that the topic being published to is tracked by the resource manager. We don't want tests to write to any topic in the project.

zhoufek (changes requested):
I think the structure is good now. It could just a little bit of cleanup and extra verification.

bvolpato:
@zhoufek thanks for the quick turnaround and thorough review. Addressed the comments, please let me know if there's anything else.

it/src/main/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManager.java
```diff
@@ -0,0 +1,253 @@
...
+
+    LOG.info("Attempting to cleanup manager.");
+
+    try {
+      for (SubscriptionName subscription : createdSubscriptions) {
+        LOG.info("Deleting subscription '{}'", subscription);
+        subscriptionAdminClient.deleteSubscription(subscription);
+      }
+    } finally {
+      subscriptionAdminClient.close();
```
zhoufek, line 195: I think the two try blocks should be combined so that we can make sure both clients are closed in the same finally block.

zhoufek (changes requested):

zhoufek:
LGTM. Thanks!

PR #449, opened 2022-08-22
Avoid test error when pipeline.apply() is called without pipeline.run()

bvolpato:
There are some tests that fail on the currently `@Rule` for TestPipeline.

Those tests have a prior call `pipeline.apply()`, but can't get into `pipeline.run` as exceptions are expected, and then this failure is triggered ([see this run](https://github.com/GoogleCloudPlatform/DataflowTemplates/runs/7955029507)):

```
org.apache.beam.sdk.testing.TestPipeline$PipelineRunMissingException: The pipeline has not been run.
```

This change will make sure to test the appropriate builders, but no need to append anything to the graph, as it is not the subject of those tests.

### DIFF
--- a/v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/AvroConvertersTest.java
+++ b/v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/AvroConvertersTest.java
@@ -75,13 +75,10 @@ public class AvroConvertersTest {
     expectedException.expect(IllegalArgumentException.class);
     expectedException.expectMessage("withSchema(schema) called with null input.");
 
-    pipeline.apply(
-        AvroConverters.ReadAvroFile.newBuilder()
-            .withInputFileSpec(AVRO_FILE_PATH)
-            .withSchema(null)
-            .build());
-
-    pipeline.run();
+    AvroConverters.ReadAvroFile.newBuilder()
+        .withInputFileSpec(AVRO_FILE_PATH)
+        .withSchema(null)
+        .build();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/AvroConvertersTest.java
```diff
@@ -111,19 +111,7 @@ public void testWriteWithoutSchema() {
...
-                .withSchema(SCHEMA_FILE_PATH)
-                .build())
-        .apply(
-            AvroConverters.WriteAvroFile.newBuilder()
-                .withOutputFile(FAKE_DIR)
-                .withSchema(null)
-                .build());
-
-    pipeline.run();
+    AvroConverters.WriteAvroFile.newBuilder().withOutputFile(FAKE_DIR).withSchema(null).build();
```
an2x, line 114: I see no substantial difference between this test and the one below, but one calls `pipeline.run();` and the other one doesn't - why?

an2x (changes requested):

v2/common/src/test/java/com/google/cloud/teleport/v2/transforms/AvroConvertersTest.java
bvolpato: @an2x fair point. Those tests don't break, but they never get to `pipeline.run();`, either, as the builder throws an exception. I have removed it to make the code consistent/clearer.
The problem is when there is actually something that succeeded to `pipeline.apply()`, but `pipeline.run()` isn't called -- which is the case for `ExpectedException` with multiple PTransforms.

an2x (approved):

PR #455, opened 2022-08-29
Add manual workflow to create Releases.

pranavbhandari24:


.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      candidateName:
+        description: 'Name of the candidate'
```
bvolpato, line 7: nit: If you can put a pattern/sample it would help newcomers.

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+      cherrypickCommits:
+        description: 'Comma separated commits to cherry pick'
+        type: string
+        required: false
+
+permissions:
+  contents: write
+
+jobs:
+  release:
```
an2x, line 24: Have you considered using an existing action like [this one](https://github.com/google-github-actions/release-please-action)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+      env:
+        RELEASE_BRANCH_NAME: ${{ steps.variables.outputs.releaseBranchName }}
+        TAG_NAME: ${{ steps.variables.outputs.tagName }}
+    - name: Push changes
+      run: |
+        git push -u origin --all
+        git push -u origin --tags
+    - name: Release
+      run: |
+        gh release create $TAG_NAME --title "Dataflow Templates $TAG_NAME" --notes ""
```
an2x, line 72: Auto-generating notes from the commit history would be nice too. But this can be done separately as a future improvement. Just FYI.

zhoufek: Long term, the goal is to have a file that people update with their change, similar to what Beam does. The problem with automatically generated notes is that it's hard to determine what impact a change had (bug fix, new template, etc.) There are ways around this, but we would need to enforce some standard for PR titles.

an2x: The idea was to auto-generate the notes but then do another manual pass. Right now if I'm writing the release notes myself I still have to go to the commit history to see what was changed.

pranavbhandari24: We can discuss this as a future improvement

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+    - name: Get releaser identity
+      run: |
+        git config --global user.name '${{github.actor}}'
+        git config --global user.email '${{github.actor}}@users.noreply.github.com'
```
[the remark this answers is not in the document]
pranavbhandari24: Discussed offline and decided to leave this as is. Will make changes as needed.

zhoufek:
Could you please make sure that this is up-to-date with the main branch?

Issue #450, opened 2022-08-22
[Bug]: Dataflow throws out data with cryptic, debug-proof error


j3pic:
### Related Template(s)

I don't know.

### What happened?

I set up a Datastream and Dataflow pipeline to move data from MySQL to BigQuery. Immediately, I ran into problems. Some tables were not copied at all, even  though the Avro files containing their data were created by Dataflow. I can't verify the contents of these files because Avro files require a schema to be read, and I don't know where to find the schema.

The following error message was logged:

    Avro File Read Failure <the path to an Avro file>

After a great deal of searching, I found this repo, and grepped for the error message, finding this:

```java
      /** A class to handle errors which occur during file reads. */
      public static class ReadFileRangesFnExceptionHandler implements Serializable {
    
        /*
         * Applies the desired handler logic to the given exception and returns
         * if the exception should be thrown.
         */
        public boolean apply(ReadableFile file, OffsetRange range, Exception e) {
          LOG.error("Avro File Read Failure {}", file.getMetadata().resourceId().toString());
          return false;
          // return true;
        }
      }
```

I had to guess where this was being called. I believe this is it:

```java
      public void process(ProcessContext c) throws IOException {
        ReadableFile file = c.element();
    
        FileBasedSource<T> source =
            CompressedSource.from(createSource.apply(file.getMetadata().resourceId().toString()))
                .withCompression(file.getCompression());
        try (BoundedSource.BoundedReader<T> reader = source.createReader(c.getPipelineOptions())) {
          for (boolean more = reader.start(); more; more = reader.advance()) {
            c.output(reader.getCurrent());
          }
        } catch (RuntimeException e) {
          if (exceptionHandler.apply(file, null, e)) {
            throw e;
          }
        }
      }
```

I looked into one possible defintiion of `reader.advance()` (the one in KafkaUnboundReader.java) and found that it's a massive method that calls many other methods on several different objects, so it could fail in numerous possible ways. To make matters worse, I can't even be sure that this is the method that is being called. It could be some other method on another class being pulled in from somewhere else.

If it wasn't for the commented-out "return true" in the `apply` method, a stack trace would have possibly been logged, and maybe I'd know something more.

I downloaded the file, removed the copy in the bucket, and then re-uploaded the file. Then Dataflow successfully read the file. 

### Beam Version

2.35.0

### Relevant log output

```shell
2022-08-04 01:24:52.616 PDT Avro File Read Failure gs://<REDACTED>
```

PR #463, opened 2022-09-06
Log Avro File Read Failure with stack trace

bvolpato:
Will help on bug https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/450, at least to understand what's going on.

Currently, the customer reports that the following log is issued:
```
2022-08-04 01:24:52.616 PDT Avro File Read Failure gs://<REDACTED>
```

But there's no way to know what exactly is going on. This one-liner change will print the stacktrace/cause to the log, so we can further troubleshoot if it happens again.


bvolpato:
Fixed (https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/a0cd5cde666d304ebc5e838a5c0690f59a193933)

PR #465, opened 2022-09-11
Run spotless:apply over the project

bvolpato:
Another round of `spotless:apply`. We'll need to improve pre-submits to avoid noncompliant code from getting merged.


bvolpato:
Will reopen soon

PR #472, opened 2022-09-26
Issue 447 - Create Bigquery Resource Manager

Polber:
Signed-off-by: Jeffrey Kinard <jeff@thekinards.com>

### DIFF
--- a/it/pom.xml
+++ b/it/pom.xml
@@ -32,7 +32,7 @@
     <maven-checkstyle-plugin.version>3.0.0</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.6.2</maven-compiler-plugin.version>
     <mockito.version>3.7.7</mockito.version>
-    <mockito-inline.version>4.2.0</mockito-inline.version>
+    <mockito-inline.version>4.5.1</mockito-inline.version>
     <puppycrawl.version>8.7</puppycrawl.version>
     <re2j.version>1.5</re2j.version>
     <slf4j.version>1.7.25</slf4j.version>
--- a/it/pom.xml
+++ b/it/pom.xml
@@ -32,6 +32,7 @@
     <maven-checkstyle-plugin.version>3.0.0</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.6.2</maven-compiler-plugin.version>
     <mockito.version>3.7.7</mockito.version>
+    <mockito-inline.version>4.2.0</mockito-inline.version>
     <puppycrawl.version>8.7</puppycrawl.version>
     <re2j.version>1.5</re2j.version>
     <slf4j.version>1.7.25</slf4j.version>
@@ -66,6 +67,10 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-storage</artifactId>
       <version>${cloud-storage.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigquery</artifactId>
     </dependency>
 
     <dependency>
@@ -112,6 +117,12 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-inline</artifactId>
+      <version>${mockito-inline.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
@@ -1,0 +1,332 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.bigquery;
+
+import static com.google.cloud.teleport.it.bigquery.BigQueryResourceManagerUtils.checkValidTableId;
+
+import com.google.api.gax.paging.Page;
+import com.google.auth.Credentials;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryError;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.DatasetInfo;
+import com.google.cloud.bigquery.FieldValueList;
+import com.google.cloud.bigquery.InsertAllRequest.RowToInsert;
+import com.google.cloud.bigquery.InsertAllResponse;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.Table;
+import com.google.cloud.bigquery.TableDefinition;
+import com.google.cloud.bigquery.TableId;
+import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.TableResult;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default class for implementation of {@link BigQueryResourceManager} interface.
+ *
+ * <p>The class supports one dataset, and multiple tables per dataset object.
+ *
+ * <p>The class is thread-safe.
+ */
+public final class DefaultBigQueryResourceManager implements BigQueryResourceManager {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultBigQueryResourceManager.class);
+  private static final String DEFAULT_DATASET_REGION = "us-central1";
+
+  private final String projectId;
+  private final String datasetId;
+
+  private final BigQuery bigQuery;
+  private Dataset dataset;
+
+  private DefaultBigQueryResourceManager(Builder builder, BigQuery bigQuery, String datasetId) {
+    this.projectId = builder.projectId;
+    this.datasetId = datasetId;
+    this.bigQuery = bigQuery;
+  }
+
+  public static DefaultBigQueryResourceManager.Builder builder(String testId, String projectId) {
+    return new DefaultBigQueryResourceManager.Builder(testId, projectId);
+  }
+
+  /**
+   * Returns the project ID this Resource Manager is configured to operate on.
+   *
+   * @return the project ID.
+   */
+  public String getProjectId() {
+    return projectId;
+  }
+
+  /**
+   * Return the dataset ID this Resource Manager uses to create and manage tables in.
+   *
+   * @return the dataset ID.
+   */
+  public String getDatasetId() {
+    return datasetId;
+  }
+
+  /**
+   * Helper method for determining if a dataset exists in the resource manager.
+   *
+   * @throws IllegalStateException if a dataset does not exist.
+   */
+  private void checkHasDataset() {
+    if (dataset == null) {
+      throw new IllegalStateException("There is no dataset for manager to perform operation on.");
+    }
+  }
+
+  /**
+   * Helper method for fetching a table stored in the dataset given a table name.
+   *
+   * @param tableId the name of the table to fetch.
+   * @return the table, if it exists.
+   * @throws IllegalStateException if the given table name does not exist in the dataset.
+   */
+  private synchronized Table getTableIfExists(String tableId) throws IllegalStateException {
+    checkHasDataset();
+    Table table = dataset.get(tableId);
+    if (table == null) {
+      throw new IllegalStateException(
+          "The table " + tableId + " does not exist in dataset " + datasetId + ".");
+    }
+    return table;
+  }
+
+  /**
+   * Helper method for logging individual errors thrown by inserting rows to a table. This method is
+   * used to log errors thrown by inserting certain rows when other rows were successful.
+   *
+   * @param insertErrors the map of errors to log.
+   */
+  private void logInsertErrors(Map<Long, List<BigQueryError>> insertErrors) {
+    for (Map.Entry<Long, List<BigQueryError>> entries : insertErrors.entrySet()) {
+      long index = entries.getKey();
+      for (BigQueryError error : entries.getValue()) {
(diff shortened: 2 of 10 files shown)


it/pom.xml
```diff
@@ -26,6 +26,7 @@
     <beam.version>2.38.0</beam.version>
     <cloud-bom.version>24.2.0</cloud-bom.version>
     <cloud-storage.version>2.2.2</cloud-storage.version>
+    <google-cloud-bigquery.version>1.108.0</google-cloud-bigquery.version>
```
bvolpato, line 29: Just curious - do we need to specify a version? isn't it declared in the `libraries-bom`?

it/pom.xml
```diff
@@ -67,6 +68,11 @@
       <artifactId>google-cloud-storage</artifactId>
       <version>${cloud-storage.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigquery</artifactId>
+      <version>${google-cloud-bigquery.version}</version>
```
bvolpato, line 74: Maybe remove this <version> if possible (see comment above)

it/src/test/java/com/google/cloud/teleport/it/bigquery/BigQueryResourceManagerUtilsTest.java
```diff
@@ -0,0 +1,45 @@
...
+package com.google.cloud.teleport.it.bigquery;
+
+import static com.google.cloud.teleport.it.bigquery.BigQueryResourceManagerUtils.checkValidTableId;
+import static org.junit.Assert.assertThrows;
+
+import java.util.Arrays;
+import org.junit.Test;
+
+/** Unit tests for {@link com.google.cloud.teleport.it.bigquery.BigQueryResourceManagerUtils}. */
+public class BigQueryResourceManagerUtilsTest {
```
bvolpato, line 25: A few valid tests for `checkValidTableId` may be good, to make sure it works as intended (we are just validating error cases now).

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+    prepareCreateDataset();
+    testManager.createDataset(DATASET_ID);
+
+    assertThrows(IllegalStateException.class, () -> testManager.createDataset(DATASET_ID));
+  }
+
+  @Test
+  public void testCreateDatasetShouldWorkWhenBigQueryDoesNotThrowAnyError() {
+    prepareCreateDataset();
+    testManager.createDataset(DATASET_ID);
```
bvolpato, line 172: A verify() to make sure it called `bigQuery.create(datasetInfo)` would be great in here. An example here https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/it/src/test/java/com/google/cloud/teleport/it/pubsub/DefaultPubsubResourceManagerTest.java#L120

bvolpato: (just saw that you've done a check below for dataset, could do it here)

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  @Test
+  public void testCreateTableShouldCreateDatasetWhenDatasetDoesNotExist() {
+    prepareCreateDataset();
+    prepareCreateTable();
+
+    testManager.createTable(TABLE_NAME, schema);
+
+    verify(bigQuery).create((DatasetInfo) any());
```
bvolpato, line 192: This may throw an unchecked cast exception, no? I usually see `any(DatasetInfo.class)` but if no warnings shown this is just a nitpick.

bvolpato: Also a verify to make sure it runs `bigQuery.create(tableInfo)` would be great here

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+  public void testCreateDatasetShouldThrowErrorWhenDatasetCreateFails() {
+    when(bigQuery.create((DatasetInfo) any())).thenThrow(RuntimeException.class);
+
+    assertThrows(
+        BigQueryResourceManagerException.class, () -> testManager.createDataset(DATASET_ID));
+  }
+
+  @Test
+  public void testCreateDatasetShouldNotCreateDatasetWhenDatasetAlreadyExists() {
+    prepareCreateDataset();
```
bvolpato, line 163: nitpick-ish, but prepareCreateDataset() is 2 lines only, could be inlined in methods to make it flow better and have more self-contained tests

bvolpato (changes requested):
Good stuff! Thanks!

I gave it a quick review, and the overall structure looks really good. Left some comments and a few nitpicks that may make it better.

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
```diff
@@ -0,0 +1,369 @@
...
+    return datasetId;
+  }
+
+  /**
+   * Return the test ID this Resource Manager uses to generate the instance and cluster(s) ID's
+   * from.
+   *
+   * @return test ID.
+   */
+  public String getTestId() {
```
an2x, line 96: Similar to BigtableResourceManager in PR #445: no need to store and expose `testId`?

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
```diff
@@ -0,0 +1,369 @@
...
+        bigQuery.delete(dataset.getDatasetId());
+      }
+    } catch (Exception e) {
+      throw new BigQueryResourceManagerException("Failed to delete resources.", e);
+    }
+    LOG.info("Manager successfully cleaned up.");
+  }
+
+  /** Builder for {@link DefaultBigQueryResourceManager}. */
+  public static final class Builder {
```
an2x, line 293: Similar concerns as for the BigtableResourceManager builder in PR #445:
1. Generate `datasetId` in `build()` only.
2. Remove getters.
3. Remove `setDatasetId`.
4. Don't modify `bigQuery` instance variable state in `build()`.

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+    assertThat(tmBuilder.getDatasetId())
+        .matches(TEST_ID.replace('-', '_') + "_\\d{8}_\\d{6}_\\d{6}");
+    tmBuilder
+        .setProjectId("fake_project_id")
+        .setTestId("fake_test_id")
+        .setDatasetId(DATASET_ID)
+        .setCredentials(credentials)
+        .setBigQuery(bigQuery);
+
+    assertThat(tmBuilder.getCredentials()).isEqualTo(credentials);
```
an2x, line 120: Builders are typically created write-only, with their state not being checked in tests. What a unit test should verify is that the builder produces a correct object it was expected to build. How/if the intermediate state is stored in the builder is its implementation details.

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+    assertThat(tmBuilder.getProjectId()).isEqualTo("fake_project_id");
+    assertThat(tmBuilder.getTestId()).isEqualTo("fake_test_id");
+    assertThat(tmBuilder.getDatasetId()).matches(DATASET_ID);
+    assertThat(tmBuilder.getBigQuery()).isEqualTo(bigQuery);
+
+    DefaultBigQueryResourceManager tm = tmBuilder.build();
+
+    assertThat(tm.getProjectId()).isEqualTo("fake_project_id");
+    assertThat(tm.getTestId()).isEqualTo("fake_test_id");
```
an2x, line 129: Do we care aboure `testId` in the resource manager? AFAIU it's really used anywhere. I'd remove the `testId` instance variable and this check in the test.

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+        DefaultBigQueryResourceManager.builder(TEST_ID, PROJECT_ID);
+    tmBuilder.setDatasetId(DATASET_ID).setCredentials(credentials);
+    DefaultBigQueryResourceManager tm = tmBuilder.build();
+
+    assertThat(tm.getDatasetId()).isEqualTo(DATASET_ID);
+  }
+
+  @Test
+  public void testGetTestIdReturnsCorrectValue() {
+    assertThat(testManager.getTestId()).isEqualTo(TEST_ID);
```
an2x, line 150: Same: why check `testId`?

an2x (changes requested):

it/pom.xml
Polber: Honestly, I wasn't sure why it needed that dependency entry in the first place when the bigtable RM didn't, but maven was complaining, so I added it. However, it seems to compile fine without the version specified as you suggested.

it/src/test/java/com/google/cloud/teleport/it/bigquery/BigQueryResourceManagerUtilsTest.java
Polber: Fixed.

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
Polber: Fixed. I also did the same for prepareCreateTable.

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
Polber: Agreed. I removed all instances and tests for testId.

it/src/test/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManagerTest.java
Polber: I removed the test since all the necessary final states are checked in other tests.

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    public DefaultBigQueryResourceManager build() {
+      // create default bigQuery client if one was not set.
+      BigQuery bigQueryToUse;
+      if (bigQuery == null) {
+        BigQueryOptions.Builder bigQueryOptions =
+            BigQueryOptions.newBuilder().setProjectId(projectId);
+
+        // set credentials, if provided
```
pranavbhandari24, line 318: Will this work if credentials are not provided?

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
```diff
@@ -0,0 +1,332 @@
...
+      this.bigQuery = bigQuery;
+      return this;
+    }
+
+    public Builder setCredentials(Credentials credentials) {
+      this.credentials = credentials;
+      return this;
+    }
+
+    public DefaultBigQueryResourceManager build() {
```
pranavbhandari24, line 311: I think everything under this methods should be done in the constructor of the class. The Builder#build method should only call the private constructor and pass the builder to it. (Similar to what is done in DefaultSpannerResourceManager)

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
pranavbhandari24: A better question: Is there a case where the credentials will not be provided?

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Default class for implementation of {@link BigQueryResourceManager} interface.
+ *
+ * <p>The class supports one dataset, and multiple tables per dataset object.
+ *
+ * <p>The class is thread-safe.
+ */
+public final class DefaultBigQueryResourceManager implements BigQueryResourceManager {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultBigQueryResourceManager.class);
+  private static final String DEFAULT_DATASET_REGION = "us-central1";
```
pranavbhandari24, line 52: Is there a reason why we have a default region and its not passed in as a builder parameter?

pranavbhandari24 (commented):
Thanks for creating this resource manager. The test coverage looks amazing! 

Overall, it looks good! Left a few comments (nitpicks)

an2x (dismissed):

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
Polber: That is just a macro constant for defining a region for the dataset that gets created by default when createTable() is called without a previous call to createDataset(). I don't see a reason why the builder would need that value unless we want to make the dataset region configurable which can be achieved by calling the createDataset() method.

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
Polber: I think there could be a case from an open-source contributor running the IT's, but in general, no I don't think there will be a case. However, since it is technically optional, I saw the typical builder pattern had setters for any optional value, so I followed that pattern. It's also consistent with the Pub/Sub RM builder.

bvolpato (approved):
LGTM

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
Polber: I agree. I moved the logic from build to the private constructor, and I created a package-private VisibleForTesting constructor for passing the mocked client instead of using a setter in the builder.

it/src/main/java/com/google/cloud/teleport/it/bigquery/DefaultBigQueryResourceManager.java
pranavbhandari24: I think Spanner Resource Manager takes the region as a builder parameter. Using default also makes sense.

pranavbhandari24 (approved):
LGTM!

PR #476, opened 2022-10-04
fix: WindowedFilenamePolicy.java : LOG.debug …

vdolez-happn:
LOG.debug was using `userDayPattern` instead of `userYearPattern`


bvolpato:
@vdolez-happn Thanks for the PR. We have reorged the repository and have taken this into account.

PR #480, opened 2022-10-11
Added JdbcToBigQueryTruncateLoad.java 

suddhasatwabhaumik:
Written new template [JdbcToBigQueryTruncateLoad.java](https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/1b2fc123ce096b6f1473b846e76d4de8598ac00f) for Customer use cases where we need to Truncate and Load from source SQL/RDBMS Tables rather than Appending to BigQuery (target) tables. Please review and approve as applicable.

### DIFF
--- a/src/main/java/com/google/cloud/teleport/templates/JdbcToBigQueryTruncateLoad.java
+++ b/src/main/java/com/google/cloud/teleport/templates/JdbcToBigQueryTruncateLoad.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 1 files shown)


bvolpato:
Thanks for contributing @suddhasatwabhaumik.

I have one fundamental question, though. Do we need an entire new template for that? Or maybe we could make `withWriteDisposition` come from a parameter, so they decide what's the appropriate disposition for their job?

Arguably, that would have to be on a Flex template (but we already have [one for JdbcToBigQuery](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java)).



Note: spotless checks are failing, you would have to run `mvn spotless:apply` on the project to format / clear it up.

PR #487, opened 2022-10-13
syndeo should only build in open source. it does not need to build internally

pabloem:
…ternally

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,10 +136,22 @@
     </plugins>
   </build>
 
+     <profiles>
+     <profile>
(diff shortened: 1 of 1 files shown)


pabloem:
r: @zhoufek 

Syndeo template tests are developed on OSS. Furthermore, Syndeo currently depends on snapshot releases or Beam, which are available in OSS, but not available in internal google build systems. This change will allow merging syndeo changes without breaking builds internally.

pabloem:
@zhoufek this should unblock all other syndeo PRs

PR #494, opened 2022-10-18
Removing useless scripts from Syndeo template

pabloem:


pabloem:
r: @damondouglas

damondouglas (approved):
@pabloem LGTM

PR #530, opened 2022-12-02
Rethrow exception from ExtractGcsFile so that Dataflow will retry the pardo

johnjcasey:


bvolpato (approved):

bvolpato (approved):

robertwb:
Should we do an audit of the codebase to ensure there are not other instances of this kind of error?

PR #531, opened 2022-12-03
Add Pub/Sub to BigQuery and Pub/Sub to Pub/Sub integration tests.

pranavbhandari24:
Added the following integration tests,

* Pub/Sub topic to BigQuery
* Pub/Sub subscription to BigQuery
* Pub/Sub to Pub/Sub

Also, changed the TemplateIntegrationTest annotation to handle the case where a template contains multiple Template annotations.

### DIFF
--- a/v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorIT.java
+++ b/v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorIT.java
@@ -54,8 +54,7 @@ public final class BulkCompressorIT extends TemplateTestBase {
   @Test
   public void testCompressGzip() throws IOException {
     // Arrange
-    String name = testName.getMethodName();
-    String jobName = createJobName(name);
+    String jobName = createJobName(testName.getMethodName());
 
     LaunchConfig.Builder options =
         LaunchConfig.builder(jobName, specPath)
--- a/v1/src/test/java/com/google/cloud/teleport/templates/PubSubSubscriptionToBigQueryIT.java
+++ b/v1/src/test/java/com/google/cloud/teleport/templates/PubSubSubscriptionToBigQueryIT.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.templates;
+
+import static com.google.cloud.teleport.it.dataflow.DataflowUtils.createJobName;
+import static com.google.cloud.teleport.it.matchers.TemplateAsserts.assertThatRecords;
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.StandardSQLTypeName;
+import com.google.cloud.bigquery.TableResult;
+import com.google.cloud.teleport.it.TemplateTestBase;
+import com.google.cloud.teleport.it.bigquery.BigQueryResourceManager;
+import com.google.cloud.teleport.it.bigquery.DefaultBigQueryResourceManager;
+import com.google.cloud.teleport.it.dataflow.DataflowOperator;
+import com.google.cloud.teleport.it.dataflow.DataflowOperator.Result;
+import com.google.cloud.teleport.it.dataflow.DataflowTemplateClient.JobInfo;
+import com.google.cloud.teleport.it.dataflow.DataflowTemplateClient.JobState;
+import com.google.cloud.teleport.it.dataflow.DataflowTemplateClient.LaunchConfig;
+import com.google.cloud.teleport.it.pubsub.DefaultPubsubResourceManager;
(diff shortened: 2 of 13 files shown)


v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorIT.java
```diff
@@ -72,7 +71,7 @@ public void testCompressGzip() throws IOException {
 
     Result result =
         new DataflowOperator(getDataflowClient())
-            .waitForCondition(
+            .waitForConditionAndFinish(
```
bvolpato, line 74: I think I tried `AndFinish` before and it may cause some problems for batch jobs depending on timing (i.e., it is already completed and we call cancel). Usually batch jobs will finish by themselves (either completed or failed after 5 retries), so just waiting for condition may be enough.

v1/src/test/java/com/google/cloud/teleport/templates/PubSubSubscriptionToBigQueryIT.java
```diff
@@ -0,0 +1,139 @@
...
+
+    AtomicReference<TableResult> records = new AtomicReference<>();
+    Result result =
+        new DataflowOperator(getDataflowClient())
+            .waitForConditionAndFinish(
+                createConfig(info),
+                () -> {
+                  TableResult values = bigQueryResourceManager.readTable(bqTable);
+                  records.set(values);
+                  return values.getTotalRows() == messages.size();
```
bvolpato, line 124: I'd take care about using `==` when using Pub/Sub, as it doesn't offer exactly once guarantees. Maybe we can `>=` (and in the future we may want to deduplicate in the assertions, but at least we'll get a faster failure and not wait for a condition that already passed, it it makes sense).

v1/src/test/java/com/google/cloud/teleport/templates/PubSubToPubSubIT.java
```diff
@@ -0,0 +1,116 @@
...
+    AtomicReference<PullResponse> records = new AtomicReference<>();
+    Result result =
+        new DataflowOperator(getDataflowClient())
+            .waitForConditionAndFinish(
+                createConfig(info),
+                () -> {
+                  PullResponse response =
+                      pubsubResourceManager.pull(outputSubscription, messages.size());
+                  records.set(response);
+                  return response.getReceivedMessagesList().size() == messages.size();
```
bvolpato, line 98: Same thing with `>=`

v1/src/test/java/com/google/cloud/teleport/templates/PubSubToPubSubIT.java
```diff
@@ -0,0 +1,116 @@
...
+                });
+
+    // Assert
+    List<String> actualMessages =
+        records.get().getReceivedMessagesList().stream()
+            .map(receivedMessage -> receivedMessage.getMessage().getData().toStringUtf8())
+            .collect(Collectors.toList());
+    assertThat(result).isEqualTo(Result.CONDITION_MET);
+    assertThat(actualMessages).containsExactlyElementsIn(messages);
```
bvolpato, line 107: We may want to use our own assertion that deduplicates, but that can be hardened later IMO

v2/common/src/test/java/com/google/cloud/teleport/v2/utils/BigQueryIOUtilsTest.java
```diff
@@ -209,15 +210,22 @@ public void testWriteResultToBigQueryInsertErrorsWithStorageApiNoErrors() {
...
-    WriteResult writeResult = pipeline.apply(
-            TestStream.create(TableRowJsonCoder.of())
-                .addElements(tableRow)
-                .advanceWatermarkToInfinity())
-        .apply(BigQueryIO.writeTableRows()
-            .withCreateDisposition(CreateDisposition.CREATE_NEVER)
-            .withExtendedErrorInfo()
-            .to(new TableReference().setProjectId(PROJECT).setDatasetId(DATASET).setTableId(TABLE))
-            .withTestServices(bigQueryServices));
+    WriteResult writeResult =
```
bvolpato, line 213: nitpick/style: this can be simply returned instead of stored

bvolpato (changes requested):
Great stuff! This is looking really good, I've left a few minor comments, please take a look if they make sense.

v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorIT.java
pranavbhandari24: Ah, that makes sense! Have been writing only streaming template tests recently, didn't think about this.

pranavbhandari24: Changed.

v1/src/test/java/com/google/cloud/teleport/templates/PubSubSubscriptionToBigQueryIT.java
pranavbhandari24: Makes sense! I did not know Pub/Sub does not guarantee exactly once

v1/src/test/java/com/google/cloud/teleport/templates/PubSubToPubSubIT.java
pranavbhandari24: Done

v1/src/test/java/com/google/cloud/teleport/templates/PubSubToPubSubIT.java
pranavbhandari24: Changed the assertion to only check if all expected messages are present in the actual messages

v2/common/src/test/java/com/google/cloud/teleport/v2/utils/BigQueryIOUtilsTest.java
pranavbhandari24: This file was reformatted when running spotless 
But, I don't know why you would store it in a variable. Changed it to return directly

bvolpato (approved):

bvolpato (dismissed):

PR #534, opened 2022-12-05
Use "verify" instead of "install" for unit tests workflow

bvolpato:
Important verifications are done (for example, maven-dependency-plugin checks)


pranavbhandari24 (dismissed):
LGTM!

PR #540, opened 2022-12-14
Move KMS Encryption utils to common directory and add Secret Manager utils

andreigurau:
This PR moves the KMS Encryption utils from the googlecloud-to-googlecloud directory to the common directory, since this set of utils would be used by templates in the googlecloud-to-splunk directory. Furthermore, this adds a Secret Manager utils, to be able to fetch secrets from Secret Manager (also to be used by the googlecloud-to-splunk templates).

Note: Even though I saw a Secret Manager ValueProvider in the v1 side, I'm not sure if I implemented it correctly on the v2 side, since now it's just a class with static methods that reads from Secret Manager. Also, the unit tests for Secret Manager in this PR don't exactly work (see [1] for error on the first unit test - only test that is failing) and feel poorly implemented to me, but I'm not sure how else to do it (use Mockito maybe?). As far as I know, there weren't any unit tests on the v1 side that directly interacts/mocks with Secret Manager (same with KMSEncryption actually - there are also no unit tests on v2 side for KMS either). I'm going to leave this PR open to feedback, but I'll need help for what to do about the unit tests especially, since I'm not sure what I should be doing here.

[1]: ``com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Permission denied on resource project my_project.
Caused by: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Permission denied on resource project my_project.``


bvolpato:
Have you tried to `gcloud auth login`, `gcloud auth application-default login`? Your project might be set to something different too, so I'd suggest `gcloud init` once you've done the above.

andreigurau:
Just tried it out, still getting the same error, even with the project not set to anything. I think I'm getting this error because I'm actually trying to fetch from a real Secret Manager from a project called 'my-project'. That's why I was wondering if I had to use mocks in the unit test.

pranavbhandari24:
I don't think we should move these files. Ideally, we would not be using ValueProviders in v2 at all. 

The best thing to do would be to create a KMSUtils file similar to the SecretManagerUtils file you created and use those in googlecloud-to-splunk. We can then refactor the files in googlecloud-to-googlecloud to move away from ValueProviders entirely

andreigurau:
Closing out this PR, as some overlapping refactor work was done to introduce KMSUtils in the common directory, but I raised a new PR [PR#545](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/545) that still adds a SecretManagerUtils class to the common directory.

PR #541, opened 2022-12-15
Add schema templates to StreamingDataGenerator template.

pranavbhandari24:


v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
```diff
@@ -396,19 +434,22 @@ static class MessageGeneratorFn extends DoFn<Long, byte[]> {
...
-          schemaLocation, "schemaLocation argument of MessageGeneratorFn class cannot be null.");
-      this.schemaLocation = schemaLocation;
+    private final String schema;
+
+    MessageGeneratorFn(SchemaTemplate schemaTemplate, String schemaLocation) {
+      checkArgument(
+          schemaTemplate != null || schemaLocation != null,
+          "Either schemaTemplate or schemaLocation argument of MessageGeneratorFn class must be provided.");
+      if (schemaLocation != null) {
+        this.schema = GCSUtils.getGcsFileAsString(schemaLocation);
```
bvolpato, line 444: Doing here instead of on `@Setup` is slightly different. Here it will need access at job creation + serialize the schema as part of the job graph, which I think could be large (not sure if an issue).

bvolpato (changes requested):
I like it, thanks Pranav!

Just left 2 comments if you can TAL

bvolpato (approved):

PR #543, opened 2022-12-17
Generate transform configs

pabloem:


pabloem:
r: @andreigurau 

This code generates the configurations, but I have not dealt with the fact that the API for transform service is not published externally. Could y'all publish it externally? Once we have that, we'll easily produce the configs.

pabloem:
I'm also happy to move this code wherever it makes sense to have it

pabloem:
sample from today: https://gist.github.com/pabloem/59d60af50c06ce8140178154e825d89e

syndeo-template/src/main/java/com/google/cloud/syndeo/GenerateConfiguration.java
```diff
@@ -0,0 +1,162 @@
...
+            .addAllTransformDescriptions(descriptions)
+            .build()
+            .toString()
+            .getBytes(StandardCharsets.UTF_8));
+  }
+
+  static com.google.cloud.datapipelines.v1.TransformDescription providerToConfiguration(
+      SchemaTransformProvider provider) {
+    LOG.info("Generating configuration for {}", provider.identifier());
+    System.out.println(String.format("Generating configuration for %s", provider.identifier()));
```
andreigurau, line 59: nit: remove the print statement

syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
```diff
@@ -68,6 +69,29 @@ public interface Options extends PipelineOptions, DataflowPipelineOptions {
...
+      Set.of(
+          // Old names:
+          "bigquery:read",
+          "kafka:read",
+          "bigtable:write",
+          "schemaIO:bigquery:read",
+          "schemaIO:bigquery:write",
+
+          // New names:
+          "beam:schematransform:org.apache.beam:sql_transform:v1",
```
andreigurau, line 82: Should we be adding all of these right now? Currently we only support Kafka Read, BigQuery Read + Write, Bigtable Write, and Pubsub Read. I'm not sure if we should be adding the unsupported ones as well, but I would defer to another opinion

andreigurau (commented):
Thanks! Overall it looks good, but some slight comments

zhoufek (commented):
Would it be possible to control which fields actually get added? Some fields may not make sense in the context of this template, and others may not be able to be feasibly supported.

pabloem:
> Would it be possible to control which fields actually get added? Some fields may not make sense in the context of this template, and others may not be able to be feasibly supported.

Can you elaborate? I suppose you mean fields like "useTestingService" which we only use to test internally, right? Yes, we can do that. Perhaps as part of the SUPPORTED_TRANSFORMS field. LMK if that sounds like what you have in mind.

zhoufek:
> > Would it be possible to control which fields actually get added? Some fields may not make sense in the context of this template, and others may not be able to be feasibly supported.
> 
> Can you elaborate? I suppose you mean fields like "useTestingService" which we only use to test internally, right? Yes, we can do that. Perhaps as part of the SUPPORTED_TRANSFORMS field. LMK if that sounds like what you have in mind.

Yeah, fields like that. We may also not be able to support a field in an expected way, like how Kafka schema registries can't be used to dynamically set the schema, so we may want to discard that field to avoid confusion. Some services may also have limitations, like not being able to understand a schema from a BigQuery query, so we may want to discard it for a similar reason.

pabloem:
@zhoufek - I've added lists of supported fields in SUPPORTED_URNS - feel free ot take a look.

@andreigurau - I've addressed your comments.

LMK if that looks fine for y'all.

pabloem:
ptal : D

syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
```diff
@@ -68,6 +69,29 @@ public interface Options extends PipelineOptions, DataflowPipelineOptions {
...
+          "bigtable:write",
+          Set.of("instanceId", "tableId", "keyColumns", "projectId", "appProfileId"),
+          "schemaIO:bigquery:write",
+          Set.of("table", "createDisposition", "writeDisposition"),
+          // New names:
+          "beam:schematransform:org.apache.beam:bigquery_storage_write:v1",
+          Set.of("instanceId", "tableId", "keyColumns", "projectId", "appProfileId"),
+          "beam:schematransform:org.apache.beam:bigquery_storage_read:v1",
+          Set.of("table"),
+          "beam:schematransform:org.apache.beam:kafka_write:v1",
```
andreigurau, line 88: Kafka Write is not a supported config at the moment, could you remove this?

syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
```diff
@@ -68,6 +69,29 @@ public interface Options extends PipelineOptions, DataflowPipelineOptions {
...
+          Set.of(),
+          "beam:schematransform:org.apache.beam:kafka_read:v1",
+          Set.of(),
+          "syndeo:schematransform:com.google.cloud:bigtable_write:v1",
```
andreigurau, line 92: nit: it might be easier to read these URNs if a blank line is added after each `Set.of` line

pabloem: the formatting is decided by spotless, not by me.

syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
pabloem: done.

andreigurau (dismissed):
LGTM

syndeo-template/src/main/java/com/google/cloud/syndeo/GenerateConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+import com.google.cloud.datapipelines.v1.BatchGetTransformDescriptionsResponse;
+import com.google.cloud.datapipelines.v1.MapType;
+import com.google.cloud.datapipelines.v1.TransformDescription;
+import com.google.cloud.datapipelines.v1.TypeName;
+import com.google.cloud.syndeo.common.ProviderUtil;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.*;
```
zhoufek, line 27: Sorry, missed this last time: Please don't star import. Only import what you intend to use.

syndeo-template/src/main/java/com/google/cloud/syndeo/GenerateConfiguration.java
```diff
@@ -0,0 +1,165 @@
...
+  public static void main(String[] args) throws IOException {
+    Collection<SchemaTransformProvider> providers = ProviderUtil.getProviders();
+    List<TransformDescription> descriptions =
+        providers.stream()
+            .filter(provider -> SyndeoTemplate.SUPPORTED_URNS.containsKey(provider.identifier()))
+            .map(GenerateConfiguration::providerToConfiguration)
+            .collect(Collectors.toList());
+
+    Files.write(
+        Paths.get("transform_configs.prototext"),
```
zhoufek, line 46: Also missed this: I think the intent is that each config gets its own file. That said, I also don't see a way to set the directory where this file is written, so is the intent that this will just write the file, and then copy it over later, potentially splitting it up into separate files then?

I just want to make sure we're on the same page here.

zhoufek (changes requested):
I'm going to remove the Google LGTM label. Sorry, but I think these should be addressed before we submit this.

pabloem:
good catch. Thanks @zhoufek - I've addressed both those comments. LMK if this is good to go.

PR #558, opened 2023-01-19
Add various performance tests.

pranavbhandari24:
Add the following performance tests:
 * Spanner to Text
 * Streaming Data generator to Spanner 
 * Spanner to Avro
 * Avro to Bigtable

### DIFF
--- a/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
@@ -15,7 +15,7 @@
  */
 package com.google.cloud.teleport.it;
 
-import static com.google.cloud.teleport.it.PerformanceBenchmarkingBase.createConfig;
+import static com.google.cloud.teleport.it.LoadTestBase.createConfig;
 import static com.google.cloud.teleport.it.PipelineUtils.createJobName;
 import static com.google.cloud.teleport.it.matchers.TemplateAsserts.assertThatPipeline;
 import static com.google.cloud.teleport.it.matchers.TemplateAsserts.assertThatResult;
--- a/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
@@ -134,6 +134,11 @@ public class DataGenerator {
       return this;
     }
 
+    public DataGenerator.Builder setSinkType(String value) {
+      parameters.put("sinkType", value);
+      return this;
+    }
+
     public Builder setWorkerMachineType(String value) {
       parameters.put("workerMachineType", value);
       return this;
@@ -154,8 +159,48 @@ public class DataGenerator {
       return this;
     }
 
+    public DataGenerator.Builder setOutputDirectory(String value) {
+      parameters.put("outputDirectory", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setOutputType(String value) {
+      parameters.put("outputType", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setNumShards(String value) {
+      parameters.put("numShards", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setAvroSchemaLocation(String value) {
+      parameters.put("avroSchemaLocation", value);
+      return this;
+    }
+
     public DataGenerator.Builder setTopic(String value) {
       parameters.put("topic", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setProjectId(String value) {
+      parameters.put("projectId", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setSpannerInstanceName(String value) {
+      parameters.put("spannerInstanceName", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setSpannerDatabaseName(String value) {
+      parameters.put("spannerDatabaseName", value);
+      return this;
+    }
+
+    public DataGenerator.Builder setSpannerTableName(String value) {
+      parameters.put("spannerTableName", value);
       return this;
     }
 
--- a/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
@@ -16,6 +16,7 @@
 package com.google.cloud.teleport.it;
 
 import static com.google.cloud.teleport.it.PerformanceBenchmarkingBase.createConfig;
+import static com.google.cloud.teleport.it.PipelineUtils.createJobName;
 import static com.google.cloud.teleport.it.matchers.TemplateAsserts.assertThatPipeline;
 import static com.google.cloud.teleport.it.matchers.TemplateAsserts.assertThatResult;
 
@@ -30,6 +31,7 @@ import java.io.IOException;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
+import org.junit.rules.TestName;
 
 /** Helper class for starting a Streaming Data generator dataflow template job. */
 public class DataGenerator {
@@ -64,6 +66,18 @@ public class DataGenerator {
     return new DataGenerator.Builder(jobName)
         .setSchemaTemplate(schemaTemplate)
         .setAutoscalingAlgorithm(AutoscalingAlgorithmType.THROUGHPUT_BASED);
+  }
+
+  public static DataGenerator.Builder builderWithSchemaLocation(
+      TestName testName, String schemaLocation) {
+    return DataGenerator.builderWithSchemaLocation(
+        createJobName(testName.getMethodName() + "-data-generator"), schemaLocation);
+  }
+
+  public static DataGenerator.Builder builderWithSchemaTemplate(
+      TestName testName, String schemaTemplate) {
+    return DataGenerator.builderWithSchemaTemplate(
+        createJobName(testName.getMethodName() + "-data-generator"), schemaTemplate);
   }
 
   /**
--- a/it/src/main/java/com/google/cloud/teleport/it/PerformanceBenchmarkingBase.java
+++ b/it/src/main/java/com/google/cloud/teleport/it/PerformanceBenchmarkingBase.java
@@ -42,6 +42,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nullable;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
@@ -51,7 +52,7 @@ import org.junit.runners.JUnit4;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Integration test for testing performance of dataflow templates. */
+/** Base class for performance tests. It provides helper methods for common operations. */
 @RunWith(JUnit4.class)
 public class PerformanceBenchmarkingBase {
   private static final Logger LOG = LoggerFactory.getLogger(PerformanceBenchmarkingBase.class);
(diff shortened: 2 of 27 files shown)


Polber (dismissed):

it/src/main/java/com/google/cloud/teleport/it/DataGenerator.java
```diff
@@ -33,8 +33,9 @@
 
 /** Helper class for starting a Streaming Data generator dataflow template job. */
 public class DataGenerator {
+  // TODO: Change this back
```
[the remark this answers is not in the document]
pranavbhandari24: Done

it/src/main/java/com/google/cloud/teleport/it/PerformanceBenchmarkingBase.java
```diff
@@ -201,6 +201,7 @@ public boolean waitForNumMessages(String jobId, String pcollectionName, Long exp
   protected Map<String, Double> getMetrics(JobInfo jobInfo, String inputPcollection)
       throws ParseException, InterruptedException, IOException {
     // Metrics take up to 3 minutes to show up
+    LOG.info("Sleeping for 4 minutes to query metrics.");
     Thread.sleep(Duration.ofMinutes(4).toMillis());
```
bvolpato, line 205: awaitility.org could be an alternative so we keep polling for metrics instead of hardcoding X minutes. but it's ok to ignore for now (or maybe add a TODO with intentions)

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtablePerformanceIT.java
```diff
@@ -0,0 +1,165 @@
...
+import org.slf4j.LoggerFactory;
+
+/** Performance tests for {@link AvroToBigtable} (GCS_Avro_to_Cloud_Bigtable) template. */
+@TemplateIntegrationTest(AvroToBigtable.class)
+@RunWith(JUnit4.class)
+public class AvroToBigtablePerformanceIT extends PerformanceBenchmarkingBase {
+  private static final Logger LOG = LoggerFactory.getLogger(AvroToBigtablePerformanceIT.class);
+  private static final String SPEC_PATH =
+      MoreObjects.firstNonNull(
+          "gs://dataflow-templates/latest/GCS_Avro_to_Cloud_Bigtable", TestProperties.specPath());
```
bvolpato, line 58: something to think / longer term - do you think we can try to mimic TemplateTestBase to stage those templates on demand? Or make that an option/property? It would be good to validate benchmark across changes of a template, instead of using a fixed baseline (both are good, but we can optionally set `-DspecPath=gs://dataflow-templates/latest/GCS_Avro_to_Cloud_Bigtable` like handled here)

v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorPerformanceIT.java
```diff
@@ -200,11 +210,74 @@ public void testGenerateBigQuery10gb() throws IOException, ParseException, Inter
...
+            .addParameter("schemaTemplate", SchemaTemplate.GAME_EVENT.name())
+            .addParameter("qps", "1000000")
+            .addParameter("messagesLimit", NUM_MESSAGES)
+            .addParameter("sinkType", "SPANNER")
+            .addParameter("projectId", PROJECT)
+            .addParameter("spannerInstanceName", spannerResourceManager.getInstanceId())
+            .addParameter("spannerDatabaseName", spannerResourceManager.getDatabaseId())
+            .addParameter("spannerTableName", testName.getMethodName())
+            .addParameter("numWorkers", "50")
+            .addParameter("maxNumWorkers", "100")
```
bvolpato, line 267: just curious - would somehow externalize QPS/numWorkers/maxNumWorkers help? or use parameterized tests to have different sets of tests?

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtablePerformanceIT.java
```diff
@@ -0,0 +1,165 @@
...
+
+  @Test
+  public void testBacklog10gb() throws IOException, ParseException, InterruptedException {
+    // Arrange
+    String name = testName.getMethodName();
+    String jobName = createJobName(name);
+    // Bigtable expects Avro files to meet a specific schema
+    DataGenerator dataGenerator =
+        DataGenerator.builderWithSchemaLocation(jobName + "-data-generator", generatorSchemaPath)
+            .setQPS("1000000")
```
bvolpato, line 111: nit: how realistic is to have 1 million/s? Is it really the intention to cause backpressure on the DataGenerator? maybe having some constants (e.g. BACKPRESSURE_QPS, HEAVY_QPS) would help understanding

bvolpato (dismissed):
Thanks Pranav.

I added a few nits/suggestions but no concerns. Feel free to skip or address separately. 

Note: Feel free to `Google LGTM` too, I'm just leaving as is so @Polber can take a second look.

Polber (approved):
I agree with all of @bvolpato comments, but they can definitely be TODO's for another PR. Great work!

Polber:
The issue that is preventing the unit tests from passing should be resolved in #561 

Once that PR is merged, the checks can be re-run.

it/src/main/java/com/google/cloud/teleport/it/PerformanceBenchmarkingBase.java
pranavbhandari24: This is a bit complex since the metrics are available but they are late. Sleeping for 4 minutes guarantees that we get the most accurate metrics. However, I agree that we shouldn't hardcode X minutes.

I will look into this and fix this later. Added a todo to do this.

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtablePerformanceIT.java
pranavbhandari24: I'm not sure we should do that since we are planning on moving the IT module to the beam repo soon? PerformanceBenchmarkingBase will be used for non-templates tests too

v2/streaming-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/StreamingDataGeneratorPerformanceIT.java
pranavbhandari24: Having parameterized tests for QPS/num workers/max num workers won't give us much information, but will increase our tests by a lot. Also, most of the time the pipeline autoscales down/up in a matter of a few minutes.

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtablePerformanceIT.java
pranavbhandari24: 1 million QPS is very unrealistic since we have a message limit of only 35 million. The intention is to cause backpressure on the data generator. The current implementation of Streaming data generator template creates stragglers since the fanout of `GenerateSequence.from(...).to(...)` is very high. This leads to costly and slow processing in the pipeline. I plan on fixing this by adding a `Reshuffle` before `Generate Fake Messages` step in the pipeline, but we are currently blocked by https://github.com/apache/beam/issues/25140.

Having a high QPS and a high number of starting workers reduces the fanout thereby improving processing time and cost (cost reduces by more than 100% for generating 10gb of data) 

Note: This issue doesn't occur for streaming pipelines from Streaming Data generator template, only for batch pipelines.

pranavbhandari24: Adding constants for clarity is a good idea. I'll do this in a separate PR since I will need to make changes to other performance tests too.

bvolpato (approved):

it/src/main/java/com/google/cloud/teleport/it/launcher/PipelineLauncher.java
```diff
@@ -173,10 +173,10 @@ public static Builder builder(TestName testName) {
     }
 
     /** Builder for the {@link LaunchConfig}. */
-    public static final class Builder {
+    public static final class Builder<T> {
       private final String jobName;
       private final String specPath;
-      private final Map<String, String> environment;
+      private final Map<String, T> environment;
```
bvolpato, line 179: nit: This feels more like a `Map<String, Object>` (to allow any value type) rather than a specific `T`. I feel <T> without specifying a type will reach the same outcome, but it may raise some raw type warnings.

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtableLT.java
```diff
@@ -0,0 +1,159 @@
...
+            artifactClient
+                .uploadArtifact(
+                    "input/bigtable.avsc",
+                    Resources.getResource("schema/avro/bigtable.avsc").getPath())
+                .name());
+  }
+
+  @After
+  public void teardown() {
+    bigtableResourceManager.cleanupAll();
```
bvolpato, line 96: nit:
I found that using `if (bigtableResourceManager != null)` around cleanups are a good practice for those tests.

If the setup/ResourceManager fails for any reason, this will be invoked and can throw a NullPointerException, so you end up getting some extra errors/red herrings on the output.

v1/src/test/java/com/google/cloud/teleport/templates/PubsubToBigQueryLT.java
```diff
@@ -124,49 +148,42 @@ public void testBacklog10gb() throws IOException, ParseException, InterruptedExc
     Result result =
         pipelineOperator.waitForConditionAndFinish(
             createConfig(info, Duration.ofMinutes(40)),
-            () -> waitForNumMessages(info.jobId(), INPUT_PCOLLECTION, numMessages));
+            () ->
+                bigQueryResourceManager.getRowCount(PROJECT, table.getDataset(), table.getTable())
+                    == NUM_MESSAGES);
```
bvolpato, line 153: `>= NUM_MESSAGES` is safer here due to Beam/Dataflow not offering exactly-once guarantees in some cases.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/pubsubtotext/PubsubToTextLT.java
```diff
@@ -72,9 +68,9 @@ public final class PubsubToTextPerformanceIT extends PerformanceBenchmarkingBase
   private static final String DEFAULT_WINDOW_DURATION = "10s";
   private static final Pattern EXPECTED_PATTERN = Pattern.compile(".*subscription-output-.*");
   private static final String INPUT_PCOLLECTION = "Read PubSub Events/PubsubUnboundedSource.out0";
-  private static final String OUTPUT_PTRANSFORM =
-      "Write File(s)/WriteFiles/WriteShardedBundlesToTempFiles/ApplyShardingKey";
-  private static final long numMessages = 35000000L;
+  private static final String OUTPUT_PCOLLECTION =
+      "Write File(s)/WriteFiles/WriteShardedBundlesToTempFiles/ApplyShardingKey.out0";
+  private static final long NUM_MESSAGES = 35000000L;
```
bvolpato, line 73: Command explaining magic number = ~10gb

v1/src/test/java/com/google/cloud/teleport/templates/PubsubToBigQueryLT.java
```diff
@@ -85,37 +96,50 @@ public void teardown() {
...
   public void testBacklog10gb() throws IOException, ParseException, InterruptedException {
-    // Generate fake data to topic
-    String jobName = createJobName(testName.getMethodName());
+    testBacklog10gb(Function.identity());
+  }
+
+  @Test
+  public void testBacklog10gbUsingStreamingEngine()
+      throws IOException, ParseException, InterruptedException {
+    testBacklog10gb(config -> config.addEnvironment("enableStreamingEngine", true));
```
bvolpato, line 105: +1, much better with params adder.

bvolpato (approved):
LGTM. Added a few nit comments, feel free to ignore.

it/src/main/java/com/google/cloud/teleport/it/launcher/PipelineLauncher.java
pranavbhandari24: I agree, changed it to `Map<String, Object>`

v1/src/test/java/com/google/cloud/teleport/bigtable/AvroToBigtableLT.java
pranavbhandari24: Added checks in all performance tests

v1/src/test/java/com/google/cloud/teleport/templates/PubsubToBigQueryLT.java
pranavbhandari24: Done

PR #559, opened 2023-01-19
Addressing NPE on bad Json config

pabloem:


syndeo-template/src/test/java/com/google/cloud/syndeo/common/SyndeoTemplateLaunchTest.java
```diff
@@ -37,11 +37,31 @@ public class SyndeoTemplateLaunchTest {
...
+  public void testBuildWithBQToBQConfigBadSink() throws JsonProcessingException {
+    SyndeoTemplate.buildFromJsonPayload(BQ_TO_BQ_CONFIG);
+  }
+
+  @Test
+  public void testBuildWithBQToBQConfig() throws JsonProcessingException {
+    // TODO(pabloem): Replace this with
+    // "beam:schematransform:org.apache.beam:bigquery_storage_write:v1" for Beam  2.45.0.
+    SyndeoTemplate.buildFromJsonPayload(
+        BQ_TO_BQ_CONFIG.replace("bigquery:write", "schemaIO:bigquery:write"));
```
manavgarg, line 65: Just to understand, it would fail without the replacement string because this transform is in beam and not in templates (unlike the read schema transform)?

manavgarg:
Also, LGTM

manavgarg (approved):

syndeo-template/src/test/java/com/google/cloud/syndeo/common/SyndeoTemplateLaunchTest.java
pabloem: the `urn` is set by the transform itself. We just don't happen to have an implementation that returns `bigquery:write` as its URN. But yes, you're correct otherwise.

manavgarg (approved):

PR #561, opened 2023-01-23
Fix intermittent test failures in TextToPubsubIT

Polber:
Signed-off-by: Jeffrey Kinard <jeff@thekinards.com>

### DIFF
--- a/v1/src/test/java/com/google/cloud/teleport/templates/TextToPubsubIT.java
+++ b/v1/src/test/java/com/google/cloud/teleport/templates/TextToPubsubIT.java
@@ -32,8 +32,8 @@ import com.google.pubsub.v1.PullResponse;
 import com.google.pubsub.v1.SubscriptionName;
 import com.google.pubsub.v1.TopicName;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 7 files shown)


v1/src/test/java/com/google/cloud/teleport/templates/TextToPubsubIT.java
```diff
@@ -81,22 +82,23 @@ public void testTextToTopic() throws IOException {
...
+                  return records.get().size() >= expectedArtifacts.size();
                 });
     // TODO: Replace this check by truth subjects (like assertThatArtifacts(), assertThatRecords())
-    List<String> actualMessages =
-        records.get().getReceivedMessagesList().stream()
-            .map(receivedMessage -> receivedMessage.getMessage().getData().toStringUtf8())
-            .collect(Collectors.toList());
+    List<String> actualMessages = records.get();
     assertThat(result).isEqualTo(Result.CONDITION_MET);
     assertThat(actualMessages).containsAtLeastElementsIn(expectedArtifacts);
```
bvolpato, line 103: Good catch. 

A bit of a nitpick, you don't need an AtomicReference here. Keeping the same reference is already atomic, so you can have `List<String> records` and just `records.addAll()` when doing a new pull

Polber: Ok I honestly wasn't sure about that and didn't want to mess with the logic - just keep it from failing. I can edit that and make sure it still passes. Thanks.

bvolpato: Yeah that's fine -- Java has some restrictions on what can be used from within the lambda not being reassigned.

So you can't have `PullResponse records` and keep reassigning it, that was the reason why we used an `AtomicReference`. But now that we can keep the same `List<String>` and will just aggregate new elements into the same list, we can just drop the `AtomicReference` as it may introduce confusion here for no benefits.

v1/src/test/java/com/google/cloud/teleport/templates/TextToPubsubIT.java
```diff
@@ -81,24 +81,27 @@ public void testTextToTopic() throws IOException {
     // Act
     JobInfo info = launchTemplate(options);
     assertThat(info.state()).isIn(JobState.ACTIVE_STATES);
-    AtomicReference<PullResponse> records = new AtomicReference<>();
+    List<String> records = new ArrayList<>();
     Result result =
         new DataflowOperator(getDataflowClient())
-            .waitForConditionAndFinish(
+            .waitForCondition(
```
bvolpato, line 87: If I got it correctly, you are not cancelling / finishing it here because it's a batch job?

bvolpato (dismissed):

bvolpato (changes requested):
Actions are failing

bvolpato (approved):

PR #562, opened 2023-01-23
Fixing schema validation for BigTableSchemaTransform

pabloem:


pabloem:
As of right now, nested schemas cannot be supported in our BigTable schema transform. This adds the necessary validation to error out at expansion time, rather than erroring at runtime.

r: @manavgarg

manavgarg (approved):
LGTM

PR #563, opened 2023-01-24
Fixing spotless on v2 templates

pabloem:
r @bvolpato


bvolpato:
Sorry just saw this, this was pushed on a separate commit. Working on some AIs to prevent this from getting pushed.

PR #567, opened 2023-01-26
Test for create never behavior on BQ

pabloem:


pabloem:
this PR depends of https://github.com/apache/beam/pull/25195 being merged

pabloem:
r: @manavgarg 

This verifies that an error will be thrown at construction time when a pipeline with `create_never` is launched against a table that does not exist (thus the pipeline cannot run because we cannot output)

syndeo-template/src/test/java/com/google/cloud/syndeo/it/KafkaToBigQueryIT.java
```diff
@@ -95,14 +98,50 @@ public class KafkaToBigQueryIT {
...
+  @Test
+  public void testErrorOnCreateNeverIfTableNotExisting() throws Exception {
+    JsonNode rootConfig =
+        KafkaToBigQueryLocalTest.generateConfigurationWithKafkaBootstrap(
+            KAFKA_BOOTSTRAP_SERVER, testName.getMethodName());
+    ((ObjectNode) rootConfig.get("sink").get("configurationParameters"))
+        .put("createDisposition", "CREATE_NEVER");
+    ((ObjectNode) rootConfig.get("sink").get("configurationParameters"))
+        .put(
+            "table",
```
manavgarg, line 113: nit: shouldn't we just set the "table" by passing this name on ln 108 instead of testName.getMethodName() ?

manavgarg (approved):
LGTM.

syndeo-template/src/test/java/com/google/cloud/syndeo/it/KafkaToBigQueryIT.java
pabloem: great catch. fixed!

PR #572, opened 2023-01-31
Updating template to generate BQ and PS configs properly

pabloem:


syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
```diff
@@ -82,9 +82,9 @@ public interface Options extends PipelineOptions, DataflowPipelineOptions {
           Set.of("table", "createDisposition", "writeDisposition"),
           // New names:
           "beam:schematransform:org.apache.beam:bigquery_storage_write:v1",
-          Set.of("instanceId", "tableId", "keyColumns", "projectId", "appProfileId"),
+          Set.of(),
           "beam:schematransform:org.apache.beam:bigquery_storage_read:v1",
-          Set.of("table"),
+          Set.of(),
```
mahagen, line 87: Is the intention here to not check the parameters against a Set() anymore and let everything through?

mahagen (approved):
One comment about the removal of the Set of parameters.

syndeo-template/src/main/java/com/google/cloud/syndeo/SyndeoTemplate.java
pabloem: yes, correct!

PR #577, opened 2023-02-02
bigtable-cdc-to-hbase flex template

georgecma:


v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,252 @@
...
+     * @param hbaseMutations row mutation to append origin info to
+     * @param cbtQualifierInput origin info string denoting mutation is from bigtable
+     * @throws IOException
+     */
+    private void appendOriginInfoToMutations(RowMutations hbaseMutations, String cbtQualifierInput)
+        throws IOException {
+      // Get last mutation to deduce last column family from
+      Mutation lastMutation =
+          hbaseMutations.getMutations().get(hbaseMutations.getMutations().size() - 1);
+      List<String> familyList = (List<String>) lastMutation.getFingerprint().get("families");
```
jackdingilian, line 183: Is there a way to do this that doesn't require the cast?

Also are we guaranteed that the families list preserves the order of the mutations?

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,252 @@
...
+      String lastEntryCf = familyList.get(familyList.size() - 1);
```
jackdingilian, line 184: nit: pulling the code for identify the lastEntryCf into a helper might help with readability here

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * Appends origin information to row mutation.
+     *
+     * @param hbaseMutations row mutation to append origin info to
+     * @param cbtQualifierInput origin info string denoting mutation is from bigtable
+     * @throws IOException
+     */
+    private void appendOriginInfoToMutations(RowMutations hbaseMutations, String cbtQualifierInput)
```
jackdingilian, line 178: nit: appendSourceInfoToMutations.  For consistency with SOURCE_HBASE/SOURCE_CBT.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,252 @@
...
+     * @return Hbase RowMutations object
+     */
+    private static RowMutations convertToRowMutations(ChangeStreamMutation mutation)
+        throws IOException {
+      byte[] hbaseRowKey = mutation.getRowKey().toByteArray();
+      RowMutations hbaseMutations = new RowMutations(hbaseRowKey);
+
+      for (Entry entry : mutation.getEntries()) {
+        if (entry instanceof SetCell) {
+
```
jackdingilian, line 208: Can we create a separate convert method for each Mutation entry type? Since there's a decent amount of logic for some of these and it'll be easier to unit test

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright (C) 2021 Google Inc.
```
jackdingilian, line 3: 2023

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/bigtable/data/v2/models/ChangeStreamMutation.java
```diff
@@ -0,0 +1,342 @@
+/*
```
jackdingilian, line 1: You don't have to copy the whole class to test will. You can do something like this: https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/test/java/com/google/cloud/spanner/FakePartitionFactory.java#L38

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
```diff
@@ -0,0 +1,224 @@
...
+
+        // TODO: we use MutateRow(RowMutations). Figure out if it'd be more efficient batch writing here
+        //  with e.g. BufferedMutator.mutate(Mutations) after grouping by rowkey.
+        table.mutateRow(mutations);
+        Metrics.counter("HbaseRepl", "mutations_written_to_hbase").inc();
+        ++recordsWritten;
+      }
+
+      @FinishBundle
+      public void finishBundle() throws Exception {
```
jackdingilian, line 197: We can probably just remove startBundle and finishBundle since we don't do the batching right. This was leftover from the HbaseIO batching in beam

jackdingilian (commented):
Nice! I'll have to go through some of this again but I made a first pass at the core parts.

Just want to note to make sure we don't merge this until after we've released the Change Stream Beam code

v2/pom.xml
```diff
@@ -714,6 +714,7 @@
         <module>pubsub-cdc-to-bigquery</module>
         <module>pubsub-to-mongodb</module>
         <module>streaming-data-generator</module>
+        <module>bigtable-cdc-to-hbase</module>
```
bvolpato, line 717: Sort this

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,187 @@
...
+    // Note that we set start and end times in actual test run
+    pipelineOptions = PipelineOptionsFactory.create().as(BigtableToHbasePipelineOptions.class);
+    // Set bigtable change stream options
+    pipelineOptions.setBigtableProjectId(args.get("bigtableProjectId"));
+    pipelineOptions.setInstanceId(args.get("instanceId"));
+    pipelineOptions.setTableId(args.get("tableId"));
+    pipelineOptions.setAppProfileId(args.get("appProfileId"));
+    // Set pipeline options
+    pipelineOptions.setStreaming(true);
+    pipelineOptions.setExperiments(Arrays.asList("use_runner_v2"));
```
bvolpato, line 88: Have some concerns with requiring `parameters` on the IT. Can we make it assume some defaults? Or maybe a few @Test variants validating different configurations?

bvolpato (commented):
Thanks for contributing! A few drive-by comments

v2/bigtable-cdc-to-hbase/pom.xml
georgecma: Resolved

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed to appendSourceTagToMutations, matching same-name function in Hbase-Bigtable replicator

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed to use cell scanner

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Done

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: added convert functions for setCells, deleteCells, and deleteFamily

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: removed

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/bigtable/data/v2/models/ChangeStreamMutation.java
georgecma: We can't create a fake object because Beam transform expects the package name to match 
 package com.google.cloud.bigtable.data.v2.models. 
 Should I strip down the class and remove all non-used functions?

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
georgecma: We expect these parameters to be unnecessary after Cdc GA. The alternative would be to expose our live persistent instances... Not sure what to do here

v2/pom.xml
georgecma: done, moved to top of modules list

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <version>1.2.17</version>
+    </dependency>
+
+
+    <dependency>
+      <groupId>com.google.auto.service</groupId>
+      <artifactId>auto-service-annotations</artifactId>
+      <version>1.0-rc6</version>
```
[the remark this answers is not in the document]
georgecma: done

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+      <name>Apache Development Snapshot Repository</name>
+      <url>https://repository.apache.org/content/repositories/snapshots/</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
```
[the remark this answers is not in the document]
georgecma: Yes, once Cdc is GA we expect to not have to use env for this repo anymore.

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
```diff
@@ -0,0 +1,284 @@
...
+        throw new IllegalStateException(
+            "The table " + tableId + " does not exist in instance " + instanceId + ".");
+      }
+    }
+  }
+
+  @Override
+  public synchronized void createTable(String newTableId, BigtableTableSpec spec) {
+    if (!Objects.equals(tableId, newTableId)) {
+      LOG.warn(
```
jackdingilian, line 147: Should this throw an exception? not sure exactly how ResourceMananger is used but seems like we should fail the test here since it would never be able to use newTableId

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
```diff
@@ -0,0 +1,284 @@
...
+          "Test tries to create table named differently from static tableId: {}.{}",
+          instanceId,
+          this.tableId);
+    }
+  }
+
+  @Override
```
jackdingilian, line 154: We can avoid a lot of the duplication here by extending `DefaultBigtableResourceManager` and only overriding the methods that change.

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
```diff
@@ -0,0 +1,284 @@
...
+import org.slf4j.LoggerFactory;
+import org.threeten.bp.Duration;
+
+/**
+ * An implementation for implementation of {@link BigtableResourceManager} interface which allows to
+ * run test as if new resources are created, but reuse pre-created resources all the time.
+ *
+ * <p>The class is thread-safe.
+ */
+public class StaticBigtableResourceManager implements BigtableResourceManager {
```
jackdingilian, line 47: Are we planning on checking this in or is this just a placeholder until we have change stream support in the admin client?

georgecma: This is a placeholder until Cdc can be enabled from Admin client.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.RowMutations;
+import org.apache.hadoop.hbase.client.Table;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Applies Hbase RowMutation objects to an Hbase table. */
```
jackdingilian, line 47: nit: probably worth adding a note and link to HBaseIO where we copied the general structure of this from

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
georgecma: Changed to throw BigtableResourceManagerException

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      if (hbaseMutations.getMutations().size() == 0) {
+        throw new Exception(
+            "hbase mutations object null, should not be possible because all mutations are supported");
+      }
+      return hbaseMutations;
+    }
+
+    private static Put convertSetCell(byte[] hbaseRowKey, Entry entry) {
```
jackdingilian, line 239: I think it's worth unit testing each of these (even though they should also be covered by the ConvertChangeStreamTest)

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,291 @@
...
+     * @return Hbase RowMutations object
+     */
+    private static RowMutations convertToRowMutations(ChangeStreamMutation mutation)
+        throws Exception {
+      byte[] hbaseRowKey = mutation.getRowKey().toByteArray();
+      RowMutations hbaseMutations = new RowMutations(hbaseRowKey);
+
+      for (Entry entry : mutation.getEntries()) {
+        if (entry instanceof SetCell) {
+
```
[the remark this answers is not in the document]
georgecma: removed extra lines

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,291 @@
...
+        } else if (entry instanceof DeleteFamily) {
+
+          hbaseMutations.add(convertDeleteFamily(hbaseRowKey, entry));
+        }
+      }
+
+      if (hbaseMutations.getMutations().size() == 0) {
+        throw new Exception(
+            "hbase mutations object null, should not be possible because all mutations are supported");
```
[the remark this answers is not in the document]
georgecma: added new exception when Entry type is not DeleteCells, DeleteFamily, or SetCells
changed this exception to empty change stream

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
```diff
@@ -0,0 +1,215 @@
...
+          connection = null;
+        }
+      }
+
+      @Override
+      public void populateDisplayData(DisplayData.Builder builder) {
+        builder.delegate(WriteRowMutations.this);
+      }
+
+      private long recordsWritten;
```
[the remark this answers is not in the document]
georgecma: removed

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,193 @@
...
```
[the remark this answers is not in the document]
georgecma: removed options

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,291 @@
...
+      DeleteCells deleteCells = (DeleteCells) entry;
+
+      // Convert timestamp to milliseconds
+      long ts = convertMicroToMilliseconds(deleteCells.getTimestampRange().getEnd());
+      // If RowMutation is created manually without a timestamp,then the timestamp end bound
+      // becomes null and unbounded.
+      // TODO: will this scenario ever happen in practice? If it will then we cast to now
+      //  Else we remove this and change integration test to always create with timestamp
+      if (deleteCells.getTimestampRange().getEndBound() == BoundType.UNBOUNDED) {
+        ts = Time.now();
```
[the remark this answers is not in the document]
georgecma: After conversation with @jackdingilian , timestamp will be set to now. This is how Hbase deals with non-timestamped cells and how Bigtable also handles unbounded timestamp cells. 

Bigtable: https://cloud.google.com/bigtable/docs/keep-only-latest-value
Hbase: https://hbase.apache.org/2.3/devapidocs/src-html/org/apache/hadoop/hbase/client/Mutation.html#line.592

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/utils/RowMutationsCoder.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.hadoop.hbase.client.RowMutations;
+import org.apache.hadoop.hbase.protobuf.ProtobufUtil;
+import org.apache.hadoop.hbase.protobuf.generated.ClientProtos.MutationProto;
+import org.apache.hadoop.hbase.protobuf.generated.ClientProtos.MutationProto.MutationType;
+
+/**
+ * Row mutations coder to provide serialization support for Hbase RowMutations object, which isn't
+ * natively serializable.
+ */
+public class RowMutationsCoder extends AtomicCoder<RowMutations> implements Serializable {
```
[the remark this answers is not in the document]
georgecma: Added

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Added explicit unit tests in separate file to test ConvertChangeStreamMutationFn.convertToRowMutations function

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/bigtable/data/v2/models/ChangeStreamMutation.java
georgecma: Done

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** End to end table that runs the pipeline from Bigtable to Hbase. */
+@Category(TemplateIntegrationTest.class)
+@TemplateIntegrationTest(BigtableToHbasePipeline.class)
+@RunWith(JUnit4.class)
+public class BigtableToHbasePipelineIT extends TemplateTestBase {
```
jackdingilian, line 56: We should test the source filtering in here too

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/HbaseRowMutationIOTest.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  @Before
+  public void setUp() throws IOException, InterruptedException {
+    // Provide custom encoder to non-serializable RowMutations class.
+    pipeline.getCoderRegistry().registerCoderForClass(RowMutations.class, RowMutationsCoder.of());
+  }
+
+  @Test
+  public void writesSingleRowPutsInOrder() throws Exception {
```
jackdingilian, line 80: It's worth noting I don't think that DirectRunner doesn't support key-ordered delivery. So I don't think these are actually guaranteed to apply in the same order unless run w dataflow

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
```diff
@@ -0,0 +1,250 @@
...
+    @Default.String("default")
+    String getAppProfileId();
+
+    void setAppProfileId(String appProfileId);
+
+    /** Hbase specific configs. Mirrors configurations on hbase-site.xml. */
+    @TemplateParameter.Text(
+        description = "Zookeeper quorum location",
+        helpText = "Zookeeper quorum location, corresponds to hbase.zookeeper.quorum")
+    String getHbaseZookeeperQuorum();
```
vermas2012, line 97: please refer to other pipelines, we usually take host and port.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/package-info.java
```diff
@@ -0,0 +1,18 @@
+/*
```
vermas2012, line 1: why is this req?

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.RowMutations;
+import org.apache.hadoop.util.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Converts Bigtable change stream RowMutations objects to their approximating HBase mutations. */
+public class ConvertChangeStream {
```
vermas2012, line 48: please use a better name. Typically we use convention like HBaseResultToMutationFn
https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/HBaseResultToMutationFn.java

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+
+    public ConvertChangeStreamMutation() {}
+
+    private ConvertChangeStreamMutation(
+        boolean enabled, String cbtQualifierInput, String hbaseQualifierInput) {
+      if (enabled) {
+        checkArgument(cbtQualifierInput != null, "cbt qualifier cannot be null.");
+        checkArgument(hbaseQualifierInput != null, "hbase qualifier cannot be null.");
+      }
+      twoWayReplication = enabled;
```
vermas2012, line 87: isEnabled?

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+  public static ConvertChangeStreamMutation convertChangeStreamMutation() {
+    return new ConvertChangeStreamMutation();
+  }
+
+  /**
+   * Change stream converter that converts change stream mutations into RowMutation objects. Input
+   * is Bigtable change stream KV.of(rowkey, changeStreamMutation), Output is KV.of(rowkey, hbase
+   * RowMutations object)
+   */
+  public static class ConvertChangeStreamMutation
```
vermas2012, line 61: For my education: do we need a PTransform? Can't we directly use a DoFn?

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+
+      if (lastEntry instanceof DeleteCells) {
+        if (((DeleteCells) lastEntry)
+            .getQualifier()
+            .equals(ByteString.copyFromUtf8(hbaseQualifierInput))) {
+          Metrics.counter("HbaseRepl", "mutations_filtered_from_hbase").inc();
+          return true;
+        }
+      }
+      Metrics.counter("HbaseRepl", "mutations_replicated_from_bigtable").inc();
```
vermas2012, line 159: What is this parameter "HbaseRepl", how is this used, should we use more verbose name?

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+      return false;
+    }
+
+    static byte[] convertUtf8String(String utf8) {
+      return utf8.getBytes(StandardCharsets.UTF_8);
```
vermas2012, line 164: You can also use Bytes object from HBase ecosystem.

vermas2012: https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/util/Bytes.html#toBytes-java.lang.String-

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /**
+     * Appends origin information to row mutation.
```
vermas2012, line 168: Please explain here. Anyone changing this code should understand what this means and implications of changing this code.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+      DeleteCells deleteCells = (DeleteCells) entry;
+
+      // Convert timestamp to milliseconds
+      long ts = convertMicroToMilliseconds(deleteCells.getTimestampRange().getEnd());
+      // If RowMutation is created with an unbounded timestamp, the mutation is meant to delete
+      // all versions of a cell up to the point of operation.
+      // This behavior is approximated by casting the Hbase delete as deleting up to current point
+      // in time.
+      if (deleteCells.getTimestampRange() == TimestampRange.unbounded()) {
+        ts = Time.now();
```
vermas2012, line 256: You should use the timestamp of CDCMutation. Basically time at which CBT accepted this delete. If you process this message after 3 days, you are deleting 3 extra days worth of mutations. This will cause divergence.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+
+      return delete;
+    }
+
+    private static Delete convertDeleteFamily(byte[] hbaseRowKey, Entry entry) {
+      // Bigtable deletefamily does not have a timestamp and relies on transaction sequence to
+      // delete everything before it.
+      // Hbase deletes operate from timestamps only. Therefore, we approximate Bigtable
+      // deletefamily to Hbase deletefamily with timestamp now().
+      long now = Time.now();
```
vermas2012, line 274: same as above. Use timestamp from CBT,  it should be the time when CBT wrote that entry in its replication/WAL logs.

vermas2012: @jackdingilian Double check my thinking here. You can process stream after "days", right? We shouldn't use stream processing time (Now) anywhere, rather we should base it on the time mutation was written to CBT WAL.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
```diff
@@ -0,0 +1,287 @@
...
+      if (mutation.getEntries().size() == 0) {
+        throw new Exception("Change stream entries list is empty.");
+      }
+
+      byte[] hbaseRowKey = mutation.getRowKey().toByteArray();
+      RowMutations hbaseMutations = new RowMutations(hbaseRowKey);
+      // Check and convert entries to hbase mutations
+      for (Entry entry : mutation.getEntries()) {
+        if (entry instanceof SetCell) {
+          hbaseMutations.add(convertSetCell(hbaseRowKey, entry));
```
vermas2012, line 220: its ok for the first version, but find out what is the cost of creating a new PUT for every cell? In replication, we keep on buffering SetCells to a single PUT, deleteCell into a single DELETE. We only create new PUT/Delete when there is mutatino type changes. 

So for example, if you have setcell, setcell, setcell -> it should be a single PUT with 3 cells
setcell, setcell, setcell, deletecell, setcell -> it should be a  PUT with 3 cells, delete with 1, put with 1

vermas2012: See how MutationBuilder and its child classes are used here. 
https://github.com/googleapis/java-bigtable-hbase/blob/main/hbase-migration-tools/bigtable-hbase-replication/bigtable-hbase-replication-core/src/main/java/com/google/cloud/bigtable/hbase/replication/CloudBigtableReplicationTask.java#L47

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
```diff
@@ -0,0 +1,215 @@
...
+    private class WriteRowMutationsFn extends DoFn<KV<byte[], RowMutations>, Void> {
+
+      WriteRowMutationsFn(WriteRowMutations writeRowMutations) {
+        checkNotNull(writeRowMutations.tableId, "tableId");
+        checkNotNull(writeRowMutations.configuration, "configuration");
+      }
+
+      @Setup
+      public void setup() throws Exception {
+        connection = ConnectionFactory.createConnection(configuration);
```
vermas2012, line 180: Why are we not using HBaseIO? We shouldn't be writing our own sink here , unless absolutely necessary

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
```diff
@@ -0,0 +1,215 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Writes Hbase RowMutations to a Hbase table. Based off of {@link
+ * org.apache.beam.sdk.io.hbase.HBaseIO}. HbaseIO applies Mutations, this class applies RowMutations
+ * objects.
+ */
+public class HbaseRowMutationIO {
```
vermas2012, line 52: need to review again, I have not reviewed this class based on my opinion that we should use HBaseIO as much as possible. Maybe udpate it to take RowMutations, if it does not take that already.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/utils/RowMutationsCoder.java
```diff
@@ -0,0 +1,97 @@
...
```
vermas2012, line 36: Make sure we add lots of tests for this coder, test all possible combinations and special cases.

vermas2012 (changes requested):
I have not seen any tests, but just the important code changes. 

Major issues with using timestamp "now" while creating mutations.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: HbaseIO only allow writing single Mutation objects.  We rewrote the IO to instead write Hbase RowMutations objects to reflect the CBT RowMutation input.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: @jackdingilian I don't see any timestamp for DeleteFamily or its parent class. If the commit timestamp exists, then we can also generalize to using that timestamp for Delete requests with unbounded timestamps.

jackdingilian: @vermas2012 Yeah that's correct, you could process it as far back as the change stream retention period if you're catching up.

@georgecma you can use the `commitTimestamp` on the `ChangeStreamMutation` that contains the DeleteFamily

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
```diff
@@ -0,0 +1,250 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.client.RowMutations;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Bigtable change stream pipeline to replicate changes to Hbase. Pipeline reads from a Bigtable
+ * change stream, converts change stream mutations to their nearest Hbase counterparts, and writes
+ * the resulting Hbase row mutations to Hbase.
```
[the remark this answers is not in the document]
jackdingilian: The language in the change stream docs is:

```
As a special case, all writes to a single row on a single cluster will be streamed in order by commitTimestamp. This is only the case for instances with a single cluster, or if all writes to a given row happen on only one cluster in a replicated instance. This order is maintained in the Dataflow connector via [key-ordered delivery](https://beam.apache.org/documentation/runtime/model/#data-ordering-in-a-pipeline-execution), as the Dataflow connector emits ChangeStreamMutation records with the row key as the record key.
```

We could add some adaption of that here

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
```diff
@@ -0,0 +1,250 @@
...
+    category = TemplateCategory.STREAMING,
+    displayName = "Bigtable CDC to HBase Replicator",
+    description = "A streaming pipeline that replicates Bigtable change stream data to HBase",
+    optionsClass = BigtableToHbasePipeline.BigtableToHbasePipelineOptions.class,
+    flexContainerName = "bigtable-cdc-to-hbase",
+    contactInformation = "https://cloud.google.com/support")
+public class BigtableToHbasePipeline {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BigtableToHbasePipeline.class);
+  private static final String USE_RUNNER_V2_EXPERIMENT = "use_runner_v2";
```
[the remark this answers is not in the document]
georgecma: No, in the dataflow template, we have to manually toggle V2 using the --unified-worker option. So to prevent confusion we manually set the experiment in the code itself since we will always use V2.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed implementation to use commitTimestamp from ChangeStreamMutation

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
georgecma: Added source filtering test

v2/bigtable-cdc-to-hbase/README.md
```diff
@@ -0,0 +1,75 @@
+# Bigtable CDC to Hbase Dataflow Template
+
+The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java) pipeline reads from a Bigtable change stream, applies a mutation converter to convert change stream mutations to HBase mutations, and writes the mutations to a specified Hbase instance.
+
+## Getting Started
+
+### Requirements
+* Bigtable table with change streams enabled
+* Hbase instance
```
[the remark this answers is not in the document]
georgecma: Changed line

v2/bigtable-cdc-to-hbase/README.md
```diff
@@ -0,0 +1,75 @@
...
+export GCS_BUCKET_NAME=<gcs-bucket-name without gs://, to stage template at>
+export REGION=<gcp-region to run dataflow job at>
+# Hbase configs
+export ZOOKEEPER_QUORUM=<zookeeper-quorum, e.g. my-zookeeper-server:2181>
+export HBASE_ROOT_DIR=<hbase-root-dir, e.g. hdfs://my-server/hbase>
+
+# optional parameters, modify as needed
+export HBASE_DISTRIBUTED=true
+export TWO_WAY_REPLICATION=true
+export CBT_QUALIFIER=SOURCE_CBT
```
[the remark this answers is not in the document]
georgecma: Commented it out and added description. Users should not change these unless needed.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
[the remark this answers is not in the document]
georgecma: Added section

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/utils/RowMutationsCoder.java
georgecma: Coder test is in RowMutationsCoderTest.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed to use MutationsBuilder

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/package-info.java
georgecma: flex-template requires it

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Added comment to explain function

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed to use Bytes.toBytes(...)

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: We can - here we wrap it so it's encapsulated in case it needs to evolve in the future:
```
Expose every major data-parallel task accomplished by your library as a composite PTransform. This allows the structure of the transform to evolve transparently to the code that uses it: e.g. something that started as a ParDo can become a more complex transform over time.
```
- https://beam.apache.org/contribute/ptransform-style-guide/#exposing-a-ptransform-vs-something-else

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Per https://beam.apache.org/releases/javadoc/2.1.0/org/apache/beam/sdk/metrics/Metrics.html guidelines, swapped "HbaseRepl" out for ConvertChnageStreamMutationFn

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed signature to bidirectionalReplicationEnabled

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Changed to use commit timestamp

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ConvertChangeStream.java
georgecma: Renamed Transformer to ChangeStreamToRowMutations to better reflect input-output transformation

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
georgecma: Changed signature to host + port

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
```diff
@@ -0,0 +1,250 @@
...
+        helpText = "Hbase root directory, corresponds to hbase.rootdir")
+    String getHbaseRootDir();
+
+    void setHbaseRootDir(String hbaseRootDir);
+
+    @TemplateParameter.Text(
+        description = "Hbase cluster distribution mode",
+        helpText = "Whether hbase is distributed, corresponds to hbase.cluster.distributed")
+    @Default.String("true")
+    String getHbaseClusterDistributed();
```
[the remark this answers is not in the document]
georgecma: Removed

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: Will move HbaseRowMutationsIO to Beam repo. However, the deployment cadence will be quite slow so we will keep this module in this repo until then.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: Will move HbaseRowMutationsIO to Beam repo. However, the deployment cadence will be quite slow so we will keep this module in this repo until then.

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/HbaseRowMutationIOTest.java
georgecma: Test naming confusing, this test does not test ordering, only writing puts. Renamed test.

v2/bigtable-cdc-to-hbase/pom.xml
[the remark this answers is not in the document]
georgecma: Will remove this before merging

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
georgecma: Will remove this before merging

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: Will remove this before merging

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: Will remove this before merging

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/ChangeStreamToRowMutationsTest.java
Polber, line 1: The unit tests that correspond to files in utils or transforms should be under those folders in test/ as well. The test folder should have the same structure as the main folder.

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,329 @@
...
+    pipelineOptions.setBidirectionalReplicationEnabled(
+        Boolean.parseBoolean(args.get("bidirectionalReplicationEnabled")));
+    pipelineOptions.setCbtQualifier(cbtQualifier);
+    pipelineOptions.setHbaseQualifier(hbaseQualifier);
+
+    // Create Hbase cluster
+    hBaseTestingUtility = new HBaseTestingUtility();
+    hBaseTestingUtility.startMiniCluster();
+    // Create HBase table that mirrors persistent Hbase table.
+    hbaseTable = HbaseUtils.createTable(hBaseTestingUtility, pipelineOptions.getTableId());
```
Polber, line 107: We have been creating an integration testing framework over the past few months that involves creating resource managers to manage resources used by templates. Since this template introduces a new resource (Hbase), this logic should be included in an hbase resource manager.

This may be something that needs to be added after this PR is submitted, but having custom resource management inside an IT is not something we plan to support.

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  @AfterClass
+  public static void tearDown() throws IOException {
+    if (hBaseTestingUtility != null) {
+      hBaseTestingUtility.shutdownMiniCluster();
+    }
+    if (bigtableResourceManager != null) {
+      bigtableResourceManager.cleanupAll();
+    }
```
Polber, line 139:
```suggestion
    ResourceManagerUtils.cleanResources(bigtableResourceManager);
```

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,329 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** End to end table that runs the pipeline from Bigtable to Hbase. */
+@Category(TemplateIntegrationTest.class)
+@TemplateIntegrationTest(BigtableToHbasePipeline.class)
```
Polber, line 59:
```suggestion
@TemplateIntegrationTest(value = BigtableToHbasePipeline.class, template = "bigtable-cdc-to-hbase")
```

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
```diff
@@ -0,0 +1,282 @@
...
+ */
+@Template(
+    name = "bigtable-cdc-to-hbase",
+    category = TemplateCategory.STREAMING,
+    displayName = "Bigtable CDC to HBase Replicator",
+    description = "A streaming pipeline that replicates Bigtable change stream data to HBase",
+    optionsClass = BigtableToHbasePipeline.BigtableToHbasePipelineOptions.class,
+    flexContainerName = "bigtable-cdc-to-hbase",
+    contactInformation = "https://cloud.google.com/support")
+public class BigtableToHbasePipeline {
```
Polber, line 62: Naming convention is typically {source}To{sink} (i.e. `BigtableToHbase`). I would remove the "Pipeline" suffix.

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
```diff
@@ -0,0 +1,329 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** End to end table that runs the pipeline from Bigtable to Hbase. */
```
Polber, line 57:
```suggestion
Integration test for {@link BigtableToHbase} (BigtableToHbase). */
```

Polber (changes requested):
Thanks for contributing a new template to out repo! I left some comments (including several nits that are up to your discretion), but let me know if you want any more clarification. Thanks.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/HbaseRowMutationIO.java
georgecma: Used to initialize light-weight connection to Hbase and to log record count for debugging purposes.

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipelineIT.java
georgecma: After talking to @Polber , we will remove IT for now until CDC GA and then we can spin up hermetic bigtables with CDC enabled

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
georgecma: After talking to @Polber , we will remove IT for now until CDC GA and then we can spin up hermetic bigtables with CDC enabled

it/src/main/java/com/google/cloud/teleport/it/bigtable/StaticBigtableResourceManager.java
georgecma: After talking to @Polber , we will remove IT for now until CDC GA and then we can spin up hermetic bigtables with CDC enabled

it/src/main/java/com/google/cloud/teleport/it/bigtable/BigtableResourceManager.java
```diff
@@ -60,7 +46,10 @@ public interface BigtableResourceManager extends ResourceManager {
    * @param columnFamilies A collection of column family names for the table.
    * @throws BigtableResourceManagerException if there is an error creating the table in Bigtable.
    */
-  void createTable(String tableId, Iterable<String> columnFamilies);
+  void createTable(String tableId, Iterable<String> columnFamilies)
+      throws BigtableResourceManagerException;
+
+  void createTable(String tableId, BigtableTableSpec spec) throws BigtableResourceManagerException;
```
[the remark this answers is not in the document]
georgecma: After talking to @Polber , we will remove IT for now until CDC GA and then we can spin up hermetic bigtables with CDC enabled

it/src/main/java/com/google/cloud/teleport/it/bigtable/DefaultBigtableResourceManager.java
```diff
@@ -238,6 +243,23 @@ public synchronized void createTable(
...
+  }
+
+  @Override
+  public synchronized void createTable(
+      String tableId, Iterable<String> columnFamilies, Duration maxAge) {
+    BigtableTableSpec spec = new BigtableTableSpec();
+    spec.setColumnFamilies(columnFamilies);
+    spec.setMaxAge(maxAge);
+    createTable(tableId, spec);
+  }
```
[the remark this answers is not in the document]
georgecma: After talking to @Polber , we will remove IT for now until CDC GA and then we can spin up hermetic bigtables with CDC enabled

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableToHbasePipeline.java
georgecma: Renamed to BigtableCdcToHbase

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ChangeStreamToRowMutations.java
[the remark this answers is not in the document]
georgecma: After discussion w/ @Polber , we will keep the current transform in the template folder because the transform and utils are fairly esoteric and we do not expect them to be shared by other templates. 
(The transform is a CDC mutation to Hbase RowMutation converter.)

v2/bigtable-cdc-to-hbase/pom.xml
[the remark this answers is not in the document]
georgecma: We will swap out the repo URL after CDC GA.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
```diff
@@ -0,0 +1,281 @@
...
+    String getInstanceId();
+
+    void setInstanceId(String instanceId);
+
+    @TemplateParameter.Text(
+        description = "Table id",
+        helpText = "The name of the Bigtable table whose changes to stream")
+    String getTableId();
+
+    void setTableId(String tableId);
```
bvolpato, line 89: Can we try to consolidate / share options class between this and Bigtable CDC to BigQuery (https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/628)? e.g.,  BigtableCdcReadOptions

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ChangeStreamToRowMutations.java
```diff
@@ -0,0 +1,212 @@
...
+            .equals(ByteString.copyFromUtf8(hbaseQualifierInput))) {
+          Metrics.counter(ConvertChangeStreamFn.class, "mutations_filtered_from_hbase").inc();
+          return true;
+        }
+      }
+      Metrics.counter(ConvertChangeStreamFn.class, "mutations_replicated_from_bigtable").inc();
+      return false;
+    }
+
+    static byte[] convertUtf8String(String utf8) {
```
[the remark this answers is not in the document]
bvolpato: +1 on .getBytes(), and passing StandardCharsets.UTF_8 since we know the encoding

v2/bigtable-cdc-to-hbase/src/test/java/com/google/cloud/teleport/v2/templates/utils/HashUtils.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Logger log = LoggerFactory.getLogger(HashHbaseRowMutationsFn.class);
+
+    public HashHbaseRowMutationsFn() {}
+
+    @ProcessElement
+    public void processElement(ProcessContext c) throws Exception {
+      RowMutations rowMutations = c.element().getValue();
+
+      if (!new String(c.element().getKey()).equals(new String(rowMutations.getRow()))) {
+        throw new Exception("Hash error, KV rowkey is not the same as rowMutations rowkey");
```
bvolpato, line 141: nit: IllegalArgumentException? IllegalStateException?

bvolpato (commented):
Thanks! This is looking good, just left a few comments.  Jeff's recommendations all make sense too.

I just think we can share some parts from Bigtable CDC to BigQuery here so we have less things to maintain in the feature (but that could likely be done afterwards)

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
georgecma: We would like to merge this in by EOM. Talked to Alexey, we can merge this first and I can add IT and rebase pipeline options from his template after his is merged in as well.

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
```diff
@@ -0,0 +1,257 @@
+/*
+ * Copyright (C) 2022 Google LLC
```
Polber, line 2: nit: 
```suggestion
 * Copyright (C) 2023 Google LLC
```

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
```diff
@@ -0,0 +1,257 @@
...
+    /** Hbase specific configs. Mirrors configurations on hbase-site.xml. */
+    @TemplateParameter.Text(
+        description = "Zookeeper quorum host",
+        helpText = "Zookeeper quorum host, corresponds to hbase.zookeeper.quorum host")
+    String getHbaseZookeeperQuorumHost();
+
+    void setHbaseZookeeperQuorumHost(String hbaseZookeeperQuorumHost);
+
+    @TemplateParameter.Text(
+        description = "Zookeeper quorum port",
```
Polber, line 112: Should this be `optional`? It has a default value for the port, so no need to set.

Polber (changes requested):
Looks really good! I left a few comments, but nothing too big. Thanks for adding this!

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+
+  <build>
+    <extensions>
+      <extension>
+        <groupId>com.google.cloud.artifactregistry</groupId>
+        <artifactId>artifactregistry-maven-wagon</artifactId>
+        <version>2.1.5</version>
+      </extension>
+    </extensions>
+  </build>
```
bvolpato, line 42: Just curious why we need the wagon here

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+      <artifactId>hbase-shaded-testing-util</artifactId>
+      <version>2.3.6</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- Needed for hbase test cluster -->
+    <dependency>
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <version>1.2.17</version>
+    </dependency>
```
bvolpato, line 77: Is this needed? Is it just for testing? Can we use `<scope>test</scope>`?

We bring logging implementations through
```
        <!-- Plain text logging for the unit tests -->
        <dependency>
            <groupId>com.google.cloud.teleport</groupId>
            <artifactId>plaintext-logging</artifactId>
            <scope>test</scope>
        </dependency>
```

So not entirely sure the need here

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+
+
+    <dependency>
+      <groupId>com.google.auto.service</groupId>
+      <artifactId>auto-service-annotations</artifactId>
+      <version>${autovalue.service.version}</version>
+    </dependency>
+
+
+    <!-- Override 2.x client because dataflowTemplates doesn't recognize major versions -->
```
bvolpato, line 87: What it means? Templates bring google-api-client-bom 2.2.0 now at 2.48.0: https://mvnrepository.com/artifact/org.apache.beam/beam-sdks-java-google-cloud-platform-bom/2.48.0

This pinning here can be risky, so curious for the motivation.

v2/bigtable-cdc-to-hbase/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+      <version>2.10.2</version>
+    </dependency>
+
+    <dependency>
+    <groupId>com.google.api</groupId>
+    <artifactId>gax</artifactId>
+    <version>2.19.2</version>
```
bvolpato, line 103: This is defined by the BOM as well (2.26.0), so we can avoid pinning here

v2/bigtable-cdc-to-hbase/pom.xml
georgecma: Actually not needed right now since we don't spin up hbase testing cluster. Will remove.

v2/bigtable-cdc-to-hbase/pom.xml
georgecma: removed. legacy from private preview days

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
georgecma: Changed to extend bigtable-common options

v2/bigtable-cdc-to-hbase/pom.xml
georgecma: removed

Polber (commented):
FYI https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/5225170072/jobs/9434360695?pr=577#step:4:1427

v2/bigtable-cdc-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableCdcToHbase.java
```diff
@@ -0,0 +1,216 @@
...
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates;
+
+import com.google.cloud.teleport.metadata.Template;
+import com.google.cloud.teleport.metadata.TemplateCategory;
+import com.google.cloud.teleport.metadata.TemplateParameter;
+import com.google.cloud.teleport.v2.bigtable.options.BigtableCommonOptions.ReadChangeStreamOptions;
+import com.google.cloud.teleport.v2.templates.transforms.ChangeStreamToRowMutations;
+import java.io.IOException;
```
Polber, line 23:
```suggestion
```

georgecma:
New PR: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/780

PR #586, opened 2023-02-09
Prevent casting to TemporalAccessor when reading datetimes from MSSQL

andreigurau:
Band-aid fix to prevent TemporalAccessor casting on datetime columns when using the MSSQL driver. Won't be merged until more validation is done

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
+++ b/v1/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
@@ -222,9 +222,18 @@ public class JdbcConverters {
                 DATE_FORMATTER.format(resultSet.getDate(i).toLocalDate()));
             break;
           case "datetime":
-            outputTableRow.set(
-                getColumnRef(metaData, i),
-                DATETIME_FORMATTER.format((TemporalAccessor) resultSet.getObject(i)));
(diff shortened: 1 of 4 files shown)


v1/src/main/java/com/google/cloud/teleport/templates/common/JdbcConverters.java
```diff
@@ -222,9 +226,16 @@ public TableRow mapRow(ResultSet resultSet) throws Exception {
                 DATE_FORMATTER.format(resultSet.getDate(i).toLocalDate()));
             break;
           case "datetime":
-            outputTableRow.set(
-                getColumnRef(metaData, i),
-                DATETIME_FORMATTER.format((TemporalAccessor) resultSet.getObject(i)));
+            if (driverClassName.get().equals(MSSQL_DRIVER_CLASS)) {
```
bvolpato, line 229: This is a good change -- I think you can replace this if with `if (resultSet.getTimestamp(i) instanceof Timestamp)` so we generalize this for other drivers (that may happen to also return java.sql.Timestamp)

andreigurau: Done, I like this change more

bvolpato (dismissed):

PR #592, opened 2023-02-15
Sketching a column-expression-based schema transform

pabloem:


syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/sql/SyndeoScalarSqlTransformSchemaTransformProvider.java
```diff
@@ -0,0 +1,87 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.syndeo.transforms.sql;
+
+import com.google.auto.service.AutoService;
+import com.google.auto.value.AutoValue;
+import java.util.Collections;
+import java.util.List;
+import org.apache.beam.sdk.extensions.sql.expansion.SqlTransformSchemaTransformProvider;
```
zhoufek, line 22: Would it be better to wrap `SyndeoSqlSchemaTransformProvider`. It seems like this would just generate the SQL query to pass to that, letting it fill in fields like the dialect.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/sql/SyndeoScalarSqlTransformSchemaTransformProvider.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  public @UnknownKeyFor @NonNull @Initialized List<@UnknownKeyFor @NonNull @Initialized String>
+      outputCollectionNames() {
+    return List.of("output", "errors");
+  }
+
+  @DefaultSchema(AutoValueSchema.class)
+  @AutoValue
+  public abstract static class SyndeoScalarSqlTransformSchemaTransformConfiguration {
+    public abstract List<String> getColumnExpressions();
```
zhoufek, line 85: Maybe rename this to something that reflects that this is for the `SELECT` statement, like `getSelectColumnExpressions` or, in case this opens up to more than just columns, `getSelectExpressions`. That way, this is clearly separated from any other expression lists that we support (e.g. if we start supporting filters).

zhoufek (changes requested):

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/sql/SyndeoScalarSqlTransformSchemaTransformProvider.java
pabloem: done

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/sql/SyndeoScalarSqlTransformSchemaTransformProvider.java
pabloem: done

zhoufek:
Looks like there are some unit test failures. Could you please fix those?

pabloem:
fyi @zhoufek added lgtm tag

Issue #477, opened 2022-10-05
[Feature Request]: Add Datastream resource manager


itodotimothy6:
### Related Template(s)

datastream-to-*

### What feature(s) are you requesting?

Adding Datastream resource manager

PR #596, opened 2023-02-20
Issue 477: Create Datastream Resource Manager

itodotimothy6:
Draft PR as suggested by @bvolpato 

#477 

This PR attempts to create Datastream Resource Manager for Dataflow Template Integration Testing

The RM would be able 
- Create Source Connection Profile (MySQL, PostgresSQL, Oracle)
- Create Destination Profile (GCS)
- Create a Stream
- Start/Stop a Stream by updating Stream State
- Clean up all resources.

Currently a work in progress

### DIFF
--- a/it/google-cloud-platform/pom.xml
+++ b/it/google-cloud-platform/pom.xml
@@ -102,6 +102,10 @@
             <artifactId>google-cloud-datastore</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-datastream</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.auto.value</groupId>
             <artifactId>auto-value</artifactId>
             <version>${autovalue.version}</version>
--- a/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/DatastreamResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/DatastreamResourceManager.java
@@ -1,0 +1,502 @@
+/*
+ * Copyright (C) 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.gcp.datastream;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.cloud.datastream.v1.AvroFileFormat;
+import com.google.cloud.datastream.v1.BigQueryDestinationConfig;
+import com.google.cloud.datastream.v1.BigQueryProfile;
+import com.google.cloud.datastream.v1.ConnectionProfile;
+import com.google.cloud.datastream.v1.ConnectionProfileName;
+import com.google.cloud.datastream.v1.CreateConnectionProfileRequest;
+import com.google.cloud.datastream.v1.CreateStreamRequest;
+import com.google.cloud.datastream.v1.DatastreamClient;
+import com.google.cloud.datastream.v1.DatastreamSettings;
+import com.google.cloud.datastream.v1.DeleteConnectionProfileRequest;
+import com.google.cloud.datastream.v1.DeleteStreamRequest;
+import com.google.cloud.datastream.v1.DestinationConfig;
+import com.google.cloud.datastream.v1.GcsDestinationConfig;
+import com.google.cloud.datastream.v1.GcsProfile;
+import com.google.cloud.datastream.v1.JsonFileFormat;
+import com.google.cloud.datastream.v1.LocationName;
+import com.google.cloud.datastream.v1.MysqlProfile;
+import com.google.cloud.datastream.v1.MysqlSourceConfig;
+import com.google.cloud.datastream.v1.OracleProfile;
+import com.google.cloud.datastream.v1.OracleSourceConfig;
+import com.google.cloud.datastream.v1.PostgresqlProfile;
+import com.google.cloud.datastream.v1.PostgresqlSourceConfig;
+import com.google.cloud.datastream.v1.SourceConfig;
+import com.google.cloud.datastream.v1.StaticServiceIpConnectivity;
+import com.google.cloud.datastream.v1.Stream;
+import com.google.cloud.datastream.v1.StreamName;
+import com.google.cloud.datastream.v1.UpdateStreamRequest;
+import com.google.cloud.teleport.it.common.ResourceManager;
+import com.google.common.base.Strings;
+import com.google.protobuf.Duration;
+import com.google.protobuf.FieldMask;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import org.apache.arrow.util.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Client for Datastream resources.
+ *
+ * <p>This class is thread safe.
+ */
+public final class DatastreamResourceManager implements ResourceManager {
+  private static final Logger LOG = LoggerFactory.getLogger(DatastreamResourceManager.class);
+  private static final String FIELD_STATE = "state";
+  private final DatastreamClient datastreamClient;
+  private final String location;
+  private final String projectId;
+  private final Set<String> createdStreamIds;
+  private final Set<String> createdConnectionProfileIds;
+
+  enum DestinationOutputFormat {
+    AVRO_FILE_FORMAT,
+    JSON_FILE_FORMAT
+  }
+
+  private DatastreamResourceManager(Builder builder) throws IOException {
+    this(
+        DatastreamClient.create(
+            DatastreamSettings.newBuilder()
+                .setCredentialsProvider(builder.credentialsProvider)
+                .build()),
+        builder);
+  }
+
+  @VisibleForTesting
+  public DatastreamResourceManager(DatastreamClient datastreamClient, Builder builder) {
+    this.datastreamClient = datastreamClient;
+    this.location = builder.location;
+    this.projectId = builder.projectId;
+    this.createdStreamIds = Collections.synchronizedSet(new HashSet<>());
+    this.createdConnectionProfileIds = Collections.synchronizedSet(new HashSet<>());
+  }
+
+  public static Builder builder(String projectId, String location) {
+    checkArgument(!Strings.isNullOrEmpty(projectId), "projectID can not be null or empty");
+    checkArgument(!Strings.isNullOrEmpty(location), "location can not be null or empty");
+    return new Builder(projectId, location);
+  }
+
+  /**
+   * @param connectionProfileId The ID of the connection profile.
+   * @param source An object representing the JDBC source.
+   * @return A Datastream JDBC source connection profile.
+   */
+  private synchronized ConnectionProfile createJDBCSourceConnectionProfile(
+      String connectionProfileId, JDBCSource source) {
+    checkArgument(
+        !Strings.isNullOrEmpty(connectionProfileId),
+        "connectionProfileId can not be null or empty");
+    LOG.info(
+        "Creating JDBC Source Connection Profile {} in project {}.",
+        connectionProfileId,
+        projectId);
+    ConnectionProfile.Builder connectionProfileBuilder = ConnectionProfile.newBuilder();
+    JDBCSource.SourceType type = source.type();
+
+    try {
+      switch (type) {
+        case MYSQL:
+          MysqlProfile.Builder mysqlProfileBuilder = MysqlProfile.newBuilder();
+          mysqlProfileBuilder
+              .setHostname(source.hostname())
+              .setUsername(source.username())
+              .setPassword(source.password())
+              .setPort(source.port());
+          connectionProfileBuilder.setMysqlProfile(mysqlProfileBuilder);
+          break;
+        case ORACLE:
+          OracleProfile.Builder oracleProfileBuilder = OracleProfile.newBuilder();
+          oracleProfileBuilder
+              .setHostname(source.hostname())
+              .setUsername(source.username())
+              .setPassword(source.password())
+              .setPort(source.port());
+          connectionProfileBuilder.setOracleProfile(oracleProfileBuilder);
+          break;
+        case POSTGRESQL:
+          PostgresqlProfile.Builder postgresqlProfileBuilder = PostgresqlProfile.newBuilder();
+          postgresqlProfileBuilder
+              .setHostname(source.hostname())
+              .setUsername(source.username())
+              .setPassword(source.password())
+              .setPort(source.port())
+              .setDatabase(((PostgresqlSource) source).database());
+          connectionProfileBuilder.setPostgresqlProfile(postgresqlProfileBuilder);
+          break;
+        default:
+          throw new DatastreamResourceManagerException(
+              "Could not recognize JDBC source type " + type.name());
+      }
+
+      ConnectionProfile connectionProfile =
+          connectionProfileBuilder
+              .setDisplayName(connectionProfileId)
+              .setStaticServiceIpConnectivity(StaticServiceIpConnectivity.getDefaultInstance())
+              .build();
+      CreateConnectionProfileRequest request =
+          CreateConnectionProfileRequest.newBuilder()
+              .setParent(LocationName.of(projectId, location).toString())
+              .setConnectionProfile(connectionProfile)
+              .setConnectionProfileId(connectionProfileId)
+              .build();
+      ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+      createdConnectionProfileIds.add(connectionProfileId);
+
+      LOG.info(
+          "Successfully created JDBC Source Connection Profile {} in project {}.",
+          connectionProfileId,
+          projectId);
+      return reference;
+    } catch (ExecutionException | InterruptedException e) {
+      throw new DatastreamResourceManagerException(
+          "Failed to create JDBC source connection profile. ", e);
+    }
+  }
+
+  /**
+   * @param sourceConnectionProfileId The ID of the connection profile.
+   * @param source An object representing the JDBC source.
+   * @return a SourceConfig object which is required to configure a Stream.
+   */
+  public synchronized SourceConfig buildSourceConfig(
+      String sourceConnectionProfileId, JDBCSource source) {
+
+    createJDBCSourceConnectionProfile(sourceConnectionProfileId, source);
+    SourceConfig.Builder sourceConfigBuilder =
+        SourceConfig.newBuilder()
+            .setSourceConnectionProfile(
+                ConnectionProfileName.format(projectId, location, sourceConnectionProfileId));
+
+    switch (source.type()) {
+      case MYSQL:
+        sourceConfigBuilder.setMysqlSourceConfig((MysqlSourceConfig) source.config());
+        break;
+      case POSTGRESQL:
+        sourceConfigBuilder.setPostgresqlSourceConfig((PostgresqlSourceConfig) source.config());
+        break;
+      case ORACLE:
+        sourceConfigBuilder.setOracleSourceConfig((OracleSourceConfig) source.config());
+        break;
+      default:
+        throw new DatastreamResourceManagerException(
+            "Could not recognize JDBC source type " + source.type().name());
+    }
+
+    return sourceConfigBuilder.build();
+  }
+
+  /**
+   * @param connectionProfileId The ID of the GCS connection profile.
+   * @param gcsBucketName The GCS Bucket to connect to.
+   * @param gcsRootPath The Path prefix to specific gcs location. Can either be empty or must start
+   *     with '/'.
+   * @return A Datastream GCS destination connection profile.
+   */
+  public synchronized ConnectionProfile createGCSDestinationConnectionProfile(
+      String connectionProfileId, String gcsBucketName, String gcsRootPath) {
+    checkArgument(
+        !Strings.isNullOrEmpty(connectionProfileId),
+        "connectionProfileId can not be null or empty");
+    checkArgument(!Strings.isNullOrEmpty(gcsBucketName), "gcsBucketName can not be null or empty");
+    checkArgument(gcsRootPath != null, "gcsRootPath can not be null");
+    checkArgument(
+        gcsRootPath.isEmpty() || gcsRootPath.charAt(0) == '/',
+        "gcsRootPath must either be an empty string or start with a '/'");
+
+    LOG.info(
+        "Creating GCS Destination Connection Profile {} in project {}.",
+        connectionProfileId,
+        projectId);
+
+    try {
+      ConnectionProfile.Builder connectionProfileBuilder =
+          ConnectionProfile.newBuilder()
+              .setDisplayName(connectionProfileId)
+              .setStaticServiceIpConnectivity(StaticServiceIpConnectivity.getDefaultInstance())
+              .setGcsProfile(
+                  GcsProfile.newBuilder().setBucket(gcsBucketName).setRootPath(gcsRootPath));
+
+      CreateConnectionProfileRequest request =
+          CreateConnectionProfileRequest.newBuilder()
+              .setParent(LocationName.of(projectId, location).toString())
+              .setConnectionProfile(connectionProfileBuilder)
+              .setConnectionProfileId(connectionProfileId)
+              .build();
+
+      ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+      createdConnectionProfileIds.add(connectionProfileId);
+      LOG.info(
+          "Successfully created GCS Destination Connection Profile {} in project {}.",
+          connectionProfileId,
+          projectId);
+
+      return reference;
+    } catch (ExecutionException | InterruptedException e) {
+      throw new DatastreamResourceManagerException(
+          "Failed to create GCS source connection profile. ", e);
+    }
+  }
+
+  /**
+   * @param connectionProfileId The ID of the connection profile.
+   * @param path The Path prefix to specific GCS location. Can either be empty or must start with
+   *     '/'.
+   * @param destinationOutputFormat The format of the files written to GCS.
+   * @return A DestinationConfig object representing a GCS destination configuration.
+   */
+  public synchronized DestinationConfig buildGCSDestinationConfig(
+      String connectionProfileId, String path, DestinationOutputFormat destinationOutputFormat) {
+
+    DestinationConfig.Builder destinationConfigBuilder =
+        DestinationConfig.newBuilder()
+            .setDestinationConnectionProfile(
+                ConnectionProfileName.format(projectId, location, connectionProfileId));
+
+    GcsDestinationConfig.Builder gcsDestinationConfigBuilder =
+        GcsDestinationConfig.newBuilder().setPath(path);
+
+    if (destinationOutputFormat == DestinationOutputFormat.AVRO_FILE_FORMAT) {
+      gcsDestinationConfigBuilder.setAvroFileFormat(AvroFileFormat.getDefaultInstance());
+    } else {
+      gcsDestinationConfigBuilder.setJsonFileFormat(JsonFileFormat.getDefaultInstance());
+    }
+
+    destinationConfigBuilder.setGcsDestinationConfig(gcsDestinationConfigBuilder);
+    return destinationConfigBuilder.build();
+  }
+
+  /**
+   * @param connectionProfileId The ID of the connection profile.
+   * @return A Datastream BigQuery destination connection profile.
+   */
+  public synchronized ConnectionProfile createBQDestinationConnectionProfile(
+      String connectionProfileId) {
+
+    LOG.info(
+        "Creating BQ Destination Connection Profile {} in project {}.",
+        connectionProfileId,
+        projectId);
+
+    try {
+      ConnectionProfile.Builder connectionProfileBuilder =
+          ConnectionProfile.newBuilder()
+              .setDisplayName(connectionProfileId)
+              .setStaticServiceIpConnectivity(StaticServiceIpConnectivity.getDefaultInstance())
+              .setBigqueryProfile(BigQueryProfile.newBuilder());
+
+      CreateConnectionProfileRequest request =
+          CreateConnectionProfileRequest.newBuilder()
+              .setParent(LocationName.of(projectId, location).toString())
+              .setConnectionProfile(connectionProfileBuilder)
+              .setConnectionProfileId(connectionProfileId)
+              .build();
+
+      ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+      createdConnectionProfileIds.add(connectionProfileId);
+
+      LOG.info(
+          "Successfully created BQ Destination Connection Profile {} in project {}.",
+          connectionProfileId,
+          projectId);
+      return reference;
+    } catch (ExecutionException | InterruptedException e) {
+      throw new DatastreamResourceManagerException(
+          "Failed to create BQ destination connection profile. ", e);
+    }
+  }
+
+  /**
+   * @param connectionProfileId The ID of the connection profile.
+   * @param stalenessLimitSeconds The desired data freshness in seconds.
+   * @param datasetId The ID of the BigQuery dataset.
+   * @return A DestinationConfig object representing a BigQuery destination configuration.
+   */
+  public synchronized DestinationConfig buildBQDestinationConfig(
+      String connectionProfileId, long stalenessLimitSeconds, String datasetId) {
+
+    DestinationConfig.Builder destinationConfigBuilder =
+        DestinationConfig.newBuilder()
+            .setDestinationConnectionProfile(
+                ConnectionProfileName.format(projectId, location, connectionProfileId));
+
+    BigQueryDestinationConfig.Builder bigQueryDestinationConfig =
+        BigQueryDestinationConfig.newBuilder()
+            .setDataFreshness(Duration.newBuilder().setSeconds(stalenessLimitSeconds).build())
+            .setSingleTargetDataset(
+                BigQueryDestinationConfig.SingleTargetDataset.newBuilder().setDatasetId(datasetId));
+
+    destinationConfigBuilder.setBigqueryDestinationConfig(bigQueryDestinationConfig);
+    return destinationConfigBuilder.build();
+  }
+
+  /**
+   * @param streamId The ID of the stream.
+   * @param sourceConfig A SourceConfig object representing the source configuration.
+   * @param destinationConfig A DestinationConfig object representing the destination configuration.
+   * @return A Datastream stream object.
+   */
+  public synchronized Stream createStream(
+      String streamId, SourceConfig sourceConfig, DestinationConfig destinationConfig) {
+
+    LOG.info("Creating Stream {} in project {}.", streamId, projectId);
+
+    try {
+      CreateStreamRequest request =
+          CreateStreamRequest.newBuilder()
+              .setParent(LocationName.format(projectId, location))
+              .setStreamId(streamId)
+              .setStream(
+                  Stream.newBuilder()
+                      .setSourceConfig(sourceConfig)
+                      .setDisplayName(streamId)
+                      .setDestinationConfig(destinationConfig)
+                      .setBackfillAll(Stream.BackfillAllStrategy.getDefaultInstance())
+                      .build())
+              .build();
+
+      Stream reference = datastreamClient.createStreamAsync(request).get();
+      createdStreamIds.add(streamId);
+
+      LOG.info("Successfully created Stream {} in project {}.", streamId, projectId);
+      return reference;
+    } catch (ExecutionException | InterruptedException e) {
+      throw new DatastreamResourceManagerException("Failed to create stream. ", e);
+    }
+  }
+
+  public synchronized Stream updateStreamState(String streamId, Stream.State state) {
+
+    LOG.info("Updating {}'s state to {} in project {}.", streamId, state.name(), projectId);
+
+    try {
+      Stream.Builder streamBuilder =
+          Stream.newBuilder()
+              .setName(StreamName.format(projectId, location, streamId))
+              .setState(state);
+
+      FieldMask.Builder fieldMaskBuilder = FieldMask.newBuilder().addPaths(FIELD_STATE);
+
+      UpdateStreamRequest request =
+          UpdateStreamRequest.newBuilder()
+              .setStream(streamBuilder)
+              .setUpdateMask(fieldMaskBuilder)
+              .build();
+
+      Stream reference = datastreamClient.updateStreamAsync(request).get();
+
+      LOG.info(
+          "Successfully updated {}'s state to {} in project {}.",
+          streamId,
+          state.name(),
+          projectId);
+      return reference;
+    } catch (InterruptedException | ExecutionException e) {
+      throw new DatastreamResourceManagerException("Failed to update stream. ", e);
+    }
+  }
+
+  public synchronized Stream startStream(String streamId) {
+    LOG.info("Starting Stream {} in project {}.", streamId, projectId);
+    return updateStreamState(streamId, Stream.State.RUNNING);
+  }
+
+  public synchronized Stream pauseStream(String streamId) {
+    LOG.info("Pausing Stream {} in project {}.", streamId, projectId);
+    return updateStreamState(streamId, Stream.State.PAUSED);
+  }
+
+  public synchronized void cleanupAll() {
+    LOG.info("Cleaning up Datastream resource manager.");
+    boolean producedError = false;
+
+    try {
+      for (String stream : createdStreamIds) {
+        datastreamClient
+            .deleteStreamAsync(
+                DeleteStreamRequest.newBuilder()
+                    .setName(StreamName.format(projectId, location, stream))
+                    .build())
+            .get();
+      }
+      LOG.info("Successfully deleted stream(s). ");
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Failed to delete stream(s).");
+      producedError = true;
+    }
+
+    try {
+      for (String connectionProfile : createdConnectionProfileIds) {
+        datastreamClient
+            .deleteConnectionProfileAsync(
+                DeleteConnectionProfileRequest.newBuilder()
+                    .setName(ConnectionProfileName.format(projectId, location, connectionProfile))
+                    .build())
+            .get();
+      }
+      LOG.info("Successfully deleted connection profile(s). ");
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Failed to delete connection profile(s).");
+      producedError = true;
+    }
+
+    try {
+      datastreamClient.close();
+    } catch (Exception e) {
+      LOG.error("Failed to close datastream client. ");
+      producedError = true;
+    }
+
+    if (producedError) {
+      throw new DatastreamResourceManagerException(
+          "Failed to delete resources. Check above for errors.");
+    }
+
+    LOG.info("Successfully cleaned up Datastream resource manager.");
+  }
+
+  /** Builder for {@link DatastreamResourceManager}. */
+  public static final class Builder {
+    private final String projectId;
+    private final String location;
+    private CredentialsProvider credentialsProvider;
+
+    private Builder(String projectId, String location) {
+      this.projectId = projectId;
+      this.location = location;
+    }
+
+    public Builder setCredentialsProvider(CredentialsProvider credentialsProvider) {
+      this.credentialsProvider = credentialsProvider;
+      return this;
+    }
+
+    public DatastreamResourceManager build() throws IOException {
+      return new DatastreamResourceManager(this);
+    }
+  }
+}
--- a/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/DatastreamResourceManagerException.java
+++ b/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/DatastreamResourceManagerException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.gcp.datastream;
+
+/** Custom exception for {@link DatastreamResourceManager} implementations. */
+public class DatastreamResourceManagerException extends RuntimeException {
+
+  public DatastreamResourceManagerException(String errorMessage, Throwable err) {
+    super(errorMessage, err);
+  }
+
+  public DatastreamResourceManagerException(String errorMessage) {
+    super(errorMessage);
+  }
+}
--- a/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/JDBCSource.java
+++ b/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/JDBCSource.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.gcp.datastream;
+
+import com.google.protobuf.MessageOrBuilder;
+import java.util.List;
+import java.util.Map;
+
+/** Parent class for JDBC Resources required for configuring Datastream. */
+public abstract class JDBCSource {
+  private final String hostname;
+  private final String username;
+  private final String password;
+  private final int port;
+  private final Map<String, List<String>> allowedTables;
+
+  enum SourceType {
+    ORACLE,
+    MYSQL,
+    POSTGRESQL,
+  }
+
+  JDBCSource(Builder<?> builder) {
+    this.hostname = builder.hostname;
+    this.username = builder.username;
+    this.password = builder.password;
+    this.port = builder.port;
+    this.allowedTables = builder.allowedTables;
+  }
+
+  public abstract SourceType type();
+
+  public abstract MessageOrBuilder config();
+
+  public String hostname() {
+    return this.hostname;
+  }
+
+  public String username() {
+    return this.username;
+  }
+
+  public String password() {
+    return this.password;
+  }
+
+  public int port() {
+    return this.port;
+  }
+
+  public Map<String, List<String>> allowedTables() {
+    return this.allowedTables;
+  }
+
+  public abstract static class Builder<T extends JDBCSource> {
+    private String hostname;
+    private String username;
+    private String password;
+    private int port;
+    private Map<String, List<String>> allowedTables;
+
+    public Builder(
+        String hostname,
+        String username,
+        String password,
+        int port,
+        Map<String, List<String>> allowedTables) {
+      this.hostname = hostname;
+      this.username = username;
+      this.password = password;
+      this.port = port;
+      this.allowedTables = allowedTables;
+    }
+
+    public abstract T build();
+  }
+}
--- a/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/MySQLSource.java
+++ b/it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/datastream/MySQLSource.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.it.gcp.datastream;
+
+import com.google.cloud.datastream.v1.MysqlDatabase;
+import com.google.cloud.datastream.v1.MysqlRdbms;
+import com.google.cloud.datastream.v1.MysqlSourceConfig;
+import com.google.cloud.datastream.v1.MysqlTable;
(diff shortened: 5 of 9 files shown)


it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,195 @@
...
+    private final String location;
+    private final String projectId;
+    private final Map<String, ArrayList<String>> allowedTables;
+    private final String gcsBucketName;
+    private final String gcsRootPath;
+    private final boolean isOracleSource;
+    private final boolean isMySqlSource;
+    private final boolean isPostgresSource;
+    private final boolean isGcsDestination;
+    private final boolean isBqDestination;
```
bvolpato, line 24: Enums for source / destination would make it better and more extensible

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,195 @@
...
+            DatastreamClient datastreamClient,
+            String location, String projectId,
+            Map<String, ArrayList<String>> allowedTables,
+            String gcsBucketName,
+            String gcsRootPath,
+            boolean isMySqlSource,
+            boolean isOracleSource,
+            boolean isPostgresSource,
+            boolean isBqDestination,
+            boolean isGcsDestination) {
```
bvolpato, line 37: Strongly encourage using a builder to make the RM easier to use. Please look at other resource managers and @AutoValue annotations

it/pom.xml
```diff
@@ -15,205 +15,222 @@
...
+            <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
bvolpato, line 117: Reformatting makes it very hard to review and will show all the file with your name on blame. Suggest keeping the original formatting and adding your deps only

```
        <dependency>
            <groupId>com.google.api.grpc</groupId>
            <artifactId>proto-google-cloud-datastream-v1</artifactId>
            <version>1.2.2</version>
            <scope>compile</scope>
        </dependency>
        <dependency>
            <artifactId>google-cloud-datastream</artifactId>
            <version>1.2.2</version>
            <scope>compile</scope>
        </dependency>
        <dependency>
            <groupId>backport-util-concurrent</groupId>
            <artifactId>backport-util-concurrent</artifactId>
            <version>3.1</version>
        </dependency>
```

pom.xml
```diff
@@ -15,1026 +15,1048 @@
   ~ the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~-->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
```
bvolpato, line 17: Same here related to formatting

bvolpato (changes requested):
Thanks! Added a few suggestions.

And don't forget to run `mvn spotless:apply` on the project to fix formatting / license headers

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done. Please review

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.cloud.datastream.v1.ConnectionProfile;
+import com.google.cloud.datastream.v1.Stream;
+
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Interface for managing datastream resources in integration tests.
+ */
+public interface DatastreamResourceManager {
```
Polber, line 12: This should extend `ResourceManager`
```suggestion
public interface DatastreamResourceManager extends ResourceManager {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,30 @@
...
+    Stream createStream(String streamId,
+                        String sourceConnectionProfile,
+                        String destinationConnectionProfile)
+            throws ExecutionException, InterruptedException;
+
+    Stream startStream(String streamId) throws ExecutionException, InterruptedException;
+
+    Stream pauseStream(String streamID) throws ExecutionException, InterruptedException;
+
+    void cleanUpAll() throws ExecutionException, InterruptedException;
```
Polber, line 28: This line can be removed since `cleanupAll()` is inherited. This is so that when we run a suite of integration tests, we can pass multiple resource managers that were used for the tests in a cleanup function that treats all of them as an instance of `ResourceManager` which is guaranteed to have a `cleanupAll()` method.
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+        return updateStreamState(streamId, Stream.State.RUNNING);
+    }
+
+    @Override
+    public Stream pauseStream(String streamID) throws ExecutionException, InterruptedException {
+        return updateStreamState(streamID, Stream.State.PAUSED);
+    }
+
+    @Override
+    public void cleanUpAll() throws ExecutionException, InterruptedException {
```
Polber, line 241: This should be named `cleanupAll()` after the inherited method in `ResourceManager.java`. 
```suggestion
    public void cleanupAll() throws ExecutionException, InterruptedException {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+    @VisibleForTesting
+    public DefaultDatastreamResourceManager(
+            DatastreamClient datastreamClient,
+            String location,
+            String projectId,
+            Map<String, ArrayList<String>> allowedTables,
+            String gcsBucketName,
+            String gcsRootPath,
+            Source source,
+            Destination destination) {
```
Polber, line 62: I didn't go through every single one, but most of these variables are only used once, in which case they do not need to be stored as instance variables. `gcsBucketName` for example is only used when creating the destination connection profile, so it can be given as a parameter to the `createDestinationConnectionProfile()` method. `projectId`, however, is used in nearly every method, so it makes sense to be given to the constructor and stored as an instance variable.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+                .setParent(LocationName.of(projectId, location).toString())
+                .setConnectionProfile(connectionProfileBuilder)
+                .setConnectionProfileId(connectionProfileId)
+                .build();
+
+        ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+        createdConnectionProfileIds.add(connectionProfileId);
+
+        return reference;
+    }
```
Polber, line 144: I think there should be separate `createGCSDestinationConnectionProfile` and `createBigQueryDestinationConnectionProfile` methods, and the enum removed entirely. These two connection profiles require different arguments (e.g. BQ needs dataset/tables, GCS needs path), and as I said in my comment about the constructor, I think those parameters should be passed to this method instead of being stored as instance variables. Both will return a `ConnectionProfile` object to be used when creating the stream, and that way the caller can decide which type they want directly which IMO helps with user-error since all of the builder parameters are optional and can get missed.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+
+    public SourceConfig buildSourceConfig(String sourceConnectionProfileId) {
+        SourceConfig.Builder sourceConfigBuilder = SourceConfig
+                .newBuilder()
+                .setSourceConnectionProfile(ConnectionProfileName
+                        .format(projectId, location, sourceConnectionProfileId));
+
+        if (this.source == Source.MYSQL) {
```
Polber, line 152: This can also be changed to a switch statement like my previous suggestion.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+                mysqlRdmsBuilder.addMysqlDatabases(mysqlDbBuilder);
+            }
+            sourceConfigBuilder.setMysqlSourceConfig(MysqlSourceConfig.newBuilder().setIncludeObjects(mysqlRdmsBuilder));
+        }
+        // TODO: Oracle & Postgres source config builders
+
+        return sourceConfigBuilder.build();
+    }
+
+    public DestinationConfig buildDestinationConfig(String destinationConnectionProfileId) {
```
Polber, line 168: Similar to my comment for `createDestinationConnectionProfile()`, I think this should be split into two methods - one for BQ and one for GCS.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+     */
+    public static final class Builder {
+        private final String projectId;
+        private final String location;
+        private DatastreamClient datastreamClient;
+        private Map<String, ArrayList<String>> allowedTables;
+        private String gcsBucketName;
+        private String gcsRootPath;
+        private DefaultDatastreamResourceManager.Source source;
+        private DefaultDatastreamResourceManager.Destination destination;
```
Polber, line 269: Like I said in my previous comment on the constructor, I think most of these can be removed and refactored to be parameters to their respective methods.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+        }
+        // TODO: BQ Destination Builder
+
+        return destinationConfigBuilder.build();
+    }
+
+    @Override
+    public Stream createStream(String streamId,
+                               String sourceConnectionProfileId,
+                               String destinationConnectionProfileId)
```
Polber, line 192: Based on my other comments, I think this method should take in a `SourceConfig` and `DestinationConfig` that were created using the other methods above. So the workflow from the caller would look something like...
```
// Create resource manager
DatastreamResourceManager rm = DefaultDatastreamResourceManager.builder(...).build();

// Create profiles
createSourceConnectionProfile(...);
createDestinationConnectionProfile(...);

// Create configs
SourceConfig source = rm.buildSourceConfig(...);
DestinationConfig dest = rm.buildDestinationConfig(...);

// Create and start stream
rm.createStream(..., source, dest);
rm.startStream(...);

// Wait...

// Close stream
rm.pauseStream(...);
rm.cleanupAll();
```

Polber (changes requested):
Hey Timothy, I added a few comments to consider as you continue work on this. Thanks for contributing to this effort!

it/pom.xml
```diff
@@ -85,6 +85,23 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsublite</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-datastream-v1</artifactId>
+      <version>1.2.2</version>
```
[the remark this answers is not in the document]
itodotimothy6: Done

it/pom.xml
```diff
@@ -85,6 +85,23 @@
...
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
itodotimothy6: done

it/pom.xml
```diff
@@ -85,6 +85,23 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-datastream</artifactId>
+      <version>1.2.2</version>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
itodotimothy6: done

it/pom.xml
```diff
@@ -85,6 +85,23 @@
...
+    </dependency>
+    <dependency>
+      <groupId>backport-util-concurrent</groupId>
+      <artifactId>backport-util-concurrent</artifactId>
+      <version>3.1</version>
```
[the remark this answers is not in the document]
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
+package com.google.cloud.teleport.it.datastream;
+
+import com.google.cloud.datastream.v1.*;
```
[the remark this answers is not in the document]
itodotimothy6: done. `mvn spotless:check` pointed out some other formatting issues. I'll rerun that when PR is almost complete

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+import com.google.protobuf.FieldMask;
+import org.apache.arrow.util.VisibleForTesting;
+
+import java.io.IOException;
+import java.util.*;
```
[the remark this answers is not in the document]
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+            mysqlProfileBuilder.setHostname(hostname).setUsername(username).setPassword(password).setPort(3306);
+            connectionProfileBuilder.setMysqlProfile(mysqlProfileBuilder);
+        } else if (this.source == Source.ORACLE) {
+            oracleProfileBuilder.setHostname(hostname).setUsername(username).setPassword(password).setPort(1521);
+            connectionProfileBuilder.setOracleProfile(oracleProfileBuilder);
+        } else if (this.source == Source.ORACLE) {
+            postgresqlProfileBuilder.setHostname(hostname).setUsername(username).setPassword(password)
+                    .setPort(5432).setDatabase(postgresDb);
+            connectionProfileBuilder.setPostgresqlProfile(postgresqlProfileBuilder);
+        }
```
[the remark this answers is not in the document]
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+        this.createdStreamIds = Collections.synchronizedSet(new HashSet<>());
+        this.createdConnectionProfileIds = Collections.synchronizedSet(new HashSet<>());
+    }
+
+    public static Builder builder(String projectId, String location) {
+        return new Builder(projectId, location);
+    }
+
+    @Override
+    public ConnectionProfile createSourceConnectionProfile(String connectionProfileId)
```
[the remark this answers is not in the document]
itodotimothy6: Done.

Using the `JDBCConnectionProfile` might be confusing bc in the context of datastream, connection profile means something else. I called it `JDBCSource` instead

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,306 @@
...
+        // TODO: Confirm from team how to handle this. make them attributes or constants?
+        String hostname = "<hostname>";
+        String username = "<username>";
+        String password = "<password>";
+        String postgresDb = "postgres";
+
+        ConnectionProfile.Builder connectionProfileBuilder = ConnectionProfile.newBuilder();
+        MysqlProfile.Builder mysqlProfileBuilder = MysqlProfile.newBuilder();
+        OracleProfile.Builder oracleProfileBuilder = OracleProfile.newBuilder();
+        PostgresqlProfile.Builder postgresqlProfileBuilder = PostgresqlProfile.newBuilder();
```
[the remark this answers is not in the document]
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done. I assume the constants(e.g `AVRO_FILE_FORMAT`) should remain as it is despite being only used once

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
itodotimothy6: done

itodotimothy6:
Added configs for oracle, postgres & BQ

Next step is to implement Exception handling.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,363 @@
...
+
+        public Builder setSourceType(SourceType sourceType) {
+            this.sourceType = sourceType;
+            return this;
+        }
+
+        public Builder setJDBCSource(JDBCSource jdbcSource) {
+            this.jdbcSource = jdbcSource;
+            return this;
+        }
```
Polber, line 357: Just leaving a comment to go alongside my other comments about removing these setters.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,363 @@
...
+
+        private Builder(String projectId, String location) {
+            this.projectId = projectId;
+            this.location = location;
+        }
+
+        public Builder datastreamClient(DatastreamClient datastreamClient) {
+            this.datastreamClient = datastreamClient;
+            return this;
+        }
```
Polber, line 347: The datastreamClient should be getting initialized in the constructor for the resource manager. It should never be passed in to the builder, so this setter should be removed. If there are variables that should be configured in the client, then those should have setters that the constructor can pull values from. For example, I assume there needs to be some sort of credential passed to the client similar to BigQuery and other RM's, so that should have a setter here.

Polber (changes requested):

itodotimothy6:
As per latest comments 
- I created `JDBCSource` abstract class and it's subclasses (`MySQLSource`, `PostgresSource`, `OracleSource`)
- I created the `DatastreamResourceManagerException` custom exception for exception handling.
- Removed `datastreamClient' from the RM builder and initialized in the constructor

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
Polber, line 1: We are changing the structure of the IT framework, and this interface file is no longer needed.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Default class for implementation of {@link DatastreamResourceManager} interface.
+ */
+public final class DefaultDatastreamResourceManager implements DatastreamResourceManager {
```
Polber, line 36: Part of my last comment, this no longer needs to implement the DatastreamResourceManager interface, but should implement ResourceManager. It also does not need the "Default" prefix.
```suggestion
public final class DatastreamResourceManager implements ResourceManager {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(DefaultDatastreamResourceManager.class);
```
Polber, line 38:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(DatastreamResourceManager.class);
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+    public DefaultDatastreamResourceManager(
+            DatastreamClient datastreamClient,
+            String location,
+            String projectId) {
+        this.datastreamClient = datastreamClient;
+        this.location = location;
+        this.projectId = projectId;
+        this.createdStreamIds = Collections.synchronizedSet(new HashSet<>());
+        this.createdConnectionProfileIds = Collections.synchronizedSet(new HashSet<>());
+    }
```
Polber, line 65: I made all the GCP RM's use a similar structure to this before adding the non-GCP RM's, but I prefer the way we have been making the non-GCP RM's that pass the builder to the testing constructor rather than individual variables. I find it cleaner and doesn't restrict which builder variables are set in unit tests. I will go back and refactor the GCP RM's too most likely.
```suggestion
    private DatastreamResourceManager(Builder builder) throws IOException {
        this(
                DatastreamClient.create(),
                builder);
    }

    @VisibleForTesting
    DatastreamResourceManager(
            DatastreamClient datastreamClient,
            DatastreamResourceManager.Builder builder) {
        this.datastreamClient = datastreamClient;
        this.location = builder.location;
        this.projectId = builder.projectId;
        this.createdStreamIds = Collections.synchronizedSet(new HashSet<>());
        this.createdConnectionProfileIds = Collections.synchronizedSet(new HashSet<>());
    }
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    .setConnectionProfileId(connectionProfileId)
+                    .build();
+            ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+            createdConnectionProfileIds.add(connectionProfileId);
+
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create JDBC source connection profile. ", e);
+        }
+
```
Polber, line 110: nit: Could use a "...generated connection profile ... successfully..." log at the end.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+
+    }
+
+    public SourceConfig buildSourceConfig(String sourceConnectionProfileId, JDBCSource source) {
+        Map<String, ArrayList<String>> allowedTables = source.getAllowedTables();
```
Polber, line 115: nit:
```suggestion
        Map<String, List<String>> allowedTables = source.getAllowedTables();
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    oracleRdmsBuilder.addOracleSchemas(oracleSchemaBuilder);
+                }
+                sourceConfigBuilder
+                        .setOracleSourceConfig(OracleSourceConfig.newBuilder().setIncludeObjects(oracleRdmsBuilder));
+        }
+
+        return sourceConfigBuilder.build();
+    }
+
+    @Override
```
Polber, line 166:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    .setParent(LocationName.of(projectId, location).toString())
+                    .setConnectionProfile(connectionProfileBuilder)
+                    .setConnectionProfileId(connectionProfileId)
+                    .build();
+
+            ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+            createdConnectionProfileIds.add(connectionProfileId);
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create JDBC source connection profile. ", e);
```
Polber, line 188:
```suggestion
            throw new DatastreamResourceManagerException("Failed to create GCS source connection profile. ", e);
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        }
```
Polber, line 189: nit: Could also use a success log here.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    public DestinationConfig buildGCSDestinationConfig(
+            String connectionProfileId,
+            String path,
+            String destinationOutputFormat) {
```
Polber, line 195: I would change `destinationOutputFormat` to be either an enum with values  AVRO_FILE_FORMAT and JSON_FILE_FORMAT, or make it a boolean `use_avro` or something similarly named since there are only 2 options. Much clearer than comparing string literals.

And as a side note, when you do compare strings like you do on line 204 with `destinationOutputFormat == AVRO_FILE_FORMAT`, I think the standard, at least in our repo, is to use `.equals()`. Technically someone could pass a value for `destinationOutputFormat` that would cause this to not equal true even if they are the same value. Here is an example of that: https://stackoverflow.com/questions/8676775/string-comparison-in-java-what-is-wrong-with

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+            gcsDestinationConfigBuilder.setAvroFileFormat(AvroFileFormat.getDefaultInstance());
+        } else {
+            gcsDestinationConfigBuilder.setJsonFileFormat(JsonFileFormat.getDefaultInstance());
+        }
+
+        destinationConfigBuilder.setGcsDestinationConfig(gcsDestinationConfigBuilder);
+        return destinationConfigBuilder.build();
+    }
+
+    // TODO: BQ profile builder
```
Polber, line 214:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    .setConnectionProfileId(connectionProfileId)
+                    .build();
+
+            ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+            createdConnectionProfileIds.add(connectionProfileId);
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create BQ destination connection profile. ", e);
+        }
+
```
Polber, line 237: nit: could also use a success log here.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+
+    }
+
+    // TODO: Confirm with Jeff whether to configure dataset for each schema
```
Polber, line 241:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        BigQueryDestinationConfig.Builder bigQueryDestinationConfig = BigQueryDestinationConfig.newBuilder()
+                .setDataFreshness(stalenessLimit)
+                .setSingleTargetDataset(BigQueryDestinationConfig
+                        .SingleTargetDataset.newBuilder().setDatasetId(datasetId));
+
+        destinationConfigBuilder.setBigqueryDestinationConfig(bigQueryDestinationConfig);
+        return destinationConfigBuilder.build();
+    }
+
+    @Override
```
Polber, line 261:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    .setUpdateMask(fieldMaskBuilder)
+                    .build();
+
+            return datastreamClient.updateStreamAsync(request).get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to update stream. ", e);
+        }
+    }
+
+    @Override
```
Polber, line 313:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+    public Stream startStream(String streamId) {
+        LOG.info("Starting Stream {} in project {}.", streamId, projectId);
+        return updateStreamState(streamId, Stream.State.RUNNING);
+    }
+
+    @Override
```
Polber, line 319:
```suggestion
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+            UpdateStreamRequest request = UpdateStreamRequest.newBuilder()
+                    .setStream(streamBuilder)
+                    .setUpdateMask(fieldMaskBuilder)
+                    .build();
+
+            return datastreamClient.updateStreamAsync(request).get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to update stream. ", e);
+        }
+    }
```
Polber, line 311: I would add a log at the end that specifies the state has been successfully updated, and to which state it was updated to.

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        this.projectId = projectId;
+        this.createdStreamIds = Collections.synchronizedSet(new HashSet<>());
+        this.createdConnectionProfileIds = Collections.synchronizedSet(new HashSet<>());
+    }
+
+    public static Builder builder(String projectId, String location) {
+        return new Builder(projectId, location);
+    }
+
+    public ConnectionProfile createJDBCSourceConnectionProfile(String connectionProfileId, JDBCSource source) {
```
Polber, line 71:
```suggestion
    public synchronized ConnectionProfile createJDBCSourceConnectionProfile(String connectionProfileId, JDBCSource source) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create JDBC source connection profile. ", e);
+        }
+
+
+    }
+
+    public SourceConfig buildSourceConfig(String sourceConnectionProfileId, JDBCSource source) {
```
Polber, line 114:
```suggestion
    public synchronized SourceConfig buildSourceConfig(String sourceConnectionProfileId, JDBCSource source) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                }
+                sourceConfigBuilder
+                        .setOracleSourceConfig(OracleSourceConfig.newBuilder().setIncludeObjects(oracleRdmsBuilder));
+        }
+
+        return sourceConfigBuilder.build();
+    }
+
+    @Override
+    public ConnectionProfile createGCSDestinationConnectionProfile(
```
Polber, line 167:
```suggestion
    public synchronized ConnectionProfile createGCSDestinationConnectionProfile(
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+
+            ConnectionProfile reference = datastreamClient.createConnectionProfileAsync(request).get();
+            createdConnectionProfileIds.add(connectionProfileId);
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create JDBC source connection profile. ", e);
+        }
+    }
+
+    public DestinationConfig buildGCSDestinationConfig(
```
Polber, line 192:
```suggestion
    public synchronized DestinationConfig buildGCSDestinationConfig(
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        } else {
+            gcsDestinationConfigBuilder.setJsonFileFormat(JsonFileFormat.getDefaultInstance());
+        }
+
+        destinationConfigBuilder.setGcsDestinationConfig(gcsDestinationConfigBuilder);
+        return destinationConfigBuilder.build();
+    }
+
+    // TODO: BQ profile builder
+    public ConnectionProfile createBQDestinationConnectionProfile(String connectionProfileId) {
```
Polber, line 215:
```suggestion
    public synchronized ConnectionProfile createBQDestinationConnectionProfile(String connectionProfileId) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create BQ destination connection profile. ", e);
+        }
+
+
+    }
+
+    // TODO: Confirm with Jeff whether to configure dataset for each schema
+    public DestinationConfig buildBQDestinationConfig(
```
Polber, line 242:
```suggestion
    public synchronized DestinationConfig buildBQDestinationConfig(
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                .setDataFreshness(stalenessLimit)
+                .setSingleTargetDataset(BigQueryDestinationConfig
+                        .SingleTargetDataset.newBuilder().setDatasetId(datasetId));
+
+        destinationConfigBuilder.setBigqueryDestinationConfig(bigQueryDestinationConfig);
+        return destinationConfigBuilder.build();
+    }
+
+    @Override
+    public Stream createStream(String streamId,
```
Polber, line 262:
```suggestion
    public synchronized Stream createStream(String streamId,
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+            Stream reference = datastreamClient.createStreamAsync(request).get();
+            createdStreamIds.add(streamId);
+
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create stream. ", e);
+        }
+    }
+
+    public Stream updateStreamState(String streamId, Stream.State state) {
```
Polber, line 291:
```suggestion
    public synchronized Stream updateStreamState(String streamId, Stream.State state) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+                    .build();
+
+            return datastreamClient.updateStreamAsync(request).get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to update stream. ", e);
+        }
+    }
+
+    @Override
+    public Stream startStream(String streamId) {
```
Polber, line 314:
```suggestion
    public synchronized Stream startStream(String streamId) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        LOG.info("Starting Stream {} in project {}.", streamId, projectId);
+        return updateStreamState(streamId, Stream.State.RUNNING);
+    }
+
+    @Override
+    public Stream pauseStream(String streamId) {
```
Polber, line 320:
```suggestion
    public synchronized Stream pauseStream(String streamId) {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+        LOG.info("Pausing Stream {} in project {}.", streamId, projectId);
+        return updateStreamState(streamId, Stream.State.PAUSED);
+    }
+
+    @Override
+    public void cleanupAll() {
```
Polber, line 326:
```suggestion
    public synchronized void cleanupAll() {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DefaultDatastreamResourceManager.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            for (String connectionProfile : createdConnectionProfileIds) {
+                datastreamClient.deleteConnectionProfileAsync(DeleteConnectionProfileRequest
+                        .newBuilder().setName(ConnectionProfileName.format(projectId, location, connectionProfile))
+                        .build()).get();
+            }
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to delete datastream resources. ", e);
+        }
```
Polber, line 342: I would put these delete loops in separate try-catch blocks to try and close as many resources as possible even if there is an error. Look at how the `MongDBResourceManager's` `cleanupAll()` works.
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/f00c2f2ccbebe56fbd9569c00aaac4736a519af1/it/mongodb/src/main/java/com/google/cloud/teleport/it/mongodb/MongoDBResourceManager.java#L242-L274

Polber: There also should be a `datastreamClient.close()` or similar function that closes the client's connection to the resource that needs to be invoked here. I've seen issues with new clients being unable to connect to resources in subsequent tests due to unclosed connection streams, so I want to avoid that here.

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,126 @@
...
+
+    enum SourceType {
+        ORACLE,
+        MYSQL,
+        POSTGRESQL,
+    }
+
+    private final SourceType type;
+
+    public abstract SourceType getType();
```
Polber, line 36: nit: I would move this below the constructors

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,126 @@
...
+
+    JDBCSource(JDBCSourceBuilder<?> builder) {
+        this(
+                builder.hostname,
+                builder.username,
+                builder.password,
+                builder.port,
+                builder.type,
+                builder.allowedTables);
+    }
```
Polber, line 46: I don't see a reason to have 2 constructors. I would set all the variables in this constructor and remove the other constructor

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public String getPassword() {
+        return this.password;
+    }
+
+    public int getPort() {
+        return this.port;
+    }
+
```
Polber, line 79: nit: I think the `get` is implied, you could just name these methods `hostname()`, `username()`, etc. Since this is more of a factory class, I think this is the format I tend to see more, but it's not overly important.

it/src/main/java/com/google/cloud/teleport/it/datastream/MySQLSource.java
```diff
@@ -0,0 +1,20 @@
+package com.google.cloud.teleport.it.datastream;
+
+public class MySQLSource extends JDBCSource {
+
+    MySQLSource(MySQLSourceBuilder builder) {
+        super(builder);
+    }
```
Polber, line 7: For all these concrete JDBCSource classes (mySQL, postgres, Oracle), I would add some input validation to constructor as I know some of these require certain variables that other don't and they all share the same builder. Checking that all required variable are set will alleviate some pain points with using this RM.

Polber (changes requested):
I added some comments - A lot of them are regarding format changes due to a refactor of the `it/` module so very minor changes and nits. However, there are a few logical changes as well. Also, if it hasn't already been conveyed, there will need to be javadocs on all the classes and methods, so take a look at the other RM's and make sure it is documented similarily.

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,126 @@
...
+        public JDBCSourceBuilder setAllowedTables(Map<String, ArrayList<String>> allowedTables) {
+            this.allowedTables = allowedTables;
+            return this;
+        }
+
+        public JDBCSourceBuilder setType(SourceType type) {
+            this.type = type;
+            return this;
+        }
```
Polber, line 121: I would remove this setter

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,126 @@
...
+    public int getPort() {
+        return this.port;
+    }
+
+
+    public Map<String, ArrayList<String>> getAllowedTables() {
+        return this.allowedTables;
+    }
+
+    public static abstract class JDBCSourceBuilder<T extends JDBCSource> {
```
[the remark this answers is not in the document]
itodotimothy6: Just dropping a note here to discuss in next sync.

"put required parameters in a constructor and leave the optional ones as setter methods"

I applied the above rule of thumb to the builders. `PostgresSource` requires about 9 params so putting it in one constructor makes it crowdy. I left it this way for now

itodotimothy6:
Unit testing is not yet complete but I'm updating the PR so you have visibility to my current progress prior to our sync tomorrow.

Items done in latest commit

- made corrections per latest review comments
- confirmed RM methods are working (creating resources) by testing creating a temporary `Test` class.

Polber (changes requested):

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+        try {
+            for (String stream : createdStreamIds) {
+                datastreamClient.deleteStreamAsync(DeleteStreamRequest
+                        .newBuilder()
+                        .setName(StreamName.format(projectId, location, stream))
+                        .build()).get();
+            }
+            LOG.info("Successfully deleted stream(s). ");
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to delete stream(s). ", e);
```
Polber, line 352: Change this to an ERROR log so the other closing blocks can be tried.
```suggestion
            LOG.error("Failed to delete stream(s).");
            producedError = true;
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+
+        try {
+            for (String connectionProfile : createdConnectionProfileIds) {
+                datastreamClient.deleteConnectionProfileAsync(DeleteConnectionProfileRequest
+                        .newBuilder().setName(ConnectionProfileName.format(projectId, location, connectionProfile))
+                        .build()).get();
+            }
+            LOG.info("Successfully deleted connection profile(s). ");
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to delete connection profile(s). ", e);
```
Polber, line 363: Same here.
```suggestion
            LOG.error("Failed to delete connection profile(s). ");
            producedError = true;
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        try {
+            datastreamClient.close();
+        } catch (Exception e) {
+            throw new DatastreamResourceManagerException("Failed to close datastream client. ", e);
```
Polber, line 369: And here
```suggestion
            LOG.error("Failed to close datastream client. ");
            producedError = true;
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+    public synchronized Stream pauseStream(String streamId) {
+        LOG.info("Pausing Stream {} in project {}.", streamId, projectId);
+        return updateStreamState(streamId, Stream.State.PAUSED);
+    }
+
+
+    public synchronized void cleanupAll() {
+        LOG.info("Cleaning up Datastream resource manager.");
+
+        try {
```
Polber, line 343: Add a booelan to keep track of errors
```suggestion
        LOG.info("Cleaning up Datastream resource manager.");

        boolean producedError = false;
        try {
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+        } catch (InterruptedException | ExecutionException e) {
+            throw new DatastreamResourceManagerException("Failed to delete connection profile(s). ", e);
+        }
+
+        try {
+            datastreamClient.close();
+        } catch (Exception e) {
+            throw new DatastreamResourceManagerException("Failed to close datastream client. ", e);
+        }
+
```
Polber, line 371: Throw the error at the end.
```suggestion
if (producedError) {
    throw new DatastreamResourceManagerException("Failed to delete resources. Check above for errors.");
}
```

it/src/main/java/com/google/cloud/teleport/it/datastream/MySQLSource.java
Polber, line 1: Needs a license header

it/src/main/java/com/google/cloud/teleport/it/datastream/OracleSource.java
Polber, line 1: Needs a license header.

it/src/main/java/com/google/cloud/teleport/it/datastream/PostgresSource.java
Polber, line 1: Needs a license header.

Polber (changes requested):
Left a few more comments.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+                for (String db : allowedTables.keySet()) {
+                    MysqlDatabase.Builder mysqlDbBuilder = MysqlDatabase.newBuilder().setDatabase(db);
+                    for (String table : allowedTables.get(db)) {
+                        mysqlDbBuilder.addMysqlTables(MysqlTable.newBuilder().setTable(table));
+                    }
+                    mysqlRdmsBuilder.addMysqlDatabases(mysqlDbBuilder);
+                }
+                sourceConfigBuilder
+                        .setMysqlSourceConfig(MysqlSourceConfig.newBuilder().setIncludeObjects(mysqlRdmsBuilder));
+                break;
```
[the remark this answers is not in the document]
itodotimothy6: same as above comment. `sourceConfigBuilder` will update multiple times in some cases. please lmk if my understanding is not correct.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+            LOG.info("Successfully created JDBC Source Connection Profile {} in project {}.", connectionProfileId, projectId);
+            return reference;
+        } catch (ExecutionException | InterruptedException e) {
+            throw new DatastreamResourceManagerException("Failed to create JDBC source connection profile. ", e);
+        }
+    }
+
+    public synchronized SourceConfig buildSourceConfig(String sourceConnectionProfileId, JDBCSource source) {
+        checkArgument(source.getAllowedTables() == null && source.getAllowedTables().isEmpty(),
+                "JDBC source allowed tables can not be null or empty");
```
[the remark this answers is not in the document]
itodotimothy6: just saw the below comments on `config()`. This might not be relevant after the recommended changes in below comment

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+        Map<String, List<String>> allowedTables = source.getAllowedTables();
+        SourceConfig.Builder sourceConfigBuilder = SourceConfig.newBuilder()
+                .setSourceConnectionProfile(ConnectionProfileName
+                        .format(projectId, location, sourceConnectionProfileId));
+
+        switch (source.type()) {
```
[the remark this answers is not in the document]
itodotimothy6: ok this sounds good. 

Removing `allowedTables` from `JDBCSource` means i'll have to pass it as an argument in `config()`. Allowed tables is not optional and is needed to configure the source. I can either have `allowedTables` as an attribute in all 3 subclasses or argument in config(). the latter seems easier. Lmk if you recommend otherwise

image

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
```diff
@@ -0,0 +1,112 @@
...
+            int port,
+            SourceType type,
+            Map<String, List<String>> allowedTables) {
+        this.hostname = hostname;
+        this.username = username;
+        this.password = password;
+        this.port = port;
+        this.type = type;
+        this.allowedTables = allowedTables;
+    }
```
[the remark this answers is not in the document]
itodotimothy6: the other constructor is dependent on this.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
```diff
@@ -0,0 +1,392 @@
...
+        ConnectionProfile.Builder connectionProfileBuilder = ConnectionProfile.newBuilder();
+        JDBCSource.SourceType type = source.type();
+
+        try {
+            switch (type) {
+                case MYSQL:
+                    MysqlProfile.Builder mysqlProfileBuilder = MysqlProfile.newBuilder();
+                    mysqlProfileBuilder.setHostname(source.hostname()).setUsername(source.username()).setPassword(source.password()).setPort(source.port());
+                    connectionProfileBuilder.setMysqlProfile(mysqlProfileBuilder);
+                    break;
```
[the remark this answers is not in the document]
Polber: Ah, you are correct. I am mixed up with Python switch statements 🙃. I thought the `break` was implied in Java, but apparently it is not.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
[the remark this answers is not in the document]
Polber: As you said, this may not be relevant, but I was going off the fact that it is an optional setter method in the `JDBCSource.Builder` which would imply the `allowedTables` is optional. I would make that variable required in the builder's constructor should it stick around after your next refactor. If it gets factored out, then disregard.

it/src/main/java/com/google/cloud/teleport/it/datastream/JDBCSource.java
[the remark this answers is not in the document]
Polber: My point is it does not need to depend on this.

e.g.
```
    JDBCSource(Builder<?> builder) {
        this(
                this.hostname = builder.hostname;
                this.username = builder.username;
                this.password = builder.password;
                this.port = builder.port;
                this.type = builder.type;
                this.allowedTables = builder.allowedTables;
    }
```

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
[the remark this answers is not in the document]
Polber: I think it could remain optional, and if it is not specified, it could default to all tables. Lmk what you think, or if that is even possible.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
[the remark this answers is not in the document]
Polber: My comments are slightly contradictory...

I think the `JDBCSource.Builder` could retain the `setAllowedTables()` method, but it would be abstract and take in a type `T` that could be defined in the subclasses so that they all define what exactly the `allowedTables` are. Since some are schemas and some are table names, I think that would make it clearer.

it/src/main/java/com/google/cloud/teleport/it/datastream/DatastreamResourceManager.java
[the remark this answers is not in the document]
Polber: You are correct.

itodotimothy6:
Updated PR as per our previous conversations.

Polber (approved):

PR #602, opened 2023-02-23
Syndeo file reads transform

pabloem:


pabloem:
based on pubsub notifications: https://cloud.google.com/storage/docs/pubsub-notifications

pabloem:
r: @manavgarg can you let me know what you think?

I still need to write a nice test for this, but for now this is what this transform will look like.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/files/SyndeoFilesReadSchemaTransformProvider.java
```diff
@@ -0,0 +1,287 @@
...
+                                })
+                            .withOutputTags(
+                                FILE_GENERATION_OUTPUT_TAG, TupleTagList.of(ERROR_TAG)));
+
+            PCollectionTuple result =
+                firstResult
+                    .get(FILE_GENERATION_OUTPUT_TAG)
+                    .apply(
+                        Deduplicate.<KV<String, Long>>values()
+                            .withDuration(Duration.standardMinutes(5L)))
```
manavgarg, line 168: is this duration enough for dedupe? what happens if the notification is delayed, would we be consuming the file again?
Also, what would be window interval for the pipeline?

manavgarg (commented):
Thanks Pablo for completing this in such a short time. looks good overall :) (had one question around dedupe and windowing)

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/files/SyndeoFilesReadSchemaTransformProvider.java
pabloem: the pipeline itself would all be on the GlobalWindow - this Duration would be a 'real time' duration of 5 minutes. If a duplicate of a notification arrives over 5 minutes after the initial message, then the file would indeed be consumed twice. We can expand this to be a larger number. I don't think a larger time would incur a significant larger cost.

pabloem:
failed tests are unrelated to Syndeo template. New tests are passing (See Syndeo integration test flow)

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/files/SyndeoFilesReadSchemaTransformProvider.java
manavgarg: i see, thanks for the explanation. Would we need to change the window duration and the dedupe both in that case to a larger number? Just changing the dedupe interval without the windowing duration would not be enough, right?

PR #603, opened 2023-02-23
updating syndeo template to proper dependencies

pabloem:
To use the proper kafka urn and proper snapshot dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,8 +45,7 @@
     <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
 
     <!-- Beam and linked versions -->
-    <beam.version>2.44.0</beam.version>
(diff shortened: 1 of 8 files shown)


pabloem:
r: @manavgarg

manavgarg (dismissed):
LGTM

Abacn:
Hi @pabloem I also opened #604; it would be nice to integrate that PR here. The problem is that the last Beam 2.46.0-SNAPSHOT has breaking changes but the fix is in 2.47.0-SNAPSHOT due to release cut

pabloem:
okay let's see...

pabloem:
fixing up...

pabloem:
finally passing I think : )

PR #595, opened 2023-02-16
Split test utilities into sub projects

Abacn:
Split `google-cloud-platform`, `kafka`, `elasticsearch`, `mongodb` from `it`. This avoids irrelevant client library dependencies.

PR #604, opened 2023-02-23
Bump syndeo-template beam snapshot version to 2.47.0

Abacn:
This should also fix syndeo integration tests


Abacn:
It appears that changing syndeo-template/pom.xml does not trigger syndeo unit test run?

Abacn:
seeing different failures than #595

PR #611, opened 2023-02-24
[BigQueryToBigtable] Add parameters for bigtable client config

kongweihan:
To allow customers fine tune bigtable client's configurations.

### DIFF
--- a/v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
+++ b/v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
@@ -17,6 +17,7 @@ package com.google.cloud.teleport.v2.templates;
 
 import com.google.cloud.bigtable.beam.CloudBigtableIO;
 import com.google.cloud.bigtable.beam.CloudBigtableTableConfiguration;
+import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
 import com.google.cloud.teleport.metadata.Template;
 import com.google.cloud.teleport.metadata.TemplateCategory;
 import com.google.cloud.teleport.metadata.TemplateParameter;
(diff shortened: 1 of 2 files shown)


v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
```diff
@@ -118,6 +119,60 @@ public interface BigQueryToBigtableOptions extends PipelineOptions {
     String getBigtableWriteColumnFamily();
 
     void setBigtableWriteColumnFamily(String value);
+
+    @TemplateParameter.Text(
```
bvolpato, line 123: `TemplateParameter.Text`/`String` being used for what could be `TemplateParameter.Integer`/`String`? I would vote to make it restrictive here as UI can do a better job at validating before gets to us / fails pipeline / etc.

v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
```diff
@@ -118,6 +119,60 @@ public interface BigQueryToBigtableOptions extends PipelineOptions {
...
+    String getBigtableRpcTimeoutMs();
+
+    void setBigtableRpcTimeoutMs(String value);
+
+    @TemplateParameter.Text(
+        order = 14,
+        optional = true,
+        description = "The additional retry codes",
+        helpText = "This sets the additional retry codes")
+    String getBigtableAdditionalRetryCodes();
```
bvolpato, line 173: Have an example of how this looks like? `401,419`? Please add here. The `@TemplateParameter` annotations accept `example = `

v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
kongweihan: Done.

v2/bigquery-to-bigtable/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtable.java
kongweihan: Done.

PR #615, opened 2023-02-28
:construction_worker: add a new parameter form template deployment mojo 

rguillome:
In order to define the artifact registry url against organization's locations constraints.

If at your organisation, some constraints about location (like low-carbon region only) are activated, your can't use gcr.io or eu.gcr.io or us.gcr.io which are multi-region could be not allowed.
This allow to define a new parameter to give a one region focused url for the artifact registry used in the flex template .


bvolpato (approved):
Good, thanks for contributing!

We are still staging most things to GCR, but there's ongoing effort for GAR and this is very helpful.

PR #616, opened 2023-02-28
:adhesive_bandage: add quotes around PK partitionning  in BigQuery

rguillome:
For BQ Merge statement to avoid issues when reserved word column names.


bvolpato:
Thanks for contributing.

Seems like it fails on unit tests `buildMergeStatement_expectedResult(com.google.cloud.teleport.v2.cdc.merge.MergeInfoTest)` now, need to adjust assertions for a little bit. Can you please take a look?

bvolpato (approved):

PR #617, opened 2023-03-01
Support Pubsub's service-supported schemas for reading

pabloem:
r: @manavgarg 

r: @pranavbhandari24


pabloem:
the pubsub schema service implementation is silly, and it decouples the schema format from the encoding format - which nobody else does. I've added AVRO+BINARY(AVRO) encoding, and AVRO+JSON encoding support.

manavgarg:
> the pubsub schema service implementation is silly, and it decouples the schema format from the encoding format - which nobody else does. I've added AVRO+BINARY(AVRO) encoding, and AVRO+JSON encoding support.
> 
> @zhoufek @manavgarg - if a user does not provide Format+Schema, then we look up the schema from Pubsub. Is that reasonable? Another option is to add a `usePubsubSchemas` parameter or something like that. Thoughts?

thanks pablo for working on this. I think this is a reasonable thing to do.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
```diff
@@ -63,43 +67,88 @@ public SchemaTransform from(SyndeoPubsubReadSchemaTransformConfiguration configu
...
+        throw new RuntimeException(
+            String.format(
+                "Unable to retrieve schema information for topic %s or subscription %s",
+                configuration.getTopic(), configuration.getSubscription()),
+            e);
       }
-    };
+    } else {
+      beamSchema =
+          Objects.equals(configuration.getDataFormat(), "JSON")
```
manavgarg, line 98: what if it is neither JSON or AVRO. Should we explicitly catch that error?

manavgarg (dismissed):
LGTM. one minor comment.

PR #620, opened 2023-03-07
Add various performance tests.

pranavbhandari24:
Add the following Load tests,

1. Bulk compressor
2. Text to BigQuery
3. Streaming data generator to Jdbc
4. Jdbc to BQ

Also, add workflows to run load tests.


.github/workflows/java-pr.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     name: Dataflow Templates Integration Tests
     needs: [spotless_check, java_build, java_unit_tests]
     timeout-minutes: 180
-    runs-on: self-hosted
+    runs-on: [self-hosted, it]
```
bvolpato, line 82: nit: Maybe adding a comment that runner has to have both can be useful here.

I [had to lookup the behavior](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on)

> If you specify an array of strings, your workflow will execute on any runner that matches all of the specified runs-on values.

.github/workflows/java-pr.yml
```diff
@@ -95,3 +95,26 @@ jobs:
           --it-artifact-bucket="cloud-teleport-testing-it-gitactions"
       - name: Cleanup Java Environment
         uses: ./.github/actions/cleanup-java-env
+  java_load_tests_templates:
+    if: ${{ github.event.label.name == 'run-load-tests' }}
+    name: Dataflow Templates Load Tests
+    needs: [ spotless_check, java_build, java_unit_tests, java_integration_tests_templates]
+    timeout-minutes: 600
+    runs-on: [self-hosted, perf]
```
bvolpato, line 103: same

cicd/cmd/run-load-tests/main.go
```diff
@@ -0,0 +1,72 @@
...
+	err := workflows.MvnCleanInstall().Run(
+		mvnFlags.IncludeDependencies(),
+		mvnFlags.IncludeDependents(),
+		mvnFlags.SkipDependencyAnalysis(),
+		mvnFlags.SkipCheckstyle(),
+		mvnFlags.SkipJib(),
+		mvnFlags.SkipTests(),
+		mvnFlags.SkipJacoco(),
+		mvnFlags.SkipShade(),
+		mvnFlags.ThreadCount(8))
```
bvolpato, line 44: Do we want to run modules in parallel? We'll probably get mixed results, how do we reduce the variance? Will `perf` machines be larger than `it`?

it/src/test/java/com/google/cloud/teleport/it/jdbc/AbstractJDBCResourceManagerIT.java
```diff
@@ -78,20 +78,20 @@ public void testDefaultMSSQLResourceManagerE2E() {
   private <T extends AbstractJDBCResourceManager<?>> void simpleTest(T rm) {
     try {
       Map<String, String> columns = new LinkedHashMap<>();
+      columns.put("id", "INTEGER");
```
bvolpato, line 81: nit This seems to be handled / added automatically based JDBCSchema.toSqlStatement

it/src/main/java/com/google/cloud/teleport/it/jdbc/JDBCResourceManager.java
```diff
@@ -87,7 +89,7 @@ public interface JDBCResourceManager extends ResourceManager {
    *     if the manager object has no dataset, if the table does not exist or if there is an
    *     Exception when attempting to insert the rows.
    */
-  boolean write(String tableName, Map<Integer, List<Object>> rows);
+  boolean write(String tableName, Map<Object, List<Object>> rows);
```
bvolpato, line 92: Required to make it object? key is the row id which seems to be always int (at least that's what we support today)

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/TextIOtoBigQueryLT.java
```diff
@@ -0,0 +1,178 @@
...
+/** Performance test for {@link TextIOToBigQuery TextIOToBigQuery} template. */
+@TemplateLoadTest(TextIOToBigQuery.class)
+@RunWith(JUnit4.class)
+public class TextIOtoBigQueryLT extends TemplateLoadTestBase {
+  private static final String SPEC_PATH =
+      MoreObjects.firstNonNull(
+          TestProperties.specPath(),
+          "gs://dataflow-templates/latest/flex/GCS_Text_to_BigQuery_Flex");
+  // 35,000,000 messages of the given schema make up approximately 10GB
+  private static final Long NUM_MESSAGES = 35000000L;
```
bvolpato, line 63: Nit but `35_000_000L` is easier to read / avoid messing up with zeroes

v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorLT.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public void testBacklog10gb(Function<LaunchConfig.Builder, LaunchConfig.Builder> paramsAdder)
+      throws IOException, ParseException, InterruptedException {
+    DataGenerator dataGenerator =
+        DataGenerator.builderWithSchemaTemplate(testName, "GAME_EVENT")
+            .setQPS("1000000")
+            .setMessagesLimit(NUM_MESSAGES)
+            .setSinkType("GCS")
+            .setOutputDirectory(getTestMethodDirPath() + "/input")
```
bvolpato, line 84: nit, in ITs we are using `getGcsPath("input")` defined in the base class. can be eventually consolidated (no need to do it now)

bvolpato (dismissed):
Looking good, just a few nits/minor comments, feel free to ignore.

Thanks Pranav!

it/src/main/java/com/google/cloud/teleport/it/jdbc/JDBCResourceManager.java
pranavbhandari24: I changed this to allow non-integer primary keys since I'll be using this for Jdbc to BigQuery load tests.

it/src/test/java/com/google/cloud/teleport/it/jdbc/AbstractJDBCResourceManagerIT.java
pranavbhandari24: I removed the method to add `id` column automatically. Instead, I made it a requirement to supply the id column in the map of fields supplied to JDBCSchema. 

I did this to allow non-integer primary key columns

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/TextIOtoBigQueryLT.java
pranavbhandari24: Done

v1/src/test/java/com/google/cloud/teleport/templates/BulkCompressorLT.java
pranavbhandari24: Sounds good

cicd/cmd/run-load-tests/main.go
pranavbhandari24: Yes, you are right. I removed the thread count flag.

.github/workflows/java-pr.yml
pranavbhandari24: Added comment.

.github/workflows/java-pr.yml
pranavbhandari24: Done

bvolpato (approved):

PR #621, opened 2023-03-08
Fix comparing row key name

krzysztofcybulski:
Hey, it looks like it was intended to check for `_id` content, not String reference


bvolpato (approved):
Good catch, thanks.

Polber (approved):

bvolpato:
Hey @krzysztofcybulski, thanks for contributing! This will eventually get in. We have some tooling that needs to re-run all the time there is a change / merge. There's no need to merge upstream -- but please use rebase instead of merge when possible.

PR #631, opened 2023-03-19
Improving artifacts being imported from syndeo pom file.

pabloem:
The base Beam expansion service includes Beam's Kafka module as a dependency, which causes the classes to clash.

Normally this is not a problem, but it caused me trouble when I was working with snapshot versions of it.

### DIFF
--- a/syndeo-template/pom.xml
+++ b/syndeo-template/pom.xml
@@ -79,6 +79,11 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
+                <groupId>org.apache.beam</groupId>
+                <artifactId>beam-sdks-java-io-kafka</artifactId>
+                <version>${beam.snapshot.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>com.google.api-client</groupId>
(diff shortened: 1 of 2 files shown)


pabloem:
r: @bvolpato 

ptal : )

syndeo-template/pom.xml
```diff
@@ -78,6 +78,11 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.apache.beam</groupId>
+                <artifactId>beam-sdks-java-io-kafka</artifactId>
+                <version>2.47.0-SNAPSHOT</version>
```
bvolpato, line 84: Should we use `${beam.snapshot.version}`?
```suggestion
                <version>${beam.snapshot.version}</version>
```

pabloem: ah yes correct!

bvolpato (dismissed):
LGTM.

We should be able to greatly simplify the POM file by removing things/hardcoded versions that are part of the BOM. But that can be done separately, just wanted to point out.

PR #638, opened 2023-03-23
Added dlq to pubsub read and write schema transform

Dippatel98:
I have added dlq support in the Pubsub Read and Pubsub Write schema transform. Two additional Nullable fields "dlqType" and "dlq" have been added to the schema transforms.

### DIFF
--- a/syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
+++ b/syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
@@ -123,6 +123,15 @@ public class ProviderUtil {
   /** Applies the given configs. */
   public static PCollectionRowTuple applyConfigs(
       Collection<TransformSpec> specs, PCollectionRowTuple tuple) {
+    TransformSpec dlqTransformSpec = null;
+    for (TransformSpec spec : specs) {
+      String[] arrOfUrn = spec.inputId.split(":", 0);
+      if (arrOfUrn[arrOfUrn.length - 2].contains("dlq")) {
+        dlqTransformSpec = spec;
+        specs.remove(dlqTransformSpec);
+        break;
+      }
+    }
     for (TransformSpec spec : specs) {
       SchemaTransform transform = spec.provider.from(spec.configuration);
       // We know we only deal with transforms with either 0 or 1 input so we know how to connect the
@@ -140,7 +149,6 @@ public class ProviderUtil {
       tuple = tuple.apply(spec.inputId, transform.buildTransform());
 
       if (tuple.getAll().containsKey(ERRORS_TAG)) {
-        // TODO(pabloem): Deal with errors.
         PCollectionRowTuple.of("input", tuple.get(ERRORS_TAG))
             .apply(
                 new SyndeoStatsSchemaTransformProvider()
@@ -148,6 +156,12 @@ public class ProviderUtil {
                         SyndeoStatsSchemaTransformProvider.SyndeoStatsConfiguration.create(
                             "errors"))
                     .buildTransform());
+        if (dlqTransformSpec != null) {
+          PCollectionRowTuple.of("errors", tuple.get(ERRORS_TAG))
+              .apply(
+                  dlqTransformSpec.inputId,
+                  dlqTransformSpec.provider.from(dlqTransformSpec.configuration).buildTransform());
+        }
       }
     }
     return tuple;
--- a/syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
+++ b/syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
@@ -45,9 +45,11 @@ import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.sdk.values.KV;
-import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionRowTuple;
+import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.sdk.values.Row;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.sdk.values.TupleTagList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Strings;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Sets;
 import org.checkerframework.checker.initialization.qual.Initialized;
@@ -63,6 +65,11 @@ public class SyndeoPubsubReadSchemaTransformProvider
   public static final String VALID_FORMATS_STR = "AVRO,JSON";
   public static final Set<String> VALID_DATA_FORMATS =
       Sets.newHashSet(VALID_FORMATS_STR.split(","));
+
+  public static final TupleTag<Row> OUTPUT_TAG = new TupleTag<>() {};
+  public static final TupleTag<Row> ERROR_TAG = new TupleTag<>() {};
+  public static final Schema ERROR_SCHEMA =
+      Schema.builder().addStringField("error").addNullableByteArrayField("row").build();
 
   @Override
   public Class<SyndeoPubsubReadSchemaTransformConfiguration> configurationClass() {
@@ -170,13 +177,19 @@ public class SyndeoPubsubReadSchemaTransformProvider
       }
 
       @ProcessElement
-      public void process(ProcessContext c) {
-        PubsubMessage message = c.element();
+      public void process(@DoFn.Element PubsubMessage message, MultiOutputReceiver receiver) {
 
         try {
-          c.output(valueMapper.apply(message.getPayload()));
+          receiver.get(OUTPUT_TAG).output(valueMapper.apply(message.getPayload()));
         } catch (Exception e) {
           errorsInBundle += 1;
+          System.out.println("Error while parsing the element" + e.toString());
+          receiver
+              .get(ERROR_TAG)
+              .output(
+                  Row.withSchema(ERROR_SCHEMA)
+                      .addValues(e.toString(), message.getPayload())
+                      .build());
         }
       }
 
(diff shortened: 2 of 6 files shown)


syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
```diff
@@ -209,6 +241,10 @@ public abstract static class SyndeoPubsubReadSchemaTransformConfiguration {
             + "For JSON data, this is a schema defined with JSON-schema syntax (https://json-schema.org/).")
     public abstract String getSchema();
 
+    public abstract @Nullable String getDlqType();
```
manavgarg, line 244: Do we need to have this and the other one as part of this transform?
I think these should reside with the dlqSchemaTransform as opposed to being declared in every IO schemaTransform.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
```diff
@@ -122,9 +159,14 @@ public abstract static class SyndeoPubsubWriteConfiguration {
         "The name of the topic to write data to. " + "Format: projects/${PROJECT}/topics/${TOPIC}")
     public abstract String getTopic();
 
-    public static SyndeoPubsubWriteConfiguration create(String format, String topic) {
+    public abstract @Nullable String getDlqType();
```
manavgarg, line 162: same comment as the readSchemaTransform.

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
```diff
@@ -0,0 +1,335 @@
...
+  @After
+  public void cleanUp() {
+    ResourceManagerUtils.cleanResources(RESOURCE_MANAGERS.toArray(new ResourceManager[0]));
+    pubsubSourceTopic = null;
+    pubsubSinkTopic = null;
+    pubsubDlqTopic = null;
+    pubsubResourceManager = null;
+  }
+
+  @Test
```
manavgarg, line 130: Eventually (not this PR), we should also add tests to stress the message size and see whats the upper limit for publishing messages to pubsub. And then we should use that value to revise our 10MB limit in the transform.

syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
```diff
@@ -141,13 +142,24 @@ public static PCollectionRowTuple applyConfigs(
...
+        // TODO(dippatel): Remove hardcoded DLQ topic
         PCollectionRowTuple.of("input", tuple.get(ERRORS_TAG))
             .apply(
                 new SyndeoStatsSchemaTransformProvider()
                     .from(
                         SyndeoStatsSchemaTransformProvider.SyndeoStatsConfiguration.create(
                             "errors"))
                     .buildTransform());
+        if (spec.configuration.getString("dlqType") == "PUBSUB"
+            || spec.configuration.getString("dlq") != null) {
```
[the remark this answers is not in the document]
Dippatel98: Yes, "dlq" field will have the topic name. I wanted to keep it generic since we wanted to support more sinks in the future, "dlq_pubsub_topic" would be specific to pubsub. We would have to add more fields or change the name in the future. How about I change it to "dlqSink" instead?

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
Dippatel98: The current PubsubDlqSchemaTransform does have this parameters. But when we are building a Syndeo pipeline in the JsonSpecPaylod we only specify the source and the sink at the top level. We don't add dlq sink as part of the JsonSpec, so in order to get the dlq parameters they need to be passed through the source and sink config parameters. In order to do what you are saying we would have to add dlq as a top level parameter. Something like this:

Old JsonSpecPayload
{
    source
    sink
}


New JsonSpecPayload
{
    source
    sink
    dlq
}

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
Dippatel98: Please refer the above comment

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
Dippatel98: Yeah, sounds good.

syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
```diff
@@ -141,13 +142,24 @@ public static PCollectionRowTuple applyConfigs(
 
       if (tuple.getAll().containsKey(ERRORS_TAG)) {
         // TODO(pabloem): Deal with errors.
```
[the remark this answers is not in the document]
manavgarg: yeah

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
manavgarg: lets add dlq configs as a top level param.

syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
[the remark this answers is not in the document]
manavgarg: I think dlqSink might be a bit misleading and might convey dlqType. So depending on the sink, we might need different  options (like topic for pubsub, bucket for gcs, etc). I think having sink specific configs should be fine IMO.

syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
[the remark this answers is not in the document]
Dippatel98: Done

syndeo-template/src/main/java/com/google/cloud/syndeo/common/ProviderUtil.java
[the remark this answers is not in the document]
Dippatel98: Because of the change from configuration parameters to top level parameters, this is irrelevant now. As now we will be including dlq information as any SchemaTransform at the top level. The parameter name will be directly taken from the relevant dlq SchemaTransform, pubsub in this case.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubReadSchemaTransformProvider.java
Dippatel98: This has been done. Now we can add dlq as a top level parameter like this: 

"dlq": {
    "urn": "syndeo:schematransform:com.google.cloud:pubsub_dlq_write:v1",
    "configurationParameters": {
        "topic": "projects/manav-jit-test/topics/dlq-test"
     }
  }

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
```diff
@@ -73,20 +80,49 @@ public SchemaTransform from(SyndeoPubsubWriteConfiguration configuration) {
...
+                                              Row.withSchema(ERROR_SCHEMA)
+                                                  .addValues(e.toString(), row)
+                                                  .build());
+                                    }
+                                  }
+                                })
+                            .withOutputTags(OUTPUT_TAG, TupleTagList.of(ERROR_TAG)));
+
+            outputTuple
+                .get(OUTPUT_TAG)
```
manavgarg, line 121: just to confirm, the next line when writing to PusubIO does not return any errors currently right?

manavgarg (approved):
Looks good, thanks for working on this. One minor clarification comment.
Also, run spotless and squash the commits before merging.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
Dippatel98: So, it throws exceptions when it encounters an issue, it does not wrap it in an error tag and return nicely. I guess that change would need to be made on the IO side. But yeah the case where the message size is too large will occur in the PubsubIO writer

manavgarg: ok, lets revisit this once IO changes are in.

manavgarg:
@pabloem would you be able to take a look at this and possibly approve if it looks good?

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
Dippatel98: Sounds good

manavgarg (approved):
LGTM

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
```diff
@@ -73,20 +80,49 @@ public SchemaTransform from(SyndeoPubsubWriteConfiguration configuration) {
...
+                                          "Error while parsing the element" + e.toString());
+                                      receiver
+                                          .get(ERROR_TAG)
+                                          .output(
+                                              Row.withSchema(ERROR_SCHEMA)
+                                                  .addValues(e.toString(), row)
+                                                  .build());
+                                    }
+                                  }
+                                })
```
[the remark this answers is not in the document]
Dippatel98: Done

Dippatel98:
> LGTM but let's make sure tests are running fine

@pabloem I have added a new IT PubsubDlqIT which will fail because the syndeo template image in cloud-teleport-testing project does not contain my changes. If we want the test to pass I can upload my template to that project

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/pubsub/SyndeoPubsubWriteSchemaTransformProvider.java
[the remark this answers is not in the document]
Dippatel98: I am guessing the serialization works now because I am seeing a different error now.

manavgarg:
FYI: @johnjcasey (Dip is working on enhancing our DLQ implementation)

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
```diff
@@ -0,0 +1,334 @@
...
+
+    return rootConfiguration;
+  }
+
+  private static JsonNode generateSchema() {
+    JsonNode rootConfiguration = JsonNodeFactory.instance.objectNode();
+    ((ObjectNode) rootConfiguration).put("$id", "https://example.com/person.schema.json");
+    ((ObjectNode) rootConfiguration).put("$schema", "https://json-schema.org/draft-07/schema");
+    ((ObjectNode) rootConfiguration).put("title", "pubsub data schema");
+    ((ObjectNode) rootConfiguration).put("type", "object");
```
bvolpato, line 306: We could cast it only once and make the type of rootConfiguration `ObjectNode`

bvolpato (approved):
LGTM. Added a few comments (can be addressed separately).

PR #643, opened 2023-03-28
Add support for remaining Splunk Write options for GCS-to-Splunk template

andreigurau:
Add the missing Splunk options, found in the Pubsub-to-Splunk template, but pulling the options from the Beam SDK

### DIFF
--- a/v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java
+++ b/v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java
@@ -26,6 +26,7 @@ import com.google.gson.JsonSyntaxException;
 import org.apache.beam.sdk.io.splunk.SplunkEvent;
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.metrics.Metrics;
+import org.apache.beam.sdk.options.Default;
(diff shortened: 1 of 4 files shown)


v2/googlecloud-to-splunk/src/main/java/com/google/cloud/teleport/v2/transforms/SplunkConverters.java
```diff
@@ -134,6 +135,95 @@ public interface SplunkOptions extends PipelineOptions {
     Integer getParallelism();
 
     void setParallelism(Integer parallelism);
+
+    @TemplateParameter.Enum(
+        order = 6,
+        enumOptions = {"PLAINTEXT", "KMS", "SECRET_MANAGER"},
```
bvolpato, line 141: Can we also make this `optional = true`?

bvolpato (changes requested):
It looks good, just one concern about a non-optional parameter that could break existing usage

andreigurau:
> It looks good, just one concern about a non-optional parameter that could break existing usage

This template isn't available for public use yet (waiting on the Splunk resource manager to be created first, and then need to write integration tests). That parameter in Pubsub-to-Splunk is required due to that reason you mentioned, but since no one is using this template, I would prefer to keep this parameter mandatory in this case

bvolpato (dismissed):
My bad, LGTM then! Thanks Andrei

PR #647, opened 2023-03-30
Fix SslConsumerFactory to localize SSL files once per worker.

pranavbhandari24:
SslConsumerFactoryFn is executed once for each connection to kafka. Saving the same file multiple times, and appending if it already exists leads to corrupt ssl files. Changing the implementation to save the file only once.


bvolpato (approved):
Good find! Thanks for the fix Pranav

PR #660, opened 2023-04-24
Differentiate between Dataflow legacy runner and Runner V2.

pranavbhandari24:


it/common/src/main/java/com/google/cloud/teleport/it/common/AbstractPipelineLauncher.java
```diff
@@ -198,8 +212,12 @@ protected JobState handleJobState(Job job) {
   }
 
   /** Creates a JobInfo object from the provided parameters. */
-  protected LaunchInfo getJobInfo(LaunchConfig options, JobState state, Job job, String runner) {
+  protected LaunchInfo getJobInfo(LaunchConfig options, JobState state, Job job) {
     Map<String, String> labels = job.getLabels();
+    String runner =
+        job.getEnvironment().getExperiments().contains("use_runner_v2")
```
bvolpato, line 218: is it null safe if there are no experiments?

bvolpato (dismissed):
LGTM, just added one question/concern regarding a possible NPE

it/common/src/main/java/com/google/cloud/teleport/it/common/AbstractPipelineLauncher.java
pranavbhandari24: Ah, good catch!

Changed the code to make it nullsafe

it/common/src/main/java/com/google/cloud/teleport/it/common/AbstractPipelineLauncher.java
```diff
@@ -198,8 +213,15 @@ protected JobState handleJobState(Job job) {
...
 
   /** Creates a JobInfo object from the provided parameters. */
-  protected LaunchInfo getJobInfo(LaunchConfig options, JobState state, Job job, String runner) {
+  protected LaunchInfo getJobInfo(LaunchConfig options, JobState state, Job job) {
     Map<String, String> labels = job.getLabels();
+    String runner = "Dataflow Legacy Runner";
+    Environment environment = job.getEnvironment();
+    if (environment != null
+        && environment.getExperiments() != null
+        && environment.getExperiments().contains("use_runner_v2")) {
```
bvolpato, line 222: We've used

```
options.addEnvironment("additionalExperiments", Collections.singletonList("use_runner_v2"));
```

To mark use_runner_v2 in some jobs. Not sure if this would've catch it

bvolpato (approved):
LGTM. Left just one minor comment regarding how we've set use_runner_v2 before

it/common/src/main/java/com/google/cloud/teleport/it/common/AbstractPipelineLauncher.java
pranavbhandari24: Expermients checked here are queried from dataflow `getJobs` API, which contain `use_runner_v2` even when it is passed from `additionalExperiments` (manually checked to make sure)

PR #669, opened 2023-04-28
Update syndeo pom file

nickuncaged1201:
Update the syndeo pom to point to Beam 2.48-SNAPSHOT. Also remove explicit dependency of pubsub Lite so to use the bom defined version. These changes are mainly needed to update the latest grpc vendor change in Beam 2.48-SNAPSHOT.

The commented out test needs an update from @Dippatel98 as the OutgoingMessages.of() function interface has changed from a recent update.

### DIFF
--- a/syndeo-template/pom.xml
+++ b/syndeo-template/pom.xml
@@ -208,7 +208,6 @@
         <dependency>
             <groupId>org.apache.beam</groupId>
             <artifactId>beam-sdks-java-io-kafka</artifactId>
-            <version>${beam.snapshot.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
@@ -218,7 +217,6 @@
         <dependency>
             <groupId>org.apache.beam</groupId>
             <artifactId>beam-sdks-java-extensions-google-cloud-platform-core</artifactId>
-            <version>${beam.snapshot.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.beam</groupId>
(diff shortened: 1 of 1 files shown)


nickuncaged1201:
r: @manavgarg , @Dippatel98 , @bvolpato

manavgarg:
Please dont comment out the test.
I have a change pending to fix the test. lets revert that change in this PR and wait for the test to be merged.

syndeo-template/pom.xml
```diff
@@ -208,6 +208,7 @@
         <dependency>
             <groupId>org.apache.beam</groupId>
             <artifactId>beam-sdks-java-io-kafka</artifactId>
+            <version>${beam.snapshot.version}</version>
```
bvolpato, line 211: nit: shouldn't use this if we are using the same version from the bom

syndeo-template/pom.xml
```diff
@@ -217,6 +218,7 @@
         <dependency>
             <groupId>org.apache.beam</groupId>
             <artifactId>beam-sdks-java-extensions-google-cloud-platform-core</artifactId>
+            <version>${beam.snapshot.version}</version>
```
bvolpato, line 221: nit: shouldn't use this if we are using the same version from the bom

syndeo-template/pom.xml
nickuncaged1201: done

syndeo-template/pom.xml
nickuncaged1201: done

bvolpato (approved):

PR #675, opened 2023-05-05
Add Smoke Tests to gitactions

Polber:
Adds `-PtemplatesIntegrationSmokeTests` maven profile to trigger IT's marked with `@Category(DirectRunnerTest.class)`.

This also adds a `java_integration_smoke_tests_templates` gitactions workflow to trigger direct runner smoke tests and verify they are successful before running the full `java_integration_tests_templates` suite on Dataflow.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,9 @@
     <integration.tests>
       com.google.cloud.teleport.metadata.TemplateIntegrationTest
     </integration.tests>
+    <direct-runner.tests>
+      com.google.cloud.teleport.metadata.DirectRunnerTest
+    </direct-runner.tests>
     <excluded.spanner.tests></excluded.spanner.tests>
 
     <licenseHeaderFile>JAVA_LICENSE_HEADER</licenseHeaderFile>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -368,6 +372,40 @@
...
+            <configuration combine.self="override">
+              <includes>
+                <include>**/*IT.java</include>
+              </includes>
+              <excludedGroups>
+                com.google.cloud.teleport.metadata.SkipDirectRunnerTest
+              </excludedGroups>
+              <groups>
+                ${direct-runner.tests}
+              </groups>
```
bvolpato, line 403: These 2 could be contradictory. Maybe we simply don't annotate SkipDirectRunnerTest and keep a comprehensive list of DirectRunnerTest?

bvolpato (approved):
LGTM, just left a clarification comment

pom.xml
Polber: I agree that would contradict, and in practice, no one is going to label a test with `@Category({DirectRunnerTest.class, SkipDirectRunnerTest.class})`, so it is essentially redundant too. I'll remove the `excludedGroups`

PR #680, opened 2023-05-09
Added a test case for max Pubsub msg size

Dippatel98:
Added a test case for max Pubsub msg size. Also a slight edit to the Syndeo Template as it was failing.

### DIFF
--- a/syndeo-template/pom.xml
+++ b/syndeo-template/pom.xml
@@ -25,6 +25,7 @@
     <properties>
         <auto-value.version>1.9</auto-value.version>
         <avro.version>1.11.1</avro.version>
+        <awaitility.version>4.2.0</awaitility.version>
         <beam.version>2.46.0</beam.version>
         <beam.snapshot.version>2.48.0-SNAPSHOT</beam.snapshot.version>
         <bigtable.version>1.15.0</bigtable.version>
@@ -353,6 +354,12 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>kafka</artifactId>
             <version>1.17.3</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>${awaitility.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
+++ b/syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
@@ -15,6 +15,8 @@
  */
 package com.google.cloud.syndeo.it;
 
+import static org.awaitility.Awaitility.await;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -28,12 +30,13 @@ import com.google.cloud.teleport.it.common.TestProperties;
 import com.google.cloud.teleport.it.common.utils.ResourceManagerUtils;
 import com.google.cloud.teleport.it.gcp.dataflow.FlexTemplateClient;
 import com.google.cloud.teleport.it.gcp.pubsub.PubsubResourceManager;
+import com.google.cloud.teleport.it.gcp.pubsub.conditions.PubsubMessagesCheck;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PullResponse;
 import com.google.pubsub.v1.SubscriptionName;
 import com.google.pubsub.v1.TopicName;
 import java.io.IOException;
-import java.time.Instant;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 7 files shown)


Dippatel98:
r: @nickuncaged1201 @bvolpato

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
```diff
@@ -243,6 +243,60 @@ public void testPubsubToPubsubFailure() throws Exception {
...
+        FlexTemplateClient.builder().setCredentials(CREDENTIALS).build();
+    PipelineOperator operator = new PipelineOperator(templateClient);
+
+    PipelineLauncher.LaunchInfo syndeoPipeline = kickstartSyndeoPipeline();
+
+    // Publish large messages to the source queue
+    publishLargeMsgToPubsub();
+
+    // Wait for the Syndeo pipeline to move all the data.
+    Thread.sleep(5 * ONE_MINUTE_MILLIS);
```
bvolpato, line 259: Is this necessary? Can we use something like http://www.awaitility.org/ to make sure we check every X seconds instead of hardcoding this time?

nickuncaged1201 (dismissed):
LGTM

syndeo-template/pom.xml
```diff
@@ -355,6 +355,12 @@
             <version>1.17.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>4.2.0</version>
```
bvolpato, line 361: very nit: our standard way has been <awaitility.version> on properties and just ${awaitility.version} here

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
```diff
@@ -146,35 +148,30 @@ public void testPubsubToPubsubSuccessful() throws Exception {
...
+    // Poll the pubsub sink until the messages are written
+    PubsubMessagesCheck pubsubCheck =
+        PubsubMessagesCheck.builder(pubsubResourceManager, pubsubSinkSubscription)
+            .setMinMessages(MESSAGE_COUNT)
+            .setMaxMessages(MESSAGE_COUNT)
+            .build();
+
+    await("Retrieving messages from Pub/Sub sink")
+        .atMost(Duration.ofMinutes(7))
+        .pollInterval(Duration.ofSeconds(15))
```
bvolpato, line 160: nit: any reason to not make this smaller? e.g., 5s? it produces more logs, but 15s may be representative, specially if we decide to plug some of those tests on DirectRunner later (as we've done for other templates, it's not planned here). for reference, we have ITs being executed in the order of ~5-6s on the DirectRunner.

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
```diff
@@ -146,35 +148,30 @@ public void testPubsubToPubsubSuccessful() throws Exception {
...
+      throw new AssertionError(
+          String.format(
+              "Expected at most %s elements to be inserted to Pubsub sink, but found %s.",
+              MESSAGE_COUNT, msgCount));
+    } else if (msgCount < MESSAGE_COUNT) {
+      throw new AssertionError(
+          String.format(
+              "Expected at least %s elements to be inserted to Pubsub sink, but found %s.",
+              MESSAGE_COUNT, msgCount));
     }
```
bvolpato, line 175: nit I think await would trigger a test failure and never arrive here. may be good to check though

bvolpato (approved):
Looks good to me! Thanks Dip! Added a few nits, feel free to ignore / address later

bvolpato:
This needs a `mvn spotless:apply` though

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
Dippatel98: Actually, I don't think it will even go through the (msgCount > MESSAGE_COUNT) condition because of the way I have setup the PubsubMessagesCheck. I have setMinMessages(MESSAGE_COUNT) and setMaxMessages(MESSAGE_COUNT). So await() will end up erroring out if it doesn't get the exact number of messages. I just added this for sanity, I can remove this if you think that would be cleaner.

syndeo-template/pom.xml
Dippatel98: Done

syndeo-template/src/test/java/com/google/cloud/syndeo/it/PubsubDlqIT.java
bvolpato: Got it. Makes sense. Maybe the check itself should be only for the minimal, since you don't want to wait 7 minutes if you already know you have more?

bvolpato:
```
Error:  Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.02 s <<< FAILURE! - in com.google.cloud.syndeo.common.SyndeoTemplateLaunchTest
Error:  testBuildWithBQToBQConfigBadSink(com.google.cloud.syndeo.common.SyndeoTemplateLaunchTest)  Time elapsed: 0.012 s  <<< ERROR!
java.lang.Exception: Unexpected exception, expected<java.lang.IllegalArgumentException> but was<java.lang.NullPointerException>
	at com.google.cloud.syndeo.common.SyndeoTemplateLaunchTest.testBuildWithBQToBQConfigBadSink(SyndeoTemplateLaunchTest.java:60)
```

FYI There's a test failure

bvolpato (approved):

PR #681, opened 2023-05-09
Refactor resource managers to remove intermediate interfaces.

pranavbhandari24:


pranavbhandari24:
It might be easier to look at the changes by commit instead of all at once

PR #700, opened 2023-05-15
Clean up active dataflow jobs after test completes.

pranavbhandari24:


it/common/src/main/java/com/google/cloud/teleport/it/common/AbstractPipelineLauncher.java
```diff
@@ -51,6 +53,8 @@ public abstract class AbstractPipelineLauncher implements PipelineLauncher {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractPipelineLauncher.class);
   private static final Pattern CURRENT_METRICS = Pattern.compile(".*Current.*");
 
+  protected static final List<String> LAUNCHED_JOBS = new ArrayList<>();
```
bvolpato, line 56: This is static but the cancel runs at @After, can't we have problems if we run surefire with parallelism?

We could probably keep this non-static and it would be fine?

pranavbhandari24: Yeah, that's true. Changed to non-static.

PR #702, opened 2023-05-15
Add a configurable parameter for refreshing UDF

zachberger:
This pull request adds a new parameter `javascriptTextTransformFunctionReloadFrequency` which when set will reload the Javascript from GCS after the period (in seconds) specified in the parameter. If the parameter is not specified then the Javascript UDF will not be reloaded


bvolpato:
Thanks for the ideas here! We are working on a solution based on this for all templates.

PR #725, opened 2023-05-22
Add static resource support to BigtableResourceManager

Polber:


it/google-cloud-platform/src/main/java/com/google/cloud/teleport/it/gcp/bigtable/BigtableResourceManager.java
```diff
@@ -370,12 +401,26 @@ public synchronized ImmutableList<Row> readTable(String tableId) {
...
   @Override
   public synchronized void cleanupAll() {
     LOG.info("Attempting to cleanup manager.");
+
+    if (usingStaticInstance) {
+      try (BigtableTableAdminClient tableAdminClient =
+          bigtableResourceManagerClientFactory.bigtableTableAdminClient()) {
+        createdTables.forEach(tableAdminClient::deleteTable);
+        LOG.info(
+            "This manager was configured to use a static instance that will not be cleaned up.");
```
alexeykukuku, line 419: Looks like the previous line deletes tables, no?

Polber: It will delete tables that were created during the RM's lifetime. i.e. any new tables that were created by calling `createTable()` but not ones that were already present

PR #716, opened 2023-05-19
Cloud Bigtable CDC to BigQuery dataflow template added

alexeykukuku:
This PR adds a new DF template to launch Cloud Bigtable->CDC->BigQuery data pipelines.

This is a clean version of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/628 already reviewed, but I found it's simpler to reapply all changes over clean head revision rather than rebase with conflicts.

PR #728, opened 2023-05-23
Add integration tests for CDC->BigQuery dataflow template

alexeykukuku:
These changes have been moved from https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/716 because the RM story how to deal with static resources hasn't been finalized yet. 

The intent is to merge #716 first, then rebase it against new main and sort out static resources approach


alexeykukuku:
This PR is obsolete

PR #730, opened 2023-05-23
Implemented SchemaIO transform for auto generating test data

xianhualiu:
The DataGeneratorSchemaTransform is a SchemaTransform to generate testing data with any given schema, data frequency and total running time.  It can be used as the source in Syndeo pipeline with urn:  syndeo:schematransform:com.google.cloud:data_generator:v1


syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
```diff
@@ -0,0 +1,103 @@
...
+    private String format;
+
+    public InstantToRowFn(Instant startTime, String schema) {
+      this.startTime = startTime;
+      this.schema = schema;
+    }
+
+    @Override
+    public List<Row> apply(Instant input) {
+      final Schema schema = Schema.parse(this.schema);
```
pabloem, line 92: might be good to avoid parsing the schema on every record? instead only parsing it in the constructor?

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.beam.sdk.values.Row;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+
+public class DataGeneratorSchemaTransform implements SchemaTransform, Serializable {
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private long recordsPerSecond = 1000l;
+  private long minutesToRun = 10;
+  private String schema;
```
pabloem, line 43: these can all be `private final` variables

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
```diff
@@ -0,0 +1,103 @@
...
+        return PCollectionRowTuple.of("output", rows);
+      }
+    };
+  }
+
+  private class InstantToRowFn extends SimpleFunction<Instant, List<Row>> {
+
+    private Instant startTime;
+    private String schema;
+    private String format;
```
pabloem, line 83: these can all be `final`, right?

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransformProvider.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @DefaultSchema(AutoValueSchema.class)
+  @AutoValue
+  public abstract static class DataGeneratorSchemaTransformConfiguration {
+
+    @SchemaFieldDescription("The number of records generated per second")
+    public abstract @Nullable Long getRecordsPerSecond();
+
+    @SchemaFieldDescription("The number of minutes to run the generator.")
+    public abstract @Nullable Long getMinutesToRun();
```
pabloem, line 92: maybe this should be in seconds rather than minutes? Minutes might be a very large resolution for someone who may want to run a 'short' test of e.g. 30 seconds

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public List<Row> apply(Instant input) {
+      final Schema schema = Schema.parse(this.schema);
+
+      final long numSplits = recordsPerSecond * 60;
+      final long periodPerSplitMsecs = 1000;
+      long ordinal =
+          (input.minus(Duration.millis(startTime.getMillis())).getMillis() / periodPerSplitMsecs);
+      return LongStream.range(ordinal * recordsPerSecond, (ordinal + 1) * recordsPerSecond)
+          .mapToObj(intVal -> RecordCreator.createRowRecord(schema))
```
pabloem, line 99: if we parse the schema into a Beam schema early, then we don't need to parse it into avro-and-then-into-beam-schema in each call here (createRowRecord)

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
```diff
@@ -0,0 +1,137 @@
...
+public final class RecordCreator {
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private static final String ALPHA_NUMBERIC_STRING =
+      "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvxyz";
+
+  public static Random random = new Random();
+
+  public static Row createRowRecord(Schema schema) {
+    org.apache.beam.sdk.schemas.Schema beamSchema = AvroUtils.toBeamSchema(schema);
```
pabloem, line 38: let's receive a Beam Schema as input to avoid this conversion

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
```diff
@@ -0,0 +1,137 @@
...
+      if (field.schema().getType() != Schema.Type.RECORD) {
+        Object value = generateRandomValue(field);
+        if (fvbuilder == null) {
+          fvbuilder = builder.withFieldValue(field.name(), value);
+        } else {
+          fvbuilder = fvbuilder.withFieldValue(field.name(), value);
+        }
+      } else {
+        if (fvbuilder == null) {
+          fvbuilder = builder.withFieldValue(field.name(), createGenericRecord(field.schema()));
```
pabloem, line 51: this is a little confusing to me. We are appending an Avro `GenericRecord` to a Beam `Row`. This is not supported. Have you tested that a nested record will work?

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+          fvbuilder = fvbuilder.withFieldValue(field.name(), createGenericRecord(field.schema()));
+        }
+      }
+    }
+    Row row = fvbuilder.build();
+    return row;
+  }
+
+  public static GenericRecord createGenericRecord(Schema schema) {
```
pabloem, line 61: this should probably return a Beam `Row` rather than an avro genericrecord

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+      } else {
+        if (fvbuilder == null) {
+          fvbuilder = builder.withFieldValue(field.name(), createGenericRecord(field.schema()));
+        } else {
+          fvbuilder = fvbuilder.withFieldValue(field.name(), createGenericRecord(field.schema()));
+        }
+      }
+    }
+    Row row = fvbuilder.build();
```
pabloem, line 57: I guess this is silly, but if the schema is empty, then this line will throw a NPE

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
xianhualiu: The reason to use avro schema is that the field can support custom attribute like size which defines the string field length. Wonder if Beam Schema has the similar support.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
xianhualiu: better to keep them not final, since class members suppose to be changeable.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.beam.sdk.transforms.SimpleFunction;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionRowTuple;
+import org.apache.beam.sdk.values.Row;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+
+public class DataGeneratorSchemaTransform implements SchemaTransform, Serializable {
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
```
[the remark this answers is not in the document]
xianhualiu: removed logger since it is not used

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransform.java
xianhualiu: Avro Schema is not serializable, which needs to be created when apply is called. Will consider using serializable schema later.

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/DataGeneratorSchemaTransformProvider.java
xianhualiu: changed to seconds

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
xianhualiu: passed beam schema as parameter

syndeo-template/src/main/java/com/google/cloud/syndeo/transforms/datagenerator/RecordCreator.java
xianhualiu: added check for null builder.

pabloem:
lgtm!

PR #751, opened 2023-05-31
[Integration Tests] Add Spanner ChangeStreams to Pubsub IT

andreigurau:
Add IT for Spanner-ChangeStreams-to-Pubsub template.

This test is flaky on DirectRunner, so annotating this with `SkipDirectRunnerTest`, along with the other Spanner-ChangeStream ITs


bvolpato (approved):
LGTM, thanks Andrei!

I think we could have a more iterative approach using different operations rather than just validating initial load -- I believe that where ITs will help the most. But we can add more tests or increase complexity in a follow up / with Changestreams team's help.

PR #754, opened 2023-05-31
Add command spec file to syndeo template

nickuncaged1201:
Maven jib plugin needs a command spec to be specified to package templates. Added this under syndeo-template/resources.


nickuncaged1201:
r: @bvolpato @manavgarg

manavgarg (approved):

PR #759, opened 2023-06-01
Add BigTableIOLT

Amar3tto:


it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,250 @@
...
+  /** Maps long number to the BigTable format record. */
+  private static class MapToBigTableFormat extends DoFn<Long, KV<ByteString, Iterable<Mutation>>>
+      implements Serializable {
+
+    @ProcessElement
+    public void processElement(ProcessContext c) {
+      int index = c.element().intValue();
+
+      ByteString key = ByteString.copyFromUtf8(String.format("key%09d", index));
+      ByteString value = ByteString.copyFromUtf8(String.format("value%09d", index));
```
Abacn, line 237: The cloud [documentation](https://cloud.google.com/blog/topics/developers-practitioners/bigtable-vs-bigquery-whats-difference) suggests ..."Bigtable can scale to billions of rows and thousands of columns" so this diagonal structured key-value might not be the typical use case for Bigtable

We possibly need a numColumns in Configuration, default set to 1 and make the column key String.format("value%09d", index%numColumn)

Abacn: Not sure if this is the reason of https://github.com/apache/beam/issues/27022. Also going to investigate it

Amar3tto: I tried to set multiple columns as you suggested, it didn't help with the issue.

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,250 @@
...
+          "small", Configuration.of(2_000_000L, 40, "DataflowRunner"), // 100 MB
+          "medium", Configuration.of(20_000_000L, 40, "DataflowRunner"), // 1 GB
+          "large", Configuration.of(200_000_000L, 100, "DataflowRunner") // 10 GB
+          );
+
+  /** Run integration test with configurations specified by TestProperties. */
+  @Test
+  public void testWriteAndRead() throws IOException {
+    tableId = generateTableId(testName);
+    resourceManager.createTable(tableId, ImmutableList.of(COLUMN_FAMILY_NAME));
```
Abacn, line 125: createTable has a default maxAge of 1h. This causes cells without setting timestamp (default to epoch) gets garbage collected. I think it should be never by default. What do you think @pranavbhandari24

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Abacn: I find the cause is
- cell does not have set timestamp, so it default to epoch (1970-01-01)
- the createTable has a garbage collection policy of 1h, so large amount data written triggers GC and some records get deleted

We need to use `.setTimestampMicros(java.time.Instant.now().toEpochMilli()*1000)` for Mutation.SetCell

Amar3tto: Thank you!

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Amar3tto: I've updated the [javadoc](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/759/files#diff-8b8008b8be3d1d84f327e4408a9d975f10c33988b276143b964595aa83a706ecR55) to make it clearer, please check

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,251 @@
...
+  private Configuration configuration;
+  private BigtableOptions bigtableOptions;
+  private String tableId;
+
+  @Rule public TestPipeline writePipeline = TestPipeline.create();
+  @Rule public TestPipeline readPipeline = TestPipeline.create();
+
+  @Before
+  public void setup() throws IOException {
+    resourceManager = DefaultBigtableResourceManager.builder(testName, project).build();
```
Abacn, line 81: There was some refactor on the main branch, `DefaultBigtableResourceManager` has been renamed to `BigtableResourceManager`. Would you mind rebase onto the latest main branch to make the test working properly

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,251 @@
...
+  /** Maps long number to the BigTable format record. */
+  private static class MapToBigTableFormat extends DoFn<Long, KV<ByteString, Iterable<Mutation>>>
+      implements Serializable {
+
+    @ProcessElement
+    public void processElement(ProcessContext c) {
+      Long index = c.element();
+
+      ByteString key = ByteString.copyFromUtf8(String.format("key%09d", index));
+      ByteString value = ByteString.copyFromUtf8(String.format("value%09d", index));
```
Abacn, line 237: We still need a rowSize in Configuration and assign the value of corresponding size here, this is because we are interested in the MB/s throughput also

For the large preset, please use 100M record and 10GB data size. We would like to compare the performance cross IOs using same size of data also

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Abacn, line 81: resource manager should be setup in `@beforeClass`, as it's teardown'd in `@afterClass`

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,251 @@
...
+  private String tableId;
+
+  @Rule public TestPipeline writePipeline = TestPipeline.create();
+  @Rule public TestPipeline readPipeline = TestPipeline.create();
+
+  @Before
+  public void setup() throws IOException {
+    resourceManager = DefaultBigtableResourceManager.builder(testName, project).build();
+
+    bigtableOptions =
```
Abacn, line 83: BigtableOptions is deprecated. Simply use `withInstanceId(instanceId)` in BigtableIO

Amar3tto: Fixed

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Amar3tto: Moved teardown to `@After` according to other tests

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Amar3tto: Done,
Used values from BigQueryIOLT

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Amar3tto: Done

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,256 @@
...
+    assertEquals(
+        PipelineLauncher.JobState.DONE,
+        pipelineLauncher.getJobStatus(project, region, readInfo.jobId()));
+    double numRecords =
+        pipelineLauncher.getMetric(
+            project,
+            region,
+            readInfo.jobId(),
+            getBeamMetricsName(PipelineMetricsType.COUNTER, READ_ELEMENT_METRIC_NAME));
+    assertEquals(configuration.getNumRows(), numRecords, 0.5);
```
Abacn, line 140: Need publish metrics to BigQuery for both write and read pipeline, can refer to here: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/6af4b6300edda4810cdf5108d8b1adff8984940a/it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/storage/FileBasedIOLT.java#L227

Abacn (commented):
Thanks for the quick iteration. Last few comments.

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * BigTableIO performance tests.
+ *
+ * <p>Example trigger command for all tests: "mvn test -pl it/google-cloud-platform -am
+ * -Dtest=BigTableIOLT \ -Dproject=[gcpProject] -DartifactBucket=[temp bucket]
+ * -DfailIfNoTests=false".
+ *
+ * <p>Example trigger command for specific test: "mvn test -pl it/google-cloud-platform -am \
+ * -Dtest="BigTableIOLT#testAvroFileLoadsWriteThenRead" -Dconfiguration=local -Dproject=[gcpProject]
```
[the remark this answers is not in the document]
Amar3tto: Fixed

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
Amar3tto: Done

it/google-cloud-platform/src/test/java/com/google/cloud/teleport/it/gcp/bigtable/BigTableIOLT.java
pranavbhandari24: That's interesting. Yeah, I'd agree that the data should never be garbage collected (if that is an option)
Will look into this.

Abacn (approved):

Abacn:
Thanks, no need to deal with "This branch is out-of-date with the base branch" in the GitHub side. This will be merged once internal test passes

PR #766, opened 2023-06-05
added support for sharding

bharadwaj-aditya:
Added support for sharded migrations in the Datastream-to-spanner template.
These changes will populate the shard id in sharded columns as per the session file.


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
```diff
@@ -294,7 +300,20 @@ public interface Options extends PipelineOptions, StreamingOptions {
     @Default.Boolean(false)
     Boolean getRoundJsonDecimals();
 
-    void setRoundJsonDecimals(Boolean value);
+    @TemplateParameter.Text(
```
Deep1998, line 303: For jsons, especially like these that can have upto 1000 keys, we read them from a gcs file. There is also a limit on the text type (not sure how much). Since we can have alot of shards it might be safer to read this from gcs. Also makes it easier to run it via gcloud.

We already do this for getSessionFilePath() template param.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
```diff
@@ -304,6 +313,21 @@ JsonNode convertTableAndColumnNames(JsonNode changeEvent, String tableName) {
     return changeEvent;
   }
 
+  JsonNode populateShardId(JsonNode changeEvent, String tableId) {
```
Deep1998, line 316: For non-sharded migrations, will shardingConfig be empty?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
```diff
@@ -304,6 +313,21 @@ JsonNode convertTableAndColumnNames(JsonNode changeEvent, String tableName) {
...
+    if (shardingConfig.isEmpty()) {
+      return changeEvent; // Nothing to do
+    }
+    CreateTable table = session.getSpSchema().get(tableId);
+    ColumnDef shardedColDef = table.getColDefs().get(table.getShardedColId());
```
Deep1998, line 321: How does the getShardedColId() behave for non-sharded migrations? Will it always be present in the session file?

v2/datastream-to-spanner/src/test/resources/session-file-sharded.json
```diff
@@ -0,0 +1,285 @@
+{
```
Deep1998, line 1: Is it a good idea to keep an inconsistent session file? t1 table does not have the shardedColId field.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
```diff
@@ -277,6 +283,9 @@ JsonNode transformChangeEventViaSessionFile(JsonNode changeEvent) {
     // Convert table and column names in change event.
     changeEvent = convertTableAndColumnNames(changeEvent, tableName);
 
+    // Add shard id to change event
```
Deep1998, line 286: nit: full stop

darshan-sj (dismissed):

PR #770, opened 2023-06-05
Refactor `it` module.

pranavbhandari24:


bvolpato (approved):
No concerns, thanks Pranav!

From a more general point of view, this feels like broken down into too many small modules (compared to what we had), but seems like a reasonable fit for Beam's approach / organization.

PR #773, opened 2023-06-07
Cleaning up syndeo-template/pom.xml

Dippatel98:
Commenting these dependencies for now. I have tested locally and ran a few syndeo pipelines, it ran fine without it. If there are any issue let me know @manavgarg, @nickuncaged1201. If there are no issues, I will remove these in my next commit.

### DIFF
--- a/syndeo-template/pom.xml
+++ b/syndeo-template/pom.xml
@@ -24,14 +24,14 @@
 
     <properties>
         <auto-value.version>1.9</auto-value.version>
-        <avro.version>1.11.1</avro.version>
+        <!-- <avro.version>1.11.1</avro.version> -->
         <awaitility.version>4.2.0</awaitility.version>
         <beam.version>2.48.0</beam.version>
         <beam.snapshot.version>2.49.0-SNAPSHOT</beam.snapshot.version>
-        <bigtable.version>1.15.0</bigtable.version>
-        <bigquery.version>2.9.0</bigquery.version>
-        <commons.version>1.8</commons.version>
-        <commons-text.version>1.10.0</commons-text.version>
+        <!-- <bigtable.version>1.15.0</bigtable.version> -->
+        <!-- <bigquery.version>2.9.0</bigquery.version> -->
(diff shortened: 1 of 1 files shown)


Dippatel98:
r: @manavgarg , @nickuncaged1201 , @bvolpato

manavgarg (commented):
what all pipelines did you run?
can we run some pipelines testing kafka, pubsub, bq and bt maybe to get more confidence?

Dippatel98:
Yeah, sure. I can do that. I ran spanner, kafka, pslite and ps pipelines

syndeo-template/pom.xml
```diff
@@ -234,12 +234,12 @@
             <artifactId>beam-runners-google-cloud-dataflow-java</artifactId>
             <version>${beam.snapshot.version}</version>
         </dependency>
-        <dependency>
+        <!-- <dependency> -->
             <!-- Direct runner included for local development and testing. -->
-            <groupId>org.apache.beam</groupId>
+            <!-- <groupId>org.apache.beam</groupId>
```
nickuncaged1201, line 239: Not sure if we should remove this? Even though I've never run a pipeline locally, but I don't know if this use case is completely irrelevant

Dippatel98: I don't think anyone runs syndeo on direct runner, not even the tests do that.

Dippatel98: If there is a valid use case, I can add it back.

Dippatel98:
I ran bq and bt pipelines as well. They seem to work fine w/o the dependencies.

manavgarg:
> I ran bq and bt pipelines as well. They seem to work fine w/o the dependencies.

Thanks. LGTM.

manavgarg (approved):

bvolpato (approved):
LGTM.

I would probably remove deps rather than commenting out, we can always go back to this PR and see how it was.

You can do it in 2 phases if it's preferable too, no concerns.

PR #781, opened 2023-06-12
[Bug]: TextToPubsubLT and TextToBigQueryLT jobs aren't drained

aleksandr-dudko:
The method waitForConditionAndCancel was used because the streaming pipeline template includes a call to Splittable DoFn. Invoking a splittable DoFn causes the job to remain in the Draining state indefinitely. See the documentation: https://cloud.google.com/dataflow/docs/guides/stopping-a-pipeline#important_information_about_draining_a_job


elizaveta-lomteva:
This PR will close issue https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/783

PR #784, opened 2023-06-12
Add integration tests for CBT->BQ pipeline, performance improvements

alexeykukuku:
Adding tests for CBT CDC->BQ pipeline along with some minor performance improvements in the pipeline critical data path.

The gist of the optimization is that previously a data-intensive path between CBT and BQ needed to decode and re-encode values multiple times to allow two data streams to be merged: mutations coming from CDC interface and Dead Letter Queue. The DLQ flow requires more re-encoding, but it is of less volume, so I made a shortcut where happy case CDC->BQ processing doesn't do excessive re-encoding.

Tests are currently @Ignored because CBT Admin API client is not released yet. I will be updating this PR when the client is out.


alexeykukuku:
This PR is superseded by https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/919

PR #790, opened 2023-06-15
PubSubToDatadog: replace `site` parameter with `url`

ash-ddog:
Preqreq to: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/791

When we originally created this template, we decided on using a more specific `site` parameter vs a generic `url` parameter, where `site` referred to a [Datadog site](https://docs.datadoghq.com/getting_started/site/). However, in order to facilitate the type of integration testing that already exists in this repo, we need to move back to a generic `url` parameter, so that integration tests can output to a localized Datadog instance (i.e. a mock server).

Let me know if you have any questions!

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/templates/common/DatadogConverters.java
+++ b/v1/src/main/java/com/google/cloud/teleport/templates/common/DatadogConverters.java
@@ -119,7 +119,7 @@ public class DatadogConverters {
     void setBatchCount(ValueProvider<Integer> batchCount);
 
     @TemplateParameter.Integer(
-        order = 5,
+        order = 4,
         optional = true,
         description = "Maximum number of parallel requests.",
         helpText = "Maximum number of parallel requests. Default 1 (no parallelism).")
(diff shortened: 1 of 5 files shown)


v1/src/main/java/com/google/cloud/teleport/templates/common/DatadogConverters.java
```diff
@@ -181,7 +173,7 @@ public interface DatadogOptions extends PipelineOptions {
     void setApiKeySecretId(ValueProvider<String> secretId);
 
     @TemplateParameter.Enum(
-        order = 9,
+        order = 8,
```
ash-ddog, line 176: These were typos from the initial contribution, not related to the parameter change.

bvolpato:
Thanks Ash, no concerns here!

bvolpato:
I mean, there's a regression if users are pointing/passing "site" now, but this is very early stage, so I don't think we need to worry yet.

PR #792, opened 2023-06-15
[Bugfix] Replace only the last occurrence of Datetime patterns in a directory path

ahmedabu98:
While using a template that utilizes DLQ, found some cases where dead-letter GCS directory names get altered unexpectedly if they contain Datetime patterns.

for example:
DLQ would take `gs://bucket/dead-letter/recommendation`  and add on to it `/YYYY/MM/DD/HH/mm` then resolve for datetime patterns. The resulting path would look like this: `gs://bucket/dead-letter/reco56endation/2023/06/15/14/56`. All occurrences of patterns get replaced. The changes in this PR make a more accurate best effort by only replacing the last occurrence in the path.

### DIFF
--- a/v2/common/src/main/java/com/google/cloud/teleport/v2/io/WindowedFilenamePolicy.java
+++ b/v2/common/src/main/java/com/google/cloud/teleport/v2/io/WindowedFilenamePolicy.java
@@ -365,8 +365,8 @@ public abstract class WindowedFilenamePolicy extends FilenamePolicy {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ahmedabu98:
R: @bvolpato

reuvenlax:
This is a risky change - are we sure that there are no customers relying on the old behavior?

ahmedabu98:
@reuvenlax agreed. pranav and I decided to just expose a parameter to configure the datetime pattern

PR #843, opened 2023-07-05
Syndeo cleanup

Dippatel98:
Cleanup:
1. SchemaTransform has been converted from an interface to an abstract class. Making the corresponding changes on the syndeo side.
2. SchemaTransform no longer supports buildTransform(), instead it requires to overriding expand() directly. Changing that and the tests to correctly call the methods.

### DIFF
--- a/syndeo-template/pom.xml
+++ b/syndeo-template/pom.xml
@@ -27,7 +27,7 @@
         <!-- <avro.version>1.11.1</avro.version> -->
         <awaitility.version>4.2.0</awaitility.version>
         <beam.version>2.48.0</beam.version>
-        <beam.snapshot.version>2.49.0-SNAPSHOT</beam.snapshot.version>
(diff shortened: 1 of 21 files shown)


Dippatel98:
r: @manavgarg, @johnjcasey, @ahmedabu98, @bvolpato

manavgarg (approved):
LGTM. Thank you for working on this. I believe we will need to submit https://github.com/apache/beam/pull/27366 before merging this PR?

manavgarg:
Also, I think the Pubsub schemaTransforms are not getting this deleted in this PR. In that case, can we rename the PR to just cleanup and in the follow-up PR we can actually delete those transforms from here. WDYT?

Dippatel98:
Yeah, [apache/beam#27366](https://github.com/apache/beam/pull/27366) needs to be merged before this. Yeah, we can rename this PR.

PR #847, opened 2023-07-06
Allow merging start/end nodes instead of matching

fbiville:


v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapperTest.java
```diff
@@ -0,0 +1,89 @@
...
+  public void ignoresSpecifiedMatchModeForEdgeTargetsInAppendMode() {
+    JSONObject object = targetObject("node", "match");
+    object.put("mode", "append");
+
+    Target target = TargetMapper.fromJson(object);
+
+    assertThat(target.getEdgeNodesMatchMode()).isNull();
+  }
+
+  @Test // TODO: discuss this - set value and reject when validation kicks in on the job spec?
```
fbiville, line 48: ditto :)

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapperTest.java
```diff
@@ -0,0 +1,89 @@
...
+  @Test
+  public void setsSpecifiedMatchModeForEdgeTargets() {
+    JSONObject object = targetObject("edge", "merge");
+
+    Target target = TargetMapper.fromJson(object);
+
+    assertThat(target.getEdgeNodesMatchMode()).isEqualTo(EdgeNodesMatchMode.merge);
+  }
+
+  @Test // TODO: discuss this - set value and reject when validation kicks in on the job spec?
```
fbiville, line 38: ditto

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/enums/EdgeNodesMatchMode.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.neo4j.model.enums;
+
+public enum EdgeNodesMatchMode {
+  match,
+  merge
```
bvolpato, line 20: very nit: any reason for enums to be lowercase? they are usually all upper

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jRowWriterTransform.java
```diff
@@ -91,6 +91,31 @@ public PCollection<Row> expand(PCollection<Row> input) {
...
+      return;
+    }
+    Neo4jConnection neo4jDirectConnect = new Neo4jConnection(neoConnection);
+    LOG.info("Adding {} indices and constraints", cyphers.size());
+    for (String cypher : cyphers) {
+      LOG.info("Executing cypher: {}", cypher);
+      try {
+        neo4jDirectConnect.executeCypher(cypher);
+      } catch (Exception e) {
+        LOG.error("Error executing cypher: {}, {}", cypher, e.getMessage());
```
bvolpato, line 106: nit This would log the entire stacktrace

```suggestion
        LOG.error("Error executing cypher: {}", cypher, e);
```

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/templates/SinglePassGraphImportIT.java
```diff
@@ -0,0 +1,163 @@
...
+    gcsClient.createArtifact(
+        "neo4j-connection.json",
+        String.format(
+            "{\n"
+                + "  \"server_url\": \"%s\",\n"
+                + "  \"database\": \"%s\",\n"
+                + "  \"auth_type\": \"basic\",\n"
+                + "  \"username\": \"neo4j\",\n"
+                + "  \"pwd\": \"%s\"\n"
+                + "}",
```
bvolpato, line 78: hopefully we can swap to Java >15 soon and use [text blocks](https://www.baeldung.com/java-text-blocks) :-)

bvolpato (changes requested):
Not a lot of concerns regarding the code here -- left minor comments.

Is there any point that you want to discuss related to the TODOs?

Marking as changes requested because there are build issues

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/enums/EdgeNodesMatchMode.java
fbiville: The existing enums in the module were already lower case, I assumed this was done specifically for easier (de)serialization

fbiville:
> Is there any point that you want to discuss related to the TODOs?

**TL;DR:**

- shall I keep a single save mode and add a new value to control how edge nodes are matched (either `MATCH` or `MERGE` aka upsert) or introduce a second mode like this PR currently does?
- if a second mode is introduced, what should happen with invalid combinations? Shall I ignore the value and proceed? Shall the execution abort as early as possible?

Two new tests cover these scenarios:

- [`ignoresSpecifiedMatchModeForEdgeTargetsInAppendMode`](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/847/files#diff-42d5bdf4e09e3c13a432cb5a45128d6a73b3c5c25ca8a5654d918a82cf2d3a57R39-R46): the target is for relationship import, the save mode is "append" (aka create the relationship and its nodes at once) and the egde node match mode is set to the non-default value "merge" (which won't have any effect, since "append" means the nodes will always be created, not upserted). The current behavior is to ignore the setting.
- [`ignoresSpecifiedMatchModeForNodeTargets`](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/847/files#diff-42d5bdf4e09e3c13a432cb5a45128d6a73b3c5c25ca8a5654d918a82cf2d3a57R49-R55): the target is for node import, not relationship, so the setting is not relevant in this context. The current behavior is to ignore the setting.

**Long explanation:**

Today, the template accepts a single "save mode" which is either "merge" (aka upsert nodes / relationships to the graph) or "append" (always create).

The way "append" is defined for relationships is to create the start and end nodes, as well as the relationship at once. 
The generated query looks roughly like this:
```cypher
// creates the whole pattern at once - regardless of existing nodes/relationships 
// labels, types, keys and properties are omitted for brevity
CREATE (startNode)-[:RELATIONSHIP]->(endNode)
```

The way upsert is currently defined for relationships is to first find (aka [match](https://neo4j.com/docs/cypher-manual/current/clauses/match/)) the start and end node of the relationship and then upsert the relationship itself. The resulting query roughly looks like this:

```cypher
MATCH (startNode) // find start node (omitting labels & keys for brevity)
MATCH (endNode) // find end node (omitting labels & keys)
MERGE (startNode)-[:RELATIONSHIP]->(endNode) // upsert relationship (omitting type, keys & properties)
```

This PR introduces a secondary mode, "edge node match mode" (better name more than welcome), specifically to control the behavior around the relationship start/end nodes.
By default, it does what's done today: MATCH.
There is a second value which allows to upsert the start/end nodes instead.
This allows users to run the import in a single pass, since nodes will be upserted with their keys as well as the relationships between them.

bvolpato:
Thanks for the explanation! I don't have a lot of the context, my inclination is going towards using upserts by default and allow users control the behavior if they want otherwise -- depending on the source / scenarios, we can have retries or at least once semantics. 

For example, if there is an exception bubbled up during `@FinishBundle`, all the `@ProcessElement` will be repeated.

bvolpato (approved):

PR #852, opened 2023-07-06
[PubSubToDatadog] Send batches before exceeding 5MB payload limit

ash-ddog:
See: https://docs.datadoghq.com/api/latest/logs/#send-logs

Relevant pieces from the link above are:
`Maximum content size per payload (uncompressed): 5MB`
`413: Payload too large (batch is above 5MB uncompressed)`

This PR attempts to change when requests are sent to Datadog to avoid sending requests that will be rejected due to payload exceeding the max limit of `5MB`, as defined in our API docs.

So with these changes, a batch is sent when:
- `currentEventSize + currentBatchSize > maxPayloadSize` (current event excluded)
- `count >= batchCount` (current event included)
- timer exceeds 2s timeout (current event N/A)

We're new to Beam @ Datadog, so we're not 100% certain this is the best way to achieve what we're looking to accomplish, so any guidance is appreciated.

Please let me know if you have questions! 🙏

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
+++ b/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
@@ -27,7 +27,6 @@ import com.google.common.collect.Lists;
 import com.google.common.net.InetAddresses;
 import com.google.common.net.InternetDomainName;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.List;
@@ -188,8 +187,8 @@ public abstract class DatadogEventWriter
     DatadogEvent event = input.getValue();
     INPUT_COUNTER.inc();
 
-    String eventPayload = publisher.getStringPayload(event);
-    long eventPayloadSize = eventPayload.getBytes(StandardCharsets.UTF_8).length;
+    String eventPayload = DatadogEventSerializer.getPayloadString(event);
+    long eventPayloadSize = DatadogEventSerializer.getPayloadSize(eventPayload);
     if (eventPayloadSize > maxBufferSize) {
       LOG.error(
           "Error processing event of size {} due to exceeding max buffer size", eventPayloadSize);
@@ -391,7 +390,7 @@ public abstract class DatadogEventWriter
     }
 
     for (DatadogEvent event : events) {
-      String payload = publisher.getStringPayload(event);
+      String payload = DatadogEventSerializer.getPayloadString(event);
       DatadogWriteError error = builder.withPayload(payload).build();
       receiver.output(error);
     }
--- a/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
+++ b/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
@@ -521,8 +521,6 @@ public abstract class DatadogEventWriter
      * @return {@link Builder}
      */
     public Builder withMaxBufferSize(Long maxBufferSize) {
-      checkArgument(
-          maxBufferSize != null, "withMaxBufferSize(maxBufferSize) called with null input.");
       return setMaxBufferSize(maxBufferSize);
     }
 
--- a/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
+++ b/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
@@ -26,9 +26,8 @@ import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
 import com.google.common.net.InetAddresses;
 import com.google.common.net.InternetDomainName;
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -161,33 +168,63 @@ public void processElement(
...
+    long eventPayloadSize = eventPayload.getBytes(StandardCharsets.UTF_8).length;
+    if (eventPayloadSize > maxBufferSize) {
+      LOG.error(
+          "Error processing event of size {} due to exceeding max buffer size", eventPayloadSize);
+      DatadogWriteError error = DatadogWriteError.newBuilder().withPayload(eventPayload).build();
+      receiver.output(error);
+      return;
+    }
+
+    timer.offset(Duration.standardSeconds(DEFAULT_FLUSH_DELAY)).setRelative();
```
ash-ddog, line 188: I'm not sure if moving this timer call makes a huge difference, but pointing it out to be double-checked.

v1/src/test/java/com/google/cloud/teleport/datadog/DatadogEventWriterTest.java
```diff
@@ -171,13 +174,41 @@ public void eventWriterCustomBatchCountAndValidation() {
...
+  }
+
   /** Test successful POST request for single batch. */
   @Test
   @Category(NeedsRunner.class)
   public void successfulDatadogWriteSingleBatchTest() {
 
     // Create server expectation for success.
-    mockServerListening(200);
+    mockServerListening(202);
```
ash-ddog, line 211: Our API returns `202` on success, so just being more realistic.

v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -161,33 +168,63 @@ public void processElement(
...
-    Long count = MoreObjects.<Long>firstNonNull(countState.read(), 0L);
     DatadogEvent event = input.getValue();
     INPUT_COUNTER.inc();
+
+    String eventPayload = publisher.getStringPayload(event);
+    long eventPayloadSize = eventPayload.getBytes(StandardCharsets.UTF_8).length;
+    if (eventPayloadSize > maxBufferSize) {
+      LOG.error(
+          "Error processing event of size {} due to exceeding max buffer size", eventPayloadSize);
+      DatadogWriteError error = DatadogWriteError.newBuilder().withPayload(eventPayload).build();
```
ash-ddog, line 183: I believe it's ok to create a `DatadogWriteError` without `statusMessage` and `statusCode`, but I'll double-check.

v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -161,33 +168,63 @@ public void processElement(
...
       @StateId(COUNT_STATE_NAME) ValueState<Long> countState,
+      @StateId(BUFFER_SIZE_STATE_NAME) ValueState<Long> bufferSizeState,
       @TimerId(TIME_ID_NAME) Timer timer)
       throws IOException {
 
-    Long count = MoreObjects.<Long>firstNonNull(countState.read(), 0L);
     DatadogEvent event = input.getValue();
     INPUT_COUNTER.inc();
+
+    String eventPayload = publisher.getStringPayload(event);
```
bvolpato, line 178: Assuming serializing to JSON 2x isn't a problem -- you could probably avoid this by estimating the overhead for the JSON fields and the actual fields on the `DatadogEvent`, but the benefits would probably be marginal.

v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -63,6 +62,7 @@ public abstract class DatadogEventWriter
   private static final Integer MAX_BATCH_COUNT = 1000;
   private static final Logger LOG = LoggerFactory.getLogger(DatadogEventWriter.class);
   private static final long DEFAULT_FLUSH_DELAY = 2;
+  private static final Long MAX_BUFFER_SIZE = 5L * 1000 * 1000; // 5MB
```
bvolpato, line 65: Is the limit actually 5mb? or a bit higher? It only considers the payload? Maybe we could have the limit a bit lower here in case there's anything else considered (a really unlikely example would be the `[` `]` as you are sending the array), but my concern is if for example, HTTP headers are taken into account.

v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -161,33 +168,63 @@ public void processElement(
...
+      @StateId(BUFFER_SIZE_STATE_NAME) ValueState<Long> bufferSizeState,
       @TimerId(TIME_ID_NAME) Timer timer)
       throws IOException {
 
-    Long count = MoreObjects.<Long>firstNonNull(countState.read(), 0L);
     DatadogEvent event = input.getValue();
     INPUT_COUNTER.inc();
+
+    String eventPayload = publisher.getStringPayload(event);
+    long eventPayloadSize = eventPayload.getBytes(StandardCharsets.UTF_8).length;
```
bvolpato, line 179: This could probably go to a utility class? And replace here + `calculateByteSize` used in tests

bvolpato:
Please take a look at the following unit tests

```
2023-07-07T17:38:39.0045859Z [ERROR] Errors: 
2023-07-07T17:38:39.0047427Z [ERROR]   DatadogIOTest.successfulDatadogIOMultiBatchNoParallelismTest:91 » IllegalArgument
2023-07-07T17:38:39.0048386Z [ERROR]   DatadogIOTest.successfulDatadogIOMultiBatchParallelismTest:120 » IllegalArgument
2023-07-07T17:38:39.0050940Z [ERROR]   DatadogIOTest.successfulDatadogIOSingleBatchParallelismTest:149 » IllegalArgument
```

v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
ash-ddog: Done!

bvolpato:
No concerns in general, please check unit tests

> DatadogEventWriterTest.successfulDatadogWriteExceedingMaxBufferSize
>
> inputBatchCount must be greater than or equal to 10

ash-ddog:
I think everything is good here! Unrelated failures in the integration tests:

```
CassandraToBigtableIT.testCassandraToBigtable
Error cleaning up resources
```

bvolpato:
Thanks! It's all good, I have a PR out for the error that you mentioned.

bvolpato (approved):

PR #854, opened 2023-07-06
[PubSubToDatadog]  Modify default batch count, add minimum

ash-ddog:
This PR changes:
- `DEFAULT_BATCH_COUNT` from `10` to `100`
- adds the concept of a `MIN_BATCH_COUNT`, set to `10` (not configurable)

I added a `minBatchCount` to a couple of the builders here, to facilitate testing (i.e. still allow testing of `batchCount=1` by setting `minBatchCount=1` as well. In production code, `minBatchCount` is always configured to `MIN_BATCH_COUNT=10`. Let me know if there's a simpler way, or if you have any questions! 🙏

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
+++ b/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
@@ -60,7 +60,7 @@ public abstract class DatadogEventWriter
     extends DoFn<KV<Integer, DatadogEvent>, DatadogWriteError> {
 
   private static final Integer MIN_BATCH_COUNT = 10;
-  private static final Integer DEFAULT_BATCH_COUNT = 10;
+  private static final Integer DEFAULT_BATCH_COUNT = 100;
   private static final Integer MAX_BATCH_COUNT = 1000;
   private static final Logger LOG = LoggerFactory.getLogger(DatadogEventWriter.class);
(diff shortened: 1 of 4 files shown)


v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java
```diff
@@ -443,6 +460,10 @@ public Builder withInputBatchCount(ValueProvider<Integer> inputBatchCount) {
       if (inputBatchCount != null && inputBatchCount.isAccessible()) {
         Integer batchCount = inputBatchCount.get();
         if (batchCount != null) {
+          checkArgument(
+              batchCount >= minBatchCount(),
+              "inputBatchCount must be greater than or equal to %s",
+              minBatchCount());
```
ash-ddog, line 466: `minBatchCount()` must not be `null` by the time this check is performed. That's why it's now part of the builder factory methods `newBuilder()` and `newBuilder(Integer minBatchCount)`

bvolpato (approved):
If I understand correctly -- smaller batches are harmful to Datadog, and you just want to enforce that to be set to at least 10, but adding a parameter to override that, so you can still have tests with 1?

Batch is just a cap, so you are not actually guaranteeing it. If you want to make sure that you accumulate more, we should probably increase `DEFAULT_FLUSH_DELAY`.

bvolpato (approved):

PR #897, opened 2023-07-20
Optimize database reset queries

fbiville:
- use `CALL IN TRANSACTIONS` for batch deletes
 - do no return results from APOC call
 - always consume results from lazily-executed autocommit queries

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
@@ -16,13 +16,12 @@
 package com.google.cloud.teleport.v2.neo4j.database;
 
 import com.google.cloud.teleport.v2.neo4j.model.connection.ConnectionParams;
-import com.google.cloud.teleport.v2.neo4j.model.enums.AuthType;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.Serializable;
-import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.function.Supplier;
 import org.apache.commons.lang3.StringUtils;
 import org.neo4j.driver.AuthTokens;
-import org.neo4j.driver.Config;
 import org.neo4j.driver.Driver;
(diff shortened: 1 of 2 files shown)


bvolpato (dismissed):

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
```diff
@@ -124,28 +95,13 @@ public void resetDatabase() {
...
-        LOG.info("Executing alternate delete cypher: {}", ddeCypher);
-        executeCypher(ddeCypher);
-        String constraintsDeleteCypher =
-            "CALL apoc.schema.assert({},{},true) YIELD label, key RETURN *";
-        executeCypher(constraintsDeleteCypher);
-      } catch (Exception dde) {
-        LOG.error("Error executing detach delete", dde);
-      }
+      String database = !StringUtils.isEmpty(this.database) ? this.database : "neo4j";
+      String cypher = String.format("CREATE OR REPLACE DATABASE `%s`", database);
```
jonathan-lemos, line 99: Isn't this subject to injection?

bvolpato: Yes, good catch. Escaping would be good to prevent issues.

The attack vector might be nonexistent though -- those parameters come from the given template options, and in theory for you to inject something / actually run the statement, the credentials (that are part of the same payload) have to work -- so they already have privileged access / no need for escalation through injection.

fbiville: note: it's as vulnerable as it was before the change, but let me switch to the safer parameterized query instead.

fbiville: thanks for the catch @jonathan-lemos

jonathan-lemos (approved):

PR #901, opened 2023-07-20
Fix support of multiple key definitions for edge source/target nodes

fbiville:
This only concerns the "transposed" spec syntax.

If the keys for the source/target node of the edge were encoded as a JSON object, the existing implementation would only pick the "first" key and define a mapping only for that one.

JSON objects are deserialized into Java's HashMap instances, and these do not provide any guarantees for the iteration order, which makes the key definition very difficult to reason about.

Moreover, the "verbose" syntax allows to define multiple keys and these are properly mapped to multiple mappings.

This commit aligns the behavior between the two syntaxes' mappers.

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TransposedMappingMapper.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TransposedMappingMapper.java
@@ -96,9 +96,11 @@ public class TransposedMappingMapper {
 
     if (edgeMappingsObject.has("source")) {
       JSONObject sourceObj = edgeMappingsObject.getJSONObject("source");
-      FieldNameTuple keyTuple = getFieldAndNameTuple(sourceObj.get("key"));
-      Mapping keyMapping = new Mapping(FragmentType.source, RoleType.key, keyTuple);
-      mappings.add(keyMapping);
+      List<FieldNameTuple> keyTuples = getFieldAndNameTuples(sourceObj.get("key"));
+      for (FieldNameTuple keyTuple : keyTuples) {
+        Mapping keyMapping = new Mapping(FragmentType.source, RoleType.key, keyTuple);
+        mappings.add(keyMapping);
+      }
 
       // support dynamic labels on source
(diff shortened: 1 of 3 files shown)


bvolpato (approved):

bvolpato:
Got a flake importing this internally

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TransposedMappingMapperTest.parsesMultipleKeyMappingsFromObjectForEdgeSourceNode:55 contents match, but order was wrong
expected: [Mapping{constant='null', role=key, type=null, name='value1', labels=[], field='key1', description='', mandatory=false, unique=false, indexed=true, fragmentType=source}, Mapping{constant='null', role=key, type=null, name='value2', labels=[], field='key2', description='', mandatory=false, unique=false, indexed=true, fragmentType=source}]
but was : [Mapping{constant='null', role=key, type=null, name='value2', labels=[], field='key2', description='', mandatory=false, unique=false, indexed=true, fragmentType=source}, Mapping{constant='null', role=key, type=null, name='value1', labels=[], field='key1', description='', mandatory=false, unique=false, indexed=true, fragmentType=source}]
[ERROR]   TransposedMappingMapperTest.parsesMultipleKeyMappingsFromObjectForEdgeTargetNode:157 contents match, but order was wrong
expected: [Mapping{constant='null', role=key, type=null, name='value1', labels=[], field='key1', description='', mandatory=false, unique=false, indexed=true, fragmentType=target}, Mapping{constant='null', role=key, type=null, name='value2', labels=[], field='key2', description='', mandatory=false, unique=false, indexed=true, fragmentType=target}]
but was : [Mapping{constant='null', role=key, type=null, name='value2', labels=[], field='key2', description='', mandatory=false, unique=false, indexed=true, fragmentType=target}, Mapping{constant='null', role=key, type=null, name='value1', labels=[], field='key1', description='', mandatory=false, unique=false, indexed=true, fragmentType=target}]
[INFO] 
[ERROR] Tests run: 42, Failures: 2, Errors: 0, Skipped: 0
```

fbiville:
Apologies for the flaky test, it's for the same reason as the previous occurrence, fixed in https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/968

PR #905, opened 2023-07-21
Make "runtime tokens" parameter optional

fbiville:


fbiville:
The CI currently fails against the Dataflow server, but not the direct runner (`mvn clean verify -am -pl v2/googlecloud-to-neo4j` locally runs file). 
Is there something missing in terms of template metadata update?

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/options/Neo4jFlexTemplateOptions.java
```diff
@@ -50,6 +50,7 @@ public interface Neo4jFlexTemplateOptions extends CommonTemplateOptions {
       description = "Options JSON",
```
bvolpato, line 50: `optional = true`

bvolpato:
I think mandatory parameters (@TemplateParameter `optional` tag) are only checked on Dataflow. It's a gap that we'll have to cover on our direct client. But if you run without direct runner (`-PtemplatesIntegrationTests`) you should see it.

bvolpato:
It seems counterintuitive because you have the `@Default` -- but you still have to tell Dataflow what are the mandatory parameters, because it validates requests based on them, without having access to the class / JAR. Perhaps we could use `@Default` to imply `optional = true`, but I'd have to give it a look for potential impact.

One concern that I have today is related to the order that things show up in the UI (this shouldn't be a blocker, but may cause somewhat bad UX).

bvolpato (approved):

PR #916, opened 2023-07-25
Minor fields rename for consistency

alexeykukuku:
GC is an acronym, we're running with isGC in other templates, hence renaming


bvolpato (approved):

bvolpato:
LGTM, waiting for tests to run before importing

PR #920, opened 2023-07-27
Add UDF Reloading to v2 streaming templates

fozzie15:
Add UDF Reloading to v2 streaming templates.

### DIFF
--- a/it/cassandra/src/main/java/org/apache/beam/it/cassandra/CassandraResourceManager.java
+++ b/it/cassandra/src/main/java/org/apache/beam/it/cassandra/CassandraResourceManager.java
@@ -4,13 +4,13 @@
  * distributed with this work for additional information
  * regarding copyright ownership.  The ASF licenses this file
  * to you under the Apache License, Version 2.0 (the
- * License); you may not use this file except in compliance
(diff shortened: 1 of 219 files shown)


v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/JavascriptTextTransformer.java
bvolpato, line 1: You have a test case covering `JavascriptDocumentTransformer`, but not `JavascriptTextTransformer`.
I'd say that `JavascriptTextTransformer` is more important to cover since it's used in many templates. Document is very specific to MongoDB templates.

It's good that we have tests covering reload for Mongo, but do you think we could also cover one that uses the main transformer?

Thanks Ozzie!

fozzie15: The Pubsub to Big Query test doesn't cover this?

bvolpato: My bad, it does -- thanks!

PR #947, opened 2023-08-03
Performance improvements and integration tests for Bigtable CDC->BQ

alexeykukuku:
This PR has many changes:

1. We address throughput and memory management problems with CDC->BQ by using BQ write API and avoiding DynamicDestinations
2. When switched to Write API, we had to rework DLQ integration because we no longer have access of original content containing metadata when TableRow fails to land to BigQuery table. Now the DLQ is write only for persistent errors.
3. This resulted in changes in Options
4. Added integration tests for the template
5. Fixed DeadLetterQueueManager bug (seems like a copypasta bug)


it/pom.xml
```diff
@@ -34,6 +34,31 @@
     <dataflow-api.version>v1b3-rev20220920-2.0.0</dataflow-api.version>
   </properties>
 
+  <!-- TODO: remove this block once libraries-bom gets past 26.19.0 -->
```
bvolpato, line 37: Good news is that Beam 2.50.0 was updated to 26.21.0 https://github.com/apache/beam/pull/27780, so this is ~3-4 weeks from now

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstobigquery/model/ChangelogColumn.java
```diff
@@ -19,38 +19,49 @@
...
-  TIMESTAMP_FROM("timestamp_from", StandardSQLTypeName.TIMESTAMP.name(), false, false),
-  TIMESTAMP_FROM_NUM("timestamp_from", StandardSQLTypeName.INT64.name(), false, false),
-  TIMESTAMP_TO("timestamp_to", StandardSQLTypeName.TIMESTAMP.name(), false, false),
-  TIMESTAMP_TO_NUM("timestamp_to", StandardSQLTypeName.INT64.name(), false, false),
-  IS_GC("is_gc", "BOOL", true, true),
-  SOURCE_INSTANCE("source_instance", StandardSQLTypeName.STRING.name(), true, true),
-  SOURCE_CLUSTER("source_cluster", StandardSQLTypeName.STRING.name(), true, true),
-  SOURCE_TABLE("source_table", StandardSQLTypeName.STRING.name(), true, true),
-  TIEBREAKER("tiebreaker", StandardSQLTypeName.INT64.name(), true, true),
+  ROW_KEY_STRING("row_key", StandardSQLTypeName.STRING.name(), true, false, null),
```
bvolpato, line 22: Very nit

Just a note that this is a bit hard to read without an IDE that has inlay parameters. Creating a dummy constant NO_DEFAULT_VALUE would make this less worse

bvolpato (commented):
A few nits -- waiting for workflow runs to approve

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstobigquery/model/ChangelogColumn.java
alexeykukuku: Let's incorporate this into other upcoming PRs, really want to merge this stuff

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigtable/BigtableResourceManager.java
```diff
@@ -262,7 +270,10 @@ private void checkHasTable(String tableId) {
    */
   public synchronized void createTable(String tableId, Iterable<String> columnFamilies)
       throws BigtableResourceManagerException {
-    createTable(tableId, columnFamilies, Duration.ofHours(1));
+    BigtableTableSpec spec = new BigtableTableSpec();
+    spec.setColumnFamilies(columnFamilies);
+    spec.setMaxAge(Duration.ofHours(1));
```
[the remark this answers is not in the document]
alexeykukuku: No, 1h is because it was 1h before. 

You can use another constructor 

``public synchronized void createTable(String tableId, BigtableTableSpec bigtableTableSpec)``

And pass ``maxAge`` via ``bigtableTableSpec``, it allows to set any ``maxAge``.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstobigquery/model/ChangelogColumn.java
alexeykukuku: I implemented this, but as you can see it's a bit awkward since I can't use static constants for enum member constructors.

it/pom.xml
alexeykukuku: ack!

bvolpato (approved):
Good to go, thanks Alexey!

PR #953, opened 2023-08-08
Add queryLocation option to bigquery to read from authrozied views

### LINKED ISSUE #952: [Feature Request]: Add bigquery queryLocation option to read from authorized views
### Related Template(s)

BigQuery_to_Bigtable, BigQuery_to_Elasticsearch

### What feature(s) are you requesting?

We need to set `queryLocation` in apache beam to read from authorized views without underlying table's permission.
> * If you are reading from an authorized view wih {@link TypedRead#fromQuery}, you need to use
> * {@link TypedRead#withQueryLocation(String)} to set the location of the BigQuery job. Otherwise,
> * Beam will ty to determine that location by reading the metadata of the dataset that contains the
> * underlying tables. With authorized views, that will result in a 403 error and the query will not
>  * be resolved.

https://github.com/apache/beam/blob/29ea352fa7c96503dfc914c8e2ef982ee37a0c01/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L210-L214

It would be great if we can support the `queryLocation` option to easily read from authorized views.

k-yomo:
Resolves #952 
Add queryLocation option to bigquery to read from authrozied views.


bvolpato:
Thanks Kanji, this looks good -- triggered actions to make sure there's no regression before approving.

bvolpato (approved):

PR #955, opened 2023-08-08
Add support for targets which only define a custom write query

fbiville:


bvolpato (dismissed):
Nice cleanup, thanks!

PR #961, opened 2023-08-10
Bump bigtable version to 2.11.0

kongweihan:


bvolpato (approved):

kongweihan:
Is the unit test failure triggered by the version change? I'm unsure which package is depending on com.google.cloud.teleport.v2:jdbc-common:jar:1.0-SNAPSHOT

bvolpato:
It's a known issue for some modules, working to get rid of this.

We try to be smart about triggering only changed modules (`-pl v2/bigtable-common -am`), but it fails to build transient dependencies.

I've triggered internal tests to make sure there's no regression.

PR #966, opened 2023-08-11
Fix worker type parameter in document snippet

fbiville:


v2/googlecloud-to-neo4j/docs/README.md
```diff
@@ -63,18 +63,20 @@ The template requires the following parameters:
...
-    --template-file-gcs-location="$TEMPLATE_IMAGE_SPEC" \
+    --project "$PROJECT" \
     --region "$REGION" \
+    --template-file-gcs-location="TEMPLATE_SPEC_GCSPATH" \
     --parameters jobSpecUri="gs://neo4j-dataflow/job-specs/testing/text/text-northwind-jobspec.json" \
     --parameters neo4jConnectionUri="gs://neo4j-dataflow/job-specs/testing/connection/auradb-free-connection.json" \
-    --max-workers=1 \
-    --worker-machine-type=${MACHINE_TYPE} 
+    --parameters workerMachineType="$WORKER_MACHINE" \
+    --max-workers=1
```
bvolpato, line 79: We want to rely more and more in the generated doc to avoid it going out of date.

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/googlecloud-to-neo4j/README_Google_Cloud_to_Neo4j.md

Perhaps you can work a bit with the "example" annotations to make it similar to what you see here.

PR #968, opened 2023-08-11
Fix flaky test - explicitly sort mappings by field

fbiville:


bvolpato:
Thanks! Just `mvn spotless:apply` and it should be good to go

bvolpato (approved):

bvolpato (approved):

PR #970, opened 2023-08-14
Add support for staging Python templates via the Templates plugins

andreigurau:
This is basically PR #863, with some additional support for Python templates, by either automatically creating a simple Dockerfile, or by allowing the user to pass in a custom Dockerfile.

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
@@ -55,8 +55,8 @@ import org.twdata.maven.mojoexecutor.MojoExecutor.Element;
  * Goal which stages the Templates into Cloud Storage / Artifact Registry.
  *
  * <p>The process is different for Classic Templates and Flex Templates, please check {@link
- * #stageClassicTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} and {@link
- * #stageFlexTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)}, respectively.
+ * #stageClassicTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} and
+ * {@link #stageFlexTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)}, respectively.
  */
 @Mojo(
     name = "stage",
@@ -106,7 +106,8 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
       required = false)
   protected String basePythonContainerImage;
 
-  public TemplatesStageMojo() {}
+  public TemplatesStageMojo() {
+  }
 
   public TemplatesStageMojo(
       MavenProject project,
@@ -197,9 +198,10 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
   }
 
   /**
-   * Stages a template based on its specific type. See {@link
-   * #stageClassicTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} and {@link
-   * #stageFlexTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} for more details.
+   * Stages a template based on its specific type. See
+   * {@link #stageClassicTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} and
+   * {@link #stageFlexTemplate(TemplateDefinitions, ImageSpec, BuildPluginManager)} for more
+   * details.
    */
   public String stageTemplate(
       TemplateDefinitions definition, ImageSpec imageSpec, BuildPluginManager pluginManager)
@@ -272,8 +274,8 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
             + imageSpec.getMetadata().getMainClass()
             + " with parameters: "
             + arguments.stream()
-                .map(element -> element.toDom().getValue())
-                .collect(Collectors.toList()));
+            .map(element -> element.toDom().getValue())
+            .collect(Collectors.toList()));
 
     executeMojo(
         plugin("org.codehaus.mojo", "exec-maven-plugin"),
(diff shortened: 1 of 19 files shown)


plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/PythonDockerfileGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+    stringBuilder.append(
+        "FROM " + basePythonContainerImage + System.lineSeparator() + System.lineSeparator());
+    stringBuilder.append("ARG WORKDIR=/template" + System.lineSeparator());
+    stringBuilder.append("RUN mkdir -p ${WORKDIR}" + System.lineSeparator());
+    stringBuilder.append("COPY " + containerName + "/main.py ${WORKDIR}" + System.lineSeparator());
+    stringBuilder.append(
+        "COPY " + containerName + "/requirements.txt ${WORKDIR}" + System.lineSeparator());
+    stringBuilder.append("WORKDIR ${WORKDIR}" + System.lineSeparator() + System.lineSeparator());
+    stringBuilder.append(
+        "ENV FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE=requirements.txt" + System.lineSeparator());
```
bvolpato, line 44: It's good to have a separate generator. I think we could keep it more maintainable having the Dockerfile separately, and using something like Freemarker to replace the variables.

Take a look at TemplateDocsGenerator / README-template.md. We could likely have "Dockerfile-template" and use `${}` to interpolate content

python/pom.xml
```diff
@@ -0,0 +1,217 @@
...
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <version>${jib.maven.version}</version>
+        <configuration>
+          <from>
+            <image>${basePythonContainerImage}</image>
+          </from>
+          <container>
+            <appRoot>/template</appRoot>
```
bvolpato, line 80: nit If we'll always use Cloud Build, we don't need to keep jib plugin

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/PythonDockerfileGenerator.java
andreigurau: done

python/pom.xml
andreigurau: done

plugins/core-plugin/src/main/resources/Dockerfile-template
```diff
@@ -0,0 +1,16 @@
+FROM ${containerImage}
```
bvolpato, line 1: I think this should be the base container image

python/pom.xml
```diff
@@ -0,0 +1,189 @@
...
+
+  <name>Templates Python Wrapper Module</name>
+
+  <url>https://github.com/GoogleCloudPlatform/DataflowTemplates</url>
+
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+
+    <jib.maven.version>2.6.0</jib.maven.version>
```
bvolpato, line 40: nit remove too

plugins/core-plugin/src/main/resources/Dockerfile-template
bvolpato: right I just meant baseContainerImage, thanks!

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -402,8 +452,78 @@ protected String stageFlexTemplate(
...
+                        ? outputClassesDirectory.getAbsolutePath()
+                        : dockerfilePath));
+
+    if (process.waitFor() != 0) {
+      throw new RuntimeException(
+          "Error staging container image using gcloud."
+              + new String(process.getInputStream().readAllBytes(), StandardCharsets.UTF_8)
+              + "\n"
+              + new String(process.getErrorStream().readAllBytes(), StandardCharsets.UTF_8));
+    }
```
bvolpato, line 527: Not entirely sure, but this got stuck to me in a test.

```
04:38:47 [INFO] Running: gcloud builds submit --tag gcr.io/{PROJECT}/2023_08_16-03/streaming-llm
06:02:36 Build timed out (after 180 minutes). Marking the build as aborted.
```

My best bet would be that we are not giving `--project` and my instance was configured elsewhere and couldn't make progress, but I'll have to test it first.

Just in case, can you provide the projectId here to avoid doing in some other place?

andreigurau: I'm a little confused on what you mean, but the imagePath/project name is coming from here (which I didn't make changes to): https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/1a77a7b2e534911510876734cb7a218baa981032/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java#L342

Could you post the full command of the test you are trying to run?

bvolpato:
```
export PROJECT=cloud-teleport-testing                                                    
export BUCKET_NAME=cloud-teleport-testing-it

mvn clean install -pl plugins/templates-maven-plugin -am -DskipTests -DskipShade

mvn clean package -PtemplatesStage  \
-DskipTests -DskipShade \
-DprojectId="$PROJECT" \
-DbucketName="$BUCKET_NAME" \
-DstagePrefix="templates" \
-DtemplateName="Word_Count_Python" \
-pl python -e
```

bvolpato: I applied a change to check what's the output (`TemplatePluginUtils.redirectLinesLog`) and the stuckness doesn't happen. I guess there's something funky about `gcloud` when the stream is not being consumed. Pushed that change here.

bvolpato:
Superseded by https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1012 with a change described in the comments

PR #1004, opened 2023-08-23
[Documentation] Auto-generate site documentation for templates

bvolpato:


wtait1-ff:
Hey @bvolpato , where do the docs need to be updated to be generated here? There's an outdated caveat for the `TextIOToBigQuery` template that says it [doesn't support `RECORD` fields](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1004/files#diff-2aca89b21da23ab7d1aaca61baaaca73e08a3f38d75647a9762bd596c8cd4217R67-R69) ([but it does](https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/91d1750ac6b91e9be01839fabeaddfb77b589dfa#diff-2aca89b21da23ab7d1aaca61baaaca73e08a3f38d75647a9762bd596c8cd4217L106))

bvolpato:
@wtait1-ff Thanks for the mention. Our plan is to make the `@Template` annotation (your first link ) the source of truth for all docs. Will file a separate PR for that

PR #1014, opened 2023-08-28
Amended template readme to include data divergence scenario

georgecma:


bvolpato (approved):

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
 
+#### Beware - out-of-order writes can occur in some scenarios
+
+The replicator writes to Hbase as a side effect in Dataflow. If Dataflow loses connection to a worker, another worker will be spun up and a change stream mutation could be duplicated. In the following scenario, a duplication could occur:
+  * Dataflow spins up worker A
+  * Worker A replicates a Put
+  * Dataflow loses connection to worker A before the Put is replicated and spins up worker B
+  * Dataflow assigns the Put to worker B, now 2 workers write the same Put
+
+The above scenario is not a problem most times. However, if a major compaction occurs and a cell had a masking Delete, then a delayed Put will be applied to a cell that should've been deleted. This scenario is theoretically possible but extremely unlikely.
```
jackdingilian, line 25: I think we should make it more clear that there are multiple scenarios here, the other important one being if you have two puts with the same timestamp and different values out of order

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+There are some strategies to deal with this scenario:
+  * Do not write in Deletes.
```
jackdingilian, line 27: We need to add an option to filter GC mutations to support this. Then we should note the need to enable that here.

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+  * Disable major compaction.
+  * Set Hbase `NEW_VERSION_BEHAVIOR = false`. A duplicated Put that arrived later will always be masked by a previous Delete.
+  * Set Hbase `KEEP_DELETED_CELLS = true` (see [here](https://hbase.apache.org/book.html#cf.keep.deleted)). Deletes continue to mask puts even after major compaction. With infinite max_versions and TTL, this is effectively the same as disabling major compaction.
+
```
jackdingilian, line 31: I would add something saying in the unexpected case that this does cause divergence you can correct it by restarting your change stream with a start time preceding the problematic mutations

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
 * Bigtable table with change streams enabled
 * Hbase instance that is accessible from Dataflow
 
-#### Beware
```
vermas2012, line 11: Lets use some neutral terms like "Caveats" rather than an alarming/warning kind of term?

vermas2012: @jackdingilian do you have a better term here?

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+#### Beware - provision Hbase resources to handle Dataflow traffic
 
 * Dataflow when strongly provisioned can translate Bigtable input QPS directly to Hbase, the user should configure Hbase so that Hbase can handle that QPS without crashing.
+
 * Bigtable change streams can handle very large single mutations. Hbase write buffers should be adequately provisioned so as not to cause exceptions when writing in large mutations.
 
+#### Beware - out-of-order writes can occur in some scenarios
+
+The replicator writes to Hbase as a side effect in Dataflow. If Dataflow loses connection to a worker, another worker will be spun up and a change stream mutation could be duplicated. In the following scenario, a duplication could occur:
```
vermas2012, line 19: I don't believe this is accurate. The replicator is a DF sink and writing to HBase is the purpose of the template.

vermas2012, line 19: passive voice. 

How about "DF will spin up another worker. In such scenarios, there maybe a duplicate mutation. Consider the following sequence of events: "

v2/bigtable-changestreams-to-hbase/README.md
vermas2012, line 25: You should explain major compaction in your example. A user unfamiliar with DF won't understand this. Our goal is for users to know that this can theoritically happen, but at the same time, emphasize that its very very unlikely.

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+  * Dataflow spins up worker A
+  * Worker A replicates a Put
```
vermas2012, line 21: I think we should call out somewhere that all mutations coming here have a timestamp assigned, so PUT is an idempotent operation by itself.

v2/bigtable-changestreams-to-hbase/README.md
vermas2012: @jackdingilian 2 mutations with same timestamp is a problem with HBase <-> HBase replication as well. See https://issues.apache.org/jira/browse/HBASE-9465

vermas2012: the only realistic problem I see is "put (ts=0), delete (ts=1), major compaction, put(ts=0)" creates a zombie put.

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
 * Hbase instance that is accessible from Dataflow
 
-#### Beware
+#### Beware - provision Hbase resources to handle Dataflow traffic
 
 * Dataflow when strongly provisioned can translate Bigtable input QPS directly to Hbase, the user should configure Hbase so that Hbase can handle that QPS without crashing.
+
 * Bigtable change streams can handle very large single mutations. Hbase write buffers should be adequately provisioned so as not to cause exceptions when writing in large mutations.
 
+#### Beware - out-of-order writes can occur in some scenarios
```
vermas2012, line 17: Please add a statement with emphasis that says something like

"this can theoretically happen but its very very unlikely"

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+  * Set Hbase `NEW_VERSION_BEHAVIOR = false`. A duplicated Put that arrived later will always be masked by a previous Delete.
+  * Set Hbase `KEEP_DELETED_CELLS = true` (see [here](https://hbase.apache.org/book.html#cf.keep.deleted)). Deletes continue to mask puts even after major compaction. With infinite max_versions and TTL, this is effectively the same as disabling major compaction.
+
 ### Notes on Bidirectional Replication
 
 This template can be configured to be used with the [Hbase-Bigtable replicator](https://github.com/googleapis/java-bigtable-hbase/blob/main/hbase-migration-tools/bigtable-hbase-replication/README.md) out of the box.
 To configure, enable `bidirectional replication` settings in the `Running Template` section below.
 
-#### Beware
+#### Beware - multi-master writes are not supported
```
vermas2012, line 37: There is no way for you to stop multi-master setup. 

You should say, there is a probability of divergence if the same row is modified on the 2 databases simultaneously.

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,17 +8,33 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
 
 Divergence could occur if simultaneous writes occur on both Hbase and Bigtable and they get replicated out at the same time. To minimize this risk, only write to one database at a time.
```
vermas2012, line 39: "to minimize this risk, write to a row from either HBase or Cloud BIgtable". 

They can potentially say, 1 row range is CBT master, other row range is HBase master.That will not be a case for divergence, as a single row is always being written by 1 database. but its technically multi-master as both HBase and CBT take writes from the application.

v2/bigtable-changestreams-to-hbase/README.md
jackdingilian: I can't think of anything obvious better than "Caveats".

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Added `In the unexpected case that divergence occurs, restarting this replicator with a start time preceding the problematic mutations will correct it.`

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Rewrote section, please check new version

georgecma: Rewrote section, please check new version

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Rewrote section, now
`* Hbase undergoes a major compaction which [removes the Delete marker](https://hbase.apache.org/book.html#_delete)`

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Will create another PR to add this option

jackdingilian (commented):
LGTM

bvolpato (dismissed):

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,19 +8,53 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
 * Bigtable table with change streams enabled
 * Hbase instance that is accessible from Dataflow
 
-#### Beware
+#### Caveat - provision Hbase resources to handle Dataflow traffic
 
 * Dataflow when strongly provisioned can translate Bigtable input QPS directly to Hbase, the user should configure Hbase so that Hbase can handle that QPS without crashing.
```
vermas2012, line 13: Can you find a better word than "Strongly provisioned"?

v2/bigtable-changestreams-to-hbase/README.md
```diff
@@ -8,19 +8,53 @@ The [BigtableToHBase](src/main/java/com/google/cloud/teleport/v2/templates/Bigta
...
+
 * Bigtable change streams can handle very large single mutations. Hbase write buffers should be adequately provisioned so as not to cause exceptions when writing in large mutations.
 
+#### Caveat - data divergence can occur in extreme edge case
+
+Data divergence can occur in the following scenario:
+* Dataflow assigns a Put to worker A
```
vermas2012, line 20: You should add something here  on the lines of (All mutations for a row will go to same worker).

vermas2012 (approved):

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Changed wording pls review again

v2/bigtable-changestreams-to-hbase/README.md
georgecma: Added

georgecma:
@bvolpato hi Bruno, these changes can be pushed in now

PR #1043, opened 2023-09-11
Verify Neo4j connectivity early

fbiville:


fbiville:
I initially wanted to unit-test-drive this, but the parsing/validation logic is currently too much tangled in `com.google.cloud.teleport.v2.neo4j.templates.GoogleCloudToNeo4j` constructor, which makes unit testing borderline impossible.

The PR would significantly grow if I were to include this restructuring here, so I decided to not do it, at least for now.

PR #1045, opened 2023-09-12
Add instructions to work around `File too large` error when staging

nithinsujir:
The error seems to be happening due to the shaded jar file size excluding some limit. Adding `-DskipShade` keeps the jar separate and small.


nithinsujir:
@bvolpato PTAL

bvolpato (approved):

PR #1052, opened 2023-09-16
Add YAML Template

Polber:


python/src/test/java/com/google/cloud/teleport/templates/python/YAMLTemplateIT.java
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.runners.JUnit4;
+
+/** Integration test for {@link YAMLTemplate}. */
+// SkipDirectRunnerTest: Python templates are not supported through DirectRunner yet.
+@Category({TemplateIntegrationTest.class, SkipDirectRunnerTest.class})
+@TemplateIntegrationTest(YAMLTemplate.class)
+@RunWith(JUnit4.class)
+public final class YAMLTemplateIT extends TemplateTestBase {
+
+  private static String yamlMessage =
```
bvolpato, line 46: Missing a spotless:apply

Also: shouldn't we maybe import the yaml from a file? It's way easier to maintain, I guess.

In the future we'll have Java blocks and it may be easier to maintain it as a literal

bvolpato (dismissed):

bvolpato:
Merged

PR #1053, opened 2023-09-18
Fix schema statement generation

fbiville:
This is a **big** PR.

The main goal is to make sure that only the required Cypher schema statements are generated.

Here are the following Neo4j schema "rules" that have guided the implementation of this PR:

- a node/rel key constraint (aka `key` or `keys` in the spec) is equivalent to the combination of non-null and unique constraints
- a node/rel key constraint is implicitly indexed (no index generation needed for the targeted properties)
- a uniqueness constraint  (aka `uniques` in the spec) is implicitly indexed (no index generation needed for the targeted properties)

This also streamlines the mapping to Cypher execution process:

- mapping (`JobSpecMapper` et al): remove all validation from there (see next point)
- validation (`InputValidator`): we want to validate as early as possible, but not too early (the mapper should map what it gets even if invalid).  Downstream components only deal with valid job specs.
- optimization (`InputRefactoring`): redundant mappings are ironed out, inactive targets are removed..  Downstream components only deal with valid, streamlined job specs.
- statement generation (`ModelUtils` / `CypherGenerator`): based on the aforementioned improvements, the generation logic is simplified. It also handles some specific cases better (like self-linking nodes that would normally lead to duplicate schema statements)


bvolpato (approved):
Not a lot of context on schemas here, but took a look and no red flags. Thanks Florent!

PR #1066, opened 2023-09-21
Fix the check for non-initialized clusters when using static instances

bvolpato:


Abacn:
I see, let's revert that specific change

Abacn:
The JUnit Test Report no longer show Bigtable related failures

Abacn (approved):

PR #1078, opened 2023-09-28
Add `requirementsFile` parameter to LaunchConfig.

pranavbhandari24:
Add `requirementsFile` parameter to LaunchConfig.

This change should allow users to pass a requirements.txt for launching python jobs.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dataflow/DefaultPipelineLauncher.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dataflow/DefaultPipelineLauncher.java
@@ -360,11 +360,22 @@ public class DefaultPipelineLauncher extends AbstractPipelineLauncher {
             options.executable() != null,
             "Cannot launch a dataflow job "
                 + "without executable specified. Please specify executable and try again!");
+        if (options.requirementsFile() != null) {
+          // install requirements
+          cmd.add(
+              "virtualenv . && source ./bin/activate && pip3 install -r "
+                  + options.requirementsFile());
+          cmd.add("&&");
+        }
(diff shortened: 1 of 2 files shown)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dataflow/DefaultPipelineLauncher.java
```diff
@@ -360,11 +360,22 @@ public LaunchInfo launch(String project, String region, LaunchConfig options) th
             options.executable() != null,
             "Cannot launch a dataflow job "
                 + "without executable specified. Please specify executable and try again!");
+        if (options.requirementsFile() != null) {
+          // install requirements
+          cmd.add(
+              "virtualenv . && source ./bin/activate && pip3 install -r "
+                  + options.requirementsFile());
+          cmd.add("&&");
+        }
```
bvolpato, line 369: This will break on Windows -- I think this should be fine, but we should probably make a note somewhere on the supported/unsupported environment

bvolpato (approved):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dataflow/DefaultPipelineLauncher.java
pranavbhandari24: That's a good point. I'll add support for windows in a separate PR.

bvolpato:
Errors

      "message": "The template parameters are invalid. Details: \nbigtableChangeStreamAppProfile: Parameter didn't match regex '[a-z][a-z0-9\\-_]+[a-z0-9]'",


Were already fixed as a separate PR, you can ignore those

PR #1091, opened 2023-10-04
[MQTT to Pub/Sub] Fix username/password validation logic.

an2x:
This is to fix a bug in the username/password validation logic: it was supposed to throw an exception when the username is provided but the password isn't, but the validation would fail even if both are provided because of the incorrect `options.getPassword() != null` check.

I also simplified the validation logic a little bit (e.g. no need to check for `password.isEmpty` if we're checking for `password.isBlank`).


pranavbhandari24 (dismissed):
LGTM

PR #1105, opened 2023-10-09
Add gitactions scripts

Polber:


.github/scripts/configure-runners.sh
```diff
@@ -0,0 +1,203 @@
...
+{
+   # Display Help
+   echo "Configure github actions runners for running"
+   echo "integration and performance tests."
+   echo
+   echo "Syntax: configure-runners [-p <PROJECT>] [-a <SA_EMAIL>] [-t <GITHUB_TOKEN>] [-n|S|r|o|s|v|V|h]"
+   echo "options:"
+   echo "p     Set the name of the Google Cloud project to use."
+   echo "a     Set the service account email for the Google Cloud project."
+   echo "t     Set the token used to authenticate the runners with the repo."
```
bvolpato, line 40: very nit but usually `[-arg <val>]` is used for optional arguments, and here those 3 are mandatory.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

> Arguments or option-arguments enclosed in the '[' and ']' notation are optional and can be omitted.

bvolpato (approved):

pranavbhandari24 (approved):

Issue #1131, opened 2023-10-24
[Feature Request]: Customize the file name suffix for bulk compress with parameter


quassy:
### Related Template(s)

Bulk Compress Files on Cloud Storage

### What feature(s) are you requesting?

Just like [Cloud PubSub to GCS Text](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/README_Cloud_PubSub_to_GCS_Text.md?plain=1#L29), [Cloud BigQuery to GCS TensorFlow Records](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/README_Cloud_BigQuery_to_GCS_TensorFlow_Records.md?plain=1#L32) or others, the bulk compress files template should optionally support setting the output filename suffix to overwrite the default `.bzip2`, `.deflate`, `.gz`. I would suggest naming the parameter `outputFilenameSuffix` as it's more precise. An empty string should be allowed to make the pipeline not change filenames at all.

> * **outputFilenameSuffix** (Output filename suffix of the files to write): The suffix to place on each windowed file. Typically a file extension such as .txt or .csv. Defaults to empty.
> * **outputSuffix** (The output suffix for TFRecord files): File suffix to append to TFRecord files. Defaults to .tfrecord.

PR #1132, opened 2023-10-25
BulkCompressor: Add optional parameter outputFilenameSuffix to overwrite default

quassy:
For the bulk compress files template this adds a parameter `outputFilenameSuffix` to optionally set the output filename suffix to overwrite the default .bzip2, .deflate, .gz. with your own suffix or an empty string. Addresses #1131

Change is manually tested and confirmed to work with no parameter set (defaults to .bzip2, .deflate, .gz depending on the compression used so it's backwards compatible), parameter set to empty string and parameter set to custom ending like `.foo`.

Code was formatted with `mvn spotless:apply` and checked with `mvn spotless:check`. Autodocs were updated with `mvn clean prepare-package -DskipTests -PtemplatesSpec`.

Co-authored by @matheusfillipe (CLA was signed)

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/templates/BulkCompressor.java
+++ b/v1/src/main/java/com/google/cloud/teleport/templates/BulkCompressor.java
@@ -53,7 +53,7 @@ import org.slf4j.LoggerFactory;
 /**
  * The {@link BulkCompressor} is a batch pipeline that compresses files on matched by an input file
  * pattern and outputs them to a specified file location. This pipeline can be useful when you need
- * to compress large batches of files as part of a perodic archival process. The supported
+ * to compress large batches of files as part of a periodic archival process. The supported
  * compression modes are: <code>BZIP2</code>, <code>DEFLATE</code>, <code>GZIP</code>. Files output
  * to the destination location will follow a naming schema of original filename appended with the
  * compression mode extension. The extensions appended will be one of: <code>.bzip2</code>, <code>
@@ -159,6 +159,18 @@ public class BulkCompressor {
     ValueProvider<Compression> getCompression();
 
     void setCompression(ValueProvider<Compression> value);
+
+    @TemplateParameter.Text(
+        order = 5,
+        optional = true,
+        regexes = {"^[A-Za-z_0-9.]*"},
+        description = "Output filename suffix",
+        helpText =
+            "Output filename suffix of the files to write. Defaults to .bzip2, .deflate or .gz depending on the compression algorithm.")
+    @Required
+    ValueProvider<String> getOutputFilenameSuffix();
+
+    void setOutputFilenameSuffix(ValueProvider<String> value);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bvolpato (approved):
LGTM, thanks!

v1/src/main/java/com/google/cloud/teleport/templates/BulkCompressor.java
```diff
@@ -159,6 +159,18 @@ public interface Options extends PipelineOptions {
...
     void setCompression(ValueProvider<Compression> value);
+
+    @TemplateParameter.Text(
+        order = 5,
+        optional = true,
+        regexes = {"^[A-Za-z_0-9.]*"},
+        description = "Output filename suffix",
+        helpText =
+            "Output filename suffix of the files to write. Defaults to .bzip2, .deflate or .gz depending on the compression algorithm.")
+    @Required
```
bvolpato, line 170: The Required here slightly contradicts `optional = true` from above. They are different constructs but may raise some confusion.

Could we please remove it for clarity before merging?

Thanks!

quassy: This confused me too and I am actually not sure what `@Required` does here or why it seems required by the code, but without it the resulting template did not know the `outputFilenameSuffix` parameter:

> ERROR: (gcloud.dataflow.jobs.run) INVALID_ARGUMENT: (451b3c759e4a9769): The workflow could not be created. Causes: (451b3c759e4a994c): Found unexpected parameters: ['outputFilenameSuffix' (perhaps you meant 'outputFailureFile')]

bvolpato: Ah, I see, my bad.

Can you test replacing it with `@Default.String("")` to see how it goes? 

(Reminder that you would have to replace the conditions on the `@ProcessElement` to also account for empty, perhaps `StringUtils.isNotEmpty(options.getOutputFilenameSuffix())`)

Thanks!

matheusfillipe: Can you test replacing it with @Default.String("") to see how it goes? 

That wouldn't be a good solution since an empty string could be a desirable suffix value if you want to avoid the renaming at all.

quassy:
Thanks for the quick review & merge!

I saw that there was a release just after merging but this does not seem to include pushing the latest changes to `gs://dataflow-templates-europe-west3/latest/Bulk_Compress_GCS_Files` am I right? Because there the parameter is not supported yet:
```
ERROR: (gcloud.dataflow.jobs.run) INVALID_ARGUMENT: The template parameters are invalid.
- '@type': type.googleapis.com/google.dataflow.v1beta3.InvalidTemplateParameters
  parameterViolations:
  - description: Unrecognized parameter
    parameter: outputFilenameSuffix
```

bvolpato:
@quassy Sorry, there's an internal rollout that needs to happen.

I will try remember / ping here when it's out

PR #1133, opened 2023-10-25
Added optional parameter to stora mongodbUri in Secret Manager. If us…

ricardolui:
Adds support to store mongodbUri in Secret Manager. 
If used, the template will ignore the mongodburl and extract the mongoDbUri from the secret path.


ricardolui:
Closing Pull request in favor of using KMS implementation from Jdbc

PR #1134, opened 2023-10-25
Support KMS for mongodburi using maybeDecrypt function

ricardolui:
Support on MongoDB template for Optional KMS parameter on MongoDBuri using maybeDecrypt

### DIFF
--- a/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
+++ b/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
@@ -71,11 +71,10 @@ public class MongoDbToBigQueryOptions {
      @TemplateParameter.KmsEncryptionKey(
       order = 5,
       optional = true,
-      groupName = "Source Parameters",
       description = "Google Cloud KMS key",
       helpText =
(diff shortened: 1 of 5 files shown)


v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
```diff
@@ -67,6 +67,19 @@ public interface MongoDbOptions extends PipelineOptions, DataflowPipelineOptions
...
     void setUserOption(String userOption);
+
+     @TemplateParameter.KmsEncryptionKey(
+      order = 5,
+      optional = true,
+      groupName = "Source Parameters",
+      description = "Google Cloud KMS key",
+      helpText =
+          "Cloud KMS Encryption Key to decrypt the username, password, and connection string. If Cloud KMS key is "
+              + "passed in, the username, password, and connection string must all be passed in encrypted.",
```
bvolpato, line 78: I think here we mean the `connectionUri` only? Since I think everything is sent in that single param

ricardolui: That's right. I've copied the parameter from the Jdbc and forgot to change the description.

bvolpato (approved):
LGTM, thanks!

PR #1146, opened 2023-11-03
test: Adding databoost test for SpannerVectorEmbeddingExport.

gauravpurohit06:


v1/src/test/java/com/google/cloud/teleport/spanner/SpannerVectorEmbeddingExportIT.java
```diff
@@ -78,7 +85,7 @@ public void teardown() {
 
   @Test
   public void testSpannerToGCSJSON() throws IOException {
-    String tableName = testName + "_Documents";
+    String tableName = "testSpannerToGCSJSON_Documents" + RandomStringUtils.randomNumeric(5);
```
bvolpato, line 88: Suggest using `dataBoostEnabled` instead of randomNumeric -- with 5 there's a high chance of clashing

gauravpurohit06: Thanks for the review @bvolpato. Actually we are using parameterised runner, so to give a different `tableName` generating the random integer of length 5. We are actually deleting the database as part of cleanup, so I don't feel clashing would be a problem.  Let me know what do you think ?

bvolpato: We start a lot of tests in parallel so there's chance for collision.

We also use a single instance to avoid deleting/creating instances quotas from being hit (`-DspannerInstanceId=teleport`) this is why I suggest making this more random.

But I agree it's not a blocking issue and we can tune later if errors show up

PR #1174, opened 2023-11-16
Fix custom query flaky test

fbiville:
This introduces two node targets, in order to deduplicate the IDs used in the tests at the source query level (with GROUP BY). Doing so eliminates the concurrent MERGE problem, that fails hard when uniqueness constraints are in place (which is the case for this test's spec).

The other test has also been deleted since it overlaps now much more with what the fixed test covers.


bvolpato (approved):
LGTM, thanks!

PR #1228, opened 2023-12-19
WIP Autogenerate Template specific terraform modules for Dataflow Job creation

damondouglas:
This PR, currently WIP (issue pending), aims to autogenerate terraform modules that use variables specific to the template metadata. It relies on [JSON Configuration Syntax](https://developer.hashicorp.com/terraform/language/syntax/json) instead of the standard [HCL based syntax](https://developer.hashicorp.com/terraform/language/syntax/configuration), typically used to code terraform modules.


damondouglas:
Closing this to split up into two smaller PRs.

PR #1230, opened 2023-12-20
Add pipeline spec parameter to YAML Template

Polber:
Refactors YAML Template to have 2 template parameters instead of one - `--pipeline_spec` and `pipeline_spec_file` which is in-line with the arguments used when executing jobs directly from `apache_beam.yaml.main`.

### DIFF
--- a/python/README_Yaml_Template.md
+++ b/python/README_Yaml_Template.md
@@ -14,10 +14,11 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 
 ### Required Parameters
 
-* **yaml** (Input YAML file in Cloud Storage.): The input YAML file Dataflow reads from.
 
 ### Optional Parameters
 
+* **yaml_pipeline** (Input YAML pipeline spec.): A yaml description of the pipeline to run.
+* **yaml_pipeline_file** (Input YAML pipeline spec file in Cloud Storage.): A file in Cloud Storage containing a yaml description of the pipeline to run.
 
 
 
@@ -39,7 +40,12 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 ### Templates Plugin
 
 This README provides instructions using
-the [Templates Plugin](https://github.com/GoogleCloudPlatform/DataflowTemplates#templates-plugin).
+the [Templates Plugin](https://github.com/GoogleCloudPlatform/DataflowTemplates#templates-plugin)
+. Install the plugin with the following command before proceeding:
+
+```shell
+mvn clean install -pl plugins/templates-maven-plugin -am
+```
 
 ### Building Template
 
@@ -65,7 +71,8 @@ mvn clean package -PtemplatesStage  \
 -DbucketName="$BUCKET_NAME" \
 -DstagePrefix="templates" \
 -DtemplateName="Yaml_Template" \
--f python
+-pl python \
+-am
 ```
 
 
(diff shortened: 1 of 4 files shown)


python/src/main/python/yaml-template/main.py
```diff
@@ -21,26 +21,10 @@
...
-        dest='yaml',
-        help='Input YAML file in Cloud Storage.')
-    known_args, pipeline_args = parser.parse_known_args(argv)
-    pipeline_spec_file = _pipeline_spec_file_from_args(known_args)
+    _, args = argparse.ArgumentParser().parse_known_args(argv)
     cache_provider_artifacts.cache_provider_artifacts()
-
-    main.run(argv=pipeline_args + [f"--pipeline_spec_file={pipeline_spec_file}",
-                                   "--save_main_session"])
+    main.run(argv=args + ["--save_main_session"])
```
robertwb, line 27: Why do we want save main session? (Generally it's best to avoid if we can.)

python/README_Yaml_Template.md
```diff
@@ -14,10 +14,11 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 
 ### Required Parameters
 
-* **yaml** (Input YAML file in Cloud Storage.): The input YAML file Dataflow reads from.
 
 ### Optional Parameters
 
+* **pipeline_spec** (Input YAML spec.): A yaml description of the pipeline to run.
```
robertwb, line 20: We should consider the names "yaml_pipeline" and "yaml_pipeline_file" or similar. (We can make aliases in main.py to keep things backwards compatible there if needed.)

python/README_Yaml_Template.md
```diff
@@ -139,7 +144,7 @@ mvn clean package -PtemplatesRun \
 -Dregion="$REGION" \
 -DjobName="yaml-template-job" \
 -DtemplateName="Yaml_Template" \
--Dparameters="yaml=$YAML" \
+-Dparameters="pipeline_spec=$PIPELINE_SPEC,pipeline_spec_file=$PIPELINE_SPEC_FILE" \
```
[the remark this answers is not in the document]
Polber: Yes, that is a limitation of the auto-generated docs, but this is not what is used for official docs at the moment. I will raise an issue about documenting mutually exclusive args.

python/src/main/java/com/google/cloud/teleport/templates/python/YAMLTemplate.java
```diff
@@ -31,10 +31,19 @@
     flexContainerName = "yaml-template",
     contactInformation = "https://cloud.google.com/support")
 public interface YAMLTemplate {
-  @TemplateParameter.GcsReadFile(
+  @TemplateParameter.Text(
```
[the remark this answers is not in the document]
Polber: This is for the annotation framework that we use to build and stage the templates (both for IT's/LT's and for release). It is also what is used to auto-generate docs.

python/src/main/python/yaml-template/main.py
Polber: I'm honestly not sure, I just noticed the other python templates use it. What is the reason to avoid it? Is there ever a reason to use it?

python/README_Yaml_Template.md
Polber: I agree those names are more friendly than using "spec" - That change will need to occur on main.py before I can change the template though. So if that name change and alias can't get cherry-picked into 2.53, I'll have to update this template after 2.54 is released.

python/src/main/java/com/google/cloud/teleport/templates/python/YAMLTemplate.java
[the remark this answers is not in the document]
bvolpato: For now, ideally we'll have a better way to express metadata for Python templates

python/README_Yaml_Template.md
Polber: Not sure why I thought this^, must have been the holiday brain. I updated the template to use those names and created https://github.com/apache/beam/pull/29916 to add the aliases to main.py which, when merged, will allow this template being to be updated to remove the alias logic.

python/src/main/python/yaml-template/main.py
Polber: Removed.

PR #1237, opened 2023-12-27
Fix pipeline name record in metrics

Abacn:
Sync with https://github.com/apache/beam/pull/29876 

At some point we renamed the pipeline from e.g. test_bigquery_write to write_bigquery, but missed this place.

In the last a few Beam versions, pipeline name override for non-default configuration actually did not work as expected, causing dashboard actually reporting an averaged number of metrics from different configurations (e.g. different number of worker, column size, legacy runner/runner v2)


Abacn:
R: @bvolpato

bvolpato (approved):

Abacn:
This PR has already submitted internally but not gets merged here. What happened to copybara?

PR #1269, opened 2024-01-11
Add support to load CSV files in GCS to BigQuery.

keyliug:
Adding template to loading CSV files stored in GCS and write to GCS.

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
+++ b/v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.templates.common;
+
+import com.google.auto.value.AutoValue;
+import com.google.cloud.teleport.metadata.TemplateParameter;
+import com.google.cloud.teleport.util.DualInputNestedValueProvider;
+import com.google.cloud.teleport.util.DualInputNestedValueProvider.TranslatorInput;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.channels.Channels;
+import java.nio.charset.Charset;
+import org.apache.beam.sdk.io.Compression;
+import org.apache.beam.sdk.io.FileIO;
+import org.apache.beam.sdk.io.FileIO.ReadableFile;
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.DefaultValueFactory;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.options.ValueProvider.NestedValueProvider;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.sdk.values.TupleTagList;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CsvConverters {
+  /* Logger for class. */
+  private static final Logger LOG = LoggerFactory.getLogger(CsvConverters.class);
+
+  /**
+   * Gets Csv format according to <a
+   * href="https://javadoc.io/doc/org.apache.commons/commons-csv">Apache Commons CSV</a>. If user
+   * passed invalid format error is thrown.
+   */
+  public static ValueProvider<CSVFormat> getCsvFormat(
+      ValueProvider<String> formatString, ValueProvider<String> delimiter) {
+    return DualInputNestedValueProvider.of(
+        formatString,
+        delimiter,
+        new SerializableFunction<TranslatorInput<String, String>, CSVFormat>() {
+          @Override
+          public CSVFormat apply(TranslatorInput<String, String> input) {
+            CSVFormat format = CSVFormat.Predefined.valueOf(input.getX()).getFormat();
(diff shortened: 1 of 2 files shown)


v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -0,0 +1,306 @@
...
+public class CsvConverters {
+  /* Logger for class. */
+  private static final Logger LOG = LoggerFactory.getLogger(CsvConverters.class);
+
+  /**
+   * Gets Csv format according to <a
+   * href="https://javadoc.io/doc/org.apache.commons/commons-csv">Apache Commons CSV</a>. If user
+   * passed invalid format error is thrown.
+   */
+  public static CSVFormat getCsvFormat(String formatString, @Nullable String delimiter) {
```
keyliug, line 61: Anyway we can avoid creating the same function twice just for ValueProvider?

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+    return  format;
+  }
+
+  /**
+   * Gets Csv format according to <a
+   * href="https://javadoc.io/doc/org.apache.commons/commons-csv">Apache Commons CSV</a>. If user
+   * passed invalid format error is thrown.
+   */
+  public static ValueProvider<CSVFormat> getCsvFormat(ValueProvider<String> formatString, @Nullable ValueProvider<String> delimiter) {
```
keyliug, line 74: Do we need do anything for this class for ValueProvider support? https://github.com/apache/commons-csv/blob/master/src/main/java/org/apache/commons/csv/CSVFormat.java

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Default value factory to get delimiter from Csv format so that if the user does not pass one
+   * in, it matches the supplied {@link CsvPipelineOptions#getCsvFormat()}.
+   */
+  public static class DelimiterFactory implements DefaultValueFactory<ValueProvider<String>> {
+
+    @Override
+    public ValueProvider<String> create(PipelineOptions options) {
+      return DualInputNestedValueProvider.of(options.as(CsvPipelineOptions.class).getCsvFormat(),
```
keyliug, line 158: do we have a single input value provider to avoid the hack with the second input null here?

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -0,0 +1,306 @@
...
+    public abstract TupleTag<String> headerTag();
+
+    public abstract TupleTag<String> lineTag();
+
+    public abstract ValueProvider<String> fileEncoding();
+
+    @Override
+    public PCollectionTuple expand(PBegin input) {
+
+      if (hasHeaders().get()) {
```
keyliug, line 205: how to perform check since it's a value provider now?

Exception in thread "main" java.lang.IllegalStateException: Value only available at runtime, but accessed from a non-runtime context: RuntimeValueProvider{propertyName=containsHeaders, default=false}
	at org.apache.beam.sdk.options.ValueProvider$RuntimeValueProvider.get(ValueProvider.java:254)
	at com.google.cloud.teleport.templates.common.CsvConverters$ReadCsv.expand(CsvConverters.java:205)
	at com.google.cloud.teleport.templates.common.CsvConverters$ReadCsv.expand(CsvConverters.java:180)
	at org.apache.beam.sdk.Pipeline.applyInternal(Pipeline.java:545)
	at org.apache.beam.sdk.Pipeline.applyTransform(Pipeline.java:496)
	at org.apache.beam.sdk.values.PBegin.apply(PBegin.java:56)
	at org.apache.beam.sdk.Pipeline.apply(Pipeline.java:190)
	at com.google.cloud.teleport.templates.CSVToBigQuery.main(CSVToBigQuery.java:203)

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -0,0 +1,306 @@
...
+
+      public abstract Builder setFileEncoding(ValueProvider<String> fileEncoding);
+
+      abstract ReadCsv autoBuild();
+
+      public ReadCsv build() {
+
+        ReadCsv readCsv = autoBuild();
+
+        checkArgument(readCsv.inputFileSpec() != null, "Input file spec must be provided.");
```
keyliug, line 245: how to perform the check since it's value provider now?

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
bvolpato: You could maybe reuse `CSVFormat getCsvFormat` from your ValueProvider implementation, instead of duplicating code

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
bvolpato: I think this looks fine, did you have any issues?

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
bvolpato: NestedValueProvider.of?

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
bvolpato: This can only be done inside a DoFn, not during pipeline/graph expansion

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
bvolpato: unfortunately, only inside a DoFn (or a flex template)

for classic template the program already ran and the graph is ready, you just inject the value in the placeholders (ValueProviders), so the model is not imperative

v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
keyliug: I see, I tried move the logic to DoFn. However, the issue is that the output tags depends on the condition. When we set the output tags, it's already outside of DoFn. All the examples I find so far is always hard coded the output tags. Do you have any suggestions?

if (condition is true) {
            return input
            .apply("MatchFilePattern", FileIO.match().filepattern(inputFileSpec()))
            .apply("ReadMatches", FileIO.readMatches())
            .apply(
                "ReadCsvWithHeaders",
                ParDo.of(
                        new GetCsvHeadersFn(
                            headerTag(), lineTag(), csvFormat(), delimiter(), fileEncoding()))
                    .withOutputTags(headerTag(), TupleTagList.of(lineTag())));
}
else {
return PCollectionTuple.of(
          lineTag(), input.apply("ReadCsvWithoutHeaders", TextIO.read().from(inputFileSpec())));
}

bvolpato:
Feel free to tag @ me when you want me to take a new look, thanks!

bvolpato (approved):
This is looking good, thanks! Let's create an IT to make sure it's doing what we expect?

keyliug:
> This is looking good, thanks! Let's create an IT to make sure it's doing what we expect?

Thanks a lot for reviewing the changes and helping solving all issues. I will add the integration test.

PR #1284, opened 2024-01-22
Remove Cypher DSL for now

fbiville:
Reinstate it once we actually use it.


bvolpato (approved):
Makes sense, thanks! LGTM

PR #1301, opened 2024-01-31
Templates Maven Plugin: Flex template staging improvements. 

an2x:
Minor adjustments to how the Flex template spec file is generated (removed unused fields).


andreigurau (approved):
LGTM

PR #1255, opened 2024-01-08
Support create disposition and assign BQ schema for JdbcToBQ template

Abacn:
* Also remove a TODO to support fetchSize argument

PR #1303, opened 2024-01-31
Expose fetchSize option to DataplexJdbcIngestion template

Abacn:
We have seen OOM in this template. Similar to JdbcToBigQuery template (#1255), expose this option here

### DIFF
--- a/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexJdbcIngestionOptions.java
+++ b/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexJdbcIngestionOptions.java
@@ -250,4 +250,15 @@ public interface DataplexJdbcIngestionOptions
   Boolean getUseColumnAlias();
 
   void setUseColumnAlias(Boolean useColumnAlias);
+
+  @TemplateParameter.Integer(
+      order = 16,
+      optional = true,
+      description = "Set the data size going to be fetched and loaded in memory per Jdbc call.",
+      helpText =
+          "It should ONLY be used if the default value throws memory errors. If not set, using Beam's default "
+              + "fetch size.")
+  Integer getFetchSize();
+
+  void setFetchSize(Integer fetchSize);
 }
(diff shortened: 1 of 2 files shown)


Abacn:
R: @ahmedabu98

Abacn:
internal tracker: 323143852

v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexJdbcIngestionOptions.java
```diff
@@ -250,4 +250,15 @@ public interface DataplexJdbcIngestionOptions
...
   void setUseColumnAlias(Boolean useColumnAlias);
+
+  @TemplateParameter.Integer(
+      order = 16,
+      optional = true,
+      description = "Set the data size going to be fetched and loaded in memory per Jdbc call.",
+      helpText =
+          "It should ONLY be used if the default value throws memory errors. If not set, using Beam's default "
+              + "fetch size.")
+  Integer getFetchSize();
```
ahmedabu98, line 261: Should we set a default here? I see from the JdbcIO connector we have a default of 50,000: https://github.com/apache/beam/blob/81513f10d60f65c0873fc2484fb27a8e38fabc56/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L373

ahmedabu98 (approved):
One suggestion but LGTM

v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexJdbcIngestionOptions.java
Abacn: I intended to use JdbcIO's default the same as here: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/1f37b59e49971e0b0c8c9864fbcee407e60aa588/v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java#L157

PR #1309, opened 2024-02-01
fix maven -am -amd issue on v2/common

Polber:


Abacn:
Or make `spanner-migrations-sdk` depends on `v2/common` instead, which makes more sense

Abacn (approved):

PR #1321, opened 2024-02-08
Enfore csvFormat and delimiter as required parameters for GCS_CSV_to_BigQuery template

keyliug:
Changing csvFormat to be required so that it's clear to user which format they're using.

Change delimiter to be required due to bug https://github.com/apache/beam/issues/30261 that causes we won't be able to get the value of delimiter dynamically. This change is fine since delimiter is a common paramter in CSV and lots of users set it anyway.


Abacn:
lgtm as this is a new template / not-launched so won't break users.

PR #1335, opened 2024-02-15
Modify TemplateTestBase to launch multiple templates

Deep1998:
Modified the TemplateTestBase class to support launching multiple templates while maintaining backward compatibility for single template ITs.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
@@ -100,7 +100,11 @@ public abstract class TemplateTestBase {
   protected static final String PROJECT = TestProperties.project();
   protected static final String REGION = TestProperties.region();
 
+  // Store the template spec path for staged template specified via @TemplateIntegrationTest.
   protected String specPath;
+  // Store the template spec paths for the staged templates specified via
+  // @MultiTemplateIntegrationTest.
+  protected List<String> multiTemplateSpecPaths;
   protected Credentials credentials;
   protected CredentialsProvider credentialsProvider;
   protected String artifactBucketName;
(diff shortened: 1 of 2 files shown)


Abacn (commented):
Thanks, this lgtm, just had a minor comment

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
```diff
@@ -180,72 +188,115 @@ public void setUpBase() throws ExecutionException {
...
+        template = getTemplateAnnotation(templateIntegrationTest, templateClass);
+        if (template == null) {
+          return;
+        }
+        if (template.placeholderClass() != void.class) {
+          templateClass = template.placeholderClass();
+        }
+        multiTemplates.add(template);
+        multiTemplateClasses.add(templateClass);
+        pipelineLauncher = buildLauncher(templateClass, template);
```
aksharauke, line 233: Do we need separate pipeline launchers per template when there are multiple templates?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
```diff
@@ -283,7 +334,8 @@ private Template getTemplateAnnotation(
    * It identifies whether the template is v1 (Classic) or v2 (Flex) to setup the Maven reactor
    * accordingly.
    */
-  private String[] buildMavenStageCommand(String prefix, File pom, String bucketName) {
+  private String[] buildMavenStageCommand(
+      String prefix, File pom, String bucketName, Template template) {
```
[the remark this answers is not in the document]
Deep1998: Correct, added field comments to clarify. Also changed the method parameters.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
Deep1998: Added null check to do this only once. This earlier would be a redundant call re-initialising to the same value each time.

aksharauke: In the launchTemplate method, when setupShutdownHook is enabled - a separate thread is created to which the pipelineLauncher object is passed. Will there be multi-threading issues if there are multiple templates and the same pipelineLauncher object is being used across multiple threads to cancel the job?

Deep1998: I went through the beam code it looks like we should be safe. I could not find documentation explicitly saying it thread-safe though nor have i tested it.

aksharauke: Cool, thank you.

aksharauke (approved):

PR #1338, opened 2024-02-16
Allow spaces in JDBC URL and PropertyName

SSolanich:
Databases such as IBM AS400 have property strings that require spaces by implementation.

Example: jdbc:as400://123.123.123.123;block size=512;extended dynamic=true;lazy close=true;query optimization goal=2;data compression=true;

Currently using such strings would lead to a regex match error.

This pull request fixes such error by allowing spaces in the regular expresion match in both JDBC URL and propertyName string values.


SSolanich:
Any thoughts on this change?

PR #1339, opened 2024-02-19
Add integration tests for PK update

Deep1998:
This PR adds an IT that insert a few rows and then updates the PK of one of the rows.


darshan-sj (commented):
You will need to update the session file as well.

PR #1344, opened 2024-02-26
Support slash in input topic regex

shunping:
This is a feature request from a customer (internal bug id: 326560432). Basically, the customer wanted to use "/" in their input topic name while using MTQQ template, but got an error.

The change is similar to https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/ae0f8201901ba9fb44833f600ba6c8b1a65b3ac7.


damondouglas (changes requested):
Would you be able to comment in the PR description, if possible, what test(s) were previously broken that this PR fixes?

shunping:
> Would you be able to comment in the PR description, if possible, what test(s) were previously broken that this PR fixes?

Added a little more background and internal bug id in the description.

damondouglas (approved):
Thank you, @shunping

PR #1346, opened 2024-02-28
Fix parallelism: add missing groupby, move batching out of Neo4j writer

fbiville:


ali-ince (approved):

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jBlockingUnwindFn.java
```diff
@@ -85,29 +76,11 @@ public void setup() {
 
   @ProcessElement
   public void processElement(ProcessContext context) {
+    KV<Integer, Iterable<Row>> rowBatch = context.element();
+    LOG.info("Processing row batch from key: {}", rowBatch.getKey());
```
liferoad, line 80: just a note: this might create too many logs.

liferoad (approved):

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jBlockingUnwindFn.java
fbiville: Shall I downgrade this to the `debug` level?
The default batching values are batches of 5000 records for node writes, and 1000 records for relationship and custom query writes.

liferoad: I would recommend debug, similarly, for other places. But eventually this is up to you to decide whether the logs are too verbose for users.

fbiville: Done in e316f23349fc4148cf21487071270fc6160c6417

liferoad (approved):

PR #1350, opened 2024-03-04
Making datasource provider singleton

VardhanThigle:
Creating a singleton Connection Pool for connecting to source database. This allows the user to configure the max number of source database connections per worker.


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
```diff
@@ -0,0 +1,103 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DataSourceProvider
+    implements org.apache.beam.sdk.transforms.SerializableFunction<Void, javax.sql.DataSource> {
+
+  private static final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+  private static final Lock writeLock = readWriteLock.writeLock();
+  private static final Lock readLock = readWriteLock.readLock();
+  private static final Logger LOG = LoggerFactory.getLogger(DataSourceProvider.class);
```
bharadwaj-aditya, line 35: nit: move the logger to the top of the class.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
```diff
@@ -0,0 +1,103 @@
...
+      ret = dataSource;
+    } finally {
+      writeLock.unlock();
+    }
+    return ret;
+  }
+
+  private javax.sql.DataSource getDataSource() {
+    javax.sql.DataSource ret = null;
+    readLock.lock();
```
bharadwaj-aditya, line 90: why do we need a read lock here ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private javax.sql.DataSource initializeDataSource() {
+    javax.sql.DataSource ret;
+    writeLock.lock();
+    try {
+      // Check if it was already initialized till the write lock was acquired.
+      if (dataSource == null) {
+        dataSource = JdbcIO.PoolableDataSourceProvider.of(config).apply(null);
+        LOG.debug("initializeDataSource dataSource" + dataSource.toString());
```
bharadwaj-aditya, line 79: Please use parameterized loggers where possible.

typo: initialize datasource

bharadwaj-aditya (changes requested):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
VardhanThigle: The read-write locking was used to achivee both - synchronization and order the double checked locking loads/stores. Thanks for pointing me to double-checked locking in the review. 
I have implemented the recommended (or idiomatic for java9+) double checked lock in this push.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
VardhanThigle: Done. Thanks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    ret = getDataSource();
+    if (ret == null) {
+      ret = initializeDataSource();
+    }
+    return ret;
+  }
+
+  private javax.sql.DataSource initializeDataSource() {
+    javax.sql.DataSource ret;
+    writeLock.lock();
```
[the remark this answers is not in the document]
VardhanThigle: Thanks. That's right it's a double checked lock, the object is constructed and apply is called during pardo operation, so the double checked lock will be needed either in constructor or apply operation.

Have implemented the  idomatic doublechecked lock in this push as we had discussed offline.

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/DataSourceProvider.java
```diff
@@ -0,0 +1,145 @@
...
+  static {
+    try {
+      MethodHandles.Lookup lookup = MethodHandles.lookup();
+      DATA_SOURCE =
+          lookup
+              .in(DataSourceProvider.class)
+              .findStaticVarHandle(DataSourceProvider.class, "dataSource", DataSource.class);
+    } catch (ReflectiveOperationException e) {
+      // Logger might not be initialized in static scope!
+      System.err.println("Error while binding VarHandle: " + e.toString());
```
liferoad, line 49: Besides showing the error, is there any actionable item the users could try to resolve this?

Abacn: I would understand the fact that it may emit error as the initialization timing is not quite accurate. It is a static member but its accesses, besides initialization are all in non-static context.

Is this a connection pool pattern (all DataSourceProvider instances share the same DATA_SOURCE)? If so, this static member should be acceesses when first DataSourceProvider constructor gets initialized or before first access, instead in static context

Abacn: I do not see the need of use of reflection here. why not just use dataSource?

VardhanThigle: We use `DATA_SOURCE` varHandle to access `dataSource`  through relaxed atomic access. Ref: [Double-checked_locking for Java9](https://en.wikipedia.org/wiki/Double-checked_locking#:~:text=Java%209%20introduced%20the%20VarHandle,synchronization%20order%2C%20the%20global%20order)

The error to bind varHandle should not occur unless the code is modified and any of the type casting fails. Since the error occurs at static initialization the code developer should be able to catch it as a part of test flow and fix it.

Abacn: lgtm as it's a well practiced pattern, good to learn, sorry for confusion

Abacn:
Test failures
```
TextImportPipelineIT.testImportCsv: TextImportPipelineIT#L93
Failed to execute statement.
TextImportPipelineIT.testImportCsvBadRows: TextImportPipelineIT#L184
Failed to create database.
TextImportPipelineIT.testImportCsvBadRows: TextImportPipelineIT#L70
Error cleaning up resources
```
probably just flaky tests. Kick of test again

liferoad (approved):

PR #1355, opened 2024-03-06
[Neo4j] Allow user to specify only a sub-set of columns.

meistermeier:
This will actually truncate the parsed rows to align with the numbers with defined columns/fields.


fbiville (approved):

Abacn (commented):
Thanks for the suggested change. What use case would one expect a truncated row? I am wondering if user feed with a row that is missing field mistakenly, it would silently get passed and drop data

meistermeier:
That's true. On the other hand, and this is the intention of this PR, users shouldn't be forced to import every field. Right now there is only a list of string but in the future, there could be e.g. some kind of definition that would allow to also define the index of the column. In this way, an import could be even more tailored to the users needs.

fbiville:
This PR is basically a first step into allowing more flexibility with text sources.
Before this PR, a text source with 100 columns per row requires users to specify 100 column names in the header, even though they might only be interested in mapping the first 5 columns.
Users have full control over the spec, maybe another part of their company manages text exports, so giving a little flexibility in the spec may be a good thing for UX.

As @meistermeier pointed out, we are planning to introduce an alternative header definition where an arbitrary subset of columns can be defined, but that will come a bit later.

PR #1357, opened 2024-03-08
Improvement in replication lag by reducing wait time

aksharauke:
The timer cycle is reduced to 1 millisec from 1 second to improve the latency


aksharauke:
@Dippatel98 please help with review, thank you!

PR #1365, opened 2024-03-14
Add BigtableChangeStreamsToVectorSearch template

meagar:
I've been plugging away at this template for quite a while now and I'm hoping to get some feedback on what I've produce before I go further down the road of trying to get integration tests written.

I have a lot of extra debug logging and a lot of TODO comments that I'll have to remove, I'm looking more generally on feedback on how I've set up the pipeline and its options. I've also called out a few places where I'm uncertain on whether I'm doing the right thing (ie, should exceptions be allowed to bubble up, or be captured).


v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/BigtableChangeStreamsToVectorSearch.java
```diff
@@ -0,0 +1,235 @@
...
+    LOG.debug("  - bigtableReadInstanceId {}", options.getBigtableReadInstanceId());
+    LOG.debug("  - bigtableReadTableId {}", options.getBigtableReadTableId());
+    LOG.debug("  - bigtableChangeStreamAppProfile {}", options.getBigtableChangeStreamAppProfile());
+    LOG.debug("  - embeddingColumn {}", options.getEmbeddingColumn());
+    LOG.debug("  - crowdingTagColumn {}", options.getCrowdingTagColumn());
+    LOG.debug("  - project {}", options.getProject());
+
+    String indexName = options.getVectorSearchIndex();
+    // TODO(meagar): Raises, do we handle this, or allow exceptions to escape?
+    String vertexRegion = Utils.extractRegionFromIndexName(indexName);
```
meagar, line 127: Should I allow exceptions to bubble out of this method, or capture them and emit some kind of error?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/BigtableChangeStreamsToVectorSearch.java
```diff
@@ -0,0 +1,235 @@
...
+    if (timestamp == null) {
+      return null;
+    } else {
+      return Instant.ofEpochMilli(timestamp.getSeconds() * 1000 + timestamp.getNanos() / 1000000);
+    }
+  }
+
+  private static int bufferSizeOption(int size) {
+    if (size < 1) {
+      // TODO(meagar): Maybe throw instead?
```
meagar, line 221: Not sure if it's idiomatically better to supply a correct value, or if I should throw something like an IllegalArgumentException

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/ChangeStreamMutationToDatapointOperationFn.java
```diff
@@ -0,0 +1,193 @@
...
+      var qualifier = m.getQualifier().toStringUtf8();
+      var col = family + ":" + qualifier;
+
+      String mappedColumn;
+
+      if (col.equals(embeddingsColumn)) {
+        var floats = Utils.bytesToFloats(m.getValue(), embeddingsByteSize == 8);
+
+        // TODO(meagar):Remove this
+        while (floats.size() != 768) {
```
meagar, line 113: The dimensions of the index I was testing against; will remove this

v2/googlecloud-to-googlecloud/src/main/resources/bigtable-changestreams-to-vector-embeddings-command-spec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "mainClass": "com.google.cloud.teleport.v2.templates.BigtableChangeStreamsToVectorSearch",
```
meagar, line 2: I have no idea what the purpose of this file is, I copied it directly from `bgitable-changestreams-to-gcs-command-spec.json` and did `s/gcs/vector-search`.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/BigtableChangeStreamsToVectorSearch.java
```diff
@@ -0,0 +1,235 @@
...
+            // .withShardedKey() // TODO(meagar): What does this do?
+            )
+        .apply("Values", Values.create())
+        .apply(
+            "Upsert Datapoints to VectorSearch",
+            ParDo.of(new UpsertDatapointsFn(vertexEndpoint, indexName)));
+
+    results
+        .get(ChangeStreamMutationToDatapointOperationFn.RemoveDatapointTag)
+        .apply(WithKeys.of("foo")) // TOOD(meagar): Better name for place-holders?
```
meagar, line 187: It wasn't clear to me why we need to introduce keys for GroupIntoBatches to work, and what value I should use for the key since it doesn't actually matter. Can I do something like `WithKeys.of(0)`?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/LogChangeStreamFn.java
```diff
@@ -0,0 +1,75 @@
+/*
```
meagar, line 1: TODO: Remove this whole file

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/UpsertDatapointsFn.java
```diff
@@ -0,0 +1,49 @@
...
+  @ProcessElement
+  public void processElement(@Element Iterable<IndexDatapoint> datapoints) {
+    LOG.debug("Upserting datapoints: {}", datapoints);
+    UpsertDatapointsRequest request =
+        UpsertDatapointsRequest.newBuilder()
+            .addAllDatapoints(datapoints)
+            .setIndex(indexPath)
+            .build();
+
+    // TODO(meagar): This may raise exceptions, which I believe kills the pipeline
```
meagar, line 44: Curious about this, when a pipeline step may raise exceptions, should I allow that? Does Dataflow recover from this gracefully?

v2/googlecloud-to-googlecloud/src/main/resources/bigtable-changestreams-to-vector-embeddings-command-spec.json
jloures: I believe this is put in place to enable this template to show up in the dropdown within the google cloud console.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/BigtableChangeStreamsToVectorSearch.java
jloures: I think this is fine. Generally speaking we want to halt execution of the pipeline the moment we encounter something like this (i.e. invalid index name)

jloures:
@meagar please also fix errors from the PR checks such as: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/8740490105/job/24033560748?pr=1365

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/VectorSearchResourceManager.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    LOG.debug("DeployedIndex {} does not exist, creating", TEST_DEPLOYED_INDEX_NAME);
+    return deployIndexToEndpoint(testIndex.getName(), testEndpoint.getName());
+  }
+
+  private Index createIndex(String indexName) throws Exception {
+    // This is somewhat of a black box, copied from an index in a good state.
+    // The resulting index will have a dimensionality of 10
+    final CharSequence indexSchema =
```
meagar, line 313: This is a little weird; As far as I could tell, most of the properties of an index are specified as part of this big metadata schema document, rather than just being exposed as properties you can set on the builder itself, not really sure why this is the case. The values I'm using here are taken from an index I created and then queried, along with the schema URI property below.

jloures (changes requested):
I did a first pass review on this PR. 
Some things require a bit more context about the downstream service, hence my questions here.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/BigtableChangeStreamsToVectorSearchOptions.java
```diff
@@ -0,0 +1,214 @@
...
+
+  @SuppressWarnings("unused")
+  void setDeleteMaxBufferDuration(String maxBufferDuration);
+
+  @TemplateParameter.Text(
+      order = 14,
+      optional = false,
+      description = "Vector Search Index Path",
+      helpText =
+          "The Vector Search Index where changes will be streamed, in the format 'projects/{projectID}/locations/{region}/indexes/{indexID}' (no leading or trailing spaces)",
```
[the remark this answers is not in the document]
meagar: Yes; do we typically call that a "project number" instead of a "project ID"?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/BigtableChangeStreamsToVectorSearchOptions.java
```diff
@@ -0,0 +1,214 @@
...
+  @TemplateParameter.Integer(
+      order = 10,
+      optional = true,
+      description = "Maximum batch size for upserts for Vector Search",
+      helpText =
+          "The maximum number of upserts to buffer before upserting the batch to the Vector Search Index. "
+              + "Batches will be sent when there are either upsertBatchSize records ready, or any record has been "
+              + "waiting upsertBatchDelay time has passed.",
+      example = "10")
+  int getUpsertMaxBatchSize();
```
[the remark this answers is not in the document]
meagar: I'm not sure what a useful default would be. When not specified, the default is to have *no* max batch size, and just rely on the maximum duration to trigger batch submits.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/ChangeStreamMutationToDatapointOperationFn.java
```diff
@@ -0,0 +1,225 @@
...
+
+      String mappedColumn;
+
+      if (col.equals(embeddingsColumn)) {
+        var floats = Utils.bytesToFloats(m.getValue(), embeddingsByteSize == 8);
+
+        datapointBuilder.addAllFeatureVector(floats);
+      } else if (col.equals(crowdingTagColumn)) {
+        LOG.info("Setting crowding tag {}", m.getValue().toStringUtf8());
+        datapointBuilder.getCrowdingTagBuilder().setCrowdingAttribute("abc").build();
```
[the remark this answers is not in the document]
meagar: Ah, sorry, this leaked through from trying to make integration tests work, "abc" was just a dummy value I was trying to use to verify the tag was syncing.

Ultimately I found that, when I write a crowding tag through to vector search and then read it back, I'm not supposed to get the original tag back, instead I get a hash. At first I thought this was a bug: http://b/335409898

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstovectorsearch/Utils.java
```diff
@@ -0,0 +1,85 @@
...
+  public static Map<String, String> parseColumnMapping(String mapstr) {
+    Map<String, String> columnsWithAliases = new HashMap<>();
+    if (StringUtils.isBlank(mapstr)) {
+      return columnsWithAliases;
+    }
+    String[] columnsList = mapstr.split(",");
+
+    for (String columnsWithAlias : columnsList) {
+      String[] columnWithAlias = columnsWithAlias.split("->");
+      if (columnWithAlias.length == 2) {
```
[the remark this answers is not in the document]
meagar: I don't think so, if you give it an input like `"foo->"`, then splitting on `->` actually yields a 1 element array, not a two element array with a second empty element.

However it looks like `->foo` actually does produce a leading empty element, which is an error I can detect and alert on.

jloures (approved):

PR #1361, opened 2024-03-12
[Do not merge] Beam 2.55.0RC1 validation

Abacn:

PR #1366, opened 2024-03-14
Test Beam 2.55.1

Abacn:


Abacn:
Still failing for the same reason as #1361, even though rc2 fixed org.hamcrest dependency. Found the cause is due to maven logic: https://stackoverflow.com/questions/75333105/maven-test-dependency-removes-transitive-compile-dependency-from-uberjar

So still need to move hamcrest dependency to from test to compile

PR #1368, opened 2024-03-15
Add session file support to bulk template

Deep1998:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SourceDbToSpanner.java
```diff
@@ -103,12 +116,33 @@ static PipelineResult run(SourceDbToSpannerOptions options) {
...
+    SpannerConfig spannerConfig =
+        SpannerConfig.create()
+            .withProjectId(ValueProvider.StaticValueProvider.of(options.getProjectId()))
+            .withHost(ValueProvider.StaticValueProvider.of(options.getSpannerHost()))
+            .withInstanceId(ValueProvider.StaticValueProvider.of(options.getInstanceId()))
+            .withDatabaseId(ValueProvider.StaticValueProvider.of(options.getDatabaseId()));
+
+    PCollection<Ddl> ddl =
+        pipeline.apply("Process Information Schema", new ProcessInformationSchema(spannerConfig));
+    PCollectionView<Ddl> ddlView = ddl.apply("Cloud Spanner DDL as view", View.asSingleton());
```
aksharauke, line 131: During some investigation for forward migration, we figured that this Ddl is not needed as side-input. It can just be made a member of the DoFn in which it will be used. It is serializable so, should be okay to add as member.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/transformer/ChangeEventTypeConvertor.java
```diff
@@ -0,0 +1,258 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.math.NumberUtils;
+
+/**
+ * Utility class with methods which converts text fields in change events represented by JSONObject
+ * to Cloud Spanner types.
+ */
+public class ChangeEventTypeConvertor {
```
aksharauke, line 37: These classes are giving me major deja-vu 😄 Maybe move to spanner-common? Same for the classes that are common with forward migration template.

PR #1369, opened 2024-03-15
Escape fields in generated sql

ali-ince:
This PR changes the logic used in generation of SQL statements to be executed towards BigQuery or Beam, and escapes the field names to protect against reserved keywords being specified by the user.

It also switches to [JSQLParser](https://github.com/JSQLParser/JSqlParser/wiki) for building dynamic queries.


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
```diff
@@ -88,7 +88,7 @@ public void resetDatabase() {
     LOG.info("Resetting database");
     try {
       String database = !StringUtils.isEmpty(this.database) ? this.database : "neo4j";
-      String cypher = "CREATE OR REPLACE DATABASE $db";
+      String cypher = "CREATE OR REPLACE DATABASE $db WAIT 60 SECONDS";
```
fbiville, line 91: note: that's not available in 4.4 so we'll have to change the query in that case (in the PR that actually implement proper Neo4j 4.4 EE)

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/utils/ModelUtilsTest.java
```diff
@@ -0,0 +1,576 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.beam.sdk.schemas.Schema;
+import org.junit.Test;
+
+public class ModelUtilsTest {
+
+  @Test
+  public void shouldGenerateCorrectSqlStatementForNodes() {
+    // No Keys
```
fbiville, line 42: nit: I would split these into their own tests

fbiville (approved):

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
ali-ince: The reason I added this is because some tests were failing because of the database not being ready in time. Let's revisit this while doing the version specific query generation work.

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/utils/ModelUtilsTest.java
ali-ince: done.

ali-ince:
@liferoad this PR is now ready for your review.

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
fbiville: Turns out, `WAIT` is supported in Neo4j 4.4 and I missed it.

ali-ince:
@liferoad would you please re-trigger, the failing test is now fixed

PR #1374, opened 2024-03-19
Adding Codecov tool integration

darshan-sj:
Adding Codecov tool integration

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,9 @@
+comment:
+  layout: "header, diff, components, files" # show component info in the PR comment
+
+coverage:
+  status:
+    patch:
+      default:
+        target: 80%
+        informational: true 
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -6,4 +6,49 @@ coverage:
     patch:
       default:
         target: 80%
-        informational: true 
+        informational: true
+
+component_management:
+  individual_components:
+  - component_id: spanner-templates
+    name: spanner-templates
+    paths:
+    - "v1/src/main/java/com/google/cloud/teleport/spanner/**"
+    - "v1/src/main/java/com/google/cloud/teleport/templates/SpannerToText.java"
+    - "v1/src/main/java/com/google/cloud/teleport/templates/common/SpannerConverters.java"
+    - "v2/datastream-to-spanner/**"
+    - "v2/spanner-common/**"
+    - "v2/spanner-change-streams-to-sharded-file-sink/**"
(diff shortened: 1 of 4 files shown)


.github/codecov.yml
```diff
@@ -0,0 +1,9 @@
+comment:
+  layout: "header, diff, components, files" # show component info in the PR comment
+
+coverage:
+  status:
+    patch:
+      default:
+        target: 80%
+        informational: true
```
manitgupta, line 9: This will enforce patch coverage for the whole repo.

Please create components as done in - https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/master/.github/codecov.yml#L14

and only add patch coverage for the Spanner relevant modules (similar to CODEOWNERS)

Other teams can add their components as they see fit. 

Also add a minimum threshold for the overall coverage of Spanner component.

pom.xml
```diff
@@ -333,6 +334,39 @@
...
+          </execution>
+          <execution>
+            <id>report-aggregate</id>
+            <phase>verify</phase>
+            <goals>
+              <goal>report-aggregate</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
manitgupta, line 369: There are typically files we do not want to include in the coverage such as ITs etc, can you exclude those from JaCoCo?

manitgupta (changes requested):

pom.xml
darshan-sj: JaCoco already takes care of excluding Unit tests, Integration tests etc. Only the source files are taken. I could only think of AutoValue classes which I have excluded in Line 344. Let me know if there are any other type of files to be excluded. -

.github/codecov.yml
```diff
@@ -0,0 +1,56 @@
...
```
manitgupta, line 9: Will this apply patch coverage to the entire project?
Please verify once. Asking because I see target of 80% at each component level as well.

darshan-sj: informational: true makes it non blocking. It just shows the coverage % for the whole project and doesn't block PRs. Should I remove it?

.github/codecov.yml
darshan-sj: Done.

.github/codecov.yml
manitgupta: Ahh, ofcourse. I think that's okay!

manitgupta (dismissed):

damccorm (dismissed):

manitgupta (dismissed):

PR #1376, opened 2024-03-19
Support FLOAT32 type in v1 templates of Spanner

arawind:
### Avro:

GoogleSQL: 
* [Export](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-21_05_01_22-9618753751624897563;graphView=0?project=span-cloud-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))
* [Import](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-21_05_38_00-8680174928377892705;graphView=0?project=span-cloud-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))

PG: 
* [Export](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-21_05_47_07-2771163877050158940;graphView=0?project=span-cloud-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))
* [Import](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-21_05_58_11-1643174307354839989;graphView=0?project=span-cloud-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))

### CSV:

GoogleSQL: 
* [Export](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-22_00_17_56-5509502875882896332?e=13802955&mods=monitoring_api_prod&project=span-cloud-testing)
* [Import](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-22_02_50_39-14326646282971458157?e=13802955&mods=monitoring_api_prod&project=span-cloud-testing)

PG: 
* [Export](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-22_03_17_13-5127588554893435640;graphView=0?project=span-cloud-testing&e=13802955&mods=monitoring_api_prod&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))
* [Import](https://console.cloud.google.com/dataflow/jobs/us-central1/2024-03-22_03_09_55-9477941764979866789?e=13802955&mods=monitoring_api_prod&project=span-cloud-testing)


manitgupta:
This is an XL sized PR...is it possible to break this up? 

Maybe Avro template changes in one and text template changes in another?

manitgupta:
I see some failures in  `TextImportIT`, can you please check?
I will also review the PR in the meantime

pom.xml
```diff
@@ -181,6 +184,18 @@
...
+      <!-- Remove after Beam gets the latest gax version -->
+      <dependency>
+        <groupId>com.google.api</groupId>
+        <artifactId>gax-grpc</artifactId>
+        <version>${gax.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.api</groupId>
+        <artifactId>gax</artifactId>
+        <version>${gax.version}</version>
```
manitgupta, line 197: Can you file an issue for this with details on why this was required?
Please also link the issue here with a TODO in a comment.

CC: @damccorm

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
```diff
@@ -374,6 +384,45 @@ static Optional<List<String>> readStringArray(
...
+      case INT:
+        // Can cause loss of precision: un-representable integers will be rounded to the nearest
+        // float.
+        {
+          List<Integer> value = (List<Integer>) fieldValue;
+          return Optional.of(
+              value.stream()
+                  .map(x -> x == null ? null : Float.valueOf(x))
+                  .collect(Collectors.toList()));
+        }
```
manitgupta, line 409: When will this case occur?  When a customer wants to read a Avro integer into a Float32 in Spanner? 

When you say `un-representable integers` do you mean bigint type of numbers which lie outside of Float32's range?

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
```diff
@@ -374,6 +384,45 @@ static Optional<List<String>> readStringArray(
...
+      case STRING:
+        {
+          List<Utf8> value = (List<Utf8>) record.get(fieldName);
+          return Optional.of(
+              value.stream()
+                  .map(x -> x == null ? null : Float.valueOf(x.toString()))
```
manitgupta, line 415: What happens when the cast is not possible?

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
```diff
@@ -579,6 +628,27 @@ private static Optional<Boolean> readBool(
...
+        return Optional.ofNullable((Integer) record.get(fieldName)).map(x -> (float) x);
+      case FLOAT:
+        return Optional.ofNullable((Float) record.get(fieldName));
+      case STRING:
+        return Optional.ofNullable((Utf8) record.get(fieldName))
+            .map(Utf8::toString)
+            .map(Float::valueOf);
+        // Avoid decoding 64 bit values into 32 bit space as this will cause a precision loss.
+      case LONG:
+      case DOUBLE:
```
manitgupta, line 646: What is the template behaviour if the user is trying to perform this conversion?

arawind:
Thanks!

> I see some failures in `TextImportIT`, can you please check? I will also review the PR in the meantime

This failure is expected today as the type isn't released yet in Spanner. We're expecting it to land soon (like soon soon); I can let you know the exact details privately.

v1/src/test/java/com/google/cloud/teleport/spanner/AvroRecordConverterTest.java
```diff
@@ -258,6 +258,103 @@ public void testParseFloat() {
     assertThrows(IllegalArgumentException.class, () -> avroRecordConverter.apply(avroRecord1));
   }
 
+  @Test
+  public void float32Array() {
```
manitgupta, line 262: Nit: Please name the method according to what it is testing

v1/src/test/java/com/google/cloud/teleport/spanner/AvroRecordConverterTest.java
```diff
@@ -258,6 +258,103 @@ public void testParseFloat() {
...
+
+    // Convert from int to Float32.
+    avroRecord = new GenericRecordBuilder(schema).set("id", 0).set(colName, intArray).build();
+    result = AvroRecordConverter.readFloat32Array(avroRecord, INT, colName);
+    assertArrayEquals(floatArray.toArray(), result.get().toArray());
+
+    // Convert from String to Float32.
+    avroRecord = new GenericRecordBuilder(schema).set("id", 0).set(colName, stringArray).build();
+    result = AvroRecordConverter.readFloat32Array(avroRecord, STRING, colName);
+    assertArrayEquals(floatArray.toArray(), result.get().toArray());
```
manitgupta, line 284: Can we also cover what happens if string cast is not possible?

manitgupta (commented):
I don't see any changes `ImportPipeline.java`?
That is the Avro record to Spanner pipeline, is no change required in that?
Similarly don't see changes in the export templates as well - `ExportPipeline.java`

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: It throws an error, and the behavior would be similar to what happens in FLOAT64. The parsing will fail when the string is not a numeric.

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: > When will this case occur? When a customer wants to read a Avro integer into a Float32 in Spanner?

Yes, that's correct.

>When you say un-representable integers do you mean bigint type of numbers which lie outside of Float32's range?

No, I'm referring to the int32 type here (IIUC avro integer is an int32)

Since not all int32 values can be represented as a float32 ([wikipedia has some details](https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Precision_limitations_on_integer_values)), rounding is inevitable during the conversion from int32 to float32.

Similar is the case when int64 values are attempted to be parsed as a float64 value -- some int64 values will be rounded off. Ideally that conversion should also have a warning.

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: I'm not sure what the template would do -- I think the exception would be caught and the template would fail. 

The alternative option would be to not fail these cases, which will result in silent conversion of INT64/FLOAT64 values to `Infinity`, which would cause unexpected data loss.

I felt that the safer path would be to fail the conversion instead of proceeding with importing lossy data. If we really want to support them, we could possibly have a flag/option that the customer can willingly set to override this failure.

FLOAT32 is the first type that sees this issue, as we were supporting LONG and DOUBLE, which support storing the smaller types (INT and FLOAT). If and when we start supporting INT32, we would have to take the same decision on whether to support conversions from a higher range type to a lower range one.

arawind:
> I don't see any changes `ImportPipeline.java`? That is the Avro record to Spanner pipeline, is no change required in that? Similarly don't see changes in the export templates as well - `ExportPipeline.java`

These files do not have any type related details -- they seem generic runners of the pipeline -- should there be any changes there?

pom.xml
arawind: Filed a task in Beam: https://github.com/apache/beam/issues/30700 and added a TODO here.

Will resolve this TODO once that task is done and dataflow's beam is updated to 2.56 (which is expected to have this change).

There are no concerning details around this -- This PR updates the Spanner Java Client library to 6.61, which requires the latest version of gax. Without updating gax, we see errors like `class com.google.cloud.spanner.v1.stub.SpannerStubSettings overrides final method com.google.api.gax.rpc.StubSettings.getEndpoint()Ljava/lang/String;`. These errors are not related to the FLOAT32 type support -- it is just correcting the dependencies here.

Since dataflow's gax dependency seems to be coming from what Beam is using [1], we will not be able to update the spanner library version in dataflow unless the gax version is updated in Beam (which should automatically happen when the spanner library version is updated in Beam too).

(sorry if the details are wrong / confusing, I'm unfamiliar with development around maven and managing its dependencies; PLMK if you have a better suggestion here)

[1] or rather, the spanner component in Dataflow templates is using a Beam version that uses an older gax version, that's incompatible with spanner client lib 6.61. Ideally, we would not see this issue if Beam already uses the latest spanner client library (which I believe is how Dataflow templates' development is expected to be done -- Beam first and then Dataflow templates, but we're trying to skip that step here for a faster rollout)

v1/src/test/java/com/google/cloud/teleport/spanner/AvroRecordConverterTest.java
arawind: Done

v1/src/test/java/com/google/cloud/teleport/spanner/AvroRecordConverterTest.java
arawind: Done

pom.xml
manitgupta: Thank you!

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
manitgupta: Thanks for the details.
So in the `DatastreamToSpanner` details, we are taking the approach of throwing exceptions on hidden lossy castings. These templates are often used for migrations, and the thought process is to force the user to make an explicit decision about such lossy casting. 

A recent example of this is #1386 where we will error out the record on lossy conversions. 


Guess I'm trying to understand - 
What is a practical usecase a customer would want to map Avro int32 bit into a float32 while being okay with rounding?

@darshan-sj you have more context than me on Import/Export, do you have any suggestions? What approach we have followed till now for other data types with similar rounding constraints?

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
manitgupta: This goes back to the earliest comment around if "rounding and lossy conversions are acceptable" or not.

manitgupta:
> > I don't see any changes `ImportPipeline.java`? That is the Avro record to Spanner pipeline, is no change required in that? Similarly don't see changes in the export templates as well - `ExportPipeline.java`
> 
> These files do not have any type related details -- they seem generic runners of the pipeline -- should there be any changes there?

@darshan-sj

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: Thanks, I've removed the INT -> FLOAT conversion for float32. I think this will have to be handled for FLOAT64 too. Not doing it in this PR.

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: Going with the safer option of loss of precision not being acceptable. We could revisit this without breaking backward compatibility if we want to change our stance

arawind:
> Can you run an Export/Import and document it in the PR description?

Done!


> Logically speaking there should be? Your other changes point to converting Avro records and data types, so why won't this change? At the minimum, the IT should change to account for this type right? What am I missing - @darshan-sj ?

The Avro conversion changes are in AvroRecordConverter and AvroSchemaToDdlConverter. The actual pipelines don't need any change.

I've added a new integration test in ExportPipelineIT. Will work on adding ImportPipelineIT too in a bit

v1/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-spanner-bom</artifactId>
-            <version>6.45.3</version>
+            <version>6.61.0</version>
```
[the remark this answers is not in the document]
arawind: Thanks, linked the same TODO here too.

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
darshan-sj: Import/Export is majorly used as Backup/restore tool and we don't support transformations. So, in majority of the cases from and to datatypes would be same since we create the schema in Spanner according to the datatype of the avro record. Only case where we would reach here would be if customer created the tables in spanner manually and changed datatype of a column from INT -> FLOAT and then run import on that. I think it's okay to have lossy conversion since customer is explicitly changing the datatype.

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
darshan-sj: Same comment as in the earlier conversation. We do not support transformations in Import/Export. So, I'm okay with throwing an exception on lossy conversion and handling it later if required.

darshan-sj (approved):

v1/src/main/java/com/google/cloud/teleport/spanner/AvroRecordConverter.java
arawind: Thanks!

darshan-sj (approved):

Abacn:
Please sync this update to Beam repo. LocalSpannerIO will be removed and this change will get lost

arawind:
@Abacn thanks for the ping. We're trying to get this change merged into 2.56 in https://github.com/apache/beam/pull/30893

PR #1381, opened 2024-03-20
load test fixes

aksharauke:
1. Does a single instantiation of the custom sharding class, fixes bug 330334591
2. Added metric in writer for failed shard
3. Made the file lookup retry logic generic - to always retry when file is expected
4. Added retry on UNAVAILABLE so that pipeline is stable
5. Corrected a DELETE bug fix for composite PK
6. MySQL added more exceptions to retry - based on stability load test observations


v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShardingLogicImplFetcher.java
```diff
@@ -0,0 +1,144 @@
...
+import org.joda.time.Instant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Singleton class to get implementation of the shard identification logic. */
+public class ShardingLogicImplFetcher {
+  private static final Logger LOG = LoggerFactory.getLogger(ShardingLogicImplFetcher.class);
+  private static IShardIdFetcher shardIdFetcher = null;
+
+  public static synchronized IShardIdFetcher getShardingLogicImpl(
```
manitgupta, line 46: This is called in the `@Setup` of the `AssignShardDoFn`, won't using `synchronized` mean that all parallel DoFn executions in different threads will block on this? 

Wondering if it will cause a performance impact 🤔

Please check this blog - 

https://medium.com/analytics-vidhya/apache-beam-a-minor-tip-for-cleaner-code-f293df3e3372

aksharauke: No performance impact, ran a load test

manitgupta: Can you evaluate the injection based approach covered in the blog?  `singleton` and `synchronised` in a massively parallel pipeline should be avoided as much as possible IMO.

aksharauke: This cannot be injected for the simple reason that worker is in a different JVM. It was tried and it failed.
There locks are short held and will return the object immediately in most cases. Same thing was observed in load test.

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dao/MySqlDao.java
```diff
@@ -90,9 +90,53 @@ public void batchWrite(List<String> batchStatements) throws SQLException {
...
+                + shardId
+                + ", will retry : "
+                + e.getMessage());
         // gives indication that the shard is being retried
         Metrics.counter(MySqlDao.class, "mySQL_retry_" + shardId).inc();
+        try {
+          Thread.sleep(1000);
+        } catch (java.lang.InterruptedException ex) {
+          throw new RuntimeException(ex);
+        }
```
manitgupta, line 104: I didn't get what we are trying to do -

If we get `CommunicationsException` , we will - 

1. Log a WARN
2. Increment a counter
3. Try to sleep for 1s

Why are we trying to sleep for 1s here? 
There should be a `backoff/retryable` `client` implemented here right? 

One which performs retries and backoffs for the `batchWrite` based on a configuration?

aksharauke: This is by design

aksharauke: I have taken some default here, don't want to over-engineer and take configuration for such things at this stage. If needed will add in future.

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShardingLogicImplFetcher.java
manitgupta: Interesting! 

My understanding of the [blog](https://medium.com/analytics-vidhya/apache-beam-a-minor-tip-for-cleaner-code-f293df3e3372) was - 

The first approach in the blog is similar to what is done in the PR - 

A `synchronised` block initializes the `RedistClient`, which is used the `ProcessElement` to fetch it.

The second approach in the blog then creates `RedisOptions` class which has a specific field called - `getRedissonClient()` - this field is annotated with a `@Default.InstanceFactory(RedisClientFactory.class)` which would use a factory to build the client.

As a result, the DoFn can cleanly get this object without the need to use `synchronisation`. Are you saying this won't work in the context of the change we are making here? Just like other `pipelineOptions` this is also just an object right (irrespective of how it was constructed).

I was thinking that analogously, there can be a factory class which builds the object of the class which is initialized using the JAR, and then that object can be injected? The `@Default.InstanceFactory(<class-name>.class)` would take care of building it 🤔

Please correct me if I am wrong!

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dao/MySqlDao.java
manitgupta: Understood, why does the `Thread.Sleep` happen inside the `batchWrite` method though? Isn't this similar having a client which contains methods which map to external APIs?

The client would configure the behaviour and method corresponding to the API would do the mark. 

Here it seems like the method which maps to the API is aware of the retry logic?

Please correct me if I am wrong!

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShardingLogicImplFetcher.java
aksharauke: Could you tell how it will work for an object that is not in the same JVM?

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dao/MySqlDao.java
aksharauke: Did not understand the comment above. I want the caller of batchWrite method to not worry about the retry. There are specific exceptions visible to batchWrite and it can decide to retry based on those.

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShardingLogicImplFetcher.java
manitgupta: Discussed this offline, will try this out sometime on my own as a thought exercise.
Please feel free to go ahead with the current approach.

shreyakhajanchi (dismissed):
LGTM. Left a minor comment

v2/gcs-to-sourcedb/src/test/resources/MultiColmPKSession.json
```diff
@@ -0,0 +1,759 @@
+{
```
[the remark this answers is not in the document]
aksharauke: Done

shreyakhajanchi (approved):

manitgupta (dismissed):

shreyakhajanchi (approved):

PR #1385, opened 2024-03-20
support vector_length column annotation in DataFlow Import/Export

ambujojha:
- Support for vector_length annotation added for both gsql and spangres.
- vector_length is treated part of the "spanner_type" or "column_type" returned by Spanner's InformationSchema
- SizedType (already existing object) has been updated to support arrayLength which is used to store vector_length param
- Associated tests

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
@@ -19,6 +19,8 @@ import static java.lang.Character.isWhitespace;
 
 import com.google.cloud.spanner.Dialect;
 import com.google.common.collect.ImmutableList;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 /** Describes a type with size. */
 public final class SizedType {
@@ -26,6 +28,13 @@ public final class SizedType {
   public final Integer size;
   // Describes the exact length for ARRAY types if needed. Used for embedding vectors.
   public final Integer arrayLength;
+
+  private static final Pattern EMBEDDING_VECTOR_PATTERN =
+      Pattern.compile(
+          "^ARRAY<([a-zA-Z0-9]+)>\\(vector_length=>(\\d+)\\)$", Pattern.CASE_INSENSITIVE);
+
+  private static final Pattern PG_EMBEDDING_VECTOR_PATTERN =
+      Pattern.compile("^(\\D+)\\[\\]\\svector\\slength\\s(\\d+)$", Pattern.CASE_INSENSITIVE);
 
   public SizedType(Type type, Integer size) {
     this.type = type;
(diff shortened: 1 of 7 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
```diff
@@ -99,10 +108,32 @@ public static String typeString(Type type, Integer size) {
     throw new IllegalArgumentException("Unknown type " + type);
   }
 
+  public static String typeString(Type type, Integer size, Integer arrayLength) {
```
jonathan-lemos, line 111: Why `Integer` and not `int` for the `arrayLength`? I don't see any places where the argument can be `null`.

v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
```diff
@@ -140,9 +171,16 @@ public static SizedType parseSpannerType(String spannerType, Dialect dialect) {
           }
           if (spannerType.startsWith("ARRAY<")) {
             // Substring "ARRAY<xxx>"
+            // Handle vector_length annotation
+            int annotationStart = spannerType.indexOf("(vector_length");
```
jonathan-lemos, line 175: Can you use regex to handle this parsing? It's not clear to a reader what the significance of this index is, or the `+16` below.

v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
```diff
@@ -196,6 +234,15 @@ public static SizedType parseSpannerType(String spannerType, Dialect dialect) {
         }
       case POSTGRESQL:
         {
+          if (spannerType.contains("[] vector length")) {
```
jonathan-lemos, line 237: Isn't `VECTOR LENGTH` also acceptable? Ignore if `spannerType` is guaranteed to be lowercase.

v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
```diff
@@ -196,6 +234,15 @@ public static SizedType parseSpannerType(String spannerType, Dialect dialect) {
...
+            // Substring "xxx[] vector length yyy"
+            // Handle vector_length annotation
+            int annotationStart = spannerType.indexOf("vector length");
```
jonathan-lemos, line 240: Ditto for using regex to parse this.

v1/src/test/java/com/google/cloud/teleport/spanner/common/SizedTypeTest.java
```diff
@@ -70,4 +70,25 @@ public void testStruct() {
         SizedType.typeString(complexStruct.type, null),
         "STRUCT<a BOOL, b ARRAY<STRUCT<c STRING(MAX), d ARRAY<FLOAT64>>>, e STRUCT<f STRUCT<g INT64>>>");
   }
+
+  @Test
+  public void testEmbeddingVector() {
```
jonathan-lemos, line 75: Nit: it would be nicer if this was 2 separate tests.

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/DdlTest.java
```diff
@@ -252,6 +252,67 @@ public void pgSimple() {
...
+
+    Ddl ddl = builder.build();
+    assertThat(
+        ddl.prettyPrint(),
+        equalToCompressingWhiteSpace("CREATE TABLE `Users` ("
+                + " `id` INT64 NOT NULL,"
+                + " `embedding_vector` ARRAY<FLOAT64>(vector_length=>128),"
+                + " ) PRIMARY KEY (`id` ASC)"
+                ));
+    assertNotNull(ddl.hashCode());
```
jonathan-lemos, line 281: Not necessary to address this because the other tests have it, but it doesn't look like this assertion does anything because an `int` will never be `null`.

jonathan-lemos (changes requested):
Looks like you need to run Spotless on this as well.

Overall content LGTM, but want some finer details to be addressed.

v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
ambujojha: Agreed.

v1/src/main/java/com/google/cloud/teleport/spanner/common/SizedType.java
ambujojha: well. This file doesn't treats spannertype as case sensitive. spannertype is reported by Spanner's Information Schema. 
however, I have made the regex case insensitive to make the code more robust.

jonathan-lemos (approved):

PR #1386, opened 2024-03-21
Fix conversion to Long

manitgupta:
The [toLong](https://github.com/manitgupta/DataflowTemplates/blob/f44f7ce818882b337b7989b504a7ff1e5a234d2d/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventTypeConvertor.java#L81) method in this class converts an incoming string from JsonNode to Double first and then converts it to a Long.

While both Double and Long are 64 bit, Double reserves extra bits to store the exponent -
3bM2MXXruW8uM2Q

Therefore, the same range of numbers as Long cannot be represented by Double.
For example, `3123556554908346902` cannot be converted to Long correctly via a Double.

The fix is to directly convert an incoming string to a Long using `Long.valueOf`.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventTypeConvertor.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventTypeConvertor.java
@@ -78,7 +78,7 @@ public class ChangeEventTypeConvertor {
     try {
       JsonNode node = changeEvent.get(key);
       if (node.isTextual()) {
-        return Double.valueOf(node.asText()).longValue();
+        return Long.valueOf(node.asText());
       }
       return node.asLong();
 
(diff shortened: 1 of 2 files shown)


bharadwaj-aditya (approved):
LGTM

darshan-sj (approved):

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventTypeConvertor.java
```diff
@@ -78,7 +78,7 @@ public static Long toLong(JsonNode changeEvent, String key, boolean requiredFiel
     try {
       JsonNode node = changeEvent.get(key);
       if (node.isTextual()) {
-        return Double.valueOf(node.asText()).longValue();
+        return Long.valueOf(node.asText());
       }
       return node.asLong();
```
krishnamoorthy-r, line 83: Is this line covered in unit test (i.e node.isTextual is false)? If not, please add tests to cover this branch as well.

krishnamoorthy-r (approved):
LGTM.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventTypeConvertor.java
manitgupta: All branches are covered, PTAL - 

https://screenshot.googleplex.com/85pADZkWq3ErpCQ

PR #1391, opened 2024-03-25
adding .withMaxReadTime()

Mauratay:


v2/mqtt-to-pubsub/src/main/java/com/google/cloud/teleport/v2/templates/MqttToPubsub.java
```diff
@@ -80,6 +81,7 @@ public static void validate(MqttToPubsubOptions options) {
   }
 
   public static PipelineResult run(MqttToPubsubOptions options) {
+    Duration tenHours = Duration.standardHours(10);
```
liferoad, line 84: should this be one option for users to specify?

liferoad:
@Abacn please take a look at this PR as well.

v2/mqtt-to-pubsub/src/main/java/com/google/cloud/teleport/v2/templates/MqttToPubsub.java
Mauratay: I set it to max as a default, but I could change it so users specify it

Mauratay (commented):
let me know if you'd like me to include it as an input

v2/mqtt-to-pubsub/src/main/java/com/google/cloud/teleport/v2/templates/MqttToPubsub.java
Abacn: Note that with this change maxReadTime is still not user configurable. If I understood correctly the goal is to prevent the pipeline / some data read stuck indefinitely so sets a large and finite timeout, which user usually does not need to config.

Abacn:
Could run `mvn spotless:apply` to the project to fix the format check

Mauratay (commented):
update after running mvn spotless:apply

Mauratay:
adding .withMaxReadTime() turns the job into a batch, closing PR to prevent user confusion

PR #1395, opened 2024-03-26
Rollback YamlTemplate Beam version to 2.54

Polber:


python/src/main/python/yaml-template/requirements.txt
```diff
@@ -1 +1 @@
-apache-beam[gcp,yaml]
+apache-beam[dataframe,gcp,test,yaml]==2.54.0
```
damccorm, line 1: Can we add a comment here - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2de9cc9b139dab126f27857f054b9d63ca6290ad/pom.xml#L48 - reminding people to update both that file and any files with 2.54.0 specified directly using code search?

Long term, we will probably want to find a way to share the version across templates, but this is fine for now

Polber: Added a comment

damccorm (approved):

PR #1401, opened 2024-03-27
Use neo4j version and edition for reset and schema generation

ali-ince:
This PR introduces neo4j version introspection so that we generate version targeted statements during reset database and schema generation.

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
@@ -230,7 +230,9 @@ public class CypherGenerator {
 
     if (capabilities.hasNodeKeyConstraints()) {
       cyphers.addAll(getEntityKeyConstraintStatements(labels, keyProperties));
+    }
 
+    if (capabilities.hasNodeUniqueConstraints()) {
       for (String uniqueProperty : uniqueProperties) {
         cyphers.add(
             "CREATE CONSTRAINT IF NOT EXISTS FOR (n:"
@@ -315,7 +317,9 @@ public class CypherGenerator {
                 + ModelUtils.makeSpaceSafeValidNeo4jIdentifier(relKeyProperty)
                 + " IS RELATIONSHIP KEY");
       }
+    }
 
+    if (capabilities.hasRelationshipUniqueConstraints()) {
       for (String uniqueProperty : uniqueProperties) {
         cyphers.add(
             "CREATE CONSTRAINT IF NOT EXISTS FOR ()-[r:"
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
@@ -27,6 +27,7 @@ import com.google.cloud.teleport.v2.neo4j.model.job.Target;
 import com.google.cloud.teleport.v2.neo4j.utils.ModelUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
```diff
@@ -295,30 +306,37 @@ private static Set<String> getRelationshipIndexAndConstraintsCypherStatements(Ta
...
+      for (String relKeyProperty : keyProperties) {
+        cyphers.add(
+            "CREATE CONSTRAINT IF NOT EXISTS FOR ()-[r:"
+                + escapedType
+                + "]-() REQUIRE r."
+                + ModelUtils.makeSpaceSafeValidNeo4jIdentifier(relKeyProperty)
+                + " IS RELATIONSHIP KEY");
+      }
+
+      for (String uniqueProperty : uniqueProperties) {
```
fbiville, line 319: shouldn't this be outside the `if (capabilities.hasRelationshipKeyConstraints())` block?

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jConnection.java
```diff
@@ -96,7 +128,52 @@ public void resetDatabase() {
...
+    }
+  }
+
+  private void deleteData() {
+    String ddeCypher = "MATCH (n) CALL { WITH n DETACH DELETE n } IN TRANSACTIONS";
+    LOG.info("Executing delete Cypher query: {}", ddeCypher);
+    executeCypher(ddeCypher, databaseResetMetadata("cit-detach-delete"));
+  }
+
+  private void dropSchema(Neo4jCapabilities capabilities) {
```
fbiville, line 142: you even got rid of APOC, nice!

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
ali-ince: key constraints and unique constraints were introduced at the same time for relationships, so that's why I tied key constraints to unique constraints.

ali-ince: done this refactor @fbiville.

ali-ince: also added a couple of integration tests against different versions and editions.

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/Neo4jCapabilities.java
```diff
@@ -0,0 +1,99 @@
...
+  public boolean hasNodeKeyConstraints() {
+    return hasConstraints();
+  }
+
+  public boolean hasNodeUniqueConstraints() {
+    return hasConstraints();
+  }
+
+  public boolean hasRelationshipKeyConstraints() {
+    return hasConstraints() && version == Neo4jVersion.V5;
```
fbiville, line 46: isn't it version >= 5.7 ?

fbiville (approved):

ali-ince:
would you please approve the workflow and review this @liferoad?

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/database/CypherGenerator.java
```diff
@@ -190,19 +191,21 @@ private static String getPropertiesListCypherFragment(
...
     switch (type) {
       case node:
-        return getNodeIndexAndConstraintsCypherStatements(target);
+        return getNodeIndexAndConstraintsCypherStatements(target, capabilities);
       case edge:
-        return getRelationshipIndexAndConstraintsCypherStatements(target);
+        return getRelationshipIndexAndConstraintsCypherStatements(target, capabilities);
       default:
-        throw new IllegalArgumentException(String.format("unexpected target type: %s", type));
+        return Collections.emptySet();
```
liferoad, line 203: is the empty set later handled correctly? Jus curious since the current behavior indicates this is not the error that can be retried.

fbiville: It used to throw before, as this method was only called for targets for which we can generate constraints (node and relationship targets).
This PR changes the call site in a way that this is called for every target so returning an empty set of constraints work well here (e.g.: query targets don't have any associated constraints to initialize).

ali-ince: Yes, that's fine. Not only caller handles an empty set correctly, but also there is already an if block that will prevent this line to execute.

PR #1403, opened 2024-03-27
Default to THROUGHPUT_BASED autoscaling for classic templates

damccorm:
Before migrating the release process to GitHub, most streaming classic templates had autoscaling enabled by default. This functionality was lost as part of the move over, and this seeks to restore that default.

This has caused user pain because there's not an easy way to enable autoscaling for classic templates if its not on by default.

Note that I chose 10 workers as the default to avoid overwhelming users projects with high scaling (this happened to our templates project with integration tests when I tried 100) and because it seems like a reasonable default. It will cause streaming engine jobs without a default set to change their behavior (currently they default to 100 max workers)

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
@@ -259,7 +259,7 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
     arguments.add(element("argument", "--project=" + projectId));
     arguments.add(element("argument", "--region=" + region));
     if (imageSpec.getMetadata().isStreaming()) {
-      // Default autoscaling to THROUGHPUT_BASED for streaming classic templates
+      // Default to THROUGHPUT_BASED autoscaling for streaming classic templates
       arguments.add(element("argument", "--autoscalingAlgorithm=THROUGHPUT_BASED"));
       arguments.add(element("argument", "--maxNumWorkers=5"));
     }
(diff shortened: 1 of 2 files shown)


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -258,6 +258,11 @@ protected String stageClassicTemplate(
     arguments.add(element("argument", "--templateLocation=" + templatePath));
     arguments.add(element("argument", "--project=" + projectId));
     arguments.add(element("argument", "--region=" + region));
+    if (imageSpec.getMetadata().isStreaming()) {
+      // Default to THROUGHPUT_BASED autoscaling for streaming classic templates
```
liferoad, line 262: so the user cannot change these settings?

damccorm: They can adjust maxNumWorkers, but not the autoscaling algorithm

liferoad: will `arguments.add(element("argument", "--maxNumWorkers=10"));` override the maxNumWorkers option specified by the user?

damccorm: Overriding maxNumWorkers still works. I updated to demonstrate this with a test below which produced this job - https://pantheon.corp.google.com/dataflow/jobs/us-central1/2024-04-04_09_33_22-17280287866962791451;graphView=0?project=cloud-teleport-testing&e=13802955&mods=logs_tg_staging&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))

an2x (dismissed):

liferoad (approved):

PR #1405, opened 2024-03-28
YamlTemplate speed improvements

Polber:
Create new Dockerfile based on Google-provided base template images that serves as a Distroless Xlang image for YAML templates.


.github/scripts/startup-script.sh
```diff
@@ -60,7 +60,7 @@ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
 
 # add runner to docker group
```
damccorm, line 61:
```suggestion
# add user to docker group
```

metadata/src/main/java/com/google/cloud/teleport/metadata/Template.java
```diff
@@ -112,4 +112,10 @@ enum TemplateType {
 
   /** Marker if the template is still in preview / pre-GA. */
   boolean preview() default false;
+
+  /**
+   * Comma-separated list of files to include in Template image when building with Dockerfile. Only
+   * works for YAML and XLANG types.
```
damccorm, line 118: Would be good to give an example - e.g. are these local files or can they be on gcs, etc...

damccorm: Actually, even better might be just giving the docker copy command here - that way its also clear where they're getting copied to

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -1,16 +1,67 @@
...
+# Copy template, python wheels and python launcher script from python-base
+COPY --from=python-base /template /template
+COPY --from=python-base /tmp/dataflow-requirements-cache /tmp/dataflow-requirements-cache
+COPY --from=python-base /opt/google/dataflow/python_template_launcher /opt/google/dataflow/python_template_launcher
+
+# Copy python and installed packages from python-base
+COPY --from=python-base /usr/local/bin/python$PYTHON_VERSION /usr/local/bin/python
+COPY --from=python-base /usr/local/lib/python$PYTHON_VERSION /usr/local/lib/python$PYTHON_VERSION
+
+# Copy required shared libraries from python-base
```
damccorm, line 51: Where do these libraries come from?

damccorm: Generally, it feels like we're taking in base container images, but making a lot of assumptions about how they'll be configured. What constraints do we have on our base images here?

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -560,15 +560,40 @@ private void stageFlexYamlTemplate(
...
-    stageYamlUsingDockerfile(imagePath, yamlTemplateName + "/Dockerfile");
+        basePythonContainerImage,
+        baseContainerImage,
+        beamVersion,
+        yamlTemplateName,
+        otherFiles,
+        outputClassesDirectory);
+
+    boolean useRootDirectory = true;
+    if (new File(outputClassesDirectory.getPath() + "/" + yamlTemplateName + "/main.py").exists()) {
```
damccorm, line 586: We're duplicating this logic in a couple places - instead of just passing in a boolean, could we pass a string like entrypointDirectory or something that resolves to the relative path here? Then all we need to worry about below is `/`s which can be handled with libraries.

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
Polber: They are being copied from Python base image which has a full install of python on it (with necessary licenses and libraries). I am copying the minimum shared c libraries that python uses to compile the graph since that is all that is done in the template launcher.

Ideally, there would be a staged distroless base image that has python, java and up-to-date libraries, but until that is possible, this provides significant speed up by copying just what is needed from both.

I don't know about "constraints" but the base image for python is only updated with the new version of beam, so nothing else should be changing other than the site-packages folder (which is not used by this anyway)

Polber: I suppose the python base image could theoretically be kept static since it is just used as a build env, but using the parameter keeps it inline with the other templates that also use it (i.e. python templates and java templates with python UDF support), so that ti only needs to be kept updated in one place.

The java base image could also theoretically be kept static, but having a parameter for it allows a user to use a custom java image (albeit this use-case may not exist in the wild), and the same argument could be made for the python image too.

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
Polber: The logic isn't exactly duplicated. There are 2 different directories we are tracking - the build directory within the plugin, and the build directory on cloud build. True YAML templates are built with a slightly different structure in both cases from the YamlTemplate.

This is step 1 in an effort to eventually migrate the YamlTemplate to the yaml module so that all yaml-related features are consolidated in one place.

Polber:
@damccorm Fixed relevant changes, and this should also fix the broken YamlTemplate IT's

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/YamlDockerfileGenerator.java
```diff
@@ -37,18 +38,27 @@ public class YamlDockerfileGenerator {
...
 
     Map<String, Object> parameters = new HashMap<>();
-    parameters.put("baseContainerImage", basePythonContainerImage);
-    parameters.put("yamlTemplateName", yamlTemplateName + ".yaml");
+    parameters.put("basePythonContainerImage", basePythonContainerImage);
+    parameters.put("baseJavaContainerImage", baseJavaContainerImage);
+    parameters.put("beamVersion", beamVersion);
+    parameters.put("pythonVersion", pythonVersion);
+    parameters.put(
+        "otherFiles", String.join("* ", otherFiles) + (otherFiles.size() > 0 ? "*" : ""));
```
damccorm, line 61: This is kind of confusing to me - why do we need the * at the end of this?

`COPY main.py` should work just as well as `COPY main.py*`, right? And there could be times you don't want this.

Should we instead just tell consumers that this follows docker syntax and you can append `*` or other special characters - https://docs.docker.com/reference/dockerfile/#copy

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/model/ImageSpecMetadata.java
```diff
@@ -213,14 +212,6 @@ public void setSupportsExactlyOnce(boolean supportsExactlyOnce) {
...
   }
 
-  public String getDefaultStreamingMode() {
-    return defaultStreamingMode;
-  }
-
-  public void setDefaultStreamingMode(String defaultStreamingMode) {
-    this.defaultStreamingMode = defaultStreamingMode;
-  }
-
```
damccorm, line 223: Looks like maybe a bad merge? Did you mean to get rid of this?

damccorm: And other streaming mode changes?

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -1,16 +1,67 @@
...
-COPY requirements.txt /template
-COPY ${yamlTemplateName} /template/template.yaml
+ARG REQUIREMENTS_FILE=requirements.txt
+ARG BEAM_VERSION=${beamVersion}
+ARG BEAM_PACKAGE=apache-beam[dataframe,gcp,test,yaml]==$BEAM_VERSION
+ARG PY_VERSION=${pythonVersion}
+
+# Copy template files to /template
+RUN mkdir -p $WORKDIR
+COPY main.py* requirements.txt* ${otherFiles} /template/
```
damccorm, line 15: Similarly, why do we need `main.py*` instead of just `main.py` here?

damccorm: Are you just not sure if these files exist? If so, I guess its fine to have here, but I still wouldn't append it to otherFiles (which presumably all should exist)

damccorm: Also, I think we are assuming main.py exists at least

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
damccorm: I think we make too many assumptions about the structure of the base image to allow arbitrary images. For example `/tmp/dataflow-requirements-cache` is not guaranteed to exist (and won't exist) in most images

damccorm: We're also assuming that the java container doesn't have python already installed (may or may not be true). Generally, this dockerfile is fit pretty tightly to the images its building on, so I don't think this should be parameterized.

> so that ti only needs to be kept updated in one place.

We could still keep the version in a single place next to the images (we could even keep image names stripped of version and version as separate vars since I imagine generally people just want to update the version)

plugins/core-plugin/src/main/resources/README-template.md
```diff
@@ -19,14 +19,14 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 
 ## Parameters
 
-### Required parameters
+### Required Parameters
 
-<#list spec.metadata.parameters as parameter><#if !parameter.optional!false>* **${parameter.name}** : ${parameter.helpText?ensure_ends_with(".")}
+<#list spec.metadata.parameters as parameter><#if !parameter.optional!false>* **${parameter.name}** (${parameter.label}): ${parameter.helpText?ensure_ends_with(".")}
```
damccorm, line 24: I think this is another bad merge

damccorm: There are more below too

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/YamlDockerfileGenerator.java
Polber: It was initially a workaround to copy `requirements.txt` only if it exists, but I suppose there is to reason to append asterisks on the other files. I can remove those.

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/model/ImageSpecMetadata.java
Polber: I did not, good catch

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
Polber: Same explanation here - you're right, no reason to wildcard main.py

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
Polber: The Java container does not have Python installed, it is Java distroless, so it doesn't have much more than a Java runtime and the template files/launcher.

`/tmp/dataflow-requirements-cache` is guaranteed to exist on python base image because that is where the python wheels are being downloaded to (also documented in https://cloud.google.com/dataflow/docs/guides/templates/configuring-flex-templates#supply-requirements-file)

I do get your point about it being tightly fitted, and while I think it would be nice that have a common definition so that they can stay in sync, I think there may be a few more iterations to this anyway, so I'll go ahead and hard code it for now.

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -1,16 +1,68 @@
...
-COPY ${yamlTemplateName} /template/template.yaml
+ARG REQUIREMENTS_FILE=requirements.txt
+ARG BEAM_VERSION=${beamVersion}
+ARG BEAM_PACKAGE=apache-beam[dataframe,gcp,test,yaml]==$BEAM_VERSION
+ARG PY_VERSION=${pythonVersion}
+
+# Copy template files to /template
+RUN mkdir -p $WORKDIR
+COPY main.py requirements.txt* /template/
+<#if copyOtherFiles??>COPY ${copyOtherFiles} /template/</#if>
```
damccorm, line 16: Why did we switch to this if block? If `copyOtherFiles` is empty shouldn't `COPY main.py requirements.txt* /template/` be valid?

damccorm: I'm also not familiar with this `<#if` syntax - is this a valid dockerfile? It looks more like terraform if statements.

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -1,16 +1,68 @@
...
+    && /venv/bin/pip install --no-cache-dir -U -r $REQUIREMENTS_FILE \
+    && /venv/bin/pip download --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $REQUIREMENTS_FILE \
+    && rm -rf /usr/local/lib/python$PY_VERSION/site-packages  \
+    && mv /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
+
+
+#============================================================#
+# Create Distroless xlang image compatible with YamlTemplate #
+#============================================================#
+FROM ${baseJavaContainerImage}
```
damccorm, line 34: Looks like you changed python but not java, but similar (if less strong) assumptions about its structure seem to be built in. I'd vote we change this, but I won't block on it

plugins/core-plugin/src/test/java/com/google/cloud/teleport/plugin/YamlDockerfileGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  public void testGenerateDockerfile() throws IOException, TemplateException {
+    new File(outputFolder.getAbsolutePath() + "/word-count").mkdirs();
+    YamlDockerfileGenerator.generateDockerfile(
+        "a java container image",
+        "beam_version",
+        "py_version",
+        "word-count",
+        List.of("other_file"),
```
damccorm, line 44: Would be good to have a test with an empty list here as well to test that it does the right thing

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
damccorm: Discussed offline, this is resolved by freemarker (this is a txt file)

plugins/core-plugin/src/test/java/com/google/cloud/teleport/plugin/YamlDockerfileGeneratorTest.java
```diff
@@ -0,0 +1,76 @@
...
+        List.of(),
+        outputFolder);
+    File outputFile = new File(outputFolder.getAbsolutePath() + "/word-count/Dockerfile");
+
+    assertTrue(outputFile.exists());
+    String fileContents = Files.toString(outputFile, Charsets.UTF_8);
+    assertThat(fileContents).contains("FROM a java container image");
+    assertThat(fileContents).contains("=beam_version");
+    assertThat(fileContents).contains("=py_version");
+    assertThat(fileContents).contains("COPY main.py requirements.txt* /template/");
```
damccorm, line 53: I think this is fine - would probably be more robust to check that this is the only copy (aka we don't have `COPY  /template/`, but I'm ok moving forward with it

damccorm (dismissed):

plugins/core-plugin/src/test/java/com/google/cloud/teleport/plugin/YamlDockerfileGeneratorTest.java
Polber: Added a test

damccorm (approved):
Thanks!

PR #1408, opened 2024-04-01
remove the deprecated reverse replication templates

aksharauke:
Removing deprecated reverse replication templates


bharadwaj-aditya (approved):
LGTM

PR #1409, opened 2024-04-01
DataFlow Export and Import Integration tests for embedding vectors. 

ambujojha:
- Add integration tests for import and export for vector embeddings.


aksharauke (commented):
I suggest to create separate integration tests for vector embeddings. Not all tests need the host to the sandbox host.

PR #1414, opened 2024-04-04
[BULK] Add schemaMapperInterface and implementations

Deep1998:
Added a schemaMapperInterface which provides the functions used for schema translations and fetching types.
Added 2 implementations:
- LazyMapper: Does not do any translation and returns the data types via Ddl object.
- SessionBasedMapper: Uses Schema (session file) for translation and Ddl for types.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/LazyMapper.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/LazyMapper.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.migrations.schema;
+
+import com.google.cloud.teleport.v2.spanner.ddl.Column;
+import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.ddl.Table;
+import com.google.cloud.teleport.v2.spanner.type.Type;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+
+public class LazyMapper implements SchemaMapperIface {
+
+    Ddl ddl;
+
+    public LazyMapper(Ddl ddl) {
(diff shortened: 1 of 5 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/LazyMapper.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.cloud.teleport.v2.spanner.ddl.Column;
+import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.ddl.Table;
+import com.google.cloud.teleport.v2.spanner.type.Type;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+
+public class LazyMapper implements SchemaMapperIface {
```
aksharauke, line 27: Will help to document the motivation/purpose of the class. Also the name does not clarify the intent, perhaps something like IdentityMapper or something to that tune would help.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaMapperIface.java
```diff
@@ -0,0 +1,48 @@
...
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.migrations.schema;
+
+import com.google.cloud.teleport.v2.spanner.type.Type;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+
+public interface SchemaMapperIface {
```
aksharauke, line 23: Could be named as ISchemaMapper

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+public class SessionBasedMapper implements SchemaMapperIface {
+
+    Ddl ddl;
+
+    Schema schema;
+
+    public SessionBasedMapper(Schema schema, Ddl ddl) {
```
aksharauke, line 34: It is named SessionBasedMapper , but there is nothing that reflects a session file. Does it make sense to pass the session file path here instead, rather than Schema?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,97 @@
...
+        if (!cols.containsKey(spannerColumn)) {
+            throw new NoSuchElementException(String.format("Spanner column %s not found for table %s", spannerColumn, spannerTable));
+        }
+        return cols.get(spannerColumn);
+    }
+
+    @Override
+    public Type getSpannerColumnType(String spannerTable, String spannerColumn) throws NoSuchElementException {
+        Table spTable = ddl.table(spannerTable);
+        if (spTable == null) {
```
aksharauke, line 78: I feel these checks should be done one time - at the startup and not for every call.

PR #1415, opened 2024-04-04
[BULK] Add schemaMapperInterface and implementations

Deep1998:
Added a schemaMapperInterface which provides the functions used for schema translations and fetching types.
Added 2 implementations:
- IdentityMapper: Does not do any translation and returns the data types via Ddl object.
- SessionBasedMapper: Uses Schema (session file) for translation and Ddl for types.


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+public class SessionBasedMapper implements ISchemaMapper {
+
+  Ddl ddl;
+
+  Schema schema;
+
+  public SessionBasedMapper(Schema schema, Ddl ddl) {
```
aksharauke, line 33: It is named SessionBasedMapper , but there is nothing that reflects a session file. Does it make sense to pass the session file path here instead, rather than Schema?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.cloud.teleport.v2.spanner.ddl.Column;
+import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.ddl.Table;
+import com.google.cloud.teleport.v2.spanner.type.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+public class SessionBasedMapper implements ISchemaMapper {
```
aksharauke, line 27: Class documentation will help.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,100 @@
...
+          String.format("Spanner column %s not found for table %s", spannerColumn, spannerTable));
+    }
+    return cols.get(spannerColumn);
+  }
+
+  @Override
+  public Type getSpannerColumnType(String spannerTable, String spannerColumn)
+      throws NoSuchElementException {
+    Table spTable = ddl.table(spannerTable);
+    if (spTable == null) {
```
aksharauke, line 81: Can this check be done once in the constructor for all tables that will be migrated?

Deep1998: Whenever we run into these situations, it will be only be due to an invalid session file. Because even table names, columns etc will be fetched from the session file itself. We only receive the the src table name in the changeEvent which is the only place where an external input can be incorrect which we should catch. 

However, we have had customers in the past who tried to modify the session file manually (and incorrectly) which caused the template to fail silently, which could've been avoided if we caught mismatches `chanveEvent.srcColumnName` and columns present in the session.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
Deep1998: We could do that, or rename this to SchemaBasedMapper?

aksharauke: Depends on if the session file or the Schema object will be used elsewhere other than this mapper. If Schema object is not needed elsewhere in the code, then no need to expose it and only a session file can be taken here as input.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
aksharauke: What I meant is - we have the Ddl object, we have the session file - this validation can be done once in the lifetime of the pipeline - at the time of constructor of this  class. We do not need to do it for every call to this function. Basically fail fast.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
Deep1998: Makes sense

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
Deep1998: SGTM will add a verification to fail fast which will verify the Spanner Table names/col names are consistent with DDL. 
Not adding verification on whether session types  and ddl types are consistent, because if I do, we will not support the use case that users can end up changing types later on which was a common use case as per the last discussion i.e. it is common for users to have different types on Spanner than on session file.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.migrations.schema;
+
+import com.google.cloud.teleport.v2.spanner.type.Type;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+public interface ISchemaMapper {
```
manitgupta, line 22: Is this only for MySQL? 
Different source databases will have different modeling concepts. For example, PG will have 

database -> schema -> table -> column

How will `ISchemaMapper` work for that case?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SessionBasedMapper.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class SessionBasedMapper implements ISchemaMapper {
+
+  Ddl ddl;
+
+  Schema schema;
+
+  public SessionBasedMapper(Schema schema, Ddl ddl) {
+    this.schema = schema;
+    this.ddl = ddl;
```
manitgupta, line 35: Same as earlier: `private final`.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/IdentityMapper.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public String getSpannerTableName(String srcTable) {
+    return srcTable;
+  }
+
+  @Override
+  public String getSpannerColumnName(String srcTable, String srcColumn) {
+    return srcColumn;
+  }
```
manitgupta, line 46: I didn't get this 🤔

IdentityMapper is supposed to perform mapping via "names".  Therefore - 

1. If a spTable exists with the same name as srcTable, spTable == srcTable should be returned.
2. If an spTable with the same name as srcTable does not exist, then NoSuchElement or null should be returned?

In your implementation, `srcTable` is always returned? 

This is clear from the test case below. 

You passed `abc` value in `srcTable` where the `Ddl` only had a table named `Users`. So the `getSpannerTableName` function should return the corresponding Spanner table name, which is nothing, so NoSuchElementFoundException? 

Instead you return `abc`, which will fail later because there is actually no table named `abc` in Spanner!

manitgupta (commented):
Love the size of the PR!

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
Deep1998: We currently do have a schema field but we just populate it with the db name instead of the namespace. The namespace is handled by adding modifying the table name to `namespace.tablename`.

 The reader plans on supporting namespaces first-hand, but the transformer will have to concat it for postgres handling as long as the session file is modified the way it is currently.

It might make sense to add namespace as an input to the interface and pass null for now.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/IdentityMapper.java
Deep1998: Yes this would fail at a mutation stage like it does in the live migration template. We could fail faster by adding the check here.

Deep1998: Done

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
Deep1998: Done

manitgupta (approved):

damccorm:
FYI - all changes from this PR were applied in https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/88edec34809df0e25a33c907f2896b810ebd7e1b - looks like it just didn't close the PR because a merge commit was applied after the internal change was approved/merged. This PR should be safe to close

damccorm:
Yeah that was reasonable; there was a bug which has since been fixed

PR #1417, opened 2024-04-05
Get consistent java-pr signal on main

damccorm:
Right now, https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/java-pr.yml?query=branch%3Amain is perma-red on main. This is because one of the underlying actions in setup-env assumes a checkout depth of 2 or greater - https://github.com/tj-actions/changed-files?tab=readme-ov-file#on-push-%EF%B8%8F - this should fix the problem


Abacn (approved):
thanks for the fix!

PR #1421, opened 2024-04-09
Fix example for GcsWriteFolder with trailing slash

quassy:
> Must end with a slash.

Fix the example to fit the help text.


quassy:
This is incorrect, the given example without a trailing dash is wrong per description of the function.

quassy:
This is incorrect, the given example without a trailing dash is wrong per description of the function.

quassy:
This is incorrect, the given example without a trailing dash is wrong per description of the function.

quassy:
This is incorrect, the given example without a trailing dash is wrong per description of the function.

derrickaw:
Reopen if still valid.

quassy:
@derrickaw Still valid, I have opened #3997 as a follow-up.

PR #1422, opened 2024-04-10
Update TextToBigQueryStreaming.java

tkchowdary:


v1/src/main/java/com/google/cloud/teleport/templates/TextToBigQueryStreaming.java
```diff
@@ -286,7 +286,7 @@ public static PipelineResult run(TextToBigQueryStreamingOptions options) {
                     .withExtendedErrorInfo()
                     .withoutValidation()
                     .withCreateDisposition(CreateDisposition.CREATE_IF_NEEDED)
-                    .withWriteDisposition(WriteDisposition.WRITE_APPEND)
+                    .withWriteDisposition(WriteDisposition.WRITE_TRUNCATE)
```
liferoad, line 289: what is the reason to change the default?

tkchowdary: It is creating duplicates in the BQ table after loading.

liferoad: this changes the behavior of this template. I do not think we should do this.

PR #1423, opened 2024-04-10
Template parameters: update old types to new ones

Amar3tto:
Changed `@TemplateParameter.Text` parameters to new types where it's needed.

### DIFF
--- a/v2/bigtable-changestreams-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableChangeStreamsToHBase.java
+++ b/v2/bigtable-changestreams-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableChangeStreamsToHBase.java
@@ -102,9 +102,9 @@ public class BigtableChangeStreamsToHBase {
         helpText =
             "Whether bidirectional replication between hbase and bigtable is enabled, adds additional logic to filter out hbase-replicated mutations")
     @Default.Boolean(false)
-    boolean getBidirectionalReplicationEnabled();
+    Boolean getBidirectionalReplicationEnabled();
 
-    void setBidirectionalReplicationEnabled(boolean bidirectionalReplicationEnabled);
+    void setBidirectionalReplicationEnabled(Boolean bidirectionalReplicationEnabled);
 
     @TemplateParameter.Text(
         optional = true,
@@ -129,18 +129,18 @@ public class BigtableChangeStreamsToHBase {
         description = "Dry run",
         helpText = "When dry run is enabled, pipeline will not write to Hbase")
     @Default.Boolean(false)
-    boolean getDryRunEnabled();
+    Boolean getDryRunEnabled();
 
-    void setDryRunEnabled(boolean dryRunEnabled);
+    void setDryRunEnabled(Boolean dryRunEnabled);
 
     @TemplateParameter.Boolean(
         optional = true,
         description = "Filter GC mutations",
         helpText = "Filters out garbage collection Delete mutations from CBT")
     @Default.Boolean(false)
-    boolean getFilterGCMutations();
+    Boolean getFilterGCMutations();
 
-    void setFilterGCMutations(boolean filterGCMutations);
+    void setFilterGCMutations(Boolean filterGCMutations);
   }
 
   /**
--- a/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
+++ b/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
@@ -75,12 +75,9 @@ public interface DataplexBigQueryToGcsOptions
 
   void setDestinationStorageBucketAssetName(String destinationStorageBucketAssetName);
 
-  @TemplateParameter.Text(
+  @TemplateParameter.DateTime(
       order = 4,
       optional = true,
-      regexes = {
-        "^([0-9]{4}-[0-9]{2}-[0-9]{2}(T[0-9]{2}:[0-9]{2}:[0-9]{2}(Z|[+-][0-9]{2}:[0-9]{2})?)?|-[pP]([0-9]+(\\.[0-9]+)?Y)?([0-9]+(\\.[0-9]+)?M)?([0-9]+(\\.[0-9]+)?W)?([0-9]+(\\.[0-9]+)?D)?(T([0-9]+(\\.[0-9]+)?H)?([0-9]+(\\.[0-9]+)?M)?([0-9]+(\\.[0-9]+)?S)?)?)$"
-      },
       description = "Move data older than the date.",
       helpText =
           "Move data older than this date (and optional time). For partitioned tables, move partitions last modified before this date/time. For non-partitioned tables, move if the table was last modified before this date/time. If not specified, move all tables / partitions. The date/time is parsed in the default time zone by default, but optional suffixes Z and +HH:mm are supported. Format: YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss or YYYY-MM-DDTHH:mm:ss+03:00. Relative date/time (https://en.wikipedia.org/wiki/ISO_8601#Durations) is also supported. Format: -PnDTnHnMn.nS (must start with -P meaning time in the past).")
--- a/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
+++ b/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
@@ -75,9 +75,12 @@ public interface DataplexBigQueryToGcsOptions
 
   void setDestinationStorageBucketAssetName(String destinationStorageBucketAssetName);
 
-  @TemplateParameter.DateTime(
+  @TemplateParameter.Text(
       order = 4,
       optional = true,
+      regexes = {
+        "^([0-9]{4}-[0-9]{2}-[0-9]{2}(T[0-9]{2}:[0-9]{2}:[0-9]{2}(Z|[+-][0-9]{2}:[0-9]{2})?)?|-[pP]([0-9]+(\\.[0-9]+)?Y)?([0-9]+(\\.[0-9]+)?M)?([0-9]+(\\.[0-9]+)?W)?([0-9]+(\\.[0-9]+)?D)?(T([0-9]+(\\.[0-9]+)?H)?([0-9]+(\\.[0-9]+)?M)?([0-9]+(\\.[0-9]+)?S)?)?)$"
+      },
       description = "Move data older than the date.",
       helpText =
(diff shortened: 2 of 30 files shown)


damccorm (approved):
Thanks!

Amar3tto:
> Running checks, will proceed with merge process once those complete

Fixed Spotless

v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
```diff
@@ -75,12 +75,9 @@ public interface DataplexBigQueryToGcsOptions
 
   void setDestinationStorageBucketAssetName(String destinationStorageBucketAssetName);
 
-  @TemplateParameter.Text(
+  @TemplateParameter.DateTime(
```
an2x, line 78: I'm not sure we should use the DateTime type.
I don't know why, but for some reason it was added as a separate type in the templates plugin. The problem is that the API doesn't actually support DateTime as the parameter type, so the plugin translates it to the regular TEXT parameter: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ffc6262b6fb66a30069b18bb1fe5c34ade117535/plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/model/ImageSpecParameter.java#L365

So by switching from Text to DateTime here we achieved nothing and only lost the regexp check (so now a user can enter anything into this text field without the format being checked).

I think we should either:
1) Keep using Text type here and in other similar places, or
2) Switch to DateTime, but add the `regexes` attribute to this annotation too, and translate it to TEXT param with a regexp in the plugin. I think different templates can expect dates to be in different format, so we can't just use a common hard-coded pattern.

an2x (changes requested):

v2/dataplex/src/main/java/com/google/cloud/teleport/v2/options/DataplexBigQueryToGcsOptions.java
Amar3tto: Could you please take a look [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ffc6262b6fb66a30069b18bb1fe5c34ade117535/metadata/src/main/java/com/google/cloud/teleport/metadata/util/MetadataUtils.java#L188), do we really lose regex with `DateTime`?

Amar3tto: I agree that it is safer not to change `DateTime` here because it had its own regex. But in other places there was only `Text` without regex, so by changing to `DateTime` we will at least add some common regex. What do you think?

an2x: > Could you please take a look [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ffc6262b6fb66a30069b18bb1fe5c34ade117535/metadata/src/main/java/com/google/cloud/teleport/metadata/util/MetadataUtils.java#L188), do we really lose regex with DateTime?

Yes, good point, I missed that. Turns out there is a default regexp for DateTime.
But still, switching from Text with a regexp that is different from this default to DateTime can be problematic.

> But in other places there was only Text without regex, so by changing to DateTime we will at least add some common regex. What do you think?

It's possible that for some of these parameters there is no regexp in the annotation (and hence the format is not checked in the UI), but a certain format is checked/expected in the pipeline itself (it'll just fail in runtime if it's wrong). So my enforcing the DateTime default regexp here we can run into problems with this format not being compatible with what a template expects.
I think we can change it to DateTime only if we review and test each of these templates to make sure the default DateTime format is actually what it expects.
I'd err on the side of just not changing this for now, it's not very important.

v2/bigtable-changestreams-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableChangeStreamsToHBase.java
```diff
@@ -102,9 +102,9 @@ public interface BigtableToHbasePipelineOptions
         helpText =
             "Whether bidirectional replication between hbase and bigtable is enabled, adds additional logic to filter out hbase-replicated mutations")
     @Default.Boolean(false)
-    boolean getBidirectionalReplicationEnabled();
+    Boolean getBidirectionalReplicationEnabled();
 
-    void setBidirectionalReplicationEnabled(boolean bidirectionalReplicationEnabled);
+    void setBidirectionalReplicationEnabled(Boolean bidirectionalReplicationEnabled);
 
     @TemplateParameter.Text(
```
an2x, line 109: Can be changed to `@TemplateParameter.Boolean`?

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSourceDb.java
```diff
@@ -134,7 +134,7 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setTimerIntervalInMilliSec(Integer value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.DateTime(
```
an2x, line 137: Revert DateTime change?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/TextIOToBigQuery.java
```diff
@@ -244,10 +242,9 @@ public interface Options
 
     void setOutputTable(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.GcsWriteFile(
```
an2x, line 245: This should be a `GcsReadFile`.
GcsWriteFile allows selecting a non-existing file. GcsReadFile requires selecting an existing file. `javascriptTextTransformGcsPath` is actually an input parameter for the template and must exist.

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToShardedFileSink.java
```diff
@@ -154,7 +154,7 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setMetadataDatabase(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.DateTime(
```
an2x, line 157: Revert DateTime?

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToShardedFileSink.java
```diff
@@ -164,7 +164,7 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setStartTimestamp(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.DateTime(
```
an2x, line 167: Revert DateTime?

v2/spanner-change-streams-to-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToSink.java
```diff
@@ -138,7 +138,7 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setMetadataDatabase(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.DateTime(
```
an2x, line 141: Revert DateTime?

v2/spanner-change-streams-to-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToSink.java
```diff
@@ -148,7 +148,7 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setStartTimestamp(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.DateTime(
```
an2x, line 151: Revert DateTime?

an2x (changes requested):

v2/bigtable-changestreams-to-hbase/src/main/java/com/google/cloud/teleport/v2/templates/BigtableChangeStreamsToHBase.java
Amar3tto: No, this is Text (e.g. "BIDIRECTIONAL_REPL_SOURCE_CBT")

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/GCSToSourceDb.java
Amar3tto: Done

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/TextIOToBigQuery.java
Amar3tto: Fixed

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToShardedFileSink.java
Amar3tto: Done

v2/spanner-change-streams-to-sharded-file-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToShardedFileSink.java
Amar3tto: Done

v2/spanner-change-streams-to-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToSink.java
Amar3tto: Done

v2/spanner-change-streams-to-sink/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToSink.java
Amar3tto: Done

an2x (commented):
The most recent version looks good to me, but the integration tests are failing: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1423/checks?check_run_id=24262306056

The first one related to Python UDF can be ignored, but there are a lot of errors complaining about wrong PubSub topic format. Can you please check why?

akashorabek:
@an2x made the changes. Can you run it again pls?

akashorabek:
@an2x 
I've reverted the type to Text for `BigtableChangeStreamsToPubSub `and `SpannerChangeStreamsToPubSub`, as the format for their Pub/Sub topics seems to require a different format, namely `someTopic`, whereas for other templates, the format is standard, namely `projects/someProject/topics/someTopic`.

an2x (approved):

damccorm (approved):

Issue #1427, opened 2024-04-11
[Feature Request]: Add new template to publish ordered spanner change records to pub/sub


forksumit:
### Related Template(s)

New template

### What feature(s) are you requesting?

There is already an template [SpannerChangeStreamsToPubSub](https://github.com/forksumit/DataflowTemplates/blob/ordered-spanner-cdc-to-pubsub/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSub.java)  which publishes spanner change records to Pub/Sub but doesn't guarantee any ordering. 

To add a new similar template which orders/sorts change records (within primary key / configurable partition) and published to pub/sub with ordering enabled.

forksumit:
Opened PR: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1428 for same

PR #1428, opened 2024-04-11
Add template SpannerOrderedChangeStreamsToPubSub

forksumit:
Add a new template (similar to [SpannerChangeStreamsToPubSub](https://github.com/forksumit/DataflowTemplates/blob/ordered-spanner-cdc-to-pubsub/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSub.java) ) which orders/sorts change records (within primary key / configurable partition) and publish to pub/sub with ordering enabled.

ISSUE: https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/1427

### DIFF
--- a/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
+++ b/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
@@ -18,6 +18,7 @@ package com.google.cloud.teleport.v2.transforms;
 import com.google.auto.value.AutoValue;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
@@ -240,8 +241,7 @@ public abstract class SpannerChangeStreamsToOrderedKV
     }
   }
 
-  private static class DataChangeRecordComparator
-      implements java.util.Comparator<DataChangeRecord> {
+  private static class DataChangeRecordComparator implements Comparator<DataChangeRecord> {
     @Override
     public int compare(DataChangeRecord r1, DataChangeRecord r2) {
       return r1.getCommitTimestamp().compareTo(r2.getCommitTimestamp());
--- a/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
+++ b/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
@@ -45,20 +45,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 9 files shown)


v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
```diff
@@ -0,0 +1,267 @@
...
+    private BufferKeyUntilOutputTimestamp(long incrementIntervalInSeconds) {
+      this.incrementIntervalInSeconds = incrementIntervalInSeconds;
+    }
+
+    @SuppressWarnings("unused")
+    @TimerId("timer")
+    private final TimerSpec timerSpec = TimerSpecs.timer(TimeDomain.EVENT_TIME);
+
+    @StateId("buffer")
+    private final StateSpec<BagState<DataChangeRecord>> buffer = StateSpecs.bag();
```
aksharauke, line 153: BagState has a 2GB limit. If this pipeline is run to handle backfill data - example to populate 1 day worth of change stream backlog to PubSub - this will fail and enter into a irrecoverable state.

aksharauke (changes requested):
What is the motivation for this template?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/SpannerChangeStreamsToOrderedKV.java
forksumit: This template has some configurable params like `orderingPartitionBucketCount`, `bufferTimerInterval` ,  which has to be configured by user based on their scale. 

`orderingPartitionBucketCount` will limit number of states and timers, `bufferTimerInterval` should help in limiting individual state buffer size. 

Idea is to clearly covey the user that these kinds of issues may occur if the params are not configured/adjusted properly for their scale.

forksumit:
> What is the motivation for this template?

I'm working with a customer who's migrating from MySQL to CloudSpanner. They have a existing CDC process (MaxwellDaemon)  which is publishing ordered change records (for limited tables) to PubSub Queue , then, downstream application of the PubSub topics are replicating data to different DBs. 

Ordering is needed to ensure consistency in target (downstream) DB. 
For example: For a primary there are multiple change records happening in source (INSERT / UPDATE  etc..) Let's say UPDATEs are received in shuffled orders by downstream subscriber , then it'll break the consistency (state of source and target DB won't match). 

Currently, there is no existing template which does that ([SpannerChangeStreamsToPubSub](https://github.com/forksumit/DataflowTemplates/blob/ordered-spanner-cdc-to-pubsub/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSub.java) can stream change records to pub/sub but doesn't ensure any kind of ordering)

forksumit:
As discussed, using bag state might be the only technical option for this specific scenario, but it's not the most reliable due to the 2GB size limitation and the potential risk of irrecoverable state due to the same. Therefore, it's not advisable to release this as a public template.

Instead, users can create their own custom flex template for this particular use case by following the guidelines provided in the public documentation: https://cloud.google.com/spanner/docs/change-streams/use-dataflow#order-by-key

Hence, closing this PR.

PR #1433, opened 2024-04-12
Bugfix for UI issue removing flex templates from drop down menu

fozzie15:
Changes the flexContainerName so that we see both templates in the dropdown menu on the UI.


an2x (dismissed):

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToBigQuery.java
```diff
@@ -68,7 +68,7 @@
       contactInformation = "https://cloud.google.com/support",
       hidden = true),
   @Template(
-      name = "Firestore_to_BigQuery_Flex",
+      name = "Firestore_to_BigQuery_Flex_Xlang",
```
Dippatel98, line 71: Do you think we should keep the naming convention consistent here? with just the `_Xlang` suffix

damccorm (approved):

PR #1437, opened 2024-04-15
Add parentName and parentTriggerValues attributes in the template plugin

Amar3tto:


an2x (approved):

Amar3tto:
@an2x I think failed checks are flaky. What can we do here?

an2x:
> @an2x I think failed checks are flaky. What can we do here?

Yes, the failures seem unrelated to the change. Let me try to merge this change as is.

PR #1439, opened 2024-04-15
Basic Contract for Bulk Reader - Part-1

VardhanThigle:
The Reader interfaces with the source-db allowing the pipeline to read from various data sources like jdbc compliant databases for example.
 
The Reader provides functionality to the pipeline to read the source schema, and, get the ReaderTransform.
 The ReaderTransform typically sits close to the pipeline root and returns two PCollections in a tagged-tuple output:
 1. PCollection&lt;SourceRows&gt; - The Sources rows of all the tables that reader is reading.
2. PCollection&lt;SourceTableReference&gt; - A reference to tables that have completed their read.

In this PR we add the basic contract of the reader interface along with a `FakeReader` implementation.

PR that will be implemented after this:
1. A PR to provide source type mappings.
2. A PR to implement skeleton of other reader interfaces (for jdbc reader) like authentication interfaces etc.
3. A PR to integrate with JDBCIO.


v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/row/SerializableGenericRecordTest.java
```diff
@@ -0,0 +1,113 @@
...
+            .build();
+    SerializableGenericRecord recordLeft = new SerializableGenericRecord(genericRecordLeft);
+    SerializableGenericRecord recordRight = new SerializableGenericRecord(genericRecordRight);
+    assertThat(recordLeft).isEqualTo(recordRight);
+    assertThat(recordLeft.hashCode()).isEqualTo(recordRight.hashCode());
+    recordRight.getRecord().put(0, 4L);
+    assertThat(recordLeft).isEqualTo(recordLeft);
+    assertThat(recordLeft).isNotEqualTo(recordRight);
+    assertThat(recordLeft.hashCode()).isNotEqualTo(recordRight.hashCode());
+    assertThat(recordLeft).isNotEqualTo(123L);
```
aksharauke, line 111: Any reason behind the 123 number?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReaderTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Test class for {@link AccumulatingTableReader}. */
+@RunWith(MockitoJUnitRunner.class)
+public class AccumulatingTableReaderTest implements Serializable {
+  @Rule public final transient TestPipeline testPipeline = TestPipeline.create();
+
+  @Test
+  public void testAccumulatingTableReader() {
```
aksharauke, line 49: nice test

aksharauke (approved):

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/row/SerializableGenericRecordTest.java
VardhanThigle: It's just to make UT coverage hit the check that left and right are of different type hence notEqual.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,124 @@
...
+  abstract TupleTag<SourceTableReference> sourceTableReferenceTag();
+
+  @Override
+  public PCollectionTuple expand(PBegin input) {
+    ImmutableMap<SourceTableReference, PCollection<SourceRow>> tablePCollections =
+        this.tableTransforms().entrySet().stream()
+            .collect(
+                ImmutableMap.toImmutableMap(
+                    Entry::getKey,
+                    e -> input.apply("SourceRowReader." + e.getKey().getName(), e.getValue())));
```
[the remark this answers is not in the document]
VardhanThigle: Thanks!

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,124 @@
...
+                    .map(
+                        e ->
+                            /*
+                             * Reduce the Collection of SourceRows to their counts in a parallel stage.
+                             * This ensures that this parallel stage waits till the table is completed,
+                             * allowing the reader to signal completion of reading a table to the pipeline controller.
+                             * As needed, more complex reduce operation can be performed providing additional information to the pipelineController.
+                             */
+                            e.getValue()
+                                .apply("TableRowCounter." + e.getKey().getName(), Count.globally())
```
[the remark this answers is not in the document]
VardhanThigle: Thanks for the review. We use a count transform there, which is a type of combine transform. So it should ideally be a parallel operation. I will add more note on the performance vs cost tradeof once we have benchmarking results.
Added a TODO for the same.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.TupleTag;
+
+/**
+ * Helper to {@link ReaderTransform} that expands the Pipeline Graph to flatten all the tableReaders
+ * and generate table completions. Please refer to {@link ReaderTransform} for greater details.
+ */
+@AutoValue
+abstract class AccumulatingTableReader extends PTransform<PBegin, PCollectionTuple> {
```
manitgupta, line 41: This does not account for reading tables in a DAG right? This call will change when we add that support?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,125 @@
...
+                             * allowing the reader to signal completion of reading a table to the pipeline controller.
+                             * As needed, more complex reduce operation can be performed providing additional information to the pipelineController.
+                             * TODO(vardhanvthigle): Add a note on cost vs performance of the Count transform after benchmarking.
+                             */
+                            e.getValue()
+                                .apply("TableRowCounter." + e.getKey().getName(), Count.globally())
+                                .apply(
+                                    "TableCompletion." + e.getKey().getName(),
+                                    MapElements.via(new SourceTableReferenceWithCount(e.getKey()))))
+                    .collect(ImmutableList.toImmutableList()))
```
manitgupta, line 79: I didn't get what this is trying to do, could you please elaborate a little bit for my understanding?

Can you also elaborate what do you mean by a signal - my understanding is that the table completion signal can be used for further processing - but how will it work - I didn't get that.

Won't the same signal also be needed for the DAG based implementation?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,125 @@
...
+                        e ->
+                            /*
+                             * Reduce the Collection of SourceRows to their counts in a parallel stage.
+                             * This ensures that this parallel stage waits till the table is completed,
+                             * allowing the reader to signal completion of reading a table to the pipeline controller.
+                             * As needed, more complex reduce operation can be performed providing additional information to the pipelineController.
+                             * TODO(vardhanvthigle): Add a note on cost vs performance of the Count transform after benchmarking.
+                             */
+                            e.getValue()
+                                .apply("TableRowCounter." + e.getKey().getName(), Count.globally())
```
manitgupta, line 75: We are creating a named stage per table - does this mean for a 1000 table database, 1000 stages will be created?

I wonder what the graph would look in the Dataflow console - will it be readable? If this will create 2000+ node graph (`TableRowCounter` and `TableCompletion`) it might not render on the console?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SerializableGenericRecord.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * <p><b>Note:</b>
+ *
+ * <ol>
+ *   <li>Based on benchmarking results, this can be further enhanced to have a singleton schema
+ *       registry and encode only the hash of the schema in the serialization.
+ *   <li>This can't be made as an {@link com.google.auto.value.AutoValue} class, as they are not
+ *       mutable and do not support custom serializations.
+ * </ol>
+ */
```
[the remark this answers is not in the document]
VardhanThigle: We do use builder to build the GenericRecord. That's the part of building `SourceRow`

GenericRecord does not implement Serializable. `SpecificRecord` does, but we cant use that since it needs a fixed schema.

This class is just to implement the serializable interface on GenericRecord.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRow.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.generic.GenericRecordBuilder;
+
+/**
+ * Encapsulates the SourceRow that is generated as a part of reading the source database tables.
+ * Provides convenience methods to get the {@code readTime} and the {@code payload}
+ */
+@AutoValue
+public abstract class SourceRow implements Serializable {
+
```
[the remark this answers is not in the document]
VardhanThigle: 1. Having a wrapper decouples the remaining code from avro.
2. `SourceRow` also has a UUID that will help the transformer to look up the schema mapping interface.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -0,0 +1,125 @@
...
+    ImmutableMap<SourceTableReference, PCollection<SourceRow>> tablePCollections =
+        this.tableTransforms().entrySet().stream()
+            .collect(
+                ImmutableMap.toImmutableMap(
+                    Entry::getKey,
+                    e -> input.apply("SourceRowReader." + e.getKey().getName(), e.getValue())));
+    PCollection<SourceRow> sourceRowPCollection =
+        PCollectionList.of(tablePCollections.values())
+            /* Flatten SourceRows read from all the tables into a single PCollection */
+            .apply("SourceRowCollector", Flatten.<SourceRow>pCollections());
```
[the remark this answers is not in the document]
VardhanThigle: That's a valid route, and that's what the POC template does. Currently the highlevel architect of the bulk template is recommending a reader per sourceDB with a flattened PCollection. 
I have carved out AccumulateTableReader to leave a possiblity of swapping this flattening logic with a pure tuple of PCollections.
We can easily implement another reader with a similar interface w/o having to refactor the rest of the reader.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SerializableGenericRecord.java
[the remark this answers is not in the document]
manitgupta: Why can't we use `GenericData.Record`? I see some code references where it is the output of a DoFn - 

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/69f26e86e095a733316f63da1af30bf77fb4f742/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/CsvConverters.java#L644


https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/69f26e86e095a733316f63da1af30bf77fb4f742/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/CsvConverters.java#L682

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: Yup, but the sequencing can be added here with a easy interace. It is not implemented to avoid complexity till it is required.

For example:
Currently AccumulatingTableReader has a `List<PTransforms>`
That can be changed to a `List<List<Ptransforms>>`
The external code, will call
with_table_reader()
with_table_reader()
and_then()
with_table_reader()

At each `and_then` we will append to a new list and before falttening, we will add `WaitOn` tranforming between outer lists.

This is currently not implemented to avoid un-necessary complexity.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: `my understanding is that the table completion signal can be used for further processing ` 
That's absolutely right!



`how will it work `:
A small summary of the other code in this file
So the reader transform returns a tuple-tagged output with 2 tags:
1. Tag for `PCollection<SourceTableReference>` - which is the completions
2. Tag for `PCollection<SourceRows>` - which is the rows
The pipeline controller can do `pcs.get(tagForCompeltions)` this gives the PCollection<SourceTableReference> an element is added here only when a table compltes.

So now the pipeline controller needs to do `pcs.get(tagForCompeltions).apply(PTrasnformThatHandlesCompltions<SourceTableReference, T>)`

Now this `PTrasnformThatHandlesCompltions` can do whatever is needed - like say send a pubsub event to live pipeline to consume cdcs, write to spanner for checkpointing, just emit a log 


On sequencing reads between tables, that can be done here too, but it's much easier to do that natively in ReaderTransform as disccused in the previous comment (assuming reader gets the topologically sorted sequence of tables from pipeline controller based on spanner schema interleaving or spanner foreign keys)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: We will need a stage per table, naming it just makes it a bit easier to debug.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceSchemaReference.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.annotation.Nullable;
+
+/** Value class to enclose the database name and PG namespace. */
+@AutoValue
+public abstract class SourceSchemaReference implements Serializable {
+
+  abstract String dbName();
+
+  @Nullable
+  abstract String nameSpace();
```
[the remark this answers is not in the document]
VardhanThigle: Changed in the current push. Thanks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: Thanks for the detailed discussion.
As discussed I will add a note on sequencing and signals (used to potentially trigger event for live pipeline)  in the design.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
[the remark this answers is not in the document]
VardhanThigle: As we discussed over call, resolving this. 
The `AccumulateTableReader ` is aligned with the current architectural choice and we could we swap this to a parallel-read implementation if needed.
I have added a note in the javaDocs of  `ReaderTransofrm.java` to summarize this.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: Will Add a note on stage per table in the main sections of design (currently there's a very passing reference in alternatives considered)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: As discussed, will add a note on this in the design.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SerializableGenericRecord.java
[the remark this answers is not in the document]
VardhanThigle: Have taken a TODO to explore a coder. Resolving for the scope of this PR.

aksharauke (approved):

manitgupta (approved):
This PR definitely raises the bar in terms of code quality and API design! Thanks for putting in the effort. 

Summarizing two things that we need to take care of - 

1. Currently records from multiple tables are flattened into a single PCollection. This seems odd. I will wait for subsequent PRs so see how that is handled, but as per offline discussion with @VardhanThigle, following things should **not** be done by the consumer of the reader's output - 

* Any kind of GPK operations which group the elements into the PCollection into batches via some criteria. 
* Relying on batchWrite APIs ability to handle mutations belonging to different tables in a single batch efficiently. 

2. Exploring coder instead of a class implementing serializable. A TODO has already been added, but it would be great if we can create a task on the Taskflow board we have setup and keep it in the backlog so that we don't forget this later. 


The last thing we should mention in the design, and the README for this template is the nuance related to the number of tables.

VardhanThigle:
Thanks. Manit


> This PR definitely raises the bar in terms of code quality and API design! Thanks for putting in the effort.
> 
> Summarizing two things that we need to take care of -
> 
> 1. Currently records from multiple tables are flattened into a single PCollection. This seems odd. I will wait for subsequent PRs so see how that is handled, but as per offline discussion with @VardhanThigle, following things should **not** be done by the consumer of the reader's output -

Sure. I have added a note in the Javadocs too and will keep this in mind while reviewing changes on transformer/writer.

> * Any kind of GPK operations which group the elements into the PCollection into batches via some criteria.
> * Relying on batchWrite APIs ability to handle mutations belonging to different tables in a single batch efficiently.
> 
> 2. Exploring coder instead of a class implementing serializable. A TODO has already been added, but it would be great if we can create a task on the Taskflow board we have setup and keep it in the backlog so that we don't forget this later.
Yup.

> The last thing we should mention in the design, and the README for this template is the nuance related to the number of tables.
Sure. Thanks.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRowTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Test class for {@link SourceRow}. */
+@RunWith(MockitoJUnitRunner.class)
+public class SourceRowTest extends TestCase {
+  @Test
+  public void testSourceRowBuilds() {
+    final String testTableUUID = "fa14a160-2841-42f2-91ee-7929d791f367";
```
Deep1998, line 32: nit: The testTableUUID can be removed from the tests in all the files.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReaderTest.java
```diff
@@ -0,0 +1,148 @@
...
+    TestAccumulatorResultTracker(
+        TupleTag<SourceRow> sourceRowTupleTag,
+        TupleTag<SourceTableReference> sourceTableReferenceTupleTag) {
+      this.sourceRowTupleTag = sourceRowTupleTag;
+      this.sourceTableReferenceTupleTag = sourceTableReferenceTupleTag;
+    }
+
+    @Override
+    @NonNull
+    public PCollection<TestResultType> expand(@NonNull PCollectionTuple input) {
```
Deep1998, line 92: Can we not do this verification by fetching the tuple elements and verifying counts (using inbuilt count methods) of the 2 pcollections, rather than mapping them to the enum and counting elementwise? Feels more readable that way?

Deep1998 (commented):
LGTM

Based on this contract, the transformer get SourceRow as input.
Regarding the tableSchemaToUUID map, that will be added in a follow up PR correct?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceTableSchema.java
```diff
@@ -0,0 +1,133 @@
...
+    return avroSchema().getField(PAYLOAD_FIELD_NAME).schema();
+  }
+
+  /* TODO(vardhanvthigle@): Add more information as needed by other modules, like:
+   * primary keys, indexing information, foreign key constraints etc.
+   */
+
+  public static Builder builder() {
+    var builder = new AutoValue_SourceTableSchema.Builder();
+    builder.setTableSchemaUUID(UUID.randomUUID().toString());
```
Deep1998, line 93: I also don't see the generation of `tableSchemaUUIDToSourceSchemaReference<>` map. Should'nt that be added in this PR as well?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReaderTest.java
VardhanThigle: Makes a ton of sense.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRowTest.java
VardhanThigle: Thanks. done.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReaderTest.java
VardhanThigle: Fixed in followup https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1449 as this got merged.

PR #1442, opened 2024-04-15
Implement StreamingMode related validations

damondouglas:
This PR adds StreamingMode related validations to guardrail configuration mismatch i.e. `StreamingMode.AT_LEAST_ONCE` when `supportsAtLeastOnce == false`.


andreigurau:
I haven't been following along with the work regarding At Least Once/Exactly Once modes (think @Dippatel98 is more familiar - Dip, could you also take a look?)

Regarding your PR, I think its possible for both `supportsAtLeastOnce = true` and `supportsExactlyOnce = true` to be present on the same TemplateDefinition. If this isn't valid, could you make sure to check for that condition as well?

damondouglas:
> Regarding your PR, I think its possible for both `supportsAtLeastOnce = true` and `supportsExactlyOnce = true` to be present on the same TemplateDefinition. If this isn't valid, could you make sure to check for that condition as well?

The PR already agrees with your case. Nonetheless I added additional tests to assert this valid case.

andreigurau (approved):
LGTM from me, but will wait for Dip to add the Google LGTM label for import

Dippatel98 (approved):

PR #1443, opened 2024-04-16
Install python3.11 on gitactions runners

Polber:
Adds installation of python 3.11 to startup script for gitactions runners

### DIFF
--- a/.github/scripts/configure-runners.sh
+++ b/.github/scripts/configure-runners.sh
@@ -181,7 +181,7 @@ gcloud compute instance-templates create $INSTANCE_TEMPLATE_NAME \
   --image-project=$IMAGE_PROJECT \
   --boot-disk-type=$BOOT_DISK_TYPE \
   --boot-disk-size=$BOOT_DISK_SIZE \
-  --machine-type="MACHINE_TYPE" \
+  --machine-type=$MACHINE_TYPE \
   --scopes=$SCOPE \
   --service-account=${SA_EMAIL} \
   --metadata-from-file=startup-script=startup-script-${NAME_SUFFIX}.sh,shutdown-script=shutdown-script-${NAME_SUFFIX}.sh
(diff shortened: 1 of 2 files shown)


damccorm:
Can we just install it using GitHub's action? like https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/55dcefc6baf0ef135bc3dea3770328dda74fe109/.github/actions/setup-env/action.yml#L53

Polber:
> Can we just install it using GitHub's action? like
> 
> https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/55dcefc6baf0ef135bc3dea3770328dda74fe109/.github/actions/setup-env/action.yml#L53

Does that work on self-hosted machines? I would assume there are permission issues with installing. Either way, this is just a quick fix for a failing test, and follows the pattern of the rest of the startup script. Perhaps refactoring all the startup actions in the future would be better, but I'm not sure that is worth the time investment right now

damccorm:
It does work (that's also how we install java https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/55dcefc6baf0ef135bc3dea3770328dda74fe109/.github/actions/setup-java-env/action.yml#L43 which runs against self-hosted, and we run the actions I linked as part of the release workflow - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/55dcefc6baf0ef135bc3dea3770328dda74fe109/.github/workflows/release.yml#L67).

> follows the pattern of the rest of the startup script. 

I don't think we install any languages here.

-----------------------

I'd vote we update to use setup-python next to setup-go in [setup-env/action.yml](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/55dcefc6baf0ef135bc3dea3770328dda74fe109/.github/actions/setup-env/action.yml#L53) in order to be consistent/have a single place where we configure language versions for CI. Its also less custom code to take care of and will handle all your things like env vars for you.

PR #1447, opened 2024-04-16
Update regex to make non-alphanumeric values optional in date-time patterns

jrmccluskey:
Legacy Pub/Sub to Text on Cloud Storage and Pub/Sub to Avro templates had regex that did not actually allow the default values listed and required a non-alphanumeric character at the front and back of the format string instead of being optional as stated. This updates the regex to allow forms like `YYYY` and `MM` without single quotes or other valid bracketing values.


jrmccluskey:
R: @Polber

PR #1455, opened 2024-04-18
fixed escape characters and timer type

aksharauke:
This PR does the following:
1. The \ is correctly escaped
2. The timer is set to processing time since the event time is not relevant


shreyakhajanchi (approved):
LGTM

aksharauke:
@krishnamoorthy-r do I need the Migrations override for this? I have added unit tests and there is increase in reverse replication coverage 
image
Note: the UTs will run locally. The flakiness fix will be done via different bug.

PR #1413, opened 2024-04-04
removed deprecated reverse replication templates

aksharauke:
removed deprecated reverse replication templates

PR #1456, opened 2024-04-18
Add parentName and parentTriggerValues attributes for templates

Amar3tto:


v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
```diff
@@ -182,6 +182,8 @@ public interface ExportPipelineOptions extends PipelineOptions {
     @TemplateParameter.Text(
         order = 10,
         optional = true,
+        parentName = "shouldExportRelatedTables",
+        parentTriggerValues = {"true"},
```
an2x, line 186: I think this needs to be reverted.  I don't know this template well, but the `shouldExportRelatedTables` parameter description says "If `tableNames` is specified but doesn't include related tables, this option must be set to true for a successful export." Which means in some cases `tableNames` parameter can be set to something with `shouldExportRelatedTables`  set to false. So we should show `tableNames` unconditionally.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -219,6 +219,8 @@ public interface PubSubToRedisOptions
     @TemplateParameter.Long(
         order = 8,
         optional = true,
+        parentName = "redisSinkType",
+        parentTriggerValues = {"HASH_SINK", "LOGGING_SINK"},
         description = "Hash key expiration time in sec (ttl)",
         helpText =
             "Key expiration time in sec (ttl, default for HASH_SINK is -1 i.e. no expiration)")
```
an2x, line 226: Can you please also update the description and mention this parameter applies only to HASH_SINK and LOGGING_SINK sink types? It's true, based on the implementation, but this is not clear from the parameter description (it won't be visible in the UI once we add parentTriggerValues anyway, but we use the parameter description to auto-generate docs in some other places).

an2x (changes requested):

aksharauke (commented):
The Spanner impacted templates will be removed once the PR is merged: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1413
So the changes can be skipped.

akashorabek:
R: @an2x

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
akashorabek: Done

akashorabek:
> The Spanner impacted templates will be removed once the PR is merged: #1413 So the changes can be skipped.

@aksharauke, removed the changes from the files that will be deleted after [this PR](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1413) will be merged. Pls approve the changes.

an2x (approved):

damondouglas (changes requested):
I see a spotless check error: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/8788623511/job/24255216239?pr=1456#step:4:852

akashorabek:
@damondouglas made a commit. Can you check again pls?

akashorabek:
@damondouglas these test failures are new and was caused after [this PR](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1453) I guess.

PR #1463, opened 2024-04-22
Add inspect variant to dlp template IT

Polber:


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dlp/DlpResourceManager.java
```diff
@@ -34,6 +35,11 @@ public class DlpResourceManager implements ResourceManager {
 
   private static final Logger LOG = LoggerFactory.getLogger(DlpResourceManager.class);
 
+  private enum DlqTemplate {
```
liferoad, line 38: do you mean DlpTemplate?

Polber: fixed.

liferoad (approved):

PR #1465, opened 2024-04-22
Allow custom endpoints for Spanner ITs

Polber:
Allows custom endpoints to be specified in Spanner IT's and adds a suite of tests that use a staged spanner instance.

### DIFF
--- a/cicd/internal/flags/it-flags.go
+++ b/cicd/internal/flags/it-flags.go
@@ -30,6 +30,7 @@ var (
 	dStageBucket         string
 	dHostIp              string
 	dPrivateConnectivity string
+	dSpannerHost         string
 	dReleaseMode         bool
 	dRetryFailures       string
 	dCloudProxyHost      string
@@ -46,6 +47,7 @@ func RegisterItFlags() {
 	flag.StringVar(&dStageBucket, "it-stage-bucket", "", "(optional) A GCP bucket to stage templates")
 	flag.StringVar(&dHostIp, "it-host-ip", "", "(optional) The ip that the gitactions runner is listening on")
 	flag.StringVar(&dPrivateConnectivity, "it-private-connectivity", "", "(optional) A GCP private connectivity endpoint")
+	flag.StringVar(&dSpannerHost, "it-spanner-host", "", "(optional) A custom endpoint to override Spanner API requests")
 	flag.BoolVar(&dReleaseMode, "it-release", false, "(optional) Set if tests are being executed for a release")
 	flag.StringVar(&dRetryFailures, "it-retry-failures", "0", "Number of retries attempts for failing tests")
 	flag.StringVar(&dCloudProxyHost, "it-cloud-proxy-host", "10.128.0.34", "Hostname or IP address of static Cloud Auth Proxy")
@@ -90,6 +92,13 @@ func PrivateConnectivity() string {
 		return "-DprivateConnectivity=" + dPrivateConnectivity
 	}
 	return ""
+}
+
+func SpannerHost() string {
+	if dSpannerHost == "" {
+		return "-DspannerHost=" + "https://staging-wrenchworks.sandbox.googleapis.com/"
+	}
+	return "-DspannerHost=" + dSpannerHost
 }
 
 func FailureMode() string {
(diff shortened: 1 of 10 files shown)


.github/workflows/java-pr.yml
```diff
@@ -163,7 +164,8 @@ jobs:
           --it-region="us-central1" \
           --it-project="cloud-teleport-testing" \
           --it-artifact-bucket="cloud-teleport-testing-it-gitactions" \
-          --it-private-connectivity="datastream-private-connect-us-central1"
+          --it-private-connectivity="datastream-private-connect-us-central1" \
+          --it-spanner-host="https://staging-wrenchworks.sandbox.googleapis.com/"
```
manitgupta, line 168: Why is this needed here?
The template just has a parameter which can be used to override the spanner host right?

manitgupta (commented):
Could you also make the same changes to the DatastreamToSpanner template IT as well?
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java

This template had the staging tests present in the internal python based UIF tests. It will be good to have parity! Thanks!

Polber:
> Could you also make the same changes to the DatastreamToSpanner template IT as well? https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
> 
> This template had the staging tests present in the internal python based UIF tests. It will be good to have parity! Thanks!

That is on my agenda of things to accomplish this week, but it will require a bit more effort since Datastream tests aren't stable yet, so that will be in a separate PR. This PR migrates all Spanner tests to use staging other than the Datastream suite unless if I'm not mistaken

cicd/internal/flags/it-flags.go
```diff
@@ -82,6 +84,13 @@ func PrivateConnectivity() string {
 	return ""
 }
 
+func SpannerHost() string {
+	if dSpannerHost == "" {
```
damccorm, line 88: Formatting looks funky here - could you run `go fmt` on the files here? Probably why the check is failing

damccorm (approved):
LGTM, but I'll defer to @manitgupta and other folks from the spanner team on this one

manitgupta (approved):
Thanks! This looks much cleaner. 👍

cicd/internal/flags/it-flags.go
```diff
@@ -84,6 +86,13 @@ func PrivateConnectivity() string {
...
+		return "-DspannerHost=" + "https://staging-wrenchworks.sandbox.googleapis.com/"
```
manitgupta, line 91: Does this make the staging instance as the default host?

Also, I don't see the changes in the Github actions that will be required to pass the `spannerHost`, will you be doing that in a separate PR?

I don't understand how the same test is being executed for both a prod and the staging Spanner host.

Polber: @manitgupta First of all, thanks for pointing that out because you made me realized I never actually call that method in the workflows (fixed).

Second of all, take the following test (part of this PR) as an example:
```
@Test
  public void testSpannerToGCSAvroStaging() throws IOException {
    spannerResourceManager =
        SpannerResourceManager.builder(testName, PROJECT, REGION, Dialect.GOOGLE_STANDARD_SQL)
            .maybeUseStaticInstance()
            .maybeUseCustomHost()
            .build();
    testSpannerToGCSAvroBase(
        paramAdder ->
            paramAdder.addParameter("spannerHost", spannerResourceManager.getSpannerHost()));
  }
```

Notice how it instantiates a separate resource manager instead of using a top-level one like the other tests. Also notice it calls `.maybeUseCustomHost()` - this method (defined in the PR) will use the custom host, if one is provided on the command line with `-DspannerHost=...`. 

We do the same thing for `maybeUseStaticInstance()` which was the ideology I was trying to follow as I like how it doesn't fail the test should a user run who can't access the endpoint (assuming the `-DspannerHost` flag was not passed).

Polber:
Test failures unrelated

PR #1471, opened 2024-04-24
Add groupName attribute where it's needed

akashorabek:


akashorabek:
@an2x I added the value Source or Target for the attribute groupName for those template parameters that are obvious from the method name or description. For parameters that are not obvious, I decided not to add the groupName attribute. Could you check if they are correctly specified?

Amar3tto:
PR was split into 3 parts:
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1539
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1541
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1542

PR #1474, opened 2024-04-24
Add scheduled workflows for test suites

Polber:
This PR will set up various automated test workflow for tracking test health through the use of status badges on the main repo page. Tests will be scheduled to run every 6 hours and test reports will be attached to each run similar to Java PRs.

### DIFF
--- a/.github/workflows/java-pr.yml
+++ b/.github/workflows/java-pr.yml
@@ -27,8 +27,6 @@ on:
       # This will make it easier to verify action changes don't break anything.
       - '.github/actions/setup-env/*'
       - '.github/workflows/java-pr.yml'
-  schedule:
-    - cron: "0 */12 * * *"
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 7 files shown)


damccorm (commented):
Are these just identical tests to the existing suites? Can we just set up schedule triggers on the existing suites so that we're not duplicating code?

.github/workflows/java-pr.yml
```diff
@@ -27,8 +27,6 @@ on:
       # This will make it easier to verify action changes don't break anything.
       - '.github/actions/setup-env/*'
       - '.github/workflows/java-pr.yml'
-  schedule:
```
Abacn, line 30: same as comments of #1475, generally we would like to have same workflow run for different triggers. Any reason this didn't work?

manitgupta:
Are we not going this anymore?

manitgupta:
Nevermind I saw the new PR

PR #1475, opened 2024-04-24
Add junit report to release workflow

Polber:


.github/workflows/release-report.yml
```diff
@@ -0,0 +1,42 @@
...
+# limitations under the License.
+
+# Checks that are intended to publish PRs test results from JUnit.
+
+name: Java PR Report
+on:
+  workflow_run:
+    workflows:
+      - Release
+    types: [ completed ]
```
damccorm, line 22: Is there a reason to have this as a separate workflow? Can we just have this as a step (or separate) at the end of the release?

Its not clear to me what we're trying to accomplish here, but it seems like this would be more valuable as a status on the release run itself rather than a different workflow

damccorm: If we want to reuse these steps across multiple workflows (which seems like the right thing to do), we should make this a composite action - https://docs.github.com/en/actions/creating-actions/creating-a-composite-action - which can be called by both the pr and release workflows. I don't think we want a separate workflow to do this at all.

In practice, this is just 2 steps so I wouldn't even bother, I'd just add these actions to both workflows directly

PR #1478, opened 2024-04-25
Add @Template config parameter to disable releasing a template

Polber:
Allow a template marked with `@Template` to be omitted from staging and release by setting the `testOnly` parameter to `true`.

This also will unblock the release that is failing due to trying to stage and release https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/4551dda65ee7dff0bbc73d8648d5525dadd2bd2f/plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/model/TemplateDefinitions.java#L550-L555

which was added in #1442


Dippatel98 (approved):

damccorm (approved):
LGTM, thanks

PR #1481, opened 2024-04-27
Jdbc table schema discovery

VardhanThigle:
Adding code for JDBC schema discovery from information system tables.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/dialectadapter/dialectadapter/mysql/MysqlDialectAdapter.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/dialectadapter/dialectadapter/mysql/MysqlDialectAdapter.java
@@ -1,0 +1,216 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.source.reader.io.jdbc.dialectadapter.dialectadapter.dialectadapter.mysql;
+
+import com.google.cloud.teleport.v2.source.reader.io.exception.exception.RetriableSchemaDiscoveryException;
+import com.google.cloud.teleport.v2.source.reader.io.exception.exception.SchemaDiscoveryException;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.dialectadapter.dialectadapter.dialectadapter.DialectAdapter;
+import com.google.cloud.teleport.v2.source.reader.io.schema.SourceSchemaReference;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SourceColumnType;
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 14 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscovery.java
```diff
@@ -0,0 +1,44 @@
...
+package com.google.cloud.teleport.v2.source.reader.io.schema;
+
+import com.google.cloud.teleport.v2.source.reader.io.exception.exception.SchemaDiscoveryException;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SourceColumnType;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import javax.sql.DataSource;
+
+/** Discover Schema of Source Tables. */
+public interface SchemaDiscovery {
```
aksharauke, line 25: Is there a use case when someone would not want to do a retry ? Can we expose only the Retriable interface and the degenerate case be no-retries?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/dialectadapter/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -0,0 +1,208 @@
...
+   *     reading information schema table directly, since:
+   *     <ol>
+   *       <li>With all other things equal, using a prepared statement to read information schema
+   *           would be faster than making individual calls.
+   *       <li>Various drivers have different properties that can affect the search scope of the
+   *           calls.
+   *     </ol>
+   */
+  @Override
+  public ImmutableMap<String, ImmutableMap<String, SourceColumnType>> discoverTableSchema(
```
aksharauke, line 65: Is it in plans to also expose additional schematic information - specifically foreign key constraints?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+                        .withExponent(1)
+                        .withMaxRetries(testRetryCount))
+                .discoverTableSchema(mockDataSource, mockSourceSchemaReference, ImmutableList.of()))
+        .isEqualTo(ImmutableMap.of());
+    verify(mockRetriableSchemaDiscovery, times(expectedCallsCount))
+        .discoverTableSchema(any(), any(), any());
+  }
+
+  @Test
+  public void testSchemaDiscoveryImplThrowsRetiesExhausted()
```
aksharauke, line 74: nit: Retries* in the function name

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscovery.java
VardhanThigle: Actually the Schema Discovery Implementation wraps the retriable interface and does the necessary retries.
The outside code will always use `SchemaDiscovery` which never throes a retriable exception.

This was done to make the retry handling logic common and avoiding the need to implement the same for say PG in future.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/dialectadapter/dialectadapter/mysql/MysqlDialectAdapter.java
VardhanThigle: In future yes. The dialect adapter can expose this information. Also for string splitting, we will need to provide range queries to JDBCIO (by default it always assumes an increasing order of the index), these range query prepare statements can be created by the dialect adapter.

aksharauke (approved):

PR #1498, opened 2024-05-01
Add random suffix to resource IDs to avoid collision

Polber:


Abacn (approved):
thank you!

let me add the label now and then approve it after test run

PR #1499, opened 2024-05-01
Xlang Dockerfile speed improvements

Polber:
Add same speed improvements that were added to YamlTemplate in #1405. This will greatly reduce startup time, remove need to specify requirements.txt in every module that uses Python UDF's (using root pom.xml for versioning beam), and is one step closer to consolidating Yaml Dockerfile and Xlang Dockerfile (will need another change on Beam before merging).


damccorm (approved):
LGTM, @fozzie15 could you please take a look as well?

fozzie15 (commented):
LGTM

v2/googlecloud-to-elasticsearch/src/main/resources/requirements.txt
```diff
@@ -1 +0,0 @@
-apache-beam[gcp,aws, azure, dataframe]==2.55.1
```
damccorm, line 1: Looks like these are now causing merge conflicts, should be an easy resolution

PR #1500, opened 2024-05-02
Add jinja preprocessing to YamlTemplate

Polber:
Adds ability to perform jinja preprocessing to Beam YAML jobs run on YamlTemplate until the functionaiity is added to vanilla Beam YAML in Beam 2.47.0

Adds template parameter `--jinja_variables `

### DIFF
--- a/python/src/main/python/yaml-template/main.py
+++ b/python/src/main/python/yaml-template/main.py
@@ -12,23 +12,96 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
-#
+
 
 import argparse
-import logging
+import json
 
+import jinja2
+import yaml
+
+import apache_beam as beam
+from apache_beam.io.filesystems import FileSystems
+from apache_beam.typehints.schemas import LogicalType
+from apache_beam.typehints.schemas import MillisInstant
 from apache_beam.yaml import cache_provider_artifacts
-from apache_beam.yaml import main
+from apache_beam.yaml import yaml_transform
+
+# Workaround for https://github.com/apache/beam/issues/28151.
+LogicalType.register_logical_type(MillisInstant)
+
+
+def _configure_parser(argv):
+  parser = argparse.ArgumentParser()
+  parser.add_argument(
+    '--yaml_pipeline',
+    '--pipeline_spec',
+    help='A yaml description of the pipeline to run.')
+  parser.add_argument(
+    '--yaml_pipeline_file',
+    '--pipeline_spec_file',
+    help='A file containing a yaml description of the pipeline to run.')
+  parser.add_argument(
+    '--json_schema_validation',
+    default='generic',
+    help='none: do no pipeline validation against the schema; '
+         'generic: validate the pipeline shape, but not individual transforms; '
+         'per_transform: also validate the config of known transforms')
+  parser.add_argument(
+    '--jinja_variables',
+    default=None,
+    type=json.loads,
+    help='A json dict of variables used when invoking the jinja preprocessor '
+         'on the provided yaml pipeline.')
+  return parser.parse_known_args(argv)
+
+
+def _pipeline_spec_from_args(known_args):
+  if known_args.yaml_pipeline_file and known_args.yaml_pipeline:
+    raise ValueError(
+      "Exactly one of yaml_pipeline or yaml_pipeline_file must be set.")
+  elif known_args.yaml_pipeline_file:
+    with FileSystems.open(known_args.yaml_pipeline_file) as fin:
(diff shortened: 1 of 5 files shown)


python/src/main/python/yaml-template/main.py
```diff
@@ -13,18 +13,52 @@
...
+
+
+def run(argv=None):
+  known_args, pipeline_args = _configure_parser(argv)
+  known_args = Namespace(
+    **vars(main._configure_parser(argv)[0]), **vars(known_args))
+  pipeline_yaml = (
+    jinja2.Environment(
+      undefined=jinja2.StrictUndefined, loader=_BeamFileIOLoader())
+    .from_string(main._pipeline_spec_from_args(known_args))
```
damccorm, line 54: If we're already reading this as a string, why do we need the _BeamFileIOLoader?

python/src/main/python/yaml-template/main.py
```diff
@@ -13,18 +13,52 @@
...
+    .render(**known_args.jinja_variables or {}))
+  pipeline_yaml = os.linesep.join(
+    [s for s in pipeline_yaml.splitlines() if s.strip()])
```
damccorm, line 57: One downside here is that we lose the original parameters in our pipeline options/display data. I think this is fine, probably a no-op for this PR, but left a comment in https://github.com/apache/beam/pull/30976/files#r1588005108

damccorm (commented):
Comments are minor, overall looks good to me.

cc/ @robertwb

python/src/main/python/yaml-template/main.py
```diff
@@ -13,18 +13,52 @@
...
+
+  pipeline_args += [
+    '--sdk_location=container', f'--yaml_pipeline={pipeline_yaml}'
```
robertwb, line 60: Why are we overrideing sdk_location?

robertwb, line 60: Won't we need to remove any `--yaml_pipeline` or --yaml_pipeline_file` arguments as well as append this one?

python/src/test/resources/YamlTemplateIT.yaml
```diff
@@ -42,21 +42,13 @@ pipeline:
...
-
-# TODO(polber) - remove with https://github.com/apache/beam/pull/30777
-providers:
-  - type: python
-    config:
-      packages: []
-    transforms:
-      'WriteToJsonPython': 'apache_beam.io.WriteToJson'
\ No newline at end of file
+        path: {{ OUTPUT_PATH_PARAM }}/bad
```
robertwb, line 54: nig: Trailing newline?

python/src/main/python/yaml-template/main.py
damccorm: We're not passing through `known_args` so I think this is fine, though I'm not actually sure why that is now that I think of it

Polber: Recommended in https://cloud.google.com/dataflow/docs/guides/build-container-image#preinstall_using_a_dockerfile since sdk is already pre-installed in image. Supposedly speeds up startup, but not sure how significantly

Polber: +1
Because the original ones are already getting processed so we don't need to reprocess

Polber: Actually, I did not consider the `--json_schema_validation` flag... maybe I should refactor to just omit the original `yaml_pipeline` and `yaml_pipeline_file` flags

python/src/main/python/yaml-template/main.py
Polber: Addressed in https://github.com/apache/beam/pull/30976#discussion_r1566324673
Essentially, allows a user to specify external files in the jinja template that are on GCS instead of restricting to local filesystem since that wouldn't be very useful in the template launcher...

python/src/test/resources/YamlTemplateIT.yaml
Polber: Fixed.

python/src/main/python/yaml-template/main.py
damccorm: Got it

Polber:
@damccorm @robertwb After reconsidering, I decided just to copy over the main from https://github.com/apache/beam/pull/30976 for the time being. There were too many issues involving overridden pipeline args, and it makes more sense to me to just copy over the same logic that will be in Beam 2.57.0 instead of trying to stitch pieces together. Once 2.57.0 is release, assuming the jinja PR is merged, I will revert the template back to how it was before.

damccorm (approved):

Polber:
> Could we add a GitHub issue to track the work to remove these pieces before proceeding? Otherwise LGTM

#1554

PR #1483, opened 2024-04-29
Add Kafka to Kafka Template

Surabhiad:

PR #1504, opened 2024-05-02
Upgrade beam version to 2.56.0

Abacn:


Abacn:
R: @damccorm

damccorm (approved):
Thanks!

Abacn:
the second run has only one test failing due to #1483, which also failing on main branch

PR #1505, opened 2024-05-02
Add custom App Profile support for Bigtable to Vector Embedding export

ron-gal:


Polber (approved):
Thanks! Will wait on checks to pass before adding LGTM label

PR #1511, opened 2024-05-03
Add support for binary AVRO messages in non confluent wire format

Dippatel98:


AnandInguva (approved):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/BinaryAvroDeserializer.java
```diff
@@ -35,7 +35,7 @@ public BinaryAvroDeserializer(Schema schema) {
     this.schema = schema;
   }
 
-  public GenericRecord deserialize(String topic, Headers header, byte[] bytes) {
+  public Object deserialize(String topic, Headers header, byte[] bytes) {
```
AnandInguva, line 38: can we typecast in this method instead and return GenericRecord?

Dippatel98: We need to use `Object` to override parent methods.

PR #1513, opened 2024-05-04
Add GCS_READ_BUCKET type to the template parameter type

akashorabek:
Add GCS_READ_BUCKET type to the template parameter type


akashorabek:
R: @an2x

damccorm (commented):
Just had one comment which applies across the board, otherwise this looks good. I'll hold off on reviewing similar PRs until that's addressed, I think we should be able to get them in pretty quickly at that point

damccorm (approved):
Thanks!

PR #1517, opened 2024-05-06
Add IT for sharded migration without migration_shard_id

manitgupta:
This IT covers the scenario where multiple Dataflow jobs are writing to the same Spanner database without using the `migration_shard_id` column. 

PK uniqueness will be guaranteed by the customer. The end result should be that data from all the shards should be processed and written to Spanner.


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithoutMigrationShardIdColumnIT.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+    ResourceManagerUtils.cleanResources(spannerResourceManager, pubsubResourceManager);
+  }
+
+  @Test
+  public void multiShardMigration() {
+    // Two dataflow jobs are running corresponding to two physical shards containing two logical
+    // shards each. Migrates Users table from 4 logical shards. Asserts data from all the shards are
+    // going to Spanner. Checks whether migration shard id column is populated properly based on the
+    // transformation context.
```
darshan-sj, line 142: "Checks whether migration shard id column is populated properly based on the transformation context."
- Stale comment

darshan-sj (approved):
LGTM

PR #1521, opened 2024-05-06
Add scheduled dispatch workflows for test suites

Polber:
PR adds various workflows to be able to run individual test suite(s) against a branch, as well as run all the workflows on a 6 hour interval for observability.

This PR does not affect existing workflows - it only adds new ones.


.github/actions/run-integration-tests/action.yml
```diff
@@ -0,0 +1,58 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Setup Environment
+      id: setup-env
+      uses: ./.github/actions/setup-env
+
+    - name: Run Build
+      shell: bash
+      run: ./cicd/run-build --changed-files="${{ inputs.changed-files }}"
```
damccorm, line 38: Should this just shell out to the `run-build` action?

damccorm: Same question applies elsewhere

.github/workflows/java-build.yml
```diff
@@ -0,0 +1,51 @@
...
+    timeout-minutes: 60
+    runs-on: [ self-hosted, it ]
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@1e31de5234b9f8995739874a8ce0492dc87873e2 # v4.0.0
+        with:
+          ref: ${{ inputs.branch }}
+          fetch-depth: 2  # setup-env action assumes fetch depth of at least 2 for https://github.com/tj-actions/changed-files?tab=readme-ov-file#on-push-%EF%B8%8F
+      - name: Run Java Build
```
damccorm, line 50: Having a build workflow seems less useful since this will be done by all the other workflows

.github/workflows/java-integration-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Checks that are intended to run on PRs containing Java code.
+
+name: Java Integration Tests
```
damccorm, line 17: Taking a step back, what is your goal with these workflows? The Java PR workflow already can be run manually, why is having separate workflows better?

It seems like maybe the goal is to have a more granular signal? Or to allow individual pieces to be run independently of the main workflow?

I think there might be better ways to accomplish this - having 2 workflow files that do more or less the same thing is somewhat of an antipattern (and is just expensive). Rather than doing this, maybe we can restructure the existing java pr workflow to have some conditional logic. For example:

1) Always run everything on schedule
2) Run everything after build in parallel instead of serially
3) Insert if conditions on the job that are basically like: `if: not is(workflow_dispatch) || input.run_its` (similar to https://github.com/apache/beam/blob/ea3c2a4dbbe20a8e7161d362102bb1753bd60bba/.github/workflows/build_release_candidate.yml#L25)

damccorm: I'm generally supportive of moving things to composite actions where helpful, but I don't like doubling our workflow count, especially since we'd now need to do more work to keep different workflows in sync.

PR #1525, opened 2024-05-07
Switch DatastreamToSpannerDataTypeIT from JSON to Avro

manitgupta:
The current ITs use JSON data from Datastream which is not used in actual migrations. 

Switching ITs to use avro files from Datastream instead of JSON to more closely resemble a production scenario. This involves generating all the test data and verifying that assertions are correct.


darshan-sj (approved):
LGTM

PR #1526, opened 2024-05-07
Add concurrency group to java-pr workflow

manitgupta:
Added concurrency so that jobs for a PR belonging to the `java-pr-${PR_NUMBER}` concurrency group get cancelled when a new execution is requested.

A new execution would be requested when new commits were added to the PR. 

Adding the `event.number` to the concurrency group name will ensure that resource isolation and in-progress job wounding will happen at a per-PR level, and each PR is treated independently.

Reference - https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-and-the-default-behavior

### DIFF
--- a/.github/workflows/java-pr.yml
+++ b/.github/workflows/java-pr.yml
@@ -32,7 +32,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: java-pr
+  group: java-pr-${{ github.event.number }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/java-pr.yml
+++ b/.github/workflows/java-pr.yml
@@ -32,7 +32,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: java-pr-${{ github.event.number }}
+  group: java-pr-${{ github.event.issue.number || github.run_id }}
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/java-pr.yml
```diff
@@ -31,6 +31,10 @@ on:
     - cron: "0 */12 * * *"
   workflow_dispatch:
 
+concurrency:
+  group: java-pr-${{ github.event.number }}
```
damccorm, line 35:
```suggestion
  group: java-pr-${{ github.event.issue.number || github.run_id }}
```

Could we do something like this instead? This will only impact pull request runs (will limit to one run per pr), but will not interfere with scheduled or manual runs (which I don't think we should cancel)

manitgupta: I think if we add a `run_id` here then this will not work as intended since multiple runs of the same PR will continue to run in parallel? 

`run_id` is effectively the thing we want to run "one at a time" I think.

damccorm: `github.event.issue.number || github.run_id` will first check if there is an issue number, which will be true for all PRs. So PRs will not use `run_id`

damccorm: See https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

manitgupta: Perfect, thanks for the ref. Making the change!

damccorm (approved):

PR #1531, opened 2024-05-07
Mark YamlTemplate as GA

Polber:


python/src/main/java/com/google/cloud/teleport/templates/python/YAMLTemplate.java
```diff
@@ -24,12 +24,21 @@
     name = "Yaml_Template",
     category = TemplateCategory.GET_STARTED,
     type = Template.TemplateType.YAML,
-    displayName = "YAML Template (Experimental)",
+    displayName = "YAML Template",
```
damccorm, line 27: Can we avoid all the append logic by just making this `YAML`? Is there any reason not to do that?

Polber: No, I just didn't like how it looked on the `@Template` tag, but idc that much - fixed.

damccorm (approved):
LGTM

PR #1535, opened 2024-05-09
Switch EventsIT and SessionIT from JSON to Avro

manitgupta:
Switches `DatastreamToSpannerEventsIT` and `DatastreamToSpannerSessionIT` to use `avro` records instead of JSON.


darshan-sj (approved):
LGTM

PR #1538, opened 2024-05-09
added unit tests to Spanner transaction writer

shreyakhajanchi:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFnTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    SpannerTransactionWriterDoFn spannerTransactionWriterDoFn =
+        new SpannerTransactionWriterDoFn(
+            spannerConfig, ddlView, schema, null, "shadow", "mysql", false, true);
+    spannerTransactionWriterDoFn.setMapper(mapper);
+    spannerTransactionWriterDoFn.setChangeEventSessionConvertor(changeEventSessionConvertor);
+    spannerTransactionWriterDoFn.setSpannerAccessor(spannerAccessor);
+    spannerTransactionWriterDoFn.processElement(processContextMock);
+
+    verify(processContextMock, times(1)).output(any(com.google.cloud.Timestamp.class));
```
darshan-sj, line 153: Can we do some assert on the transaction which got created? Otherwise, this is just testing processElement doesn't throw any exceptions, which I think gives a false sense of coverage.

darshan-sj: It can be simple like, spannerAccessor.run is called with a transaction.   ChangeEventSequenceFactory.createChangeEventSequenceFromShadowTable is called 
etc.

shreyakhajanchi: Makes sense, added PTAL

darshan-sj (approved):
Looks good

PR #1543, opened 2024-05-10
Convert sharded migration ITs to use Avro

manitgupta:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
```diff
@@ -270,80 +282,80 @@ private void assertUsersTableContents() {
     row.put("migration_shard_id", "L1");
     events.add(row);
 
-    row = new HashMap<>();
-    row.put("id", 2);
-    row.put("name", "Tester2");
-    row.put("age_spanner", 33);
+    row.clear();
```
darshan-sj, line 285: events.add(row) adds pointer to the row hashmap into the list
row.clear() clears the map and populates new data to the same row object
and then events.add(row) adds pointer to the same hashmap into the list

At the end the list will contain n hashmaps all pointing to the same row hashmap and the values will be of the last row. 
The earlier way row=new HashMap<>() is the correct way of doing this.

darshan-sj (changes requested):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
manitgupta: Then why does it work? 
The assertion should fail then right? 🤔

darshan-sj: My guess is the assert structs method iterates over these row objects one by one and checks if it is present in the structs returned from spanner. So, it is checking if the last row hashmap object exists in struct n times.

manitgupta:
You're right! I will fix this and other tests where I've made the same
mistake!
Thank you so much!

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithoutMigrationShardIdColumnIT.java
```diff
@@ -263,21 +287,9 @@ private void assertUsersTableContents() {
...
-    row.put("age", 14);
-    events.add(row);
-
-    row = new HashMap<>();
-    row.put("id", 15);
-    row.put("name", "Tester15");
-    row.put("age", 15);
+    row.put("id", 16);
+    row.put("name", "Tester16");
+    row.put("age", 116);
```
darshan-sj, line 292: id 12 needs to be asserted?

darshan-sj (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithoutMigrationShardIdColumnIT.java
manitgupta: Done

darshan-sj (approved):
LGTM

PR #1545, opened 2024-05-10
Add spanner module build override based on changeset for cicd

manitgupta:
This is a hack to only enable module only builds for spanner modules. 

Currently all modules get built which cause significant increase in the time for build and IT workflows to complete.

### DIFF
--- a/cicd/internal/op/maven.go
+++ b/cicd/internal/op/maven.go
@@ -17,6 +17,7 @@
 package op
 
 import (
+	"slices"
 	"strings"
 )
 
@@ -28,8 +29,14 @@ func RunMavenOnPom(pom string, cmd string, args ...string) error {
 	fullArgs = append(fullArgs, strings.Split(cmd, " ")...)
 	fullArgs = append(fullArgs, "-f", pom)
 	fullArgs = append(fullArgs, "-e")
+	//filter out modules parameter if it is not specified. This has to be filtered out this
+	//way because we use variable length args, and returning "" for changed modules also means
+	//an empty value in the args. This empty value ends up creating an extra blank space
+	//in the final maven command constructed in RunCmdAndStreamOutput.
+	//Instead of trimming spaces, deleting the empty arg so that it is not considered in mvn command generation
+	//is less error prone.
+	args = slices.DeleteFunc(args, func(s string) bool { return s == "ALL"})
 	fullArgs = append(fullArgs, args...)
-
 	return RunCmdAndStreamOutput("mvn", fullArgs)
 }
 
@@ -39,5 +46,6 @@ func RunMavenOnPom(pom string, cmd string, args ...string) error {
 func RunMavenOnModule(pom string, cmd string, module string, args ...string) error {
 	// fullArgs := []string{"-pl", module}
 	// fullArgs = append(fullArgs, args...)
+	args = slices.DeleteFunc(args, func(s string) bool { return s == "ALL"})
 	return RunMavenOnPom(pom, cmd, args...)
 }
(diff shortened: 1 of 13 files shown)


damccorm:
Can we give this a day or two to understand why this was disabled in the first place (@Polber)? I'm hoping we can just make a simple fix so we don't need this change

Polber:
@damccorm We used to build only affected modules, but this led to frequent race conditions where not all dependencies would get built - instead of adding custom logic each time this happened (which seemed to be increasing in frequency each passing week), we decided to just build all modules on each PR.

Perhaps there is a "better" fix (maybe building all common modules, but only running IT's on affected modules) but I don't think it is necessarily a simple fix

cicd/internal/op/maven.go
```diff
@@ -36,8 +43,14 @@ func RunMavenOnPom(pom string, cmd string, args ...string) error {
 // Runs the given Maven command on a specified module. Considering the input, this is equivalent to:
 //
 //	mvn -B {cmd} -f {pom} -pl {module} {args...}
-func RunMavenOnModule(pom string, cmd string, module string, args ...string) error {
-	// fullArgs := []string{"-pl", module}
-	// fullArgs = append(fullArgs, args...)
-	return RunMavenOnPom(pom, cmd, args...)
+func RunMavenOnModule(pom string, cmd string, module []string, args ...string) error {
+	fullArgs := args
+	for _, spanner_module := range SPANNER_MODULES {
```
damccorm, line 48: I don't really like this change as it stands because it has the potential to filter out real failures in non-spanner templates. For example, imagine I make a change in 2 templates - one spanner, one non-spanner (this is common in large-scale full-repo changes, or when doing something like regenerating docs as part of a PR) - the tests will not run on the non-spanner template. We also don't seem to have a clear path to fix the underlying problem, so this hack will remain in place indefinitely.

I'd like to propose a different path:

1) Get rid of the whole concept of only building changed files (basically greatly simplify https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0952eecc097a06ffcc4977927311edfbe5a8e2bc/cicd/internal/workflows/maven-workflows.go#L204)
2) Instead, allow individual workflows to pass in a set of modules to run against, such that you can trigger tests with `./cicd/run-unit-tests --changed-modules="v2/datastream-to-spanner,v2/sourcedb-to-spanner,v2/spanner-change-streams-to-sharded-file-sink"`. If helpful, this could probably even accept aliases (e.g. `--changed-modules="spanner_all"`)
3) Set up a spanner-only GitHub Actions workflow which passes in these modules and only triggers on spanner changes.
4) Exclude spanner from https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0952eecc097a06ffcc4977927311edfbe5a8e2bc/.github/workflows/java-pr.yml#L24

Thoughts?

manitgupta: Having some issues in getting the build to work in isolation - 

The following command is generated - 

```
mvn -B clean install -f pom.xml -e -Dmdep.analyze.skip -Djib.skip -DskipShade -Dspotless.check.skip -Dmaven.test.skip -T3 -pl=v2/datastream-to-spanner/,v2/spanner-common/,v2/spanner-change-streams-to-sharded-file-sink/,v2/gcs-to-sourcedb/,v2/spanner-migrations-sdk/,v2/spanner-custom-shard/,v2/sourcedb-to-spanner/
```

This works locally, however fails on Github actions with - 

https://paste.googleplex.com/6486993152507904

I tried adding the `common` module as a dependency to `datastream-to-spanner` but that does not seem to help. 
Do you have any idea on how to solve this?

manitgupta:
Closing in favour of #1573

PR #1549, opened 2024-05-13
optimized the processing for lean Spanner write intervals

aksharauke:
The PR does the following:
1. When there are no writes to Spanner, there are no files in GCS - to handle this scenario : the writer template is optimized to keep scanning 'empty' intervals until a file is found - in one go rather than per timer. This saves ~1 second per interval scan and if the Spanner writes happen after days of duration - this can be hours of time saved.
2. The data_seen table retention is kept to 30 days. This should be sufficient to handle the gaps between the Spanner reader and writer templates.


shreyakhajanchi (approved):
LGTM

aksharauke:
@bharadwaj-aditya please note, I have added the migrations override for this PR. Please let me know if there is a concern.

PR #1561, opened 2024-05-15
added integration test for rev replication reader job - custom shard id

shreyakhajanchi:


v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -0,0 +1,296 @@
...
+    IORedirectUtil.redirectLinesLog(exec.getInputStream(), LOG);
+    IORedirectUtil.redirectLinesLog(exec.getErrorStream(), LOG);
+
+    if (exec.waitFor() != 0) {
+      throw new RuntimeException("Error staging template, check Maven logs.");
+    }
+    gcsResourceManager.uploadArtifact(
+        "input/customShard.jar",
+        "../spanner-custom-shard/target/spanner-custom-shard-1.0-SNAPSHOT.jar");
+  }
```
manitgupta, line 284: Can this be reused? Should this be in a test utility class?

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -0,0 +1,296 @@
...
+
+  private void assertFileContentsInGCSForMultipleShards() throws java.lang.InterruptedException {
+    List<Artifact> artifactsShardA = null;
+    List<Artifact> artifactsShardB = null;
+    Thread.sleep(
+        180000); // wait sufficiently for the file to be generated. It takes about 3 minutes
+    // at-least. If not present wait additional 3 minutes before failing
+    for (int i = 0; i < 10; i++) {
+      Thread.sleep(18000); // wait for total 3 minutes over an interval of 18 seconds
```
manitgupta, line 155: Should this be modeled as a `conditionCheck`? Maybe something like `GCSRowCheck`?

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/c98fed7d7bd7cd2a8a2ac51c97be1f0706d79178/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/conditions/SpannerRowsCheck.java#L27

@darshan-sj 

Seems like the goal here is to wait for a condition to become true, and we should keep checking if the condition becomes true until the timeout limit, which is what we do in SpannerRowCheck.

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -0,0 +1,296 @@
...
+    String[] shellCommand = {"/bin/bash", "-c", String.join(" && ", commands)};
+
+    Process exec = Runtime.getRuntime().exec(shellCommand);
+
+    IORedirectUtil.redirectLinesLog(exec.getInputStream(), LOG);
+    IORedirectUtil.redirectLinesLog(exec.getErrorStream(), LOG);
+
+    if (exec.waitFor() != 0) {
+      throw new RuntimeException("Error staging template, check Maven logs.");
+    }
```
manitgupta, line 280: Should we follow the same approach of having a `TestBaseIT` class like we have defined for `DataStreamToSpanner`? 

Do I repeat all of this instantiation logic if I need to write another IT for `SpannerChangeStreamToGcs*`?

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -0,0 +1,296 @@
...
+
+    for (String shardName : shardNames) {
+      Shard shard = new Shard();
+      shard.setLogicalShardId(shardName);
+      shard.setUser("dummy");
+      shard.setHost("dummy");
+      shard.setPassword("dummy");
+      shard.setPort("3306");
+      JsonObject jsObj = (JsonObject) new Gson().toJsonTree(shard).getAsJsonObject();
+      jsObj.remove("secretManagerUri"); // remove field secretManagerUri
```
manitgupta, line 191: Why is removing this required? Why can't this be empty?

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
shreyakhajanchi: Yes, makes sense. Will follow similar approach of having TestBaseIT

shreyakhajanchi: Added

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
shreyakhajanchi: Makes sense. Added

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
shreyakhajanchi: Done

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -0,0 +1,259 @@
...
+      if (jobInfo == null) {
+        gcsResourceManager = createGcsResourceManager(getClass().getSimpleName());
+        spannerResourceManager = createSpannerResourceManager();
+        createSpannerDatabase(spannerResourceManager, spannerDdl);
+        uploadSessionFileToGcs(gcsResourceManager, sessionFileResourceName);
+        spannerMetadataResourceManager = createSpannerMetadataResourceManager();
+        createSpannerMetadataDatabase(spannerMetadataResourceManager);
+        createAndUploadJarToGcs(gcsResourceManager);
+        createAndUploadShardConfigToGcs();
+        launchReaderDataflowJob();
```
manitgupta, line 100: Can we further encapsulate this into a parameterized `launch` or `prelaunch` method? Looks like a lot of method calls for things that will essentially always need to be done. 

@darshan-sj You did something similar for live migration, do you have a suggestion here?

darshan-sj: Looks good. Currrently all the methods in Base class are like utility methods. If we have to compress this further, we will have to move the fields gcsResourceManager, and other resource managers to base class. I would prefer this implementation over that.

darshan-sj (approved):

v2/spanner-custom-shard/src/main/java/com/custom/CustomShardIdFetcherForIT.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package com.custom;
+
+import com.google.cloud.teleport.v2.spanner.utils.IShardIdFetcher;
+import com.google.cloud.teleport.v2.spanner.utils.ShardIdRequest;
+import com.google.cloud.teleport.v2.spanner.utils.ShardIdResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CustomShardIdFetcherForIT implements IShardIdFetcher {
```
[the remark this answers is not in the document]
shreyakhajanchi: It cannot be moved inside, it needs to be inside a java package so that JAR generation happens correctly, will add unit test to cover these, if that's okay?

PR #1565, opened 2024-05-15
Add datetime -> timestamp mapping to query help text

ahmedabu98:


v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/JdbcToBigQueryOptions.java
```diff
@@ -107,7 +107,8 @@ public interface JdbcToBigQueryOptions
       regexes = {"^.+$"},
       groupName = "Source",
       description = "JDBC source SQL query",
-      helpText = "The query to run on the source to extract the data.",
+      helpText = "The query to run on the source to extract the data. Important SQL to BigQuery type mappings to keep in mind:\n" +
```
damccorm, line 110: It would probably be good to provide a little more context. Maybe something like:

`Note that JDBC SQL and BigQuery types have some differences, so some type casting may be required if your schemas do not match. Some important SQL -> BigQuery type mappings to keep in mind are: ...`

ahmedabu98: Done, PTAL

damccorm (commented):
LGTM, could you fix the spotless violations with `mvn spotless:apply` though?

damccorm (approved):
Thanks!

PR #1569, opened 2024-05-16
Replace changed-files with modules-to-build

damccorm:
Remove the changed-files flag since we end up ignoring it anyways. Replace it with a mechanism for individual workflows to declare their preferred modules to build for arbitrary commands

### DIFF
--- a/cicd/internal/flags/common-flags.go
+++ b/cicd/internal/flags/common-flags.go
@@ -18,53 +18,25 @@ package flags
 
 import (
 	"flag"
-	"fmt"
-	"log"
-	"regexp"
 	"strings"
+)
+
+const (
+	ALL     = ""
+	SPANNER = "v2/datastream-to-spanner,v2/sourcedb-to-spanner,v2/spanner-change-streams-to-sharded-file-sink"
 )
 
 // Avoid making these vars public.
 var (
-	changedFiles string
+	modulesToBuild string
 )
 
 // Registers all common flags. Must be called before flag.Parse().
 func RegisterCommonFlags() {
-	flag.StringVar(&changedFiles, "changed-files", "", "List of changed files as a comma-separated string")
+	flag.StringVar(&modulesToBuild, "modules-to-build", ALL, "List of modules to build/run commands against")
 }
 
-// Returns all changed files with regexes. If no regexes are passed, all files are returned. If even one
-// is passed, then only file paths with a match anywhere in the file will be returned. If multiple are
-// passed, it is equivalent to (regex1|regex2|...|regexN)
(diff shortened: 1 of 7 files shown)


manitgupta (approved):

cicd/internal/flags/common-flags.go
```diff
@@ -18,53 +18,33 @@ package flags
...
+const (
+	ALL     = "ALL"
+	SPANNER = "SPANNER"
+)
+
 // Avoid making these vars public.
 var (
-	changedFiles string
+	modulesToBuild string
+	moduleMap      = map[string]string{ALL: "", SPANNER: "v2/datastream-to-spanner/,v2/spanner-common/,v2/spanner-change-streams-to-sharded-file-sink/,v2/gcs-to-sourcedb/,v2/spanner-migrations-sdk/,v2/spanner-custom-shard/,v2/sourcedb-to-spanner/,v2/common/,metadata"}
```
Polber, line 32: What about v1 spanner templates?

Polber, line 32: What about v1 spanner templates?

Polber: Also, there have been issues in the past with not specifying `plugins/templates-maven-plugin` since that plugin is not explicitly depended on, but is needed for staging templates for IT's

Polber (commented):
What is the plan to decouple Spanner workflows from Java PR's if the default is to build ALL modules?

cicd/internal/flags/common-flags.go
damccorm: This is a good callout - @manitgupta probably makes sense to add v1 in here.

@Polber the plan here is also to evolve this over time once we add the spanner workflow, I'll add the plugins/templates-maven-plugin dep, but it is also known that this isn't quite an exhaustive list. @manitgupta will have a follow up PR to control this once we have the main infra in.

damccorm:
> What is the plan to decouple Spanner workflows from Java PR's if the default is to build ALL modules?

The plan is to:
1) Add a spanner workflow that only runs on spanner filepaths
2) Remove spanner filepaths from the main workflow trigger

I'd eventually be interested in moving more templates out into their own workflows, but in the meantime this should (a) simplify our infra and (b) move governance to individual teams rather than requiring core infrastructure maintenance to keep this up to date. It also makes it pretty easy for any interested template owners to opt their templates in to a similar setup

cicd/internal/flags/common-flags.go
Polber: In that case LGTM

Polber (approved):

PR #1570, opened 2024-05-16
Add Kafka sink DLQ for kafka to BQ template, kafka to gcs template

AnandInguva:


v2/common/pom.xml
```diff
@@ -160,6 +160,15 @@
...
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.beam</groupId>
+            <artifactId>beam-sdks-java-io-kafka</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>3.7.0</version>
```
AnandInguva, line 170: todo: Check the latest version?

AnandInguva:
cc: @johnjcasey

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -284,6 +305,7 @@ public static PipelineResult runAvroPipeline(
                     .via(KafkaToBigQueryFlex::wrapBigQueryInsertError))
             .setCoder(FAILSAFE_ELEMENT_CODER);
 
+    // TODO: Add errorHandler to BQ or route the failedInserts to BQ router.
```
AnandInguva, line 308: I will continue work here on the next PR.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
```diff
@@ -90,19 +99,47 @@ public WriteResult expand(PCollection<KafkaRecord<byte[], byte[]>> kafkaRecords)
...
+                    SUCESS_GENERIC_RECORDS, TupleTagList.of(BadRecordRouter.BAD_RECORD_TAG)));
+
+    PCollection<BadRecord> failedGenericRecords =
+        genericRecords.get(BadRecordRouter.BAD_RECORD_TAG);
+    for (ErrorHandler<BadRecord, ?> errorHandler : errorHandlers) {
+      errorHandler.addErrorCollection(
+          failedGenericRecords.setCoder(BadRecord.getCoder(kafkaRecords.getPipeline())));
+    }
+    // TODO: Change the name.
+    PCollectionTuple genericRecords1;
```
Dippatel98, line 120: Can this be renamed to something meaningful?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -268,7 +287,9 @@ public static PipelineResult runAvroPipeline(
       }
 
       if (options.getSchemaRegistryConnectionUrl() != null && options.getOutputDataset() != null) {
-        writeResult = kafkaRecords.apply(AvroDynamicTransform.of(options));
+        writeResult =
+            kafkaRecords.apply(
+                AvroDynamicTransform.of(options).withBadRecordErrorHanlder(badRecordErrorHandlers));
```
Dippatel98, line 292: It looks like currently DLQ is only added in the case when we use BQ Dynamic Destinations. Can you also add the DLQ to AvroTransform which covers the other paths.

AnandInguva: Done pending testing

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
AnandInguva: Done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
```diff
@@ -79,6 +96,7 @@ public WriteResult expand(PCollection<KafkaRecord<byte[], byte[]>> kafkaRecords)
     Schema schema = SchemaUtils.getAvroSchema(avroSchema);
     WriteResult writeResult;
 
+    // TODO: Add support for error handler during BQ writes.
     Write<TableRow> writeToBQ =
```
an2x, line 100: `writeToBQ` -> `writeToBq` or `writeToBigQuery`

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
```diff
@@ -125,12 +171,17 @@ private static class KafkaRecordToGenericRecordFailsafeElementFn
...
     KafkaRecordToGenericRecordFailsafeElementFn(
-        Schema schema, String topicName, String useConfluentWireFormat) {
+        Schema schema,
+        String topicName,
+        String useConfluentWireFormat,
+        BadRecordRouter badRecordRouter) {
       this.schema = schema;
       this.topicName = topicName;
       this.useConfluentWireFormat = useConfluentWireFormat;
+      this.badRecordRouter = badRecordRouter;
```
an2x, line 184: If these fields are only set in the constructor, declare them final.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
```diff
@@ -61,15 +66,27 @@
...
+  private BadRecordRouter badRecordRouter = BadRecordRouter.RECORDING_ROUTER;
+  private List<ErrorHandler<BadRecord, ?>> errorHandlers;
+  private static final TupleTag<FailsafeElement<KafkaRecord<byte[], byte[]>, GenericRecord>>
+      SUCCESS_GENERIC_RECORDS = new TupleTag<>();
+  private static final TupleTag<FailsafeElement<KafkaRecord<byte[], byte[]>, TableRow>>
+      SUCCESS_TABLE_ROW = new TupleTag<>();
+  private static final KafkaRecordCoder kafkaRecordCoder =
+      KafkaRecordCoder.of(
+          NullableCoder.of(ByteArrayCoder.of()), NullableCoder.of(ByteArrayCoder.of()));
   private KafkaToBigQueryFlexOptions options;
```
an2x, line 78: In general, transforms shouldn't have a dependency on pipeline options.

1) On high level, from the design perspective, transforms are supposed to be "building blocks", reusable across multiple pipelines. Making them dependent on a specific Options class makes them unusable with other Options classes and other pipelines (or you'll end up with some overly complicated logic that tries to construct the transform from many different Options classes).
2) On low level, it's also better to keep the amount of class variable to a minimum because Transforms are serialized, i.e. it's better to keep only those variables that the transform actually needs instead of the whole Options class because it contains many other options not used in the transform. This is to reduce the size of the serialized transform. Not very important as its size shouldn't be huge anyway, but storing the whole Options class is excessive.

Typical pattern is to read Options int the pipeline class and pass the option value to the transform, example: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/4b06b8b244043565c9aa4b689e10a4be887c400a/v2/dataplex/src/main/java/com/google/cloud/teleport/v2/templates/DataplexBigQueryToGcs.java#L328

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
```diff
@@ -61,15 +66,27 @@
  */
 public class AvroTransform
     extends PTransform<PCollection<KafkaRecord<byte[], byte[]>>, WriteResult> {
-
-  private static final Logger LOG = LoggerFactory.getLogger(AvroTransform.class);
-
+  private BadRecordRouter badRecordRouter = BadRecordRouter.RECORDING_ROUTER;
+  private List<ErrorHandler<BadRecord, ?>> errorHandlers;
```
an2x, line 70: Code style nit: don't mix class variable and static variables, i.e. badRecordHandler, errorHandlers, and options should be declared together at the bottom.

an2x: Also make all of them final if possible.

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
```diff
@@ -58,6 +70,9 @@ public abstract class AvroWriteTransform
         PCollection<KafkaRecord<byte[], byte[]>>, WriteFilesResult<AvroDestination>> {
   private static final String subject = "UNUSED";
   static final int DEFAULT_CACHE_CAPACITY = 1000;
```
an2x, line 72: Can be made `private`?

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
```diff
@@ -196,19 +250,26 @@ public WriteFilesResult<AvroDestination> writeToGCS(PCollection<GenericRecord> g
...
 
   public static class ConvertBytesToGenericRecord
-      extends DoFn<KafkaRecord<byte[], byte[]>, GenericRecord> {
+      extends DoFn<
+          KafkaRecord<byte[], byte[]>,
+          FailsafeElement<KafkaRecord<byte[], byte[]>, GenericRecord>> {
     private transient SchemaRegistryClient schemaRegistryClient;
     private String schemaRegistryURL = null;
     private String schemaPath = null;
     private boolean useMock;
```
an2x, line 258: Seems like all 3 can be made `final`?

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
```diff
@@ -196,19 +250,26 @@ public WriteFilesResult<AvroDestination> writeToGCS(PCollection<GenericRecord> g
...
+    public ConvertBytesToGenericRecord(String schemaRegistryURL, BadRecordRouter badRecordRouter) {
       this.schemaRegistryURL = schemaRegistryURL;
+      this.badRecordRouter = badRecordRouter;
     }
 
-    public ConvertBytesToGenericRecord(String schemaPath, boolean useMock) {
+    public ConvertBytesToGenericRecord(
+        String schemaPath, boolean useMock, BadRecordRouter badRecordRouter) {
       this.schemaPath = schemaPath;
       this.useMock = useMock;
```
an2x, line 270: I'm not sure I understand how `useMock` is used and why it's needed.

What if someone calls ` new ConvertBytesToGenericRecord(schemaPath, useMock = false, badRecordRouter)`? AFAIU, `setup()` method will try to use `schemaRegistryURL` in this case, which is not initialized.

I'd implement it like this:
```
  public static ConvertBytesToGenericRecord withSchemaRegistry(String schemaRegistryUrl, BadRecordRouter badRecordRouter) {
    return new ConvertBytesToGenericRecord(schemaRegistryUrl, null, badRecordRouter);
  }

  public static ConvertBytesToGenericRecord withSchemaPath(String schemaPath, BadRecordRouter badRecordRouter) {
    return new ConvertBytesToGenericRecord(null, schemaPath, badRecordRouter);
  }

  private ConvertBytesToGenericRecord(String schemaRegistryUrl, String schemaPath, BadRecordRouter badRecordRouter) {
      this.schemaRegistryUrl = schemaRegistryUrl;
      this.schemaPath = schemaPath;
      this.badRecordRouter = badRecordRouter;
    }
```

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
```diff
@@ -196,19 +250,26 @@ public WriteFilesResult<AvroDestination> writeToGCS(PCollection<GenericRecord> g
...
+      this.badRecordRouter = badRecordRouter;
     }
 
     @DoFn.Setup
```
an2x, line 274: This `@Setup` implementation will result in the schema being loaded from GCS every time a new DoFn instance is constructed (could be hundreds/thousands times). I suggest loading schema once in the template code, when the pipeline is constructed, and passing the loaded schema string to the DoFn.

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/WriteTransform.java
```diff
@@ -62,7 +68,8 @@ public POutput expand(PCollection<KafkaRecord<byte[], byte[]>> kafkaRecord) {
                     .setSchemaRegistryURL(options().getSchemaRegistryURL())
                     .setSchemaPath(options().getSchemaPath())
                     .setWindowDuration(options().getWindowDuration())
-                    .build());
+                    .build()
+                    .withBadRecordErrorHandlers(errorHandlers()));
```
an2x, line 72: Calling `withX()` after `build()` seems like a strange pattern to me. Should `setErrorHandlers()` be part of the builder?

an2x (changes requested):

v2/common/pom.xml
```diff
@@ -160,6 +160,15 @@
               </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.beam</groupId>
+            <artifactId>beam-sdks-java-io-kafka</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
```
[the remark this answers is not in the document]
AnandInguva: Done

v2/common/src/main/java/com/google/cloud/teleport/v2/dlq/DeadLetterQueueOptions.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.cloud.teleport.metadata.TemplateParameter;
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+/**
+ * {@link DeadLetterQueueOptions} is used for any Dead letter queue sinks for the failed records.
+ */
+public interface DeadLetterQueueOptions extends PipelineOptions {
+  String KAFKA_DLQ_GROUP_NAME = "Kafka Dead Letter Queue";
```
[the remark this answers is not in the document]
AnandInguva: Interface is not allowing to declare final

v2/common/src/main/java/com/google/cloud/teleport/v2/dlq/DlqUtils.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Utility methods for working with dead letter queues and {@link BadRecord} objects. These methods
+ * are designed to help retrieve the original element or a JSON representation of the original
+ * element that caused an error during processing.
+ *
+ * <p>To successfully extract the original element, a coder must be defined. This class is primarily
+ * designed to operate on {@link BadRecord} objects and expects the encoded object is a {@link
+ * KafkaRecord}.
+ */
+public class DlqUtils {
```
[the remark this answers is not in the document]
AnandInguva: Done

v2/common/src/main/java/com/google/cloud/teleport/v2/dlq/KafkaDeadLetterQueue.java
```diff
@@ -0,0 +1,76 @@
...
+  public abstract String topic();
+
+  public abstract Map<String, Object> config();
+
+  public static KafkaDLQBuilder newBuilder() {
+    return new AutoValue_KafkaDeadLetterQueue.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class KafkaDLQBuilder {
```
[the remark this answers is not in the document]
AnandInguva: Done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
```diff
@@ -121,13 +157,16 @@ private static class KafkaRecordToGenericRecordFailsafeElementFn
...
-    KafkaRecordToGenericRecordFailsafeElementFn(String schemaRegistryConnectionUrl) {
+    KafkaRecordToGenericRecordFailsafeElementFn(
+        String schemaRegistryConnectionUrl, BadRecordRouter badRecordRouter) {
       this.schemaRegistryConnectionUrl = schemaRegistryConnectionUrl;
+      this.badRecordRouter = badRecordRouter;
     }
 
     @Setup
-    public void setup() throws IOException, RestClientException {
+    public void setup() {
```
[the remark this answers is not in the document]
AnandInguva: I don't think so. Changed the comment a bit

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
```diff
@@ -139,19 +178,21 @@ public void setup() throws IOException, RestClientException {
...
-      KafkaRecord<byte[], byte[]> element = context.element();
+    public void processElement(
+        @Element KafkaRecord<byte[], byte[]> kafkaRecord, MultiOutputReceiver receiver)
+        throws Exception {
       GenericRecord result = null;
       try {
-        // Serialize to Generic Record
         result =
             (GenericRecord)
                 this.deserializer.deserialize(
```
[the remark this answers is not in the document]
AnandInguva: if null then it raise an exception and we won't reach this code path right?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
```diff
@@ -121,13 +157,16 @@ private static class KafkaRecordToGenericRecordFailsafeElementFn
     private transient KafkaAvroDeserializer deserializer;
     private transient SchemaRegistryClient schemaRegistryClient;
     private String schemaRegistryConnectionUrl = null;
```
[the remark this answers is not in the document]
AnandInguva: this is getting modified and final is not getting accepted.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
AnandInguva: Done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
AnandInguva: getting errors when declaring them final

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
AnandInguva: Done

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
AnandInguva: Added a TODO.

v2/common/pom.xml
AnandInguva: Remove these file changes

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
AnandInguva: Getting error when changing this to final

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/AvroWriteTransform.java
AnandInguva: Added a TODO to remove mock

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/transforms/WriteTransform.java
AnandInguva: Done

v2/common/src/main/java/com/google/cloud/teleport/v2/dlq/DeadLetterQueueOptions.java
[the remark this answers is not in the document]
an2x: Ah, I missed that it's an interface. LG then!

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroTransform.java
AnandInguva: I changed this AvroDynamicTransform and forgot to change here. I changed it now.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/AvroDynamicTransform.java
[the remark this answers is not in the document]
an2x: Let's fix this in a separate PR.

an2x (approved):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/dlq/KafkaDeadLetterQueueOptions.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.cloud.teleport.metadata.TemplateParameter;
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+/**
+ * {@link KafkaDeadLetterQueueOptions} is used for any Dead letter queue sinks for the failed
+ * records.
+ */
+public interface KafkaDeadLetterQueueOptions extends PipelineOptions {
+  String KAFKA_DLQ_GROUP_NAME = "Kafka Dead Letter Queue";
```
Dippatel98, line 27: nit: Do you think it makes more sense, if we keep the name as just `Dead Letter Queue`. I am thinking so that wherever this is inherited it automatically goes into the `Dead Letter Queue` group and not a separate `Kafka Dead Letter Queue` group

an2x: +1, I'd make it "Dead Letter Queue" (so that when we have both Kafka DLQ and BQ DLQ they would go to the same parameter group).

PR #1572, opened 2024-05-16
Named schema release

jin-jj:
Support named schema import/export

1. Add some identifiler utility for quoting and fully qualified name(FQN) combination.
2. Refactor and use the new utility function for quoting operations in the code.
3. Add a avro schema property to set/get spanner database object FQN. Can't not set avro schema name using FQN, we hardcoded avro schema namespace which conflict with FQN.
4. Changes import/export to support spanner named schema.
    4.1: Make the table name parameter support dot(.)
    4.2: Improve the information schema scanner to support multiple named schema.
    4.3: Import/export pipeline and transformer support.

Also fix some bugs
1. Fix a bug that when table name is not existed, it will cause NPE.
2. Fix a bug that index interleave in missing quoting.
3. Fix a bug that optimizer version support is missing.


v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Index.java
```diff
@@ -15,6 +15,8 @@
  */
 package com.google.cloud.teleport.spanner.ddl;
 
+import static com.google.cloud.teleport.spanner.common.NameUtils.quoteIdentifier;
+
 import com.google.auto.value.AutoValue;
 import com.google.cloud.spanner.Dialect;
 import com.google.common.collect.ImmutableList;
```
darshan-sj, line 22: Please add a comment on line 34 stating that name() will be fully qualified name if the Database contains named schema and just name otherwise, with examples.

darshan-sj: Also, can you add similar comment to all the other ddl classes like, Table, View etc.,

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -945,8 +978,8 @@ private void listChangeStreams(Ddl.Builder builder) {
                     + " AND cst.table_catalog = csc.table_catalog"
                     + " AND cst.table_schema = csc.table_schema"
                     + " AND cst.table_name = csc.table_name"
-                    + " GROUP BY cs.change_stream_name, cs.all, cst.table_name, cst.all_columns"
-                    + " ORDER BY cs.change_stream_name, cs.all, cst.table_name"));
+                    + " GROUP BY cs.change_stream_name, cs.all, cst.table_schema, cst.table_name, cst.all_columns"
+                    + " ORDER BY cs.change_stream_name, cs.all, cst.table_schema, cst.table_name"));
```
darshan-sj, line 982: I see many queries have ` WHERE t.table_schema NOT IN ('INFORMATION_SCHEMA', 'SPANNER_SYS')` where clause added. But some of the queries, like the above don't have this new where clause. I hope they are correct and are tested.

darshan-sj (approved):
Added a few comments. Overall changes look good.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
jin-jj: ype, because change stream can not show up in system catalog, that is why it works. 

But you are right, to keep consistent, and the import/export semantic should be: List user created/creatable objects, no matter what object it is. Let me think about it to see how to improve/enhance.

v1/src/main/java/com/google/cloud/teleport/spanner/BuildReadFromTableOperations.java
```diff
@@ -88,17 +94,21 @@ public void processElement(ProcessContext c) {
                           ReadOperation.create()
                               .withQuery(
                                   String.format(
-                                      "SELECT \"%s\" AS _spanner_table, %s FROM `%s` AS t",
-                                      table.name(), columnsListAsString, table.name()))
+                                      "SELECT \"%s\" AS _spanner_table, %s FROM %s AS t",
+                                      table.name(),
+                                      columnsListAsString,
```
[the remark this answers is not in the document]
jin-jj: 1. right.
2. no, column names does not need to be qualified. the columns are referenced using t.xx already.

PR #1573, opened 2024-05-17
Add Spanner PR workflow and exclude Spanner from global workflow

manitgupta:


.github/workflows/java-pr.yml
```diff
@@ -27,6 +27,14 @@ on:
       # This will make it easier to verify action changes don't break anything.
       - '.github/actions/setup-env/*'
       - '.github/workflows/java-pr.yml'
+      # Exclude spanner paths from global run
```
damccorm, line 30:
```suggestion
      # Exclude spanner paths from global run (covered in https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/spanner-pr.yml)
```

.github/workflows/spanner-pr.yml
```diff
@@ -0,0 +1,201 @@
...
+    paths:
+      # Include spanner paths only
+      - '.github/workflows/spanner-pr.yml'
+      - 'v2/datastream-to-spanner/**'
+      - 'v2/spanner-common/**'
+      - 'v2/spanner-change-streams-to-sharded-file-sink/**'
+      - 'v2/gcs-to-sourcedb/**'
+      - 'v2/spanner-migrations-sdk/**'
+      - 'v2/spanner-custom-shard/**'
+      - 'v2/sourcedb-to-spanner/**'
```
damccorm, line 32: I would recommend including `v2/common/**` here as well since changes there could break spanner

cicd/internal/op/maven.go
```diff
@@ -29,18 +31,18 @@ func RunMavenOnPom(pom string, cmd string, args ...string) error {
 	fullArgs = append(fullArgs, "-f", pom)
 	fullArgs = append(fullArgs, "-e")
 	fullArgs = append(fullArgs, args...)
-
+	modules := flags.ModulesToBuild()
+	if len(modules) != 0 {
+		moduleArgs := []string{"-pl", strings.Join(modules, ",")}
+		fullArgs = append(fullArgs, moduleArgs...)
+		fullArgs = append(fullArgs, "-am")
```
damccorm, line 38: Oh nice - this is cleaner (both `-am` and moving it into `RunMavenOnPom`). Thanks!

damccorm (commented):
2 minor recommended changes, otherwise LGTM

cicd/internal/op/maven.go
manitgupta: Thanks!

.github/workflows/spanner-pr.yml
manitgupta: The default path builds all the templates so this would be caught irrespective of who (Spanner team or someone else) did the change. Do you think we still need to add it?

damccorm: That's a good point, I think you're fine without it. I also added this comment before I saw the `-am` change (which reduces the dependency here).

damccorm (approved):
Thanks!

PR #1574, opened 2024-05-17
[BULK] Misc changes across types, logging  and options

Deep1998:
This PR consists of

1. Support for INTERVAL type
2. Adding a Log Level parameter 
3. Removing regex constraint for source database
4. Deleting the deprecated ResultSetToMutation package


v2/common/src/main/java/com/google/cloud/teleport/v2/options/CommonTemplateOptions.java
```diff
@@ -49,4 +51,23 @@ public interface CommonTemplateOptions extends PipelineOptions {
...
+      optional = true,
+      enumOptions = {
+        @TemplateParameter.TemplateEnumOption("OFF"),
+        @TemplateParameter.TemplateEnumOption("ERROR"),
+        @TemplateParameter.TemplateEnumOption("WARN"),
+        @TemplateParameter.TemplateEnumOption("INFO"),
+        @TemplateParameter.TemplateEnumOption("TRACE"),
+        @TemplateParameter.TemplateEnumOption("DEBUG")
+      },
+      description = "Log level in the workers, defaults to INFO",
```
bharadwaj-aditya, line 66: What is the current default ? Let's ensure that we retain that.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -71,7 +71,6 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
   /* TODO: (support Sharding, PG namespaces) */
   @TemplateParameter.Text(
       order = 5,
-      regexes = {"(^[a-zA-Z0-9]+$)"},
```
bharadwaj-aditya, line 74: why is this removed ? In what cases will it fail ?

v2/common/src/main/java/com/google/cloud/teleport/v2/options/CommonTemplateOptions.java
Deep1998: Default is info only.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
Deep1998: This regex does not allow special characters in source db name, which is allowed in mysql. For ex, migrating a mysql db test_db will fail because _ is not allowed in the regex.

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya:
CodeCove says - Project coverage is 40.93%

Deep1998:
> Codecov

Yes that is for entire teleport repository.

PR #1575, opened 2024-05-17
Reverse replication reader IT - broken down into single vs multi shard

shreyakhajanchi:


v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsCustomShardIT.java
```diff
@@ -81,7 +82,7 @@ public class SpannerChangeStreamToGcsCustomShardIT extends SpannerChangeStreamTo
   @Before
   public void setUp() throws IOException, InterruptedException {
     skipBaseCleanup = true;
-    synchronized (SpannerChangeStreamToGcsSimpleIT.class) {
+    synchronized (SpannerChangeStreamToGcsMultiShardIT.class) {
```
aksharauke, line 85: This does not match the class name

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
```diff
@@ -0,0 +1,224 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Integration test for simple test of single shard,single table. */
+@Category({TemplateIntegrationTest.class, SkipDirectRunnerTest.class})
+@TemplateIntegrationTest(SpannerChangeStreamsToShardedFileSink.class)
+@RunWith(JUnit4.class)
+public class SpannerChangeStreamToGcsMultiShardIT extends SpannerChangeStreamToGcsITBase {
```
aksharauke, line 56: Did not understand the motivation for a new class. From what I understand - this is only needed when there is difference in the Dataflow input parameters.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/conditions/GCSArtifactsCheck.java
```diff
@@ -38,6 +40,9 @@ public abstract class GCSArtifactsCheck extends ConditionCheck {
   @Nullable
   abstract Integer maxSize();
 
+  @Nullable
+  abstract String tableName();
```
aksharauke, line 44: If we are writing a generic GCS checker - then this should be replaced with something generic- like a pattern to check. As such table name has no meaning for a GCS artifact.

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
```diff
@@ -0,0 +1,224 @@
...
+      ja.add(jsObj);
+    }
+
+    String shardFileContents = ja.toString();
+    LOG.info("Shard file contents: {}", shardFileContents);
+    // -DartifactBucket has the bucket name
+    gcsResourceManager.createArtifact("input/shard.json", shardFileContents);
+  }
+
+  private void writeSpannerDataForSingers(int singerId, String firstName, String shardId) {
```
aksharauke, line 142: We would need one more test case - even if there are multiple shards, Spanner writes are happening to one shard only.

aksharauke: Can be marked as TODO

aksharauke (changes requested):

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
shreyakhajanchi: Added a new class to segregate the tests requiring session file in order to remove the session file dependency as much as possible. Also parameters are different for single and multi shard tests as session file is not specified in the single shard flow

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
shreyakhajanchi: Spanner writes are happening for two shards : testShardB and testShardC. Am I missing something here ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/conditions/GCSArtifactsCheck.java
shreyakhajanchi: makes sense , updated PTAL

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
aksharauke: I meant that shard file has two shards, while the writes are happening only on one shard. But it's not a P0, resolving the comment.

v2/spanner-change-streams-to-sharded-file-sink/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamToGcsMultiShardIT.java
```diff
@@ -0,0 +1,224 @@
...
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(6)), conditionCheck);
+
+    // Assert Conditions
+    assertThatResult(result).meetsConditions();
+
+    List<Artifact> artifactsShardB =
+        gcsResourceManager.listArtifacts("output/testShardB/", Pattern.compile(".*\\.txt$"));
+    List<Artifact> artifactsShardC =
+        gcsResourceManager.listArtifacts("output/testShardC/", Pattern.compile(".*\\.txt$"));
+    assertThatArtifacts(artifactsShardB).hasFiles();
```
aksharauke, line 216: Is this check needed after the GCSArtifacts check - which itself checks the count of files?

aksharauke (approved):

PR #1578, opened 2024-05-20
Fixing value mapping for blob and year in Reader

VardhanThigle:
1. The value mapping for blob and year in the reader was not mapping the output to the expected avro type (`string` instead of `char[]` and `integer` instead of `long`), leading to a crash. Fixing the same.
Along with the UT, this PR has been tested on a live db with these two types.
2. With the code review progression, also added UT to cover null values for all types.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java
@@ -67,11 +67,11 @@ public class MysqlJdbcValueMappings implements JdbcValueMappingsProvider {
 
   /* Hex Encoded string for bytes type. */
   private static final ResultSetValueMapper<byte[]> bytesToHexString =
-      (value, schema) -> Hex.encodeHex(value);
+      (value, schema) -> new String(Hex.encodeHex(value));
 
   /* Hex Encoded string for blob types. */
   private static final ResultSetValueMapper<Blob> blobToHexString =
-      (value, schema) -> Hex.encodeHex(value.getBytes(1, (int) value.length()));
+      (value, schema) -> new String(Hex.encodeHex(value.getBytes(1, (int) value.length())));
 
   /* Extract UTC Values for date and time related types */
   private static final Calendar utcCalendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
@@ -189,7 +189,7 @@ public class MysqlJdbcValueMappings implements JdbcValueMappingsProvider {
           .put("TINYTEXT", Pair.of(ResultSet::getString, valuePassThrough))
           .put("VARBINARY", Pair.of(ResultSet::getBytes, bytesToHexString))
           .put("VARCHAR", Pair.of(ResultSet::getString, valuePassThrough))
-          .put("YEAR", Pair.of(ResultSet::getLong, valuePassThrough))
+          .put("YEAR", Pair.of(ResultSet::getInt, valuePassThrough))
           .build()
           .entrySet()
           .stream()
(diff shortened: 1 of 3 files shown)


v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/JdbcSourceRowMapperTest.java
```diff
@@ -240,7 +240,7 @@ private static ImmutableList<Col> getTestCols() {
                 .build())
         .add(
             Col.builder("blob_col", "BLOB", new SourceColumnType("BLOB", new Long[] {10L}, null))
-                .withValue(new byte[] {0x65, 0x66}, new char[] {'6', '5', '6', '6'})
+                .withValue(new byte[] {0x65, 0x66}, "6566")
```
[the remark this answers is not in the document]
VardhanThigle: Thanks.
I have added a null case for all the types.
A few other edge cases are covered in the exception test cases.

We are not checking with a real DB in the UT, so it might not be possible to test with the limits of blob length, would cover that in integration tests.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java
```diff
@@ -71,7 +71,7 @@ public class MysqlJdbcValueMappings implements JdbcValueMappingsProvider {
 
   /* Hex Encoded string for blob types. */
   private static final ResultSetValueMapper<Blob> blobToHexString =
-      (value, schema) -> Hex.encodeHex(value.getBytes(1, (int) value.length()));
```
Deep1998, line 74: We might have to do this for binary and varbinary types also?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/JdbcSourceRowMapperTest.java
```diff
@@ -323,7 +333,7 @@ private static ImmutableList<Col> getTestCols() {
                     "medium_blob_col",
                     "BLOB",
                     new SourceColumnType("MEDIUMBLOB", new Long[] {10L}, null))
-                .withValue(new byte[] {0x65, 0x66}, new char[] {'6', '5', '6', '6'})
+                .withValue(new byte[] {0x65, 0x66}, "6566")
                 .build())
```
Deep1998, line 337: Similarly UTs for binary, varbinay? And maybe YEAR too to catch this in UTs early on?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java
VardhanThigle: Thanks for the catch! Done.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/JdbcSourceRowMapperTest.java
VardhanThigle: We have a UT for bianry and varbinary, I have updated the same. The UT could not catch the issue earlier since we use derby and not actual db for testing.
For `year` we dont have a corresponding type in derby but we have an existing UT that maps it to dummy derby type (this wont be as effective as an actual DB run).

bharadwaj-aditya (approved):
LGTM

PR #1581, opened 2024-05-20
Refactor KafkaToBigQueryFlexOptions and choose defaults for the KafkaToBigQueryFlex template

Dippatel98:
Refactor KafkaToBigQueryFlexOptions and default the KafkaToBigQueryFlex template to Streaming Engine and Resource-based billing, also Storage Write API is required to write to BigQuery.

### DIFF
--- a/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
+++ b/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
@@ -27,6 +27,7 @@ public interface KafkaReadOptions extends PipelineOptions {
 
   @TemplateParameter.Text(
       order = 1,
+      groupName = "Source",
       optional = true,
       regexes = {"[,:a-zA-Z0-9._-]+"},
       description = "Kafka Bootstrap Server list",
@@ -38,9 +39,10 @@ public interface KafkaReadOptions extends PipelineOptions {
 
   @TemplateParameter.Text(
       order = 2,
+      groupName = "Source",
       optional = true,
       regexes = {"[,a-zA-Z0-9._-]+"},
-      description = "Kafka topic(s) to read input from.",
+      description = "Kafka Topic(s) to read input from",
       helpText = "Kafka topic(s) to read input from.",
       example = "topic1,topic2")
   String getKafkaReadTopics();
@@ -49,38 +51,61 @@ public interface KafkaReadOptions extends PipelineOptions {
 
   @TemplateParameter.Enum(
       order = 3,
-      description = "The Kafka offset to read from.",
+      groupName = "Source",
       enumOptions = {
-        @TemplateParameter.TemplateEnumOption("latest"),
         @TemplateParameter.TemplateEnumOption("earliest"),
+        @TemplateParameter.TemplateEnumOption("latest"),
       },
-      helpText = "Set the Kafka offset to earliest or latest(default)",
-      optional = true)
+      optional = true,
+      description = "Start Offset",
+      helpText = "The Kafka Offset to read from.")
   @Default.String("latest")
   String getKafkaReadOffset();
 
   void setKafkaReadOffset(String value);
 
+  @TemplateParameter.Enum(
+      order = 4,
+      name = "kafkaReadAuthenticationMode",
+      groupName = "Source",
+      enumOptions = {
+        @TemplateParameter.TemplateEnumOption("NONE"),
+        @TemplateParameter.TemplateEnumOption("SASL_PLAIN"),
+      },
+      optional = false,
+      description = "Authentication Mode",
+      helpText = "Kafka read authentication mode. Can be NONE or SASL_PLAIN")
+  @Default.String("NONE")
+  String getKafkaReadAuthenticationMode();
+
+  void setKafkaReadAuthenticationMode(String value);
+
   @TemplateParameter.Text(
       order = 4,
-      optional = true,
-      description =
-          "Username to be used with SASL_PLAIN mechanism for reading from Kafka, stored in Google Cloud Secret Manager.",
+      groupName = "Source",
+      parentName = "kafkaReadAuthenticationMode",
+      parentTriggerValues = {"SASL_PLAIN"},
+      optional = false,
+      description = "Username",
       helpText =
           "Secret Manager secret ID for the SASL_PLAIN username. Should be in the format projects/{project}/secrets/{secret}/versions/{secret_version}.",
       example = "projects/your-project-id/secrets/your-secret/versions/your-secret-version")
+  @Default.String("")
   String getKafkaReadUsernameSecretId();
 
   void setKafkaReadUsernameSecretId(String value);
 
   @TemplateParameter.Text(
       order = 5,
-      optional = true,
-      description =
-          "Password to be used with SASL_PLAIN mechanism for reading from Kafka, stored in Google Cloud Secret Manager.",
+      groupName = "Source",
+      parentName = "kafkaReadAuthenticationMode",
+      parentTriggerValues = {"SASL_PLAIN"},
+      optional = false,
+      description = "Password",
       helpText =
           "Secret Manager secret ID for the SASL_PLAIN password. Should be in the format projects/{project}/secrets/{secret}/versions/{secret_version}",
       example = "projects/your-project-id/secrets/your-secret/versions/your-secret-version")
+  @Default.String("")
   String getKafkaReadPasswordSecretId();
 
   void setKafkaReadPasswordSecretId(String value);
--- a/v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
+++ b/v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
@@ -17,6 +17,7 @@ package com.google.cloud.teleport.v2.options;
 
 import com.google.cloud.teleport.metadata.TemplateParameter;
 import com.google.cloud.teleport.v2.kafka.options.KafkaReadOptions;
+import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
 import org.apache.beam.sdk.options.Default;
 
 /**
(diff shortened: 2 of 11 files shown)


v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -156,24 +152,31 @@ public static void main(String[] args) throws IOException, RestClientException {
...
     List<String> topicsList;
-    if (options.getKafkaReadTopics() != null) {
-      topicsList = new ArrayList<>(Arrays.asList(options.getKafkaReadTopics().split(",")));
-    } else {
-      throw new IllegalArgumentException("Please Provide --kafkaReadTopic");
-    }
     String bootstrapServers;
-    if (options.getReadBootstrapServers() != null) {
-      bootstrapServers = options.getReadBootstrapServers();
+    if (options.getReadBootstrapServerAndTopic() != null) {
```
an2x, line 174: getReadBootstrapServerAndTopic() can be a GMK topic too, right?
Should we check if this matches the GMK topic pattern first, and only then try to parse it as <bootstrap_server>:<topic>?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -156,24 +152,31 @@ public static void main(String[] args) throws IOException, RestClientException {
...
+      String[] l = options.getReadBootstrapServerAndTopic().split(":");
```
an2x, line 175: @AnandInguva IIRC we decided to make the server and topic separator different?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -156,24 +152,31 @@ public static void main(String[] args) throws IOException, RestClientException {
...
+      topicsList = List.of(l[2]);
+      bootstrapServers = l[0] + ":" + l[1];
     } else {
-      throw new IllegalArgumentException("Please Provide --bootstrapServers");
+      throw new IllegalArgumentException("Please provide bootstrap server and topic.");
```
an2x, line 179: I'd make the the error message more specific. Mention that the parameter value doesn't match a specific pattern (otherwise you may set it to something in an invalid format and still get "please provide the param" error, which is confusing).

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -248,27 +252,74 @@ public static PipelineResult runAvroPipeline(
...
 
-    if (options.getAvroFormat().equals("NON_WIRE_FORMAT") && options.getAvroSchemaPath() != null) {
-
-      writeResult = kafkaRecords.apply(AvroTransform.of(options));
+    if (options.getMessageFormat().equals("AVRO_BINARY_ENCODING")
+        && options.getBinaryAvroSchemaPath() != null) {
+      writeResult =
+          kafkaRecords.apply(
+              AvroTransform.of(
+                  options.getReadBootstrapServerAndTopic().split(":")[2],
```
an2x, line 260: nit: There is a check that verifies the format of this parameter in a different place, and then the same parameter is parsed again here. This will require modifying the code in 2 different places if the format changes. A better pattern would be to check the format and parse this option above, into 2 separate variables (bootstrapServers and topics), and then use these already parsed variables here.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -248,27 +252,74 @@ public static PipelineResult runAvroPipeline(
...
-
-      if (options.getAvroSchemaPath() != null && options.getOutputTableSpec() != null) {
-        writeResult = kafkaRecords.apply(AvroTransform.of(options));
-      }
-
-      if (options.getSchemaRegistryConnectionUrl() != null && options.getOutputDataset() == null) {
+      if (options.getSchemaFormat().equals("SINGLE_SCHEMA_FILE")) {
+
+        if (options.getConfluentAvroSchemaPath() != null && options.getOutputTableSpec() == null) {
+          throw new IllegalArgumentException("An output BigQuery table is required.");
```
an2x, line 277: I'd add more details to the error message.
"outputTableSpec parameter is required when using SINGLE_SCHEMA_FILE schema format and <... what are the other conditions?...>"

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -248,27 +252,74 @@ public static PipelineResult runAvroPipeline(
 
     WriteResult writeResult = null;
 
-    if (options.getAvroFormat().equals("NON_WIRE_FORMAT") && options.getAvroSchemaPath() != null) {
-
-      writeResult = kafkaRecords.apply(AvroTransform.of(options));
+    if (options.getMessageFormat().equals("AVRO_BINARY_ENCODING")
```
an2x, line 255: nit: The if clauses are very long and hard to read/understand the logic. Consider splitting this method into others even further, e.g.
```
  if (options.getMessageFormat().equals("AVRO_BINARY_ENCODING") {
    runAvroBinaryEncodingPipeline(...);    
  } else if (options.getMessageFormat().equals("AVRO_CONFLUENT_WIRE_FORMAT") {
    runAvroConfluentEncodingPipeline(...);    
  } else {
    throw UnsupportedOperationException("Message format " + options.getMessageFormat() + " is unspported.");
  }
```

I also suggest following the above "if/else-if/else throw exception" pattern instead of just if/else because if the latter may hide problems in the future when a 3rd format is added (which is very easy to forget to add here).

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -248,27 +252,74 @@ public static PipelineResult runAvroPipeline(
...
-      }
-
-      if (options.getAvroSchemaPath() != null && options.getOutputTableSpec() != null) {
-        writeResult = kafkaRecords.apply(AvroTransform.of(options));
-      }
-
-      if (options.getSchemaRegistryConnectionUrl() != null && options.getOutputDataset() == null) {
+      if (options.getSchemaFormat().equals("SINGLE_SCHEMA_FILE")) {
+
+        if (options.getConfluentAvroSchemaPath() != null && options.getOutputTableSpec() == null) {
```
an2x, line 276: What if getConfluentAvroSchemaPath() == null? Nothing happens?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
an2x: Also, what if (options.getMessageFormat().equals("AVRO_BINARY_ENCODING") && options.getBinaryAvroSchemaPath() == null)? Nothing happens?

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -248,27 +252,74 @@ public static PipelineResult runAvroPipeline(
...
+                      options.getUseAutoSharding()));
+        }
+      } else {
         throw new IllegalArgumentException(
-            "An output BigQuery dataset is required. It will be used to create tables per schema.");
-      }
-
-      if (options.getSchemaRegistryConnectionUrl() != null && options.getOutputDataset() != null) {
-        writeResult = kafkaRecords.apply(AvroDynamicTransform.of(options));
+            "There is a parameter misconfiguration, please set Schema Format parameter.");
```
an2x, line 322: I'd add more details to the error message, this message can be confusing if the schemaFormat param is set to something, but that value is not supported.

```
throw new IllegalArgumentException("Unsupported schemaFormat parameter value: " + options.getSchemaFormat())
```

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
```diff
@@ -58,8 +59,18 @@ public GenericRecord getDestination(ValueInSingleWindow<KV<GenericRecord, TableR
 
   @Override
   public TableDestination getTable(GenericRecord element) {
-    // tablename + record name (same across schemas) + schema id?
-    String bqQualifiedFullName = element.getSchema().getFullName().replace(".", "-");
+    Pattern sanitizePattern = Pattern.compile("[^\\w- ]", Pattern.UNICODE_CHARACTER_CLASS);
```
an2x, line 62: `Pattern.compile` is a relatively expensive operation, it's usually done only once:

```
  private static final Pattern SANITIZE_PATTERN = Pattern.compile("[^\\w- ]", Pattern.UNICODE_CHARACTER_CLASS);

  @Override
  public TableDestination getTable(GenericRecord element) {
    String sanitizedNamespace =
        SANITIZE_PATTERN.matcher(element.getSchema().getNamespace()).replaceAll("-");
    ....
```

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
```diff
@@ -58,8 +59,18 @@ public GenericRecord getDestination(ValueInSingleWindow<KV<GenericRecord, TableR
...
+    String sanitizedNamespace =
+        sanitizePattern.matcher(element.getSchema().getNamespace()).replaceAll("-");
+    String sanitizedName = sanitizePattern.matcher(element.getSchema().getName()).replaceAll("-");
+
+    Pattern multiHyphenRemovalPattern = Pattern.compile("-+", Pattern.UNICODE_CHARACTER_CLASS);
```
an2x, line 67: I think multiHyphenRemovalPattern doesn't have to be a separate pattern and both replacements can be done in a single regexp if you change the first regexp above from `[^\\w- ]` to `[^\\w- ]+` (i.e. replace 1+ characters matching this pattern with a single hyphen).

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/utils/BigQueryAvroUtils.java
```diff
@@ -108,7 +108,7 @@ public class BigQueryAvroUtils {
   private static final DateTimeFormatter DATE_AND_SECONDS_FORMATTER =
       DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss").withZoneUTC();
 
-  private static final String kafkaKeyField = "_key";
+  private static final String KAFKA_KEY_FIELD = "_key";
```
an2x, line 111: I think this constant is used in at least 3 different places. Probably makes sense to keep it only in one place. I can't think of a good place though. Maybe a separate `BigQueryConstants` class.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
```diff
@@ -58,8 +59,18 @@ public GenericRecord getDestination(ValueInSingleWindow<KV<GenericRecord, TableR
...
+    sanitizedNamespace = multiHyphenRemovalPattern.matcher(sanitizedNamespace).replaceAll("-");
+    sanitizedName = multiHyphenRemovalPattern.matcher(sanitizedName).replaceAll("-");
```
an2x, line 69: Consider moving the sanitation logic above to a utils class.

an2x (changes requested):

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: I was under the assumption that both GMK and Kafka would yield similar results `<bootstrap_server>:topic`

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: I will change it to if/else-if/else pattern. 

> Also, what if (options.getMessageFormat().equals("AVRO_BINARY_ENCODING") && options.getBinaryAvroSchemaPath() == null)? Nothing happens?

That is handled above, in this case it will never reach here.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: That case is caught at the beginning of this method.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -17,128 +17,134 @@
...
 
   @TemplateParameter.GcsReadFile(
-      order = 8,
-      parentName = "messageFormat",
-      parentTriggerValues = {"AVRO"},
+      order = 4,
+      groupName = "Source",
+      parentName = "schemaFormat",
+      parentTriggerValues = {"SINGLE_SCHEMA_FILE"},
       optional = true,
```
AnandInguva, line 111: make optional false

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -17,128 +17,134 @@
...
   String getSchemaRegistryConnectionUrl();
 
   void setSchemaRegistryConnectionUrl(String schemaRegistryConnectionUrl);
 
+  @TemplateParameter.GcsReadFile(
+      order = 12,
+      groupName = "Source",
+      parentName = "messageFormat",
+      parentTriggerValues = {"AVRO_BINARY_ENCODING"},
+      optional = true,
```
AnandInguva, line 137: make optional false

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -17,128 +17,134 @@
...
-      optional = true,
-      description = "The dead-letter table name to output failed messages to BigQuery",
-      helpText =
-          "BigQuery table for failed messages. Messages failed to reach the output table for different reasons "
-              + "(e.g., mismatched schema, malformed json) are written to this table. If it doesn't exist, it will"
-              + " be created during pipeline execution. If not specified, \"outputTableSpec_error_records\" is used instead.",
-      example = "your-project-id:your-dataset.your-table-name")
-  String getOutputDeadletterTable();
+  @TemplateParameter.KafkaTopic(
+      order = 1,
```
AnandInguva, line 59: Order 1 is repeated 2 times. idk the impact but lets make it unique

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -17,128 +17,134 @@
...
+      description = "Source Kafka Topic",
+      helpText = "Kafka Topic to read the input from.")
+  String getReadBootstrapServerAndTopic();
 
-  void setOutputDeadletterTable(String outputDeadletterTable);
+  void setReadBootstrapServerAndTopic(String value);
 
   @TemplateParameter.Enum(
-      order = 6,
+      order = 2,
```
AnandInguva, line 70: same for order 2

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -17,128 +17,134 @@
...
-      description = "BigQuery output table",
-      helpText =
-          "BigQuery table location to write the output to. The name should be in the format "
-              + "`<project>:<dataset>.<table_name>`. The table's schema must match input objects.")
-  String getOutputTableSpec();
+      regexes = {"[,:a-zA-Z0-9._-]+"},
+      description = "Kafka Bootstrap Server",
+      helpText = "Kafka Bootstrap Server list, separated by commas.",
+      example = "localhost:9092,127.0.0.1:9093",
+      hiddenUi = true)
```
AnandInguva, line 38: Why is this hidden? looking at template, we don't even need this one right? same goes for `getKafkaReadTopics`

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -95,9 +91,10 @@
     category = TemplateCategory.STREAMING,
     displayName = "Kafka to BigQuery",
     description =
-        "The Apache Kafka to BigQuery template is a streaming pipeline which ingests text data from Apache Kafka, executes a user-defined function (UDF), and outputs the resulting records to BigQuery. "
-            + "Any errors which occur in the transformation of the data, execution of the UDF, or inserting into the output table are inserted into a separate errors table in BigQuery. "
-            + "If the errors table does not exist prior to execution, then it is created.",
+        "The Apache Kafka to BigQuery template is a streaming pipeline which ingests text data from Apache Kafka, and outputs the resulting records to BigQuery. "
+            + "Any errors which occur in the transformation of the data, or inserting into the output table are inserted into a separate errors table in BigQuery. "
+            + "For any errors which occur in the transformation of the data, the roiginal records can be inserted into a seperate Kafka topic. The template supports "
```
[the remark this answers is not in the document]
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -156,24 +152,31 @@ public static void main(String[] args) throws IOException, RestClientException {
   public static PipelineResult run(KafkaToBigQueryFlexOptions options)
       throws IOException, RestClientException {
 
+    // Enable Streaming Engine
+    options.setEnableStreamingEngine(true);
+
+    // Enable Resource based billing
+    options.setDataflowServiceOptions(List.of("enable_streaming_engine_resource_based_billing"));
```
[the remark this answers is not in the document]
Dippatel98: Changed to append

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -222,13 +225,14 @@ public static PipelineResult runAvroPipeline(
...
       throw new IllegalArgumentException(
           "Avro schema is needed in order to read non confluent wire format messages.");
     }
-    if (options.getAvroFormat().equals("CONFLUENT_WIRE_FORMAT")
+    if (options.getMessageFormat().equals("AVRO_CONFLUENT_WIRE_FORMAT")
         && options.getSchemaRegistryConnectionUrl() == null
-        && options.getAvroSchemaPath() == null) {
+        && options.getConfluentAvroSchemaPath() == null) {
       throw new IllegalArgumentException(
           "Schema Registry Connection URL OR Avro schema is needed in order to read confluent wire format messages.");
```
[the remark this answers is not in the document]
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryDynamicDestination.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/utils/BigQueryAvroUtils.java
Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
Dippatel98: I am inheriting `KafkaReadOptions`, so I have to override and hide this, otherwise it would show up in the UI which would be confusing.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
Dippatel98: Its fine, since the other one is hidden anyway.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
Dippatel98: same as above

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -176,8 +234,91 @@ public interface KafkaToBigQueryFlexOptions
...
+  void setCreateDisposition(String createDisposition);
+
   @TemplateParameter.Boolean(
-      order = 13,
+      order = 21,
+      groupName = "Destination",
+      optional = true,
+      description = "Use BigQuery Storage Write API",
+      helpText =
+          "Do not use this parameter, Storage Write API is the only available method to write to BigQuery. "
```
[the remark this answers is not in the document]
Dippatel98: We have to override and mark it as hidden. And also, we do use it, we just don't respect what is set here, we by default set it to true.

AnandInguva:
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1581#discussion_r1610063225 - scratch this comment. I will ask the UI team and confirm

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
```diff
@@ -49,38 +51,61 @@ public interface KafkaReadOptions extends PipelineOptions {
...
-      description = "The Kafka offset to read from.",
+      groupName = "Source",
       enumOptions = {
-        @TemplateParameter.TemplateEnumOption("latest"),
         @TemplateParameter.TemplateEnumOption("earliest"),
+        @TemplateParameter.TemplateEnumOption("latest"),
       },
-      helpText = "Set the Kafka offset to earliest or latest(default)",
-      optional = true)
+      optional = false,
```
AnandInguva, line 59: make this optional??

Dippatel98: done

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
Dippatel98: Changed the logic based on the output.

an2x (approved):

AnandInguva (approved):

PR #1583, opened 2024-05-21
Add SERVICE_ACCOUNT to template parameter type

Amar3tto:


plugins/core-plugin/src/test/java/com/google/cloud/teleport/plugin/sample/AtoBOk.java
```diff
@@ -114,5 +114,13 @@ public interface AtoBOptions {
         example =
             "projects/project-foo/locations/us-central1/clusters/cluster-bar/topics/topic-baz")
     String getInputKafkaTopic();
+
+    @TemplateParameter.ServiceAccount(
+        order = 11,
```
damccorm, line 119: This is going to conflict with https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1513 (same order number)

Amar3tto: Resolved conflicts in all PRs (with different order numbers)

Amar3tto: Сreated a single PR

PR #1585, opened 2024-05-21
[BULK] Review configs

Deep1998:
Review config options for the bulk template.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
@@ -49,53 +49,18 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
 
   @TemplateParameter.Text(
       order = 3,
-      regexes = {"(^jdbc:[a-zA-Z0-9/:@.]+$)"},
+      regexes = {"(^jdbc:mysql://[a-zA-Z0-9/:@.]+$)"},
       groupName = "Source",
       description =
-          "Connection URL to connect to the source database host. Port number and connection properties must be supplied separately.",
-      helpText = "The JDBC connection URL string. For example, `jdbc:mysql://some-host`.")
-  String getSourceHost();
+          "Connection URL to connect to the source database host. Must contain the host, port and source db name. Can optionally contain connection properties Format: `jdbc:mysql://{host}:{port}/{dbName}?{parameters}`",
+      helpText =
+          "The JDBC connection URL string. For example, `jdbc:mysql://127.4.5.30:3306/my-db?unicode=true&characterEncoding=UTF-8`.")
+  String getSourceDbURL();
 
-  void setSourceHost(String host);
+  void setSourceDbURL(String url);
 
   @TemplateParameter.Text(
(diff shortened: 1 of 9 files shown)


shanemeister:
Te

VardhanThigle (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -23,13 +23,11 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
...
   @TemplateParameter.Text(
       order = 1,
-      optional = true,
       regexes = {"^.+$"},
       groupName = "Source",
       description = "Comma-separated Cloud Storage path(s) of the JDBC driver(s)",
       helpText = "The comma-separated list of driver JAR files.",
       example = "gs://your-bucket/driver_jar1.jar,gs://your-bucket/driver_jar2.jar")
-  @Default.String("")
   String getJdbcDriverJars();
```
VardhanThigle, line 31: Please retain the default.

The default "" implies, a user has compiled the jars into the build via say mvn imports, so no need to external class loading.

This is done for say UT, where we use derby dependency from mvn.

Deep1998: Done

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -71,6 +71,7 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
   /* TODO: (support Sharding, PG namespaces) */
   @TemplateParameter.Text(
       order = 5,
+      regexes = {"(^[^\\/?%*:|\"<>.]{1,64}$)"},
```
Deep1998, line 74: This regex is too expensive for dataflow. 
`ERROR: (gcloud.dataflow.flex-template.run) FAILED_PRECONDITION: (406846f638a637e8): Template metadata regex '(^[^\/?%*:|"<>.]{1,64}$)' was too large. Max size is 1000 but was 1031.`

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/JdbcIOWrapperConfig.java
```diff
@@ -134,12 +103,6 @@ public abstract static class Builder {
 
     public abstract Builder setJdbcDriverClassName(String value);
 
-    public abstract Builder setConnectionProperties(String value);
-
-    public abstract Builder setReconnectAttempts(Long value);
-
-    public abstract Builder setAutoReconnect(Boolean value);
```
bharadwaj-aditya, line 141: update the javadocs to suggest to add these parameters in the jdbc url

bharadwaj-aditya (changes requested):
There seem to be a few methods added here without tests. Can you add tests for those ?

PR #1590, opened 2024-05-22
Fix Neo4j query parameter creation

fbiville:
In cases where node label or relationship type mappings were colliding with source field names, the constant value of the label or type would overwrite the source field value.


ali-ince (approved):
LGTM

fbiville:
@liferoad this is an old bug that was recently reported by a colleague. It would be great if you could review it. Thanks a lot!

liferoad (approved):

PR #1593, opened 2024-05-22
Revise datastream to spanner ITs

Polber:
De-flakes the previously ignored Datastream to Spanner IT's and adds several variations to test various configurations of the template.

Also adds a CloudSQLResource manager for creating and operating on CloudSQL resources. The MySQL variants of these tests (oracle on the way) use a static CloudSQL instance and connect using a static Cloud Auth Proxy server (see #1591). This should help alleviate flakes related to Datastream being unable to connect to certain ports on the runners which was seemingly non-deterministic and difficult to debug.

---

Depends on #1465.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.it.gcp.cloudsql;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Custom class for the MySQL implementation of {@link CloudSqlResourceManager} abstract class.
+ *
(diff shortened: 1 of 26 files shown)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Custom class for the MySQL implementation of {@link CloudSqlResourceManager} abstract class.
+ *
+ * <p>The class supports one database, and multiple tables per database object. A database is *
+ * created when the container first spins up, if one is not given.
+ *
+ * <p>The class is thread-safe.
+ */
+public class CloudMySQLResourceManager extends CloudSqlResourceManager {
```
manitgupta, line 30: Do you want to add a unit test for this?
It's probably a good idea to have the framework components unit tested as well since multiple tests would use them in the future.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java
```diff
@@ -77,11 +75,15 @@ public class DataStreamToSpannerIT extends TemplateTestBase {
   private static final String MEMBER = "member";
   private static final String ENTRY_ADDED = "entry_added";
 
+  private String gcsPrefix;
+  private String dlqGcsPrefix;
+
   private static final List<String> COLUMNS = List.of(ROW_ID, NAME, AGE, MEMBER, ENTRY_ADDED);
 
-  private CustomMySQLResourceManager jdbcResourceManager;
+  private CloudSqlResourceManager jdbcResourceManager;
```
manitgupta, line 83: Nit: We should probably rename the variable as well for clarity that this CloudSQL being used behind the scenes.

manitgupta (approved):
Thanks for adding this, looks good! Added some minor comments.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java
Polber: Good point!

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
Polber: Yeah I'll add one

Polber:
Closing this in favor of #1661

PR #1598, opened 2024-05-23
Add SSL authentication to read from Kafka and Schema Registry

Dippatel98:


v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -361,4 +445,32 @@ public interface KafkaToBigQueryFlexOptions
   String getOutputDeadletterTable();
 
   void setOutputDeadletterTable(String outputDeadletterTable);
+
+  @TemplateParameter.Boolean(
```
AnandInguva, line 449: left over code?

Dippatel98: Oh! I added that initially for staging the template, let me remove that

an2x (approved):

PR #1606, opened 2024-05-28
auto table inference fix

bharadwaj-aditya:
Fix to ensure tables not present at source are not picked up during the migration.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -130,7 +130,7 @@ public final class JdbcIoWrapper implements IoWrapper {
     return sourceSchema.tableSchemas().stream()
         .filter(schema -> schema.tableName().equals(tableConfig.tableName()))
         .findFirst()
-        .orElseGet(null);
+        .orElseThrow();
   }
 
   static SourceSchema getSourceSchema(
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -31,7 +31,9 @@ import com.google.cloud.teleport.v2.source.reader.io.schema.SourceTableSchema;
 import com.google.cloud.teleport.v2.spanner.migrations.schema.SourceColumnType;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import javax.sql.DataSource;
 import org.apache.beam.sdk.io.jdbc.JdbcIO;
 import org.apache.beam.sdk.io.jdbc.JdbcIO.DataSourceConfiguration;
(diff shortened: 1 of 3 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -220,6 +221,21 @@ private static ImmutableList<TableConfig> autoInferTableConfigs(
...
 
+  static ImmutableList<String> getTablesToMigrate(
+      ImmutableList<String> configTables, ImmutableList<String> discoveredTables) {
+    List<String> tables = null;
+    if (configTables.isEmpty()) {
+      tables = discoveredTables;
+    } else {
+      tables =
+          configTables.stream()
+              .filter(t -> discoveredTables.contains(t))
```
VardhanThigle, line 232: [super nit] It might help with debugging, if we add a log in the filter lambda to indicate something on the lines of  `skipping table %s, as it was not discovered on the source.`

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -220,6 +221,21 @@ private static ImmutableList<TableConfig> autoInferTableConfigs(
     return tableConfigsBuilder.build();
   }
 
+  static ImmutableList<String> getTablesToMigrate(
```
VardhanThigle, line 224: [nit] would it be possible to make `getTablesToMigrate` private (and Visible for testing) or protected?
As of now this is called only from a private method within the same class.

VardhanThigle (approved):
Looks good to me, a couple of nits.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
bharadwaj-aditya: there is a logger right after this which prints out the table list for that purpose.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -174,10 +176,9 @@ static SourceSchema getSourceSchema(
    */
   private static ImmutableList<TableConfig> autoInferTableConfigs(
       JdbcIOWrapperConfig config, SchemaDiscovery schemaDiscovery, DataSource dataSource) {
-    ImmutableList<String> tables =
-        (config.tables().isEmpty())
-            ? schemaDiscovery.discoverTables(dataSource, config.sourceSchemaReference())
-            : config.tables();
+    ImmutableList<String> discoveredTables =
+        schemaDiscovery.discoverTables(dataSource, config.sourceSchemaReference());
+    ImmutableList<String> tables = getTablesToMigrate(config.tables(), discoveredTables);
```
aksharauke, line 181: This can probably be done earlier, and if the list is empty - then no need to even discover source schema and fail/stop fast. But just a micro-optimization, won't press on this.

aksharauke (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
bharadwaj-aditya: If the list is empty, we need to fallback to migrating all the possible tables.

VardhanThigle (approved):

PR #1609, opened 2024-05-28
Refactor pipeline options for Kafka to GCS template

AnandInguva:
Refactor KafkaToGcsFlex template pipeline options


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaDeadLetterQueueOptions.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
```
AnandInguva, line 7: remove this file

an2x (approved):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
```diff
@@ -76,42 +68,42 @@ public interface KafkaReadOptions extends PipelineOptions {
...
 
   @TemplateParameter.Enum(
-      order = 6,
+      order = 5,
       name = "kafkaReadAuthenticationMode",
       groupName = "Source",
       enumOptions = {
-        @TemplateParameter.TemplateEnumOption("NONE"),
-        @TemplateParameter.TemplateEnumOption("SASL_PLAIN"),
+        @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.SASL_MECHANISM),
```
an2x, line 91: I see SASL_MECHANISM and SASL_PLAIN used in different places in this PR, but I think it's the same thing?
Should we remove SASL_MECHANISM and keep SASL_PLAIN everywhere?

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
```diff
@@ -126,7 +126,7 @@ public static KafkaIO.Read<byte[], byte[]> readBytesFromKafka(
                 ByteArrayDeserializer.class, NullableCoder.of(ByteArrayCoder.of()))
             .withValueDeserializerAndCoder(ByteArrayDeserializer.class, ByteArrayCoder.of())
             .withConsumerConfigUpdates(config);
-    if (sslConfig != null) {
+    if (!sslConfig) {
```
an2x, line 129: We shouldn't need the `sslConfig` parameter at all. I suggest just setting `.withConsumerFactoryFn` unconditionally. This Fn only post-processed the config and replaces remote files/secrets with local files. If there is nothing to replace in the config, it's going to be a no-op. If there is something to replace, it'll replace that.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/utils/KafkaConfig.java
```diff
@@ -0,0 +1,108 @@
...
+      String authMode,
+      String keystoreLocation,
+      String truststoreLocation,
+      String truststorePasswordSecretId,
+      String keystorePasswordSecretId,
+      String keyPasswordSecretId,
+      String usernameSecretId,
+      String passwordSecretId) {
+    Map<String, Object> properties = new HashMap<>();
+    if (authMode == null) {
```
an2x, line 72: Need this check here:
```suggestion
    if (authMode == null || authMode.equals(KafkaAuthenticationMethod.NONE)) {
```
Otherwise it'll throw the "Authentication method not supported" exception below for authMode = NONE.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/values/KafkaAuthenticationMethod.java
```diff
@@ -20,7 +20,7 @@ public class KafkaAuthenticationMethod {
   // in @Template.Parameter annotations, and annotations can only have constant attributes.
 
   public static final String NONE = "NONE";
-  public static final String SASL_PLAIN = "SASL_PLAIN";
+  public static final String SASL_PLAIN = "SASL_SSL";
```
an2x, line 23: Is "SASL_SSL" a typo? I haven't seen this constant before.
If it is, please fix it.
If it's not, I suggest keeping the variable name the same (`String SASL_SSL = "SASL_SSL"`).
Otherwise SASL_PLAIN = SASL_SSL and SASL_MECHANISM = PLAIN is very confusing.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
```diff
@@ -198,30 +200,25 @@ public static PipelineResult run(KafkaToBigQueryFlexOptions options)
...
      */
 
-    ImmutableMap<String, Object> kafkaConfig =
+    ImmutableMap<String, Object> kafkaConfig = ImmutableMap.<String, Object>builder().build();
+
+    kafkaConfig =
         ImmutableMap.<String, Object>builder()
-            .putAll(KafkaCommonUtils.configureKafkaOffsetCommit(options))
+            .putAll(kafkaConfig)
+            .putAll(KafkaConfig.fromReadOptions(options))
```
an2x, line 208: I think this whole block can be replaced with `Map<String, Object> kafkaConfig = KafkaConfig.fromReadOptions(options)`?

`.putAll(kafkaConfig)` currently doesn't do anything because `kafkaConfig` is an empty map, if I'm reading this correctly.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
```diff
@@ -136,4 +128,81 @@ public interface KafkaReadOptions extends PipelineOptions {
...
+      optional = true,
+      name = "sourceSSL",
+      groupName = "Source",
+      parentName = "kafkaReadAuthenticationMode",
+      parentTriggerValues = {KafkaAuthenticationMethod.SSL},
+      description = "Truststore File Location",
+      helpText =
+          "Location of the jks file in Cloud Storage with SSL certificate to verify identity.",
+      example = "gs://your-bucket/truststore.jks")
+  String getSourceTruststoreLocation();
```
an2x, line 158: Should all getSrouce* params be renamed to getKafkaRead* params for consistency?

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToGcsFlex.java
```diff
@@ -251,23 +160,20 @@ public static PipelineResult run(KafkaToGcsOptions options) throws UnsupportedOp
...
+    if ((options.getKafkaReadUsernameSecretId().isBlank()
+            && !options.getKafkaReadPasswordSecretId().isBlank())
+        || (options.getKafkaReadPasswordSecretId().isBlank()
+            && !options.getKafkaReadUsernameSecretId().isBlank())) {
       throw new IllegalArgumentException(
           "Both username secret ID and password secret ID should be provided together or left null.");
     }
 
-    if (!SecretVersionName.isParsableFrom(options.getUserNameSecretID())) {
+    if (!SecretVersionName.isParsableFrom(options.getKafkaReadUsernameSecretId())) {
```
an2x, line 171: I think this check will throw the "Provided Secret Username ID must be in the form ..." exception even if the authentication mode doesn't require a username (NONE or SSL).

an2x (changes requested):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/values/KafkaAuthenticationMethod.java
AnandInguva: From the documentation, GMK supports SASL/PLAIN - which is plain username/password encrypted with TLS/SSL. It should be set like this - 
```
sasl.mechanism=PLAIN
# Configure SASL_SSL if TLS/SSL encryption is enabled, otherwise configure SASL_PLAINTEXT
security.protocol=SASL_SSL
```

We were setting SASL_PLAIN which not a valid value. I agree we need a good naming.

AnandInguva: More info - https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_plain.html

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToGcsFlex.java
AnandInguva: This method was not used. Removed this method.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlex.java
AnandInguva: Done

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/utils/KafkaConfig.java
AnandInguva: Done

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
AnandInguva: Done

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
AnandInguva: Based on the offline discussion
The enum would contain 
1. SASL_PLAIN
2. TLS
3. NONE

Is this right?

AnandInguva:
cc : @Surabhiad

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/utils/KafkaConfig.java
```diff
@@ -87,10 +87,9 @@ private static Map<String, Object> from(
           SslConfigs.SSL_KEY_PASSWORD_CONFIG,
           FileAwareFactoryFn.SECRET_MANAGER_VALUE_PREFIX + keyPasswordSecretId);
       properties.put(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, "");
-    } else if (authMode.equals(KafkaAuthenticationMethod.SASL_MECHANISM)) {
-      properties.put(SaslConfigs.SASL_MECHANISM, KafkaAuthenticationMethod.SASL_MECHANISM);
-      properties.put(
-          CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, KafkaAuthenticationMethod.SASL_PLAIN);
+    } else if (authMode.equals(KafkaAuthenticationMethod.PLAIN)) {
+      properties.put(SaslConfigs.SASL_MECHANISM, KafkaAuthenticationMethod.PLAIN);
```
an2x, line 91:
```suggestion
      properties.put(SaslConfigs.SASL_MECHANISM, "PLAIN");
```
I suggest just hard-coding "PLAIN" here, or introducing another constant. What's set as `sasl.mechanism` property value in Kafka config doesn't have to match the enum value we use in the template's Authentication Mode parameter (and as discussed offline, it won't: we'll use `SASL_PLAIN` as the enum value, but here we have to set the property to `PLAIN`).

an2x (changes requested):

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/options/KafkaToBigQueryFlexOptions.java
```diff
@@ -25,127 +27,26 @@
...
-      parentTriggerValues = {"AVRO_BINARY_ENCODING"},
-      optional = true,
-      description = "Cloud Storage path to the Avro schema file",
-      helpText = "Cloud Storage path to Avro schema file. For example, gs://MyBucket/file.avsc.")
-  @Default.String("")
-  String getBinaryAvroSchemaPath();
-
-  void setBinaryAvroSchemaPath(String schemaPath);
+      description = "Authentication Mode",
+      helpText = "Kafka read authentication mode. Can be NONE or PLAIN")
```
AnandInguva, line 44:
```suggestion
      helpText = "Kafka read authentication mode. Can be NONE or SASL_PLAIN")
```

v2/kafka-to-gcs/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToGcsFlex.java
```diff
@@ -54,53 +48,35 @@
...
         },
-        description = "File format of the desired output files. (TEXT, AVRO or PARQUET)",
-        helpText =
-            "The file format of the desired output files. Can be TEXT, AVRO or PARQUET. Defaults to TEXT")
-    @Default.String("TEXT")
-    String getOutputFileFormat();
-
-    void setOutputFileFormat(String outputFileFormat);
+        description = "Authentication Mode",
+        helpText = "Kafka read authentication mode. Can be NONE or PLAIN")
```
AnandInguva, line 72:
```suggestion
        helpText = "Kafka read authentication mode. Can be NONE or SASL_PLAIN")
```

an2x (approved):

PR #1619, opened 2024-05-31
[BULK] Handle BIT type,Timestamp overflow  and add stack trace to DLQ.

Deep1998:


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java
```diff
@@ -156,8 +158,15 @@ static Boolean avroFieldToBoolean(Object recordValue, Schema fieldSchema) {
     if (recordValue == null) {
       return null;
     }
+    String val = recordValue.toString();
```
VardhanThigle, line 161: What would happen if we skip this stringification, and just cast the object to a java boolean?
When we stringify, there is a possibility that the underlying library might return "0" or "false" (similarly "1" or "true"), 
For the documentation, it looks like avro's  boolean stringifies to "0" and "1" and avro's boolean within a json (does not affect this part of the code) stringifies to "false" or "true"


If we depend on the avro stringification (no harm), and why not just return `val.equals("1")`?

```
 if (Arrays.asList("0", "1").contains(val)) {
      return val.equals("1");
    }
    // BooleanUtils.toBoolean() never throws exception, so we don't need to catch it.
    return BooleanUtils.toBoolean(val);
    ```
    Makes some one not so familier (like me) feel there's some kind of undefined behavior in the last return.

Comment (Deep1998) [reply v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java:161]: Typecasting to boolean works for boolean source types only. Converting it to string then boolean allows handling long, string, boolean, char et. This stringification handles more cases.

`why not just return val.equals("1")` We could do that also, but boolean utils also handle cases like true, false, t, f etc. 

The combination of the 2 means if we can infer a 0 or 1 use that to decide else rely on boolean utils.

I get it is not intuitive but it supports alot of cases this way. Alot of these cases might be unnecessary cases as well, but did it because its a small effort to cover them.

I'll add a detailed clarifying comment if we decide to keep it this way.

File: v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/RowContext.java [hunk 2]
```diff
@@ -48,4 +50,10 @@ public abstract static class Builder {
 
     public abstract RowContext build();
   }
+
+  public String getStackTraceString() {
+    StringWriter sw = new StringWriter();
+    err().printStackTrace(new PrintWriter(sw));
```
bharadwaj-aditya, line 56: check of error not null

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -169,7 +169,7 @@ protected FailsafeElement<String, String> rowContextToDlqElement(RowContext r) {
       json.put(f.name(), value == null ? null : value.toString());
     }
     return FailsafeElement.of(json.toString(), json.toString())
-        .setErrorMessage("TransformationFailed: " + String.valueOf(r.err()));
+        .setErrorMessage("TransformationFailed: " + r.err() + "\n" + r.getStackTraceString());
```
bharadwaj-aditya, line 172: it would be better to have a different field called error stacktrace and just save this there

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java
```diff
@@ -152,12 +154,23 @@ static Map<Type, AvroToValueFunction> getPgMap() {
...
-    return BooleanUtils.toBoolean(recordValue.toString());
+    String val = recordValue.toString();
+    // If the value can be converted to a "0" or a "1", rely on this to map to boolean. This could
+    // be used for strings and long data types.
+    // For ex: BIT(1) -> Unified type long -> Spanner Boolean uses this path.
+    if (Arrays.asList("0", "1").contains(val)) {
+      return val.equals("1");
+    }
+    // Rely on booleanUtils to map the string to a bool if it was not a "0" or "1".
+    return BooleanUtils.toBoolean(val);
```
bharadwaj-aditya, line 173: nit: it would be good to add a description of the behaviour in these scenarios

bharadwaj-aditya (commented):
Looks fine overall - but some description on the cases handled and when this conversion is done would help to clarify

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
Deep1998: Hmm but we already have that in our err throwable, storing twice would be redundant right

VardhanThigle (approved):

bharadwaj-aditya (approved):
LGTM

PR #1623, opened 2024-06-01
Adding init seq to set session timezone.

VardhanThigle:
Adding Initialization Sequence for the JDBC connection.
## Summary of Changes
Session Timezone: The PR sets the session time zone to UTC by default to retrieve timestamp in UTC. This is asper the contract exposed by the Reader to always output timestamp in UTC (irrespective of the source settings).The most idomatic way to achieve this via jdbc would be to pass a Calendar object initialized to UTC to [java.sql.ResultSet.getTimestamp(String, Calendar)](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/sql/ResultSet.html#getTimestamp-java.lang.String-java.util.Calendar-) api (which is what we do in [MysqlJdbcValueMappings](https://github.com/VardhanThigle/DataflowTemplates/blob/main/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java#L80)), but due to bugs like [Bug#95644](https://bugs.mysql.com/bug.php?id=95644) , [Bug#96276](https://bugs.mysql.com/bug.php?id=96276), [Bug#93444](https://bugs.mysql.com/bug.php?id=93444), etc. in the older drivers, we are also setting the session timezone to UTC.
## Testing
This change has been tested on an end to end migration setup where MySQL server was in IST timezone and the timestamp was migrated to spanner by verifying that the migrated timestamp obtained from Spanner was as per the correct timezone offsets applied.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -280,6 +280,10 @@ public final class JdbcIoWrapper implements IoWrapper {
                 StaticValueProvider.of(config.sourceDbURL()))
             .withMaxConnections(Math.toIntExact(config.maxConnections()));
 
+    if (!config.sqlInitSeq().isEmpty()) {
+      dataSourceConfig = dataSourceConfig.withConnectionInitSqls(config.sqlInitSeq());
+    }
+
     if (config.jdbcDriverJars() != null && !config.jdbcDriverJars().isEmpty()) {
       dataSourceConfig = dataSourceConfig.withDriverJars(config.jdbcDriverJars());
     }
@@ -289,6 +293,8 @@ public final class JdbcIoWrapper implements IoWrapper {
     if (!config.dbAuth().getPassword().get().isBlank()) {
       dataSourceConfig = dataSourceConfig.withPassword(config.dbAuth().getPassword().get());
     }
+
+    LOG.info("Final DatasourceConfiguration: {}", dataSourceConfig);
     return dataSourceConfig;
   }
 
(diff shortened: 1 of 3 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -289,6 +293,8 @@ private static DataSourceConfiguration getDataSourceConfiguration(JdbcIOWrapperC
     if (!config.dbAuth().getPassword().get().isBlank()) {
       dataSourceConfig = dataSourceConfig.withPassword(config.dbAuth().getPassword().get());
     }
+
+    LOG.info("Final DatasourceConfiguration: " + dataSourceConfig);
```
Deep1998, line 297: Does this print the DataSourceConfig contents or just the address? SOme log statements i added in the past just print the memory address. 
Even if it prints the address we shouldnt fix it now ig due to timelines. We should however track it.

Deep1998 (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
bharadwaj-aditya: avoid using + in logs.
use the logging structure of coding the string with {} and passing the parameter.

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: This does print the config correctly and not just the address (have verified it in a real migration).
Fixed the `+` to use `{}` in current push.

bharadwaj-aditya (approved):
LGTM

PR #1630, opened 2024-06-05
[BULK] Integration tests for all data types

Deep1998:


bharadwaj-aditya (approved):
LGTM

nit: The tests would be easier to read if the loading and reading was done in the same format, i.e. code to load data into MySQL as opposed to the SQL file.

PR #1641, opened 2024-06-07
[BULK] Pass through unsupported types as null

Deep1998:
Currently, unsupported types (like spatial types in mysql) throw error and are written to DLQ. This is inconsistent with the live template. This change passes a null values for those types and tries to write the row to Spanner anyways.


VardhanThigle (approved):

Deep1998:
Changes merged.

PR #1643, opened 2024-06-07
Fix bug in Kafka To BigQuery Template

Dippatel98:


AnandInguva (approved):

AnandInguva:
> @AnandInguva @Dippatel98 - when do you expect this change to be released?

We expect it to be available on June 18th.

PR #1644, opened 2024-06-07
Revise Datastream to BigQuery IT's

Polber:


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigquery/conditions/BigQueryRowsCheck.java
```diff
@@ -77,7 +90,15 @@ public static Builder builder(BigQueryResourceManager resourceManager, TableId t
   }
 
   public Long getRowCount() {
-    return resourceManager().getRowCount(tableId().getTable());
+    try {
+      return resourceManager().getRowCount(tableId().getTable());
+    } catch (Exception e) {
+      if (ExceptionUtils.containsMessage(e, "Not found: Table")) {
```
damccorm, line 96: It seems kind of odd to handle this exception at this level, especially since there's only one caller who does conditional logic on top anyways. Could we more cleanly just handle this in `check` instead?

damccorm (commented):
Overall LGTM, just one minor suggestion

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigquery/conditions/BigQueryRowsCheck.java
Polber: Are you saying like this?

damccorm: Yeah, I think that is much cleaner - thanks

damccorm (approved):

PR #1649, opened 2024-06-11
Add OauthBearer auth support for Managed Kafka templates.

AnandInguva:


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/auth/package-info.java
Dippatel98, line 1: Add the copyright comment in this file as well.

AnandInguva: Done

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
```diff
@@ -53,18 +53,19 @@ public interface SchemaRegistryOptions extends PipelineOptions {
...
-          "The Kafka schema format. Can be provided as SINGLE_SCHEMA_FILE or SCHEMA_REGISTRY. "
-              + "If SINGLE_SCHEMA_FILE is specified, all messages should have the schema mentioned in the avro schema file. "
-              + "If SCHEMA_REGISTRY is specified, the messages can have either a single schema or multiple schemas.")
+          "The Kafka schema source. It can be provided as SINGLE_SCHEMA_FILE or SCHEMA_REGISTRY. "
+              + "If SINGLE_SCHEMA_FILE is specified, all the kafka messages should have the schema mentioned in the avro schema file. "
+              + "If SCHEMA_REGISTRY is specified, the kafka messages can have either a single schema or multiple schemas that are registered"
+              + "in the schema registry.")
   @Default.String(SchemaFormat.SINGLE_SCHEMA_FILE)
-  String getSchemaFormat();
+  String getSchemaSource();
```
Dippatel98, line 61: Once the templates are released which they are now, I don't think we can change the parameters like this, since it breaks backwards compatibility. You would have to mark the older parameter as `@Deprecated` and add the newer one, something like [this](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQuery.java#L204).

AnandInguva: Okay, Makes sense. I will decouple these changes from Oauth changes.

Dippatel98 (approved):

Dippatel98 (approved):

PR #1652, opened 2024-06-12
[BULK] Remove log param

Deep1998:
The existing log setting params already work.
They can be set via `--defaultSdkHarnessLogLevel` and `--defaultWorkerLogLevel` param.

Note: when launching via gcloud, these params need to be passed under the `--parameters` field. These are not independant gcloud flags.


bharadwaj-aditya (approved):
LGTM

Abacn:
The .md files are not updated accordingly after this change, and this also breaks existing user who used these params.

PR #1653, opened 2024-06-12
CassandraToBigtable classic template feature: Cassandra Writetime replication 

georgecma:
Add Cassandra writetime replication feature to CassandraToBigtable classic template. 

Cassandra writetime can now be replicated to Bigtable as Bigtable cell timestamp. For this feature to work, users have to upload a json-like schema file to a GCS bucket location first for the table they wish to replicate writetimes for. The command to generate this schema file is:

```
cqlsh -e "select json * from system_schema.columns
                where keyspace_name='$CASSANDRA_KEYSPACE'
                and table_name='$CASSANDRA_TABLE'"
    > column_schema.json
```
Then, to upload the file:
```
gcloud storage cp column_schema $WRITETIME_CASSANDRA_COLUMN_SCHEMA
```
The schema GCS file path should then be set as $WRITETIME_CASSANDRA_COLUMN_SCHEMA to be parsed by the template at run time. 
 
The template replication behavior now sets Bigtable cell time as replication time (i.e. now) instead of epoch start time previously. SetZeroTimestamp is the backwards compatibility option to set 0 as the timestamp. 

@GoogleCloudPlatform/dataflow-templates-wg


v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraToBigtable.java
```diff
@@ -213,7 +238,12 @@ public static void main(String[] args) {
             .withTable(options.getCassandraTable())
             .withMapperFactoryFn(cassandraObjectMapperFactory)
             .withEntity(Row.class)
-            .withCoder(SerializableCoder.of(Row.class));
+            .withCoder(SerializableCoder.of(Row.class))
+            .withQuery(
+                new CassandraWritetimeQueryProvider(
```
georgecma, line 243: CassandraWritetimeQueryProvider is not getting unpacked in directRunner.

georgecma: There's a bug in Beam cassandraIO: https://github.com/apache/beam/issues/31622

I'm working on a fix.

georgecma: https://github.com/apache/beam/pull/31623 fix waiting for review and Beam release. Integration testing can continue after this beam release.

georgecma:
@GoogleCloudPlatform/dataflow-templates-wg

v1/src/main/java/com/google/cloud/teleport/util/GCSAwareValueProvider.java
```diff
@@ -51,17 +58,19 @@ public boolean isAccessible() {
...
 
-    if (originalValue != null && originalValue.startsWith("gs://")) {
+    if (returnValue != null && returnValue.startsWith("gs://")) {
       try {
-        return new String(GCSUtils.getGcsFileAsBytes(originalValue), StandardCharsets.UTF_8);
+        returnValue = new String(GCSUtils.getGcsFileAsBytes(returnValue), StandardCharsets.UTF_8);
       } catch (IOException e) {
         throw new RuntimeException(
-            "Error resolving ValueProvider from Cloud Storage path " + originalValue, e);
+            "Error resolving ValueProvider from Cloud Storage path " + returnValue, e);
```
Polber, line 68: This error message was intended to print failed path upon errors, but with this change, it will likely print null or empty string rather than the original path value.

This should be refactored to store `originalValue`

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigtable/matchers/BigtableAsserts.java
```diff
@@ -67,4 +68,28 @@ public static List<Map<String, Object>> bigtableRowsToRecords(Iterable<Row> rows
...
+  }
+
+  public static String doubleQuote(Object string) {
+    return '"' + (String) string + '"';
+  }
+
+  // Convert Instant timestamp to epoch microsecond format.
+  public static long toEpochMicros(Instant time) {
+    return time.toEpochMilli() * 1000;
+  }
```
Polber, line 94: This class is intended for truthmatchers for making resource-dependent (in this case Bigtable) asserts.

Let's move these helper methods to the IT file they are being used in, or at the very least, create a utils class in the module that these methods are intended for (`v1/src/test/bigtable/...`)

Polber (changes requested):
Thanks for the contribution! I left a couple small comments for core templates related files. Please tag a member from the BigTable team to review BigTable-related changes

v1/src/main/java/com/google/cloud/teleport/util/GCSAwareValueProvider.java
georgecma: I refactored originalValue to returnValue to better reflect the variable potentially being changed in resolve(). 
This error will print out the GCS path - happy to change this variable name if there's some better name!

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigtable/matchers/BigtableAsserts.java
georgecma: Moved functions to CassandraToBigtableIT since that's the only place that they are used in.

v1/src/main/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFn.java
```diff
@@ -148,41 +180,116 @@ public void processElement(
...
+        Matcher matcher = writetimeColumn.matcher(field.getName());
+        matcher.matches();
+        String columnName = matcher.group(1);
+
+        // Get timestamp and normalize as necessary.
+        Long rawTimestamp = (Long) row.getValue(field.getName());
+        Long normalizedTimestamp = normalizeToBigtableTimestamp(rawTimestamp);
+
+        columnTimestamps.put(columnName, normalizedTimestamp);
+      }
```
vermas2012, line 252: we are silently ignoring the timestamp column here, should we be raising the flag here? There are 2 cases where this is not good:

1. User's column names have special write time token which should be avoided, how do we inform them that its not a good practice? 
2. There is some bug in the regex or column name has special character that we are not handling well. We should raise a warning here, instead of silently ignoring the timestmap.

v1/src/main/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFn.java
```diff
@@ -148,41 +180,116 @@ public void processElement(
...
+      throw new IllegalArgumentException("Timestamp " + timestamp + " is smaller than 0.");
+    }
+    Long newTimestamp = timestamp;
+    Long nowInMillis = Instant.now().toEpochMilli();
+    Long nowInMicros = nowInMillis * 1000;
+
+    if (timestamp <= nowInMillis) {
+      // Timestamp is in millisecond format. If so, pad 3 more 0s to the end.
+      newTimestamp *= 1000;
+    } else if (timestamp <= nowInMicros) {
```
vermas2012, line 277: This is a flawed assumption, we can't be sure of this. If I have a cell timestamp in in millis for future (lets say a day), you assume its in micro seconds and trim its last 3 digits. we lose last 3 digits of the millis and format is still wrong.

v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraColumnSchema.java
```diff
@@ -0,0 +1,139 @@
...
+   * <p>This class expects only one table's columns as input.
+   *
+   * @param schemaString The parsed Cassandra columns schema string for one table.
+   */
+  CassandraColumnSchema(String schemaString) {
+    primaryKeyColumns = new ArrayList<>();
+    regularColumns = new ArrayList<>();
+
+    for (String line : schemaString.split("\\r?\\n")) {
+      // The schema file may not be a strict JSON. Skip lines that are not JSON rows.
```
vermas2012, line 68: why? does cassandra not guarantee the output file format?

v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraColumnSchema.java
```diff
@@ -0,0 +1,139 @@
...
+
+      // Parse primary and regular columns.
+      if (!row.has(POSITION) || !row.has(COLUMN_NAME_COLUMN)) {
+        throw new IllegalArgumentException("Cassandra schema row has no position or column_name.");
+      }
+      String rowColumnName = row.get(COLUMN_NAME_COLUMN).getAsString();
+      if (row.get(POSITION).getAsInt() > -1) {
+        primaryKeyColumns.add(rowColumnName);
+      } else {
+        regularColumns.add(rowColumnName);
```
vermas2012, line 99: so regular column names have negative position values, and only the primary key have non negative position values?

v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraColumnSchema.java
vermas2012, line 1: I am very concerned about us using the regex for this parsing, if CQL evolves the output of the schema command, how do we adapt to it before a customer complains? Can you check if there is a cassandra library to parse the schema?

v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraToBigtable.java
```diff
@@ -182,17 +184,39 @@ public interface Options extends PipelineOptions {
...
+        helpText =
+            "GCS path to schema to copy Cassandra writetimes to Bigtable. The command to generate this schema is ```cqlsh -e \"select json * from system_schema.columns where keyspace_name='$CASSANDRA_KEYSPACE' and table_name='$CASSANDRA_TABLE'`\" > column_schema.json```. Set $WRITETIME_CASSANDRA_COLUMN_SCHEMA to a GCS path, e.g. `gs://$BUCKET_NAME/column_schema.json`. Then upload the schema to GCS: `gcloud storage cp column_schema.json $WRITETIME_CASSANDRA_COLUMN_SCHEMA`. Requires Cassandra version 2.2 onwards for JSON support.")
+    ValueProvider<String> getWritetimeCassandraColumnSchema();
+
+    void setWritetimeCassandraColumnSchema(ValueProvider<String> writetimeCassandraColumnSchema);
+
+    @TemplateParameter.Boolean(
+        order = 12,
+        optional = true,
+        description = "If true, sets Bigtable cell timestamp to 0 if writetime is not present.",
```
vermas2012, line 202: add the else: defaults to writetime=now.

v1/src/test/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFnTest.java
```diff
@@ -586,6 +670,375 @@ public void processElementNullColumn() {
...
+        ImmutableList.of(KV.of(ByteString.copyFrom(Bytes.toBytes("rowkeyvalue")), mutations));
+
+    PAssert.that(bigtableRows).containsInAnyOrder(expectedBigtableRows);
+
+    // Run the pipeline
+    pipeline.run();
+  }
+
+  @Test
+  public void processElementWithBackfillNowTimestamp() {
```
vermas2012, line 824: Add a test with 0 timestamp
Add a test with timestamp = System.currentTimeInMillis() + 1000 and see what the timestamp comes out in CBT, is it same?

vermas2012 (changes requested):

v1/src/main/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFn.java
georgecma: Added a warning

v1/src/main/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFn.java
georgecma: Added a 100 years of buffer tiem to parse millisecond timestamps.

v1/src/main/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFn.java
```diff
@@ -148,41 +180,116 @@ public void processElement(
...
+      // Time is in microsecond format. If so, set last 3 digits to 0.
+      newTimestamp /= 1000;
+      newTimestamp *= 1000;
+    } else {
+      // In this case, the timestamp is set in the future. It is not recommended to do this due to
+      // implications around Bigtable garbage collection. However, there is nothing wrong with this
+      // technically so this action is not blocked.
+      LOG.warn(
```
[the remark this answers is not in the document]
georgecma: The case here now is that this time is > 100 years into the future in microsecond timestamps. At this point, I think it'll be more important to call out this behavior than to modify it since we're not sure what the intention is anymore.

v1/src/main/java/com/google/cloud/teleport/bigtable/CassandraColumnSchema.java
georgecma: Correct - documentation on this is non-existent but the code points to this: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/schema/ColumnMetadata.java#L92
This has been the case at least for the last 9 years (at least by blame history)

v1/src/test/java/com/google/cloud/teleport/bigtable/BeamRowToBigtableFnTest.java
georgecma: Added tests for these scenarios.

v1/README_Cassandra_To_Cloud_Bigtable.md
```diff
@@ -36,6 +36,8 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 * **defaultColumnFamily** : The name of the column family of the Bigtable table. The default value is default.
 * **rowKeySeparator** : The separator used to build row-keys. The default value is '#'.
 * **splitLargeRows** : The flag for enabling splitting of large rows into multiple MutateRows requests. Note that when a large row is split between multiple API calls, the updates to the row are not atomic. .
+* **writetimeCassandraColumnSchema** : GCS path to schema to copy Cassandra writetimes to Bigtable. The command to generate this schema is ```cqlsh -e "select json * from system_schema.columns where keyspace_name='$CASSANDRA_KEYSPACE' and table_name='$CASSANDRA_TABLE'`" > column_schema.json```. Set $WRITETIME_CASSANDRA_COLUMN_SCHEMA to a GCS path, e.g. `gs://$BUCKET_NAME/column_schema.json`. Then upload the schema to GCS: `gcloud storage cp column_schema.json $WRITETIME_CASSANDRA_COLUMN_SCHEMA`. Requires Cassandra version 2.2 onwards for JSON support.
```
vermas2012, line 39: > cqlsh -e "select json * from system_schema.columns where keyspace_name='$CASSANDRA_KEYSPACE' and table_name='$CASSANDRA_TABLE'

What happens if you execute this query from the cassandra client and interpret the results from resultset? I find it hard to behelive that we can query using cqlsh but can't fetch this data via an API. 

I have suffered a lot from maintaining compatiblity of hbase clients. I don't want to add similar pain by parsing strings in cassandra. Lets see if there is any way to do this via API contract

georgecma: Discussed offline - will keep the current approach. 

The result of the cqlsh command will be like [the following](https://docs.datastax.com/en/cql-oss/3.x/cql/cql_using/useQueryJSON.html): 
```
cqlsh:cycling> select json name, checkin_id, timestamp from checkin;
 [json]
------------------------------------------------------------------------------------------------------------------
 {"name": "BRAND", "checkin_id": "50554d6e-29bb-11e5-b345-feff8194dc9f", "timestamp": "2016-08-28 21:45:10.406Z"}
  {"name": "VOSS", "checkin_id": "50554d6e-29bb-11e5-b345-feff819cdc9f", "timestamp": "2016-08-28 21:44:04.113Z"}
(2 rows)
```

Note the prefix `[json] ---...` and suffix `(2 rows)` strings that will interfere with strict parsing. I wasn't able to find a library that does fuzzy JSON matching, so we have our own JSON parser for this.

Alternatively, datastax does have a systems querier to get system column table info. The ideal scenario would be to slot a datastax connector as the first step in the pipeline to process Cassandra table schema, and then feed the results to CassandraIO. However:
1. CassandraIO [does not allow any processors before itself in a pipeline](https://beam.apache.org/releases/javadoc/2.29.0/org/apache/beam/sdk/io/cassandra/CassandraIO.Read.html)
2. It is not possible to preprocess steps and dynamically construct a pipeline in a V1 template

In a future iteration, we should either loosen CassandraIO.Read or upgrade this template to a V2 template so we can have the writetimeCassandraColumnSchema automatically processed. Added a TODO for myself to explore these approached at a future time.

vermas2012 (approved):
Please go through all the warnings of lines not covered by tests and add any relevant Tests.

Abacn (commented):
ack. will merge after fixing the format (mv. spotless:apply)

vermas2012 (approved):

Abacn:
cc: @Polber looks like merge is still blocked pending your approval

vermas2012:
@Polber Is there anything pending from the CBT side? can we merge this change?

Polber (approved):
Thanks!

PR #1654, opened 2024-06-13
[Bulk] Add table name to source row

Deep1998:
Add table name field to source row


bharadwaj-aditya (approved):
Looks a lot cleaner now. Thanks !
LGTM

PR #1656, opened 2024-06-13
[BULK] Add SourceSchemaReference to source row

Deep1998:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRow.java
```diff
@@ -29,6 +30,20 @@
 @AutoValue
 public abstract class SourceRow implements Serializable {
 
+  /**
+   * Get the namespace of the database for the row.
+   *
+   * @return namespace
+   */
+  public abstract String namespace();
```
VardhanThigle, line 38: Please mark this nullable.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRow.java
```diff
@@ -29,6 +30,20 @@
...
+
+  /**
+   * Get the database name for the row. This is the logical database name for sharded setups.
+   *
+   * @return dbName
+   */
+  public abstract String dbName();
```
VardhanThigle, line 45: Just exploring, was there a problem you faced with using `SourceSchemaReference` here as in what's the reason behind separate out the name space and `dbName`?

`SourceSchemaReference` is intended to be a unique reference to source db across mySQL, PG etc If there's a slight addition needed for other sources, it will go in there.

Deep1998: No issues really, I just separated out because I didnt want to introduce SourceSchemaReference into SourceRow (If tomorrow we want to dump entire sourceRow somewhere like dlq etc). Even datastream structures (like datastream json) has these metadata fields separate (or clubbed inside a metadata key). Can change if needed

VardhanThigle: As we synced over chat, `SourceSchemaReference` is intended to be a unique reference to source db across mySQL, PG etc If there's a slight addition needed for other sources, it will go in there, so it's might be better not to separate it out.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/row/SourceRow.java
Deep1998: No longer needed since we using SourceSchemaRef now.

VardhanThigle (approved):

PR #1657, opened 2024-06-14
Adding support of uniform partitioning for numeric types and composite keys.

VardhanThigle:
# Adding support for `ReadWithUniformPartition` for numeric types and composite keys.
## Summary
`ReadWithUniformPartition` is almost equivalent in the basic contract with JDBCIO.readWithPartition.

In addition to `JDBCIO.readWithPartition`, this transforms supports

1. Near uniform splitting of the input key space based on range counts. No partition will have a count greater than twice the expected mean.
2. Uses composite keys for splitting when necessary.
3. Allows injection of type-mapper for making it easier to support strings in future.
## Overview of commits.
This change composes of mainly these parts (in separate commits)
1. Basic Range and boundary classes. This part implements basic classes to represent a splittable boundary and range. An unsplittable range can have child ranges as columns get added to the splitting process.
2. DBAdapter and statement preparator implementation to get count and boundary  (min, max) of a range.
3. Transforms to iteratively split the ranges till a near-uniform split is  achieved.
4. Integration with larger reader under a feature flag.
## Feature Flag.
Currently there is a feature flag in `JdbcIOWrapperConfig` named `readWithUniformPartitionsFeatureEnabled` which controls if the new partitioning logic run in the migration or not.
1. As of now the flag is default to enabled.
2. It's not exposed as a pipeline option (which unfortunately means tooggle need rebuild) so that options don't get added and reverted.

## Performance
1. The splitting takes ~ 2 to 3 mins per table (1 TB table).
2. If the job is running on multiple parallel tables, please consider dding `DATAFLOW_SERVICE_OPTIONS="min_num_workers="` to the dataflow job  as dataflow tends to scale down quickly.

Note - unless we have the entire flow from the basic range class to integration, its hard to test this on a real migration.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
@@ -406,6 +406,79 @@ public final class MysqlDialectAdapter implements DialectAdapter {
     return mySQlTypeAliases.getOrDefault(normalizedType, normalizedType);
   }
 
+  private String addWhereClause(String query, ImmutableList<String> partitionColumns) {
+
+    // Implementation detail, using StringBuilder since we are generating the query in a loop.
+    StringBuilder queryBuilder = new StringBuilder(query);
+
+    boolean firstDone = false;
+    for (String partitionColumn : partitionColumns) {
+
+      if (firstDone) {
+        // Add AND for iteration after first.
+        queryBuilder.append(" AND ");
+      } else {
+        // add `where` only for first iteration.
+        queryBuilder.append(" where ");
+      }
+
+      // Include the column?
+      queryBuilder.append("((? = FALSE) OR ");
+      // range to define the where clause. `col >= range.start() AND (col < range.end() OR
+      // (range.isLast() = TRUE AND col = range.end()))`
+      queryBuilder.append(
+          String.format("(%1$s >= ? AND (%1$s < ? OR (? = TRUE AND %1$s = ?)))", partitionColumn));
+      queryBuilder.append(")");
+      firstDone = true;
+    }
+    return queryBuilder.toString();
+  }
+
+  /**
+   * Get query for the prepared statement to read columns within a range.
+   *
+   * @param tableName name of the table to read
+   * @param partitionColumns partition columns.
+   * @return Query Statement.
+   */
+  @Override
+  public String getReadQuery(String tableName, ImmutableList<String> partitionColumns) {
+    return addWhereClause("select * from " + tableName, partitionColumns);
+  }
+
+  /**
+   * Get query for the prepared statement to count a given range.
+   *
+   * @param tableName name of the table to read.
+   * @param partitionColumns partition columns.
+   * @param timeoutMillis timeout of the count query in milliseconds. Set to 0 to disable timeout.
+   * @return Query Statement.
+   */
+  @Override
+  public String getCountQuery(
+      String tableName, ImmutableList<String> partitionColumns, long timeoutMillis) {
+    return addWhereClause(
+        String.format(
+            "select /*+ MAX_EXECUTION_TIME(%s) */ COUNT(*) from %s", timeoutMillis, tableName),
+        partitionColumns);
+  }
+
+  /**
+   * Get query for the prepared statement to get min and max of a given column, optionally in the
+   * context of a parent range.
+   *
+   * @param tableName name of the table to read.
+   * @param partitionColumns if not-empty, partition columns. Set empty for first column of
+   *     partitioning.
+   */
+  @Override
+  public String getBoundaryQuery(
+      String tableName, ImmutableList<String> partitionColumns, String colName) {
+    return addWhereClause(
+        String.format("select MIN(%s),MAX(%s) from %s", colName, colName, tableName),
+        partitionColumns);
+  }
+
   /**
    * Version of MySql. As of now the code does not need to distinguish between versions of Mysql.
    * Having the type allows the implementation do finer distinctions if needed in the future.
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
@@ -419,7 +419,7 @@ public final class MysqlDialectAdapter implements DialectAdapter {
         queryBuilder.append(" AND ");
       } else {
         // add `where` only for first iteration.
-        queryBuilder.append(" where ");
+        queryBuilder.append(" WHERE ");
       }
 
       // Include the column?
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -17,9 +17,13 @@ package com.google.cloud.teleport.v2.source.reader.io.jdbc.iowrapper;
 
 import com.google.cloud.teleport.v2.source.reader.io.IoWrapper;
 import com.google.cloud.teleport.v2.source.reader.io.exception.SuitableIndexNotFoundException;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.dialectadapter.mysql.MysqlDialectAdapter;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.dialectadapter.mysql.MysqlDialectAdapter.MySqlVersion;
 import com.google.cloud.teleport.v2.source.reader.io.jdbc.iowrapper.config.JdbcIOWrapperConfig;
 import com.google.cloud.teleport.v2.source.reader.io.jdbc.iowrapper.config.TableConfig;
 import com.google.cloud.teleport.v2.source.reader.io.jdbc.rowmapper.JdbcSourceRowMapper;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range.PartitionColumn;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms.ReadWithUniformPartitions;
 import com.google.cloud.teleport.v2.source.reader.io.row.SourceRow;
 import com.google.cloud.teleport.v2.source.reader.io.schema.SchemaDiscovery;
 import com.google.cloud.teleport.v2.source.reader.io.schema.SchemaDiscoveryImpl;
@@ -50,6 +54,10 @@ import org.checkerframework.checker.nullness.qual.UnknownKeyFor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/*
+ * TODO(vardhanvthigle): Towards M3, make this a reconfigurable class, and expose (if required) the approxRowCounts
+ *  and maxPartitionHints (auto inferred) to the pipeline controller helping a better sequencing of tables.
+ */
 public final class JdbcIoWrapper implements IoWrapper {
   private static final Logger LOG = LoggerFactory.getLogger(JdbcIoWrapper.class);
 
@@ -122,12 +130,19 @@ public final class JdbcIoWrapper implements IoWrapper {
                       .setSourceTableName(sourceTableSchema.tableName())
                       .setSourceTableSchemaUUID(sourceTableSchema.tableSchemaUUID())
                       .build(),
-                  getJdbcIO(
-                      config,
-                      dataSourceConfiguration,
-                      sourceSchema.schemaReference(),
-                      tableConfig,
-                      sourceTableSchema));
+                  (config.readWithUniformPartitionsFeatureEnabled())
+                      ? getReadWithUniformPartitionIO(
+                          config,
+                          dataSourceConfiguration,
+                          sourceSchema.schemaReference(),
+                          tableConfig,
+                          sourceTableSchema)
+                      : getJdbcIO(
+                          config,
+                          dataSourceConfiguration,
+                          sourceSchema.schemaReference(),
+                          tableConfig,
+                          sourceTableSchema));
             })
         .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
   }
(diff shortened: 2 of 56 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -0,0 +1,153 @@
...
+    if (start == null) {
+      start = Long.MIN_VALUE;
+    }
+    if (end == null) {
+      end = Long.MIN_VALUE;
+    }
+
+    // This is same as (start + (end - start)/2) but that the width calculation can overflow if
+    // start is LONG.MIN and end is LONG.MAX.
+    return (start & end) + ((start ^ end) >> 1);
```
bharadwaj-aditya, line 115: this logic is not very clear. Please add an example if possible.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetter.java
```diff
@@ -0,0 +1,65 @@
...
+package com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range;
+
+import java.sql.PreparedStatement;
+import javax.annotation.Nullable;
+import org.apache.beam.sdk.io.jdbc.JdbcIO.PreparedStatementSetter;
+import org.checkerframework.checker.initialization.qual.Initialized;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.UnknownKeyFor;
+
+public class RangePreparedStatementSetter implements PreparedStatementSetter<Range> {
```
bharadwaj-aditya, line 25: Is there a strong reason to have this class, it seems like it could have been a method.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
```diff
@@ -0,0 +1,146 @@
...
+  private final SerializableFunction<Void, DataSource> dataSourceProviderFn;
+  private final long timeoutMillis;
+
+  private final String countQuery;
+
+  private final long numColumns;
+
+  @JsonIgnore private transient @Nullable DataSource dataSource;
+
+  RangeCountDoFn(
```
bharadwaj-aditya, line 58: is this meant to be default ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetter.java
VardhanThigle: We are implementing jdbcIO's `PreparedStatementSetter` interface here.
It is needed since we call JDBCIO.readall() with the ranges ultimately.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
VardhanThigle: Sorry I am not getting what we are referring to `this` here?

On the current line 58 where the comment is anchored, Its the constructor of the DoFn

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
VardhanThigle: Sure. I will add. 



At a high level it's the average between start and end. But,

(Context on going this route) - a general long column say if the user has chosen to put in random integers , it could get values from too low on the negative spectrum to too high on positive spectrum.

if we do (start + end)/2 - it overflows if the sum goes too high , both start and end are near Long.Max.
If this was unsigned , we could do with start + (end - start)/2 - the `half-width` method. But that overflows when start and end are on 2 ends of positive and negative long.
for example Long.Min + (Long.Max - Long.Min)/2 will now work as (Long.Max. - Long.Min) will overflow (even if we half them individually, it comes dangerously close to 2 * (Long.Max/2)

Sum of 2 numbers can be written as 

`(start ^ end)  + ((start & end)<<1)` basically xor of bits and add the carries.
So average simply becomes (start + end)/2
`(start ^ end)>>1  + ((start & end))`

I have added UT at all the corners of Long.MAX and Long.Min to verify that this does not overflow and does give an average rounded down (rounding down consistently is also important pice for us)

VardhanThigle: Added a comment explaining this in more detail.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -406,6 +406,79 @@ private String normalizeColumnType(String columnType) {
...
+
+    boolean firstDone = false;
+    for (String partitionColumn : partitionColumns) {
+
+      if (firstDone) {
+        // Add AND for iteration after first.
+        queryBuilder.append(" AND ");
+      } else {
+        // add `where` only for first iteration.
+        queryBuilder.append(" where ");
```
bharadwaj-aditya, line 422: nit - use same case for all clauses for uniformity.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -277,6 +330,49 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+            .setRowMapper(
+                new JdbcSourceRowMapper(
+                    config.valueMappingsProvider(),
+                    sourceSchemaReference,
+                    sourceTableSchema,
+                    config.shardID()))
+            .setWaitOnSignals(config.waitOnSignals())
+            // TODO(vardhanvthigle): if index is not of the type of a single auto incrementing key,
+            // don't set this.
+            .setSplitStageCountHint(4L)
```
bharadwaj-aditya, line 366: what is this hint ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -277,6 +330,49 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+            .setRangesPeek(config.rangesPeek());
```
bharadwaj-aditya, line 367: nit: Method name does not clarify what it does. maybe call it - additionalOperationsOnRange or something less verbose.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQuery.java
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * @return column details.
+   */
+  public abstract PartitionColumn partitionColumn();
+
+  /**
+   * @return Parent Range. Null indicates first column for splitting. Defaults to Null.
+   */
+  @Nullable
+  public abstract Range parentRange();
```
bharadwaj-aditya, line 40: can you elaborate on the docs here. It is not clear from the method name.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Implement {@link PreparedStatementSetter} to set {@link PreparedStatement} parameters for getting
+ * boundary of a column.
+ */
+public class ColumnForBoundaryQueryPreparedStatementSetter
+    implements PreparedStatementSetter<ColumnForBoundaryQuery> {
+
+  /** List of partition columns. */
+  ImmutableList<String> partitionCols;
```
bharadwaj-aditya, line 35: does this need to be default access ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/InitialSplitRangeDoFn.java
```diff
@@ -0,0 +1,103 @@
...
+   * @param out output receiver for a list of initial split of ranges. The ranges are not counted as
+   *     yet.
+   * @param c process context
+   */
+  @ProcessElement
+  public void processElement(
+      @Element Range input, OutputReceiver<ImmutableList<Range>> out, ProcessContext c) {
+
+    logger.info(
+        "RWUPT - Began split process for table {} with initia range as {}", tableName(), input);
```
bharadwaj-aditya, line 58: nit: initial range. Also please elaborate on the abbreviation

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MergeRangesDoFn.java
```diff
@@ -0,0 +1,140 @@
...
+    out.output(mergeRanges(input, c));
+  }
+
+  public static Builder builder() {
+    return new AutoValue_MergeRangesDoFn.Builder();
+  }
+
+  private ImmutableList<Range> mergeRanges(ImmutableList<Range> input, ProcessContext c) {
+
+    long totoalCount = approxTotalRowCount();
```
bharadwaj-aditya, line 70: nit: totalCount

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param input Details for the column and parent range for which a boundary is requested.
+   * @param out new Range with column boundary.
+   * @param c process context.
+   * @throws SQLException - since this is in the run time, beam will auto retry the exception.
+   */
+  @ProcessElement
+  public void processElement(
+      @Element ColumnForBoundaryQuery input, OutputReceiver<Range> out, ProcessContext c)
+      throws SQLException {
+    String boundaryQuery =
```
bharadwaj-aditya, line 90: It is unclear to me where the limits are to come, but please check that values lower than min and larger than max detected at startup are not excluded while reading.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
bharadwaj-aditya: I meant default access - not public or private

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/SplitRangeDoFn.java
```diff
@@ -0,0 +1,34 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms;
+
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range.Range;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.commons.lang3.tuple.Pair;
+
+/** Split a given range into 2 halves, typically after adding a new column. */
```
bharadwaj-aditya, line 22: Docs not clear- adding new column to ?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetterTest.java
```diff
@@ -0,0 +1,195 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Test class for {@link RangePreparedStatementSetter}. */
+@RunWith(MockitoJUnitRunner.class)
+public class RangePreparedStatementSetterTest {
```
bharadwaj-aditya, line 38: nit: Please name the class or methods as MySQL wherever queries are being verified.

bharadwaj-aditya (commented):
Reviewed the method signatures and class names and tests.

The actual PR is very large so unable to review all of it in detail. Considering this is behind a flag, let's proceed once these are addressed.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: This is the number of stages we provision to split the ranges.

We have java doc [at the source](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1657/files/2f1562c543def759334e7abc570222937c0b67fe#diff-63967ef4a0e72c5ff642b8a7351e3eca71bf77e9d9929eabd70ec91b4d34aea2R125)

This is not a compulsory parameter, it gets auto inferred, but given that we are dealing with only auto incrementing keys till we get in strings, we don't need a large split depth even for tall tables.

I will add more details in the comment in JDBCIOWrapper too.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: makes sense. I borrowed the `peek` suffix from the java stream api. It was chosen to underscore that ptransform is meant to be a consumer like transform.

Will change `additionalOperationsOnRange

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
VardhanThigle: The limits are fetched using a query of the type `select min(col), max(col)...` returned by the dbAdapter.

```
but please check that values lower than min and larger than max detected at startup are not excluded while reading.
```
Currently they are excluded. Wouldn't CDC cover for newer rows?
This behavior is consistent with JDBCIO#readWithPartitions, as in it reads only the slice between what it initially got as min and max.

I thought of keeping it consistent as I was a bit worried that reading more than max might be unbounded as - in a batch job might never fully go to a success state, for a table with auto incrementing keys and insert heavy workload.

Let's have a small catchup to discuss this more.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/SplitRangeDoFn.java
VardhanThigle: Adding a new column to an existing unsplittable range. Thanks, I will clarify in the doc.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
VardhanThigle: Got you. Thanks.
It is package private and not constructed outside the package.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java
VardhanThigle: Thanks for the catch! made it private.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQuery.java
VardhanThigle: Added more details. Thanks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/InitialSplitRangeDoFn.java
VardhanThigle: Fixed. Thanks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MergeRangesDoFn.java
VardhanThigle: Fixed Thanks.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetterTest.java
VardhanThigle: This test is verifying the behavior of the query produced and not exactly the query.
`MySqlDialectAdapterTest` does actually verify the query syntax (and the enclosing test class is named accordingly)

In this test(`RangePreparedStatementSetterTest`) we are testing `RangePreparedStatementSetter` (as needed by JDBCIO#readAll) for queries  on a derby table and the `RangePreparedStatementSetter` should not change across MySql or PG. 
For the UT we do use MySQl dialect adapter to get the prepared statement, currently (derby matches mySQL in a lot of aspects).

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
VardhanThigle: Fixed. Thanks.

bharadwaj-aditya (approved):
LGTM

PR #1660, opened 2024-06-14
Add CloudSql resource manager

Polber:
PR adds the CloudSqlResourceManager to manage SQL resources on GCP.

For MySQL - expects a static CloudSQL instance running MySQL
For Oracle - expects a static GCE VM running Oracle XE

### DIFF
--- a/cicd/cmd/run-it-smoke-tests/main.go
+++ b/cicd/cmd/run-it-smoke-tests/main.go
@@ -65,7 +65,11 @@ func main() {
 		flags.StageBucket(),
 		flags.PrivateConnectivity(),
 		flags.FailureMode(),
-		flags.RetryFailures())
+		flags.RetryFailures(),
+		flags.StaticOracleInstance(),
+		flags.CloudProxyHost(),
+		flags.CloudProxyPort(),
+		flags.CloudProxyPassword())
 	if err != nil {
 		log.Fatalf("%v\n", err)
 	}
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.it.gcp.cloudsql;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Custom class for the MySQL implementation of {@link CloudSqlResourceManager} abstract class.
+ *
+ * <p>The class supports one database, and multiple tables per database object. A database is *
+ * created when the container first spins up, if one is not given.
+ *
+ * <p>The class is thread-safe.
+ */
+public class CloudMySQLResourceManager extends CloudSqlResourceManager {
+
+  private CloudMySQLResourceManager(Builder builder) {
+    super(builder);
+  }
+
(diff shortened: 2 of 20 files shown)


cicd/cmd/run-it-smoke-tests/main.go
```diff
@@ -59,6 +59,10 @@ func main() {
 		mvnFlags.IntegrationTestParallelism(4),
 		mvnFlags.StaticBigtableInstance("teleport"),
 		mvnFlags.StaticSpannerInstance("teleport"),
+		mvnFlags.StaticOracleInstance("10.128.0.90"),
+		mvnFlags.CloudProxyHost("10.128.0.34"),
+		mvnFlags.CloudProxyPort("33134"),
+		mvnFlags.CloudProxyPassword("t>5xl%J(&qTK6?FaZ"),
```
Abacn, line 65: I understand that this is internal network only so this "password" is fine. However can we put most flags into a config file, e.g. yaml file and read them on runtime? Currently it is hardcoded in source code in several places

Abacn (commented):
Thanks, reviewed cicd change first. Will add comments for the resource manager shortly after

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.it.gcp.cloudsql;
+
+import org.jetbrains.annotations.NotNull;
```
Abacn, line 20: This is one of the annoying thing of intellij, should use checkerframeworks NonNull

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudSqlResourceManager.java
```diff
@@ -0,0 +1,185 @@
+package org.apache.beam.it.gcp.cloudsql;
+
+import static org.apache.beam.it.gcp.cloudsql.CloudSqlResourceManagerUtils.generateDatabaseName;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.beam.it.jdbc.AbstractJDBCResourceManager;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.NotNull;
```
Abacn, line 9: Same here, even both NonNull and NotNull are used

it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManagerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  private CloudMySQLResourceManager testManager;
+
+  private static final String TEST_ID = "test_id";
+
+  @Before
+  public void setUp() {
+    System.setProperty("cloudProxyHost", "127.0.0.1");
```
Abacn, line 42: Set System property in test or at runtime general is a side effect and general not recommended. If we cannot compact configurations in builder (I do not see a reason), use a static object to config is better to set jvm wide system property

Abacn (changes requested):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
Polber: fixed.

it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManagerTest.java
Polber: Decided to go with builder

cicd/cmd/run-it-smoke-tests/main.go
Polber: I refactored to use the flags module instead and set default values for the flags - config file at that point seemed like overkill but let me know

Abacn:
The Dataflow Integration Test failed on

```
JdbcToBigQueryIT.testMySqlToBigQueryWithStorageWriteApi
```

looks relevant

Polber:
> The Dataflow Integration Test failed on
> 
> ```
> JdbcToBigQueryIT.testMySqlToBigQueryWithStorageWriteApi
> ```
> 
> looks relevant

Shouldn't be relevant since that test does not use CloudSqlResourceManager, but running again just to be safe

Abacn (approved):

PR #1661, opened 2024-06-14
Add full suite of Datastream to Spanner IT's

Polber:
Depends on CloudSqlResourceManager from #1660 and Custom spanner endpoints from #1465

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datastream/DatastreamResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datastream/DatastreamResourceManager.java
@@ -51,13 +51,17 @@ import com.google.cloud.datastream.v1.StreamName;
 import com.google.cloud.datastream.v1.UpdateStreamRequest;
 import com.google.protobuf.Duration;
 import com.google.protobuf.FieldMask;
+import dev.failsafe.Failsafe;
+import dev.failsafe.RetryPolicy;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datastream/DatastreamResourceManager.java
```diff
@@ -445,7 +456,9 @@ public synchronized Stream createStream(
                       .build())
               .build();
 
-      Stream reference = datastreamClient.createStreamAsync(request).get();
+      Stream reference =
+          Failsafe.with(retryOnCancellationException())
+              .get(() -> datastreamClient.createStreamAsync(request).get());
```
manitgupta, line 461: Are these retries idempotent? This is test suite so maybe it doesn't matter, but there is a possibility of creation of orphaned streams if the retries are not idempotent.

manitgupta (approved):
LGTM, one minor comment.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datastream/DatastreamResourceManager.java
Polber: It will only retry if the request was cancelled, but even if there was a case where the stream creation is retried even after a successful creation, it would fail since it would have the same name. There is also a reaper script on the GCP project these tests run that removes old datastream streams, so a stream leak is very unlikely.

manitgupta: Sounds good, thanks.

PR #1662, opened 2024-06-14
Add PubSubCdcToBigQueryIT

Polber:


v2/pubsub-cdc-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/PubSubCdcToBigQuery.java
```diff
@@ -264,6 +264,12 @@ public static void main(String[] args) {
     Options options = PipelineOptionsFactory.fromArgs(args).withValidation().as(Options.class);
     BigQueryIOUtils.validateBQStorageApiOptionsStreaming(options);
 
+    if (options.getDeadLetterQueueDirectory() != null
```
Abacn, line 267: in case there were user specified empty string for deadLetterQueueDirectory, consider use `com.google.common.base.Strings.isNullOrEmpty` and treat empty string as null ?

v2/pubsub-cdc-to-bigquery/src/test/java/com/google/cloud/teleport/v2/templates/PubSubCdcToBigQueryIT.java
```diff
@@ -0,0 +1,236 @@
...
+  public void testSubscriptionToBigQuery() throws IOException {
+    testSubscriptionToBigQueryBase(false);
+  }
+
+  @Test
+  public void testSubscriptionToBigQueryWithUDf() throws IOException {
+    testSubscriptionToBigQueryBase(true);
+  }
+
+  public void testSubscriptionToBigQueryBase(boolean useUdf) throws IOException {
```
Abacn, line 120: the suffix "test" is usually for junit test. Just name "subscriptionToBigQueryBase" ?

Abacn (approved):
thanks, just have two minor comments. After addressed feel free to trigger code import

v2/pubsub-cdc-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/PubSubCdcToBigQuery.java
```diff
@@ -72,6 +72,7 @@
 import org.apache.beam.sdk.values.PCollectionList;
 import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.vendor.guava.v32_1_2_jre.com.google.common.collect.ImmutableList;
+import org.apache.parquet.Strings;
```
Abacn, line 75: probably intellij issue, use

```
import org.apache.beam.vendor.guava.v32_1_2_jre.com.google.common.base.Strings;
```

Polber: Ah oops, good catch

PR #1663, opened 2024-06-17
Add terraform validator

manitgupta:
- Adds a Terraform Validator for Terraform samples being added to the repo. It verifies the sample structure and the `plan` generated for it via `terraform validate`.
- Formats the existing bulk migration example to comply with Terraform validator using `terraform fmt`.

### DIFF
--- a/.github/workflows/spanner-terraform-validator.yml
+++ b/.github/workflows/spanner-terraform-validator.yml
@@ -45,15 +45,23 @@ jobs:
           terraform -chdir="$SAMPLE" validate
         
           echo "Verifying sample structure..."
-          for FILE in $FILE_LIST; do
+          MISSING_FILES=() # Initialize empty list for missing files in each sample
+          for FILE in ${FILE_LIST[@]}; do
             if [[ ! -f "$SAMPLE/$FILE" ]]; then
               echo "ERROR: Missing file: $SAMPLE/$FILE"
-              FAILED_SAMPLES+=("$SAMPLE")
-              break
+              MISSING_FILES+=("$FILE") # Add missing file to list
             fi
           done
         
-          if [[ ! " ${FAILED_SAMPLES[*]} " =~ " ${SAMPLE} " ]]; then # Check if sample is in failed list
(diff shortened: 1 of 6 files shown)


.github/workflows/spanner-terraform-validator.yml
```diff
@@ -0,0 +1,69 @@
...
+          echo "Verifying terraform configuration..."
+          terraform -chdir="$SAMPLE" init
+          terraform -chdir="$SAMPLE" validate
+        
+          echo "Verifying sample structure..."
+          for FILE in $FILE_LIST; do
+            if [[ ! -f "$SAMPLE/$FILE" ]]; then
+              echo "ERROR: Missing file: $SAMPLE/$FILE"
+              FAILED_SAMPLES+=("$SAMPLE")
+              break
```
asthamohta, line 52: What if multiple files are missing? Should we print all missing files?

manitgupta: It was just being lazy 😛
Okay will collect the list and print it!

manitgupta: Done!

asthamohta (approved):

bharadwaj-aditya:
Can you add some comments/ docs with use cases on the create sample TF scipt?

manitgupta:
> Can you add some comments/ docs with use cases on the create sample TF scipt?

As in comments/docs on the samples that will follow?

bharadwaj-aditya (approved):
LGTM

damondouglas (changes requested):
How does this work with the existing method to autogenerate terraform based on the template spec? If the existing solution isn't compatible with what already exists in the rest of the repository is fine but would be helpful to know either way.

manitgupta:
This is independent of that and is only localised to Spanner specific templates.

The existing auto-generated template only focuses on Dataflow while the samples we will add will focus on other resources relevant to an IaaC configuration with respect to the Spanner templates.

damondouglas (approved):

PR #1664, opened 2024-06-17
Enable PR reports for Spanner PR

manitgupta:


damccorm (approved):
LGTM, though I think we may want to move to a model where these are embedded in the workflow itself (instead of being its own separate workflow). Regardless, I'll merge for now.

PR #1665, opened 2024-06-17
base64 handling simplified

aksharauke:
For BYTES when Spanner stores it as base64 encoded, no need to decode it and escape - rather use the built in constructs to handle base64. Similar use of built in constructs to handle hex.

### DIFF
--- a/v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
+++ b/v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
@@ -25,9 +25,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.apache.commons.codec.DecoderException;
-import org.apache.commons.codec.binary.Base64;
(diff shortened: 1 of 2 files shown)


v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
```diff
@@ -336,31 +333,19 @@ private static String getMappedColumnValue(
...
-        && ("binary".equals(sourceColDef.getType().getName())
-            || "varbinary".equals(sourceColDef.getType().getName()))) {
-
-      // Spanner has the hex string in this case
-      try {
-        colInputValue = new String(Hex.decodeHex(valuesJson.getString(colName)));
-      } catch (DecoderException e) {
-        // return the same string value
-        colInputValue = valuesJson.getString(colName);
-      }
```
shreyakhajanchi, line 349: why are we removing this? Shouldn't we have a HEX() on this?

aksharauke: Hex handling is handled in the getHexString method. It adds HEX there.

shreyakhajanchi (approved):
LGTM

PR #1669, opened 2024-06-18
Terraform Sample: Live migration all infrastructure setup

manitgupta:
This PR adds samples for common scenarios users might have while trying to run a live migration to Spanner.

Each sample contains the following (and potentially more) files - 

1. `main.tf` - This contains the Terraform resources which will be created.
2. `outputs.tf` - This declares the outputs that will be output as part of running the terraform example.
3. `variables.tf` - This declares the input variables that are required to configure the resources.
4. `terraform.tf` - This contains the required providers and APIs/project configurations for the sample.
5. `terraform.tfvars` - This contains the dummy inputs that need to be populated to run the example.
6. `terraform_simple.tfvars` - This contains the minimal list of dummy inputs that need to be populated to run the example.

**_SCENARIO:_** This Terraform example illustrates launching a live migration job for a MySQL source, setting up all the required cloud infrastructure. **Only the source details are needed as input.**

Creates the following resources -

1. **Datastream private connection** - If configured, a Datastream private connection will be deployed for your configured VPC. If not configured, IP whitelisting will be assumed as the mode of Datastream access.
2. **Source datastream connection profile** - This allows Datastream to connect to the MySQL instance (using IP whitelisting).
3. **GCS bucket** - A GCS bucket to for Datastream to write the source data to.
4. **Target datastream connection profile** - The connection profile to configure the created bucket in Datastream.
5. **Pubsub topic and subscription** - This contains GCS object notifications as files are written to GCS for consumption by the Dataflow job.
6. **Datastream stream** - A datastream stream which reads from the source specified in the source connection profile and writes the data to the bucket specified in the target connection profile. Note that it uses a mandatory prefix path inside the bucket where it will write the data to. The default prefix path is `data` (can be overridden).
7. **Bucket notification** - Creates the GCS bucket notification which publish to the pubsub topic created. Note that the bucket notification is created on the mandatory prefix path specified for the stream above.
8. **Dataflow job** - The Dataflow job which reads from GCS and writes to Spanner.
9. **Permissions** - It adds the required roles to the specified (or the default) service accounts for the live migration to work.


.github/workflows/spanner-terraform-validator.yml
```diff
@@ -11,15 +12,14 @@ on:
...
-    - 'v2/sourcedb-to-spanner/terraform/samples/**'
 
 jobs:
   verify-terraform-samples:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         terraform_paths:
-        - v2/sourcedb-to-spanner/terraform/samples
+        - v2/datastream-to-spanner/terraform/samples
```
bharadwaj-aditya, line 22: is this expected ? It seems a little arbitrary to change

v2/datastream-to-spanner/terraform/samples/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Other Sample Repositories
+
+The following sample repositories provide additional examples -
+
+1. [Sample environment setups](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-common/terraform/samples)
+2. [Bulk migration](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/sourcedb-to-spanner/terraform/samples)
+
+## List of examples
+
+1. [MySQL source to Spanner end-to-end](all-infrastructure/README.md)
```
bharadwaj-aditya, line 16: nit: i would rename this mysql-end-to-end/mysql-production-migration instead

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
+## Sample Scenario: MySQL to Spanner using MySQL source configuration
+
+> **_SCENARIO:_** This Terraform example illustrates launching a live migration
```
bharadwaj-aditya, line 3: nit: please rename this as a mysql-production-migration. 

That is closer to what people would be looking for which this example aims to answer.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+## Terraform permissions
+
+In order to create the resources in this sample,
+the`Service account`/`User account` being used to run Terraform
+should have the
+required [permissions](https://cloud.google.com/iam/docs/manage-access-service-accounts#multiple-roles-console).
+There are two ways to add permissions -
+
+1. Adding pre-defined roles to the service account running Terraform.
+2. Creating a custom role with the granular permissions and attaching it to the
```
bharadwaj-aditya, line 16: is there a way to verify if these permissions exist before folks start ?
Can we add it in the terraform init/plan ?

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+
+[This](#adding-access-to-terraform-service-account) section in the FAQ
+provides instructions to add these roles to an existing service account.
+
+## Assumptions
+
+It takes the following assumptions -
+
+1. MySQL source is accessible via Datastream either via
+   [IP Whitelisting guide](https://cloud.google.com/datastream/docs/network-connectivity-options#ipallowlists)
```
bharadwaj-aditya, line 124: IP Allowlisting

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+  "dataflow_job" = "2024-06-14_03_01_00-3421054840094926119"
+  "datastream_source_connection_profile" = "source-mysql-thorough-wombat"
+  "datastream_stream" = "mysql-stream-thorough-wombat"
+  "datastream_target_connection_profile" = "target-gcs-thorough-wombat"
+  "gcs_bucket" = "live-migration-thorough-wombat"
+  "pubsub_subscription" = "live-migration-thorough-wombat-sub"
+  "pubsub_topic" = "live-migration-thorough-wombat"
+}
+resource_urls = {
+  "dataflow_job" = "https://console.cloud.google.com/dataflow/jobs/us-central1/2024-06-14_03_01_00-3421054840094926119?project=your-project-here"
```
bharadwaj-aditya, line 229: nit: let's replace the params with a clearer seperator. Something like <your-project-here> or so ?

bharadwaj-aditya: Also, the dataflow job url's will depend on the region where this job is being run.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+If you want to exclude the GCS bucket from deletion due to any reason, you
+can exclude it from the state file using `terraform state rm` command.
+
+## FAQ
+
+### Configuring to run using a VPC
+
+#### Datastream
+
+> **_NOTE:_** By default, **IP Whitelisting** based connectivity is assumed.
```
bharadwaj-aditya, line 262: Allowlisting.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+  ]
+  private_connectivity = {
+    private_connectivity_id = "abc"
+    range                   = "xyz"
+    vpc_name                = "pqr"
+  }
+  ...
+```
+
+Note that `vpc_name` and `range` are mandatory and for the private connectivity
```
bharadwaj-aditya, line 292: Please add links to the GCP documentation for some of these

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
```diff
@@ -0,0 +1,554 @@
...
+Verify that the custom role is attached to the service account - 
+
+```shell
+gcloud projects get-iam-policy <YOUR-PROJECT-ID>  \
+--flatten="bindings[].members" \
+--format='table(bindings.role)' \
+--filter="bindings.members:<YOUR-SERVICE-ACCOUNT>@<YOUR-PROJECT-ID>.iam.gserviceaccount.com"
+```
+
+Verify that the role has the correct set of permissions
```
bharadwaj-aditya, line 507: Is the verification manual ? Can it be automated ?

v2/datastream-to-spanner/terraform/samples/all-infrastructure/terraform.tfvars
```diff
@@ -0,0 +1,71 @@
...
+  region                          = "<YOUR_GCP_REGION>"   # Replace with your desired GCP region
+  migration_id                    = "<YOUR_MIGRATION_ID>" # Will be used as a prefix for all resources, auto-generated if not specified
+  add_policies_to_service_account = "<TRUE/FALSE>"        # This will decide if roles will be attached to service accounts or not.
+}
+
+# Datastream Parameters
+datastream_params = {
+  source_connection_profile_id  = "source-mysql"   # Or provide a custom ID
+  target_connection_profile_id  = "target-gcs"     # Or provide a custom ID
+  target_gcs_bucket_name        = "live-migration" # Or provide a custom bucket name
```
bharadwaj-aditya, line 13: nit: if possible, it would be better to have a random id attached after these names. To ensure that repeated runs of this job do not overwrite existing config or lead to unexpected behaviour.

bharadwaj-aditya (changes requested):

v2/datastream-to-spanner/terraform/samples/README.md
manitgupta: Done!

.github/workflows/spanner-terraform-validator.yml
manitgupta: I added a number of additional validations so I've excluded the terraform example for bulk from here. That sample would be updated in a separate PR and then we can enable the validator check on it.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: Done!

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: This can't be done with terraform init/plan.
I tried without permissions it fails in 5 seconds and the error is pretty intuitive. TF complains of the specific permissions that are missing.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: Done in all places.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: > Also, the dataflow job url's will depend on the region where this job is being run.

No, it doesn't.

> nit: let's replace the params with a clearer seperator. Something like or so ?

These are clickable hyperlinks and are pretty intuitive to understand from the output. Keeping them as is.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: Added

v2/datastream-to-spanner/terraform/samples/all-infrastructure/README.md
manitgupta: These are pretty sensitive permissions and should not be automated, in general. That is why we provide the recipes to the user when dealing with Terraform permissions themselves. This is a one-time setup generally done by someone with elevated privileges.

Can probably be automated by writing a shell script over gCloud I guess. Will leave that to the user for now and see if it's a consistent concern and we need to do something about it. 

Overall, the feedback is pretty consistent from terraform side in my investigations.

v2/datastream-to-spanner/terraform/samples/all-infrastructure/terraform.tfvars
manitgupta: It's already there. If the user does not provide a `migration_id`, one is auto-generated!

bharadwaj-aditya (approved):
LGTM

PR #1670, opened 2024-06-18
Terraform Sample: Live migration pre-configured profiles

manitgupta:
This PR adds samples for common scenarios users might have while trying to run a live migration to Spanner.

Each sample contains the following (and potentially more) files - 

1. `main.tf` - This contains the Terraform resources which will be created.
2. `outputs.tf` - This declares the outputs that will be output as part of running the terraform example.
3. `variables.tf` - This declares the input variables that are required to configure the resources.
4. `terraform.tf` - This contains the required providers and APIs/project configurations for the sample.
5. `terraform.tfvars` - This contains the dummy inputs that need to be populated to run the example.
6. `terraform_simple.tfvars` - This contains the minimal list of dummy inputs that need to be populated to run the example.

**_SCENARIO:_** This Terraform example illustrates launching a live migration job for a MySQL source, **given pre-created Datastream source and target connection profiles**. As a result, it does not create any new buckets in the GCP account.

Creates the following resources -

1. **Pubsub topic and subscription** - This contains GCS object notifications as files are written to GCS for consumption by the Dataflow job.
2. **Datastream stream** - A datastream stream which reads from the source specified in the source connection profile and writes the data to the bucket specified in the target connection profile. Note that it uses a mandatory prefix path inside the bucket where it will write the data to. The default prefix path is `data` (can be overridden).
3. **Bucket notification** - Creates the GCS bucket notification which publish to the pubsub topic created. Note that the bucket notification is created on the mandatory prefix path specified for the stream above.
4. **Dataflow job** - The Dataflow job which reads from GCS and writes to Spanner.
5. **Permissions** - It adds the required roles to the specified (or the default) service accounts for the live migration to work.

### DIFF
--- a/v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
+++ b/v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
@@ -1,0 +1,86 @@
+# Common Parameters
+common_params = {
+  project = "<YOUR_PROJECT_ID>"
+  region  = "<YOUR_GCP_REGION>"
+}
+
+# Datastream Parameters
+datastream_params = {
+  source_connection_profile_id  = "<YOUR_SOURCE_CONNECTION_PROFILE_ID>"
+  target_connection_profile_id  = "<YOUR_TARGET_CONNECTION_PROFILE_ID>"
+  target_gcs_bucket_name        = "<YOUR_GCS_BUCKET_NAME>"
+  pubsub_topic_name             = "live-migration" # Or your custom topic name
+  stream_id                     = "mysql-stream"   # Or your custom stream ID
+  max_concurrent_cdc_tasks      = 50               # Adjust as needed
+  max_concurrent_backfill_tasks = 50               # Adjust as needed
+  mysql_databases = [
+    {
+      database = "<YOUR_DATABASE_NAME>"
+      tables   = []
+      # Optionally list specific tables, or remove "tables" all together for all tables
+    },
+    # Add more database objects if needed
+  ]
+}
+
+# Dataflow Parameters
+dataflow_params = {
+  template_params = {
+    shadow_table_prefix = "<YOUR_SHADOW_TABLE_PREFIX>"
+    # e.g., "shadow_"
+    create_shadow_tables = true # true or false
+    rfc_start_date_time  = "<YYYY-MM-DDTHH:MM:SSZ>"
+    # e.g., "2023-12-31T12:00:00Z" (optional)
+    file_read_concurrency = 10 # Adjust as needed
+    session_file_path     = "<YOUR_SESSION_FILE_PATH>"
+    # Path to your session file (optional)
+    spanner_instance_id = "<YOUR_SPANNER_INSTANCE_ID>"
+    spanner_database_id = "<YOUR_SPANNER_DATABASE_ID>"
+    spanner_host        = "https://<YOUR_REGION>-spanner.googleapis.com"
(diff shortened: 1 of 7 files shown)


v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
```diff
@@ -0,0 +1,88 @@
...
+}
+
+# Datastream Parameters
+datastream_params = {
+  source_connection_profile_id  = "<YOUR_SOURCE_CONNECTION_PROFILE_ID>"
+  target_connection_profile_id  = "<YOUR_TARGET_CONNECTION_PROFILE_ID>"
+  target_gcs_bucket_name        = "<YOUR_GCS_BUCKET_NAME>"
+  pubsub_topic_name             = "live-migration" # Or your custom topic name
+  stream_id                     = "mysql-stream"   # Or your custom stream ID
+  max_concurrent_cdc_tasks      = 50               # Adjust as needed
```
bharadwaj-aditya, line 16: I think this is 8 or so by default. Lets set sensible defaults here as folks probably won't change this unless required.

v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
```diff
@@ -0,0 +1,88 @@
...
+    dead_letter_queue_directory = "<YOUR_DLQ_DIRECTORY>"
+    # e.g., "gs://<YOUR_BUCKET>/dlq" (optional)
+    dlq_retry_minutes   = 10 # Adjust as needed
+    dlq_max_retry_count = 3  # Adjust as needed
+    datastream_root_url = "<YOUR_DATASTREAM_ROOT_URL>"
+    # Base URL of your Datastream API (optional)
+    datastream_source_type           = "MYSQL"
+    round_json_decimals              = false
+    run_mode                         = "STREAMING"
+    transformation_context_file_path = "<YOUR_TRANSFORMATION_FILE_PATH>"
```
bharadwaj-aditya, line 52: how do folks know that this is optional ? Can we comment in the tf file to make it clearer without having to look into documentation ?

bharadwaj-aditya (commented):
Largely looks fine. Only suggestion would be to make the examples a little more independent so folks can update values without really going through documentation.

v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
manitgupta: I added 50 because that's what we use in SMT and wanted to keep the configurations same. 
Do you have a different default in mind? If we want to use Datastream defaults (20), we can just remove this from here.

v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
manitgupta: This can be determined from `variables.tf` which declares if each variable is optional or mandatory, along with defaults.

There is a `terraform_simple.tfvars` as well which demonstrates the minimal configuration the user needs to do.

v2/datastream-to-spanner/terraform/samples/pre-configured-conn-profiles/terraform.tfvars
manitgupta: Resolved

bharadwaj-aditya (approved):
LGTM

PR #1671, opened 2024-06-18
Fix return for Kafka Avro deserializer

AnandInguva:


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaRecordToGenericRecordFailsafeElementFn.java
```diff
@@ -102,10 +98,10 @@ public void processElement(ProcessContext context) {
             (GenericRecord)
                 kafkaDeserializer.deserialize(
                     element.getTopic(), element.getHeaders(), element.getKV().getValue());
+        context.output(FailsafeElement.of(element, result));
```
Dippatel98, line 101: There is a bug here, this will only output if the `messageFormat != "AVRO_BINARY_ENCODING"`. You need to move the output method call outside the else block.

AnandInguva: Done

PR #1673, opened 2024-06-19
Spanner to MySQL reverse replication Load test 100 TPS

darshan-sj:


v2/gcs-to-sourcedb/src/test/resources/SpannerToMySql100TpsLT/mysql-schema.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE `person1` (
```
aksharauke, line 1: Since the main aim of reverse replication is to give ordering guarantees (at the cost of throughput) , I suggest to use schema on mysql side that is parent-child related via foreign key relation and on Spanner side via INTERLEAVE. This way with random writes being reverse replicated under load, we are sure that the order is maintained.

darshan-sj: Yes. This is a sample Load test. we will write Load tests with parent-child relationship and other data size and shapes later. One thing is we need to first add support to write multiple rows with parent child relationship in the Data generator.

aksharauke (approved):

PR #1674, opened 2024-06-19
fix: deprecating inputFilePattern

asthamohta:


darshan-sj (dismissed):
LGTM

manitgupta (approved):

manitgupta (approved):

manitgupta (changes requested):
Lets also deprecate/remove the directory watch minutes param along with its associated configuration as well.

manitgupta:
I think we discussed that there was another flag that had to be removed?

asthamohta:
Yes removed, I hadn't pushed that commit. You can take a look now

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
```diff
@@ -372,21 +375,6 @@ public interface Options extends PipelineOptions, StreamingOptions {
...
-        helpText =
-            "The Duration for which the pipeline should keep polling a directory in GCS. Datastream"
-                + "output files are arranged in a directory structure which depicts the timestamp "
-                + "of the event grouped by minutes. This parameter should be approximately equal to"
-                + "maximum delay which could occur between event occurring in source database and "
-                + "the same event being written to GCS by Datastream. 99.9 percentile = 10 minutes")
-    @Default.Integer(10)
-    Integer getDirectoryWatchDurationInMinutes();
-
-    void setDirectoryWatchDurationInMinutes(Integer value);
```
manitgupta, line 388: Can you validate if we set this in SMT?  Even if its optional, if we are setting it, then removing it will cause SMT (and other customers using SMT) to break.

asthamohta: I checked, we only do it in a unit test so no problems there

manitgupta (approved):

VardhanThigle:
This conflicts with the bulk templates retry flow.
The current strategy is to have a common DLQ retry for bulk and live by running live in retryMode. That won't have pub/sub enabled as it will be run after the bulk completes.
cc: @bharadwaj-aditya

PR #1676, opened 2024-06-20
Add Kafka read topic and kafka write topic pickers replacing kafka topic picker

AnandInguva:


Dippatel98 (commented):
We should not be replacing `KafkaTopic`, since we are not dropping support for the older parameter, just adding support for the new ones. So I think `KafkaTopic`, `KafkaReadTopic` and `KafkaWriteTopic` should all be present.

AnandInguva:
> We should not be replacing `KafkaTopic`, since we are not dropping support for the older parameter, just adding support for the new ones. So I think `KafkaTopic`, `KafkaReadTopic` and `KafkaWriteTopic` should all be present.

KafkaReadTopic is the new KafkaTopic now. since we don't use KafkaTopic anywhere, It would be better to remove it. From the UX, this is not a breaking change.

v2/kafka-to-kafka/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToKafka.java
```diff
@@ -82,7 +82,10 @@ public static PipelineResult run(KafkaToKafkaOptions options) {
       checkArgument(
           options.getKafkaReadKeyPasswordSecretId().trim().length() > 0,
           "KafkaReadKeyPasswordSecretId version for key password required for SSL authentication");
-    } else if (options.getKafkaReadAuthenticationMode().equals(KafkaAuthenticationMethod.NONE)) {
+    } else if (options.getKafkaReadAuthenticationMode().equals(KafkaAuthenticationMethod.NONE)
```
an2x, line 85: nit: I'd add a comment explaining why the if block is empty (e.g. "no extra validation needed for these auth mechanisms as they don't require any extra parameters").

an2x (approved):

Dippatel98 (approved):

v2/kafka-to-kafka/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToKafka.java
AnandInguva: Done

PR #1684, opened 2024-06-24
Fix BQ deadletter queue for Kafka template

AnandInguva:
The current BQ dlq fails if the table is already created with empty schema.  Change the BQ dlq pipline option template annotation to Text and ask the user for the fully qualified name of the BQ table.

### DIFF
--- a/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/dlq/BigQueryDeadLetterQueueOptions.java
+++ b/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/dlq/BigQueryDeadLetterQueueOptions.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/dlq/BigQueryDeadLetterQueue.java
```diff
@@ -0,0 +1,51 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.kafka.dlq;
+
+import com.google.cloud.teleport.metadata.TemplateParameter;
+import org.apache.beam.sdk.options.Default;
+
+public interface BigQueryDeadLetterQueue {
```
an2x, line 21: By convention, all pipeline options interfaces end with "Options", i.e. this one should be "BigQueryDeadLetterQueueOptions".

Dippatel98 (approved):
LGTM, but lets wait for the ITs to pass.

an2x (changes requested):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/dlq/BigQueryDeadLetterQueue.java
AnandInguva: Done

AnandInguva:
> LGTM, but lets wait for the ITs to pass.

Ran the affected test locally and it passed.

PR #1687, opened 2024-06-26
Binary datatype need to be saved with BINARY casting

aksharauke:
Binary datatype need to be saved with BINARY casting

### DIFF
--- a/v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
+++ b/v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
@@ -390,7 +390,7 @@ public class DMLGenerator {
       case "binary":
       case "varbinary":
       case "bit":
-        response = getHexString(colValue, spannerColType);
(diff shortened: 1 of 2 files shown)


v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
```diff
@@ -390,7 +390,7 @@ private static String getColumnValueByType(
       case "binary":
       case "varbinary":
       case "bit":
-        response = getHexString(colValue, spannerColType);
+        response = getBinaryString(colValue, spannerColType);
```
darshan-sj, line 393: Is this not required for blobs?

darshan-sj (approved):
LGTM

v2/gcs-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/processing/dml/DMLGenerator.java
aksharauke: No, blobs is stored as string only

PR #1688, opened 2024-06-26
ignoring views from ddl tables

bharadwaj-aditya:
Ignore views while performing bulk migrations.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScanner.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScanner.java
@@ -110,12 +110,13 @@ public class InformationSchemaScanner {
             Statement.of(
                 "SELECT t.table_name, t.parent_table_name, t.on_delete_action"
                     + " FROM information_schema.tables AS t"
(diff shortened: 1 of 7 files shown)


Deep1998:
LGTM

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -66,6 +67,10 @@ public Ddl scan() {
 
     for (Map.Entry<String, NavigableMap<String, Index.Builder>> tableEntry : indexes.entrySet()) {
       String tableName = tableEntry.getKey();
+      if (builder.getTable(tableName) == null) {
+        continue; // Skipping as table does not exist
```
Deep1998, line 71: Why do we need to add these checks? its unclear why we would get a table name in the indexes map? The index map seems to store index_name -> index struct.

bharadwaj-aditya: The index query does not join with the tables. So the filters on table do not apply here when this result is returned. Same for all the other queries here.

PR #1689, opened 2024-06-26
[BULK] Add type coverage for custom transformations

Deep1998:
Also add json coverage for basic all data types test


bharadwaj-aditya (approved):
LGTM

PR #1691, opened 2024-06-27
Mongo to BQ - advanced filters

stankiewicz:
Mongo to BQ has poor support for filters - they are on dataflow side with help of UDFs. This makes it expensive to run for Mongo Database and is expensive to filter on dataflow side. 

BSON filters are pushed to MongoDB which allows easy filtering, especially for daily ETLs.

### DIFF
--- a/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
+++ b/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
@@ -89,7 +89,6 @@ public class MongoDbToBigQueryOptions {
         optional = true,
         helpText = "Bson filter in json format.",
         example = "{ \"val\": { $gt: 0, $lt: 9 }}")
-    @Default.String("{}")
     String getFilter();
 
     void setFilter(String jsonFilter);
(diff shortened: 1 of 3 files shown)


v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
```diff
@@ -128,6 +134,7 @@ public static boolean run(Options options)
             MongoDbIO.read()
                 .withUri(mongoDbUri)
                 .withDatabase(options.getDatabase())
+                .withQueryFn(FindQuery.create().withFilters(filter))
```
Abacn, line 137: let's do not add this config to config when the filter is null, empty, or {}. There might be performance implications.

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbToBigQuery.java
```diff
@@ -106,6 +109,9 @@ public static boolean run(Options options)
     // Get MongoDbUri plain text or base64 encrypted with a specific KMS encryption key
     String mongoDbUri = maybeDecrypt(options.getMongoDbUri(), options.getKMSEncryptionKey()).get();
 
+    String filterJson = options.getFilter();
```
Abacn, line 112: good to consider the edge case of null or empty string

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
```diff
@@ -81,6 +81,18 @@ public interface MongoDbOptions extends PipelineOptions, DataflowPipelineOptions
...
     void setKMSEncryptionKey(String keyName);
+
+    @TemplateParameter.Text(
+        order = 6,
+        groupName = "Source",
+        description = "Bson filter",
+        optional = true,
+        helpText = "Bson filter in json format.",
+        example = "{ \"val\": { $gt: 0, $lt: 9 }}")
+    @Default.String("{}")
```
Abacn, line 92: If we handles the null, the Default here is not necessary.

Abacn (approved):

PR #1697, opened 2024-06-28
Add datastream to sql IT's

Polber:


Polber:
@damccorm Do you mind taking a look when you get a chance?

it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/cloudsql/CloudSqlResourceManagerTest.java
```diff
@@ -88,6 +88,11 @@ private MockCloudSqlResourceManager createTestManager(boolean useCustomDb) {
           public @NonNull CloudSqlResourceManager build() {
             return new MockCloudSqlResourceManager(this);
           }
+
+          @Override
+          protected void configurePort() {
+            this.setPort(1234);
```
damccorm, line 94: Why do we need this? Is it just to silence the warning or are we depending on a static port number?

damccorm (commented):
Overall LGTM, just had one question

it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/cloudsql/CloudSqlResourceManagerTest.java
Polber: Since I am testing an abstract class, I have to implement the abstract methods, so I just created a dummy method. But it also serves the purpose of ensuring the correct port is passed in the tests which isn't necessarily trivial given that it is based on the Testcontainers framework which generates random port by default (not what we want for this static-based RM)

damccorm (approved):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudMySQLResourceManager.java
```diff
@@ -49,6 +53,15 @@ public Builder(String testId) {
       super(testId);
     }
 
+    @Override
+    protected void configurePort() {
+      if (System.getProperty("cloudProxyMySqlPort") != null) {
```
mstorer3772: 1) Lots of magic strings

2) Getting all these properties twice strikes me as inefficient.  Not a **big** deal, but trivial to avoid:
```
String portProp = System.getProperty("cloudProxyMySqlPort");
if (portProp != null) {
  this.setPort(Integer.parseInt(portProp));
} else {...}
```

I realize this method is used often, but all that does is emphasize how often these magic strings turn up, and introduce more chances to accidentally have the two calls look up different properties.

PR #1698, opened 2024-06-28
Interleave and FK IT

aksharauke:
Testing of following scenarios:
1. Ordered writes due to parent child (foreign key) relationship
2. Interleave tables on Spanner - testings insert,update delete
3. Primary key on source is different from Spanner


darshan-sj (approved):
LGTM

PR #1705, opened 2024-07-01
Add spanner change streams to big query base case IT test

ShuranZhang:
Add a base case spanner change streams to big query integration test. Follow up integration tests on more complex schema update scenarios will be added later in separate prs.

### DIFF
--- a/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
+++ b/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Google LLC
+ * Copyright (C) 2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -21,25 +21,16 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 import com.google.cloud.bigquery.BigQueryException;
-import com.google.cloud.bigquery.Field;
-import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.FieldValueList;
-import com.google.cloud.bigquery.LegacySQLTypeName;
-import com.google.cloud.bigquery.Schema;
-import com.google.cloud.bigquery.StandardTableDefinition;
-import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableResult;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.teleport.metadata.SkipDirectRunnerTest;
 import com.google.cloud.teleport.metadata.TemplateIntegrationTest;
 import java.io.IOException;
-import java.time.Duration;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -116,7 +107,6 @@ public class SpannerChangeStreamsToBigQueryIT extends TemplateTestBase {
     bigQueryResourceManager.createDataset(REGION);
 
     Function<LaunchConfig.Builder, LaunchConfig.Builder> paramsAdder = Function.identity();
-    ;
 
     launchInfo =
         launchTemplate(
@@ -151,19 +141,20 @@ public class SpannerChangeStreamsToBigQueryIT extends TemplateTestBase {
       assertTrue(validateChangeLogTableRow(row, colValPairs));
     }
   }
+
   boolean validateChangeLogTableRow(FieldValueList row, List<Pair> pairs) {
     for (Pair pair : pairs) {
       try {
-          if (!row.get(pair.col).getStringValue().equals(pair.val)) {
-            LOG.info(
-                "Changelog table value for col "
-                    + pair.col
-                    + " is: "
-                    + row.get(pair.col).getStringValue()
-                    + ", which is different from spanner table's value: "
-                    + pair.val);
-            return false;
-          }
+        if (!row.get(pair.col).getStringValue().equals(pair.val)) {
+          LOG.info(
+              "Changelog table value for col "
+                  + pair.col
+                  + " is: "
+                  + row.get(pair.col).getStringValue()
+                  + ", which is different from spanner table's value: "
+                  + pair.val);
+          return false;
+        }
       } catch (BigQueryException e) {
         LOG.info("Error when trying to read the value for column " + pair.col + ": " + e);
         return false;
@@ -203,6 +194,7 @@ public class SpannerChangeStreamsToBigQueryIT extends TemplateTestBase {
         + "`"
         + String.format(" WHERE Id = %d", key);
   }
+
   @NotNull
   private Supplier<Boolean> dataShownUp(String query, int minRows) {
     return () -> {
@@ -218,13 +210,13 @@ public class SpannerChangeStreamsToBigQueryIT extends TemplateTestBase {
     };
   }
 
-  private void waitForQueryToReturnRows(
-      String query, int resultsRequired, boolean cancelOnceDone) throws IOException {
+  private void waitForQueryToReturnRows(String query, int resultsRequired, boolean cancelOnceDone)
+      throws IOException {
     Config config = createConfig(launchInfo);
     Result result =
         cancelOnceDone
             ? pipelineOperator()
-            .waitForConditionAndCancel(config, dataShownUp(query, resultsRequired))
+                .waitForConditionAndCancel(config, dataShownUp(query, resultsRequired))
             : pipelineOperator().waitForCondition(config, dataShownUp(query, resultsRequired));
     assertThatResult(result).meetsConditions();
   }
--- a/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
+++ b/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
@@ -1,0 +1,231 @@
+/*
+ * Copyright (C) 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates.spannerchangestreamstobigquery;
+
+import static org.apache.beam.it.truthmatchers.PipelineAsserts.assertThatPipeline;
+import static org.apache.beam.it.truthmatchers.PipelineAsserts.assertThatResult;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.FieldList;
+import com.google.cloud.bigquery.FieldValueList;
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.Table;
+import com.google.cloud.bigquery.TableResult;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.teleport.metadata.SkipDirectRunnerTest;
+import com.google.cloud.teleport.metadata.TemplateIntegrationTest;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import org.apache.beam.it.common.PipelineLauncher.LaunchConfig;
+import org.apache.beam.it.common.PipelineLauncher.LaunchInfo;
+import org.apache.beam.it.common.PipelineOperator.Config;
+import org.apache.beam.it.common.PipelineOperator.Result;
+import org.apache.beam.it.common.utils.ExceptionUtils;
+import org.apache.beam.it.common.utils.ResourceManagerUtils;
+import org.apache.beam.it.gcp.TemplateTestBase;
+import org.apache.beam.it.gcp.bigquery.BigQueryResourceManager;
+import org.apache.beam.it.gcp.spanner.SpannerResourceManager;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.jetbrains.annotations.NotNull;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Integration test for {@link SpannerChangeStreamsToBigQuery Spanner Change Streams to BigQuery}
+ * template.
+ */
+@Category({TemplateIntegrationTest.class, SkipDirectRunnerTest.class})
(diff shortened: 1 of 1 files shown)


v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,224 @@
...
+    spannerResourceManager.executeDdlStatement(createTableStatement);
+    String createChangeStreamStatement =
+        String.format(
+            "CREATE CHANGE STREAM %s_stream FOR %s OPTIONS (value_capture_type = 'NEW_ROW')",
+            testName, spannerTable);
+    spannerResourceManager.executeDdlStatement(createChangeStreamStatement);
+    bigQueryResourceManager.createDataset(REGION);
+
+    Function<LaunchConfig.Builder, LaunchConfig.Builder> paramsAdder = Function.identity();
+    ;
```
Polber, line 110: remove
```suggestion
```

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,224 @@
...
+
+    int key = nextValue();
+    String firstName = UUID.randomUUID().toString();
+    String lastName = UUID.randomUUID().toString();
+    List<Pair> colValPairs =
+        Arrays.asList(new Pair("FirstName", firstName), new Pair("LastName", lastName));
+    Mutation expectedData = generateTableRow(spannerTable, key, colValPairs);
+    spannerResourceManager.write(Collections.singletonList(expectedData));
+    String query = queryCdcTable(cdcTable, key);
+    waitForQueryToReturnRows(query, 1, true);
```
Polber, line 137: We already have helper test classes to perform this function, in this case, `BigQueryRowsCheck`
```suggestion
    Result result =
        pipelineOperator()
            .waitForConditionsAndFinish(
                createConfig(info),
                BigQueryRowsCheck.builder(bigQueryResourceManager, cdcTable)
                    .setMinRows(1)
                    .build());

    assertThatResult(result).meetsConditions();
```

The only difference is the query you were using specified an Id equality clause, but that doesn't seem relevant as there should only be one row in the table anyway since it us spun up as a fresh instance, but correct me if I'm wrong

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,224 @@
...
+
+    TableResult tableResult = bigQueryResourceManager.runQuery(query);
+    assertEquals(1, tableResult.getTotalRows());
+    for (FieldValueList row : tableResult.iterateAll()) {
+      assertTrue(validateChangeLogTableRow(row, colValPairs));
+    }
```
Polber, line 143: Same here
```suggestion
    TableResult tableResult = bigQueryResourceManager.readTable(cdcTable);
    List<Map<String, Object>> expectedRecords = 
        SpannerAsserts.mutationsToRecords(List.of(expectedData));
    assertThatBigQueryRecords(tableResult).hasRecordsUnordered(expectedRecords);
```

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,224 @@
...
+@TemplateIntegrationTest(SpannerChangeStreamsToBigQuery.class)
+@RunWith(JUnit4.class)
+public class SpannerChangeStreamsToBigQueryIT extends TemplateTestBase {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerChangeStreamsToBigQueryIT.class);
+
+  private SpannerResourceManager spannerResourceManager;
+  private BigQueryResourceManager bigQueryResourceManager;
+
+  private static final AtomicInteger counter = new AtomicInteger(0);
```
Polber, line 71: Not sure why counter and relevant methods are necessary when only row is being written

Polber (changes requested):

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,224 @@
+/*
+ * Copyright (C) 2023 Google LLC
```
Polber, line 2: nit:
```suggestion
 * Copyright (C) 2024 Google LLC
```

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: Done.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: Thanks for sharing this test util function! Yes you are correct that for this base case test, it doesn't really matter because only one row will be written to changelog table. However this pr is only the first step for more complex tests in later prs. I should have provided more context on the follow up prs, sorry for the confusions.

 For example, this is the draft of the follow up test pr(in a separate forked repo): https://github.com/ChangyuLi28/DataflowTemplates/commit/a1c2872a700a410fed7c95193a0c4103a3fa55df#diff-28b3790277fb958cc3acef6775b2e789ad279e97d8f962068915c9800a647869 
For example, we will have new test case that tests the schema update handling during pipeline running with the above process:
1. Start the pipeline
2. Insert a new row with id = 1
3. Verify table row in BigQuery with id=1 is correct
4. Add a new column to the spanner table
5. Insert a new row with id = 2 and the newly added column specified
6. Verify table row in BigQuery with id=2 is correct and the newly added column is correctly populated.

In this scenario, we will rely on this `queryCdcTable` function to only query the row with id=2, which is the row after the schema update. Thus I thought it would be more consistent for all tests in this file to use this same test util function to query the changelog table.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: Because we will have more complex test cases coming up which will write more than one rows in one test case(see comment above). Draft follow up pr: https://github.com/ChangyuLi28/DataflowTemplates/commit/a1c2872a700a410fed7c95193a0c4103a3fa55df#diff-28b3790277fb958cc3acef6775b2e789ad279e97d8f962068915c9800a647869

Let me know if you still prefer to remove this atomic counter and just use hard coded random uuid key value instead. Thanks!

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: Thanks for sharing this test util function! I tried using this but ended up with above failure
image
As you can see we have a bunch of metadata columns in cddTable, and we only care about the values for the actual mapping columns with the source spanner table. In addition, the metadata columns `_metadata_spanner_commit_timestamp` and `_metadata_bigquery_commit_timestamp` are in-place generated by spanner and bigquery. We can't really get the exact expect values for them ahead of time. Thus I thought it would be easier to just compare the column values of the actual spanner table columns. Let me know if you have better suggestions on handling this, thank!

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
Polber: I did something very similar in the Datastream IT's (also does CDC)

Take for example https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java

In particular https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0fef11cae7308a6389bda6cafb6fc7c9b68987ce/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java#L406-L424

For your use case, you could do something like
```
ChainedConditionCheck conditionCheck = 
     ChainedConditionCheck.builder( 
             List.of( 
                 writeSpannerData(...), 
                 BigQueryRowsCheck.builder(bigQueryResourceManager, cdcTable)
                    .setMinRows(1)
                    .build(), 
                 writeSpannerData(...), 
                 BigQueryRowsCheck.builder(bigQueryResourceManager, cdcTable)
                    .setMinRows(2)
                    .build())) 
         .build(); 
```

where `writeSpannerData` is a `ConditionCheck` so that it can be chained in a particular order with the actual checks.

Ultimately, as long as the test doesn't break PR submissions or add burden to the core template infra, it is up to you and your team how the IT's are implemented. I just wanted to lay out some patterns that we like to use to possibly make maintenance a bit easier

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
Polber: Could also do
```
TableResult tableResult = bigQueryResourceManager.readTable(cdcTable);
    List<Map<String, Object>> expectedRecords = 
        SpannerAsserts.mutationsToRecords(List.of(expectedData))
            .forEach(record -> assertThatBigQueryRecords(tableResult).hasRecordSubset(record);
```
Which does a partial check to make sure each BQ row contains _at least_ the columns/values from the expected spanner row

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
Polber: No that makes sense if you plan to test a chain of events in a single test case. I would still look at my comment above because you may be able to make that chaining easier, but in general a counter makes more sense than UUID because it guarantees to not have row collisions

ShuranZhang: I see, to avoid further modifications I think it makes sense keep the increment counters in this base pr.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: Thanks for the code pointer! I definitely agree that the chained conditional check is a more elegant and clean way for this scenario. However there is a key difference between datastream to spanner it and spanner cdc to big query it. 

`BigQueryRowsCheck.builder` takes in a TableId as parameter.
 For our case, the corresponding cdc table in big query sink is implicitly created by the template:
image
If we directly get the table id like this for the cdcTable, it failed with table not exist error, since it could take couple minutes to start the pipeline and internally create the cdc table(I verified this locally). Thus we will not only need to wait until the row is populated in the cdc table, but also need to wait for the table itself is created behind the scene. That's why I caught the table not found exception in the `dataShownUp` function:
image
Where as in DatastreamToSpannerIT, the sink tables in Spanner are [manually created at first](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0fef11cae7308a6389bda6cafb6fc7c9b68987ce/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java#L367) and the spanner row checker only takes a table name string parameter.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: I tried `.hasRecordSubset` as you suggested but still encounters the same failure:
image
Seems like `mutationsToRecords` does not guaranteed the ordinal position for the corresponding column. I built the spanner mutation in the order of id, first name and last name. However the mutationsToRecords returns it in the order of first name, id and last name, which is in alphabetical order. `hasRecordSubset` seems can't handle unordered situation like `hasRecordsUnordered `. Let me know if you are ok with keeping current column validation logic or I can also manually re-sort `List<Map<String, Object>>` in original column ordinal position order in spanner table. Thanks!

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
```diff
@@ -0,0 +1,223 @@
...
+        Arrays.asList(new Pair("FirstName", firstName), new Pair("LastName", lastName));
+    Mutation expectedData = generateTableRow(spannerTable, key, colValPairs);
+    spannerResourceManager.write(Collections.singletonList(expectedData));
+    String query = queryCdcTable(cdcTable, key);
+    waitForQueryToReturnRows(query, 1, true);
+
+    TableResult tableResult = bigQueryResourceManager.runQuery(query);
+    assertEquals(1, tableResult.getTotalRows());
+    for (FieldValueList row : tableResult.iterateAll()) {
+      assertTrue(validateChangeLogTableRow(row, colValPairs));
```
nancyxu123, line 141: Is there a way to also validate that there are no elements in the permanent DLQ?

nancyxu123:
LGTM to me. Adding a comment, since don't know how to add myself as a reviewer to the PR.

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
ShuranZhang: I disabled disableDlqRetries for the test pipeline so it is guaranteed that permanent dlq is empty if we received the eligible record successfully earlier.

ShuranZhang:
Hi @Polber May I get another review when you get a chance? This pr is blocking some follow up prs. Thank you so much!

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
Polber: Hm that's unfortunate. I Should probably update that. But for this PR, that's totally fine

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/spannerchangestreamstobigquery/SpannerChangeStreamsToBigQueryIT.java
Polber: You also don't have to call it that way. 

Instead of 
```
TableId cdcTableId = bigQueryResourceManager.getTableIfExists(cdcTable).getTableId();
```

you could just use
```
TableId cdcTableId = TableId.of(bigQueryResourceManager.getDatasetId(), cdcTable);
```

The `BigQueryRowsCheck` catches exceptions and throws a WARN instead, so as it tries to query the table size, it will throw a warning that the table does not exist each cycle. We have lots of examples of templates that implicitly create BQ tables (especially DLQ tables), and they use the check like that.

PR #1709, opened 2024-07-02
Use Beam YAML main.py

Polber:
Also adds caching of yaml providers to Dockerfile as well as hides the `yaml_pipeline` provider from UI in the `@Template` tag

### DIFF
--- a/python/src/main/python/yaml-template/main.py
+++ b/python/src/main/python/yaml-template/main.py
@@ -17,20 +17,10 @@
 import argparse
 import json
 
-import jinja2
-import yaml
-
-import apache_beam as beam
-from apache_beam.io.filesystems import FileSystems
-from apache_beam.typehints.schemas import LogicalType
-from apache_beam.typehints.schemas import MillisInstant
-from apache_beam.yaml import cache_provider_artifacts
-from apache_beam.yaml import yaml_transform
-
-# Workaround for https://github.com/apache/beam/issues/28151.
-LogicalType.register_logical_type(MillisInstant)
+from apache_beam.yaml import main
 
 
+# TODO(polber) - remove _preparse_jinja_flags with Beam 2.58.0
 def _preparse_jinja_flags(argv):
   """Promotes any flags to --jinja_variables based on --jinja_variable_flags.
   This is to facilitate tools (such as dataflow templates) that must pass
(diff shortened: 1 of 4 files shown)


Polber:
R: @robertwb

python/src/main/python/yaml-template/main.py
```diff
@@ -69,74 +60,12 @@ def _preparse_jinja_flags(argv):
...
-            **yaml_transform.SafeLineLoader.strip_metadata(pipeline_spec.get(
-              'options', {}))),
-          display_data={'yaml': pipeline_yaml}) as p:
-    print("Building pipeline...")
-    yaml_transform.expand_pipeline(
-      p, pipeline_spec, validate_schema=known_args.json_schema_validation)
-    print("Running pipeline...")
+  parser = argparse.ArgumentParser()
+  _, pipeline_args = parser.parse_known_args(_preparse_jinja_flags(argv))
+  pipeline_args += ['--sdk_location=container']
```
robertwb, line 66: Why do we want this flag? Was it just for testing the dev release?

python/src/main/python/yaml-template/main.py
```diff
@@ -69,74 +60,12 @@ def _preparse_jinja_flags(argv):
...
+  cache_provider_artifacts.cache_provider_artifacts()
```
robertwb, line 67: This should be called at container build time, not at runtime. (In particular, it will attempt to cache all artifacts, whether or not they're used.

python/src/main/python/yaml-template/main.py
Polber: This should help with container startup speed according to docs

https://cloud.google.com/dataflow/docs/guides/build-container-image#pre-build_a_container_image_when_submitting_the_job

"This option prevents the workers from downloading the SDK when your job launches. Instead, the SDK is retrieved directly from the container image."

python/src/main/python/yaml-template/main.py
Polber: How can it be called at runtime? The template is not expanded during container build time. Building the container just involves running the Dockerfile that creates the launcher image by taking the base image and copying in relevant files like the template file itself (main.py). 

This would get run during expansion in the template launcher which I thought then cached artifacts for the workers to use

python/src/main/python/yaml-template/main.py
robertwb: We should look at the code, but I'm pretty sure this doesn't apply to released versions that already assume the SDK is installed in the container.

python/src/main/python/yaml-template/main.py
robertwb: I was suggesting that artifact caching be done at container build time, via a RUN command in the docker file. 

`cache_provider_artifacts.cache_provider_artifacts` installs the artifacts into the current environment (which is the container the template launcher uses and will not impact or help the workers).

python/src/main/python/yaml-template/main.py
Polber: Just ran a speed test and it ran in about the same time with and without the flag specified, so I will remove

python/src/main/python/yaml-template/main.py
Polber: Ah ok after looking at the source code, I was misunderstanding what that function was actually doing. I will add that logic to the Dockerfile in this PR

Polber: This reduced startup by ~30 seconds, so thanks for the call out!

python/src/main/python/yaml-template/main.py
```diff
@@ -69,78 +60,13 @@ def _preparse_jinja_flags(argv):
...
-            pipeline_args,
-            pickle_library='cloudpickle',
-            **yaml_transform.SafeLineLoader.strip_metadata(pipeline_spec.get(
-              'options', {}))),
-          display_data={'yaml': pipeline_yaml}) as p:
-    print("Building pipeline...")
-    yaml_transform.expand_pipeline(
-      p, pipeline_spec, validate_schema=known_args.json_schema_validation)
-    print("Running pipeline...")
+  parser = argparse.ArgumentParser()
```
robertwb, line 64: I don't think you need this empty ArgumentParser anymore; you can just do

```
main.run(_preparse_jinja_flags(argv))
```

robertwb (approved):
You might want to update the description with the other changes as well (e.g. making the yaml argument hidden, caching in the docker build.)

python/src/main/python/yaml-template/main.py
Polber: Good call. Developing python on IntelliJ is a painful experience...

PR #1711, opened 2024-07-03
Terraform: Improvements to existing Terraform templates for live migrations

manitgupta:
Made some changes to the single shard templates - 

1. Updated datastream configuration to make it easier to specify the database and tables. 
2. Made it easier to configure session file in the job.
3. Updated `README` with notes about private connectivity related IAM permissions etc.
4. Fixed a bug with label always being auto-generated id.
5. Remove redundant dependencies in `depends_on` and instead rely on Terraform dependency resolution.

### DIFF
--- a/v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/variables.tf
+++ b/v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/variables.tf
@@ -29,8 +29,8 @@ variable "datastream_params" {
     stream_prefix_path            = optional(string, "data")
     pubsub_topic_name             = optional(string, "live-migration")
     stream_id                     = optional(string, "mysql-stream")
-    max_concurrent_cdc_tasks      = optional(number, 50)
-    max_concurrent_backfill_tasks = optional(number, 50)
+    max_concurrent_cdc_tasks      = optional(number, 5)
+    max_concurrent_backfill_tasks = optional(number, 15)
     mysql_database = object({
       database = string
       tables   = optional(list(string))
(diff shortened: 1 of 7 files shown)


v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/variables.tf
```diff
@@ -29,17 +29,18 @@ variable "datastream_params" {
     stream_prefix_path            = optional(string, "data")
     pubsub_topic_name             = optional(string, "live-migration")
     stream_id                     = optional(string, "mysql-stream")
-    max_concurrent_cdc_tasks      = optional(number, 50)
-    max_concurrent_backfill_tasks = optional(number, 50)
-    mysql_databases = list(object({
+    max_concurrent_cdc_tasks      = optional(number, 5)
+    max_concurrent_backfill_tasks = optional(number, 15)
```
bharadwaj-aditya, line 33: Any reasoning behind selecting 15 as the backfill tasks ?

bharadwaj-aditya (commented):
Minor query. Rest looks fine.

v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/variables.tf
manitgupta: Is the default in Datastream. We had made a similar change to the other sample previously. 
5 for CDC and 15 for backfill are the defaults.

bharadwaj-aditya: I was under the impression 8 was the default and 50 is what we set for max parallelism (considering concurrent backfill). Let's just make sure the default being set here is being set based on some benchmarks. You can take this up in a follow up PR if required.

bharadwaj-aditya (approved):
LGTM

v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/variables.tf
manitgupta: I confused your comment with using Datastream defaults, will change this all of this to 5 and 50 (as per HB defaults) as part of the sharding PR.

PR #1712, opened 2024-07-03
added fk and referenced tables

bharadwaj-aditya:
Expose foreign key details and references in Spanner DDL

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Ddl.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Ddl.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NavigableSet;
 import java.util.Set;
+import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Ddl.java
```diff
@@ -81,6 +82,25 @@ public Table apply(@Nullable String input) {
...
+   * provides the immediate references, not the whole tree.
+   *
+   * @param tableName
+   * @return
+   */
+  public List<String> tablesReferenced(String tableName) {
+    List<String> tablesReferenced = new ArrayList<>();
+    Table table = tables.get(tableName.toLowerCase());
+    if (table.interleaveInParent() != null) {
+      tablesReferenced.add(table.interleaveInParent());
```
Deep1998, line 96: Shouldnt we put interleave parent in the same set and convert the final set ot list? This could cause duplicate table names if a referenced table is also an interleave parent.

Deep1998 (approved):

PR #1713, opened 2024-07-03
ordered writes multishard

aksharauke:
Multishard ordered writes test

### DIFF
--- a/v2/gcs-to-sourcedb/src/test/resources/GCSToSourceDbInterleaveMultiShardIT/spanner-schema.sql
+++ b/v2/gcs-to-sourcedb/src/test/resources/GCSToSourceDbInterleaveMultiShardIT/spanner-schema.sql
@@ -1,0 +1,41 @@
+CREATE TABLE parent1 (
+  id INT64 NOT NULL,
+  update_ts TIMESTAMP,
+  in_ts TIMESTAMP,
+  migration_shard_id STRING(50),
+) PRIMARY KEY(id);
+
+CREATE TABLE parent2 (
+  id INT64 NOT NULL,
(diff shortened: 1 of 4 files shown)


v2/gcs-to-sourcedb/src/test/resources/GCSToSourceDbInterleaveMultiShardIT/spanner-schema.sql
```diff
@@ -0,0 +1,41 @@
...
+) PRIMARY KEY(id);
+
+CREATE TABLE parent2 (
+  id INT64 NOT NULL,
+  update_ts TIMESTAMP,
+  in_ts TIMESTAMP,
+  migration_shard_id STRING(50),
+) PRIMARY KEY(id);
+
+CREATE TABLE child11 (
```
darshan-sj, line 15: child11 has foreign key in MySQL schema. Is this intended? Or should we have either interleaving or FK here in spanner as well?

darshan-sj (approved):
LGTM

v2/gcs-to-sourcedb/src/test/resources/GCSToSourceDbInterleaveMultiShardIT/spanner-schema.sql
aksharauke: It was intended. Actually I took the same schema that SMT generated, looks like Spanner FK were missed. Spanner FK will in schema will guarantee ordered writes : which am doing programmatically, so keeping it such.

darshan-sj (approved):

PR #1715, opened 2024-07-03
Update codeowners and relevant workflows

Polber:
This will help auto-assign reviewers to core templates infra, with reviews for individual templates being left to author to decide.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -58,8 +58,8 @@ v2/spanner-custom-shard/ @GoogleCloudPlatform/spanner-migrations-team
 v2/spanner-to-sourcedb/ @GoogleCloudPlatform/spanner-migrations-team
 
 # Spanner testing infra
-it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner @GoogleCloudPlatform/dataflow-templates-wg
-lt/
+it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner @GoogleCloudPlatform/spanner-migrations-team
+lt/ @GoogleCloudPlatform/spanner-migrations-team
 
 # Spanner workflow files
 .github/workflows/spanner-pr.yml @GoogleCloudPlatform/spanner-migrations-team
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,41 @@
 # review when someone opens a pull request
 
+# Core templates cicd
+.github/ @GoogleCloudPlatform/dataflow-templates-wg
+cicd/ @GoogleCloudPlatform/dataflow-templates-wg
+
+# Core templates infra
+metadata/ @GoogleCloudPlatform/dataflow-templates-wg
+plaintext-logging/ @GoogleCloudPlatform/dataflow-templates-wg
+plugins/ @GoogleCloudPlatform/dataflow-templates-wg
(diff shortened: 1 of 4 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,41 @@
...
+
+# Core templates infra
+metadata/ @GoogleCloudPlatform/dataflow-templates-wg
+plaintext-logging/ @GoogleCloudPlatform/dataflow-templates-wg
+plugins/ @GoogleCloudPlatform/dataflow-templates-wg
+structured-logging/ @GoogleCloudPlatform/dataflow-templates-wg
+checkstyle/ @GoogleCloudPlatform/dataflow-templates-wg
+
+# Templates testing infra
+it/ @GoogleCloudPlatform/dataflow-templates-wg
```
damccorm, line 15: This is pretty broad and impacts most templates, lets remove this

damccorm (commented):
Thanks! This overall LGTM, just had one comment

.github/CODEOWNERS
Polber: The intention of this PR was to add `@GoogleCloudPlatform/dataflow-templates-wg` as owners of core infra code that impacts the majority of templates - including the IT framework. 

The `it/` module does not include individual tests, but does include the RM's, conditions, truthmatchers, pipeline handlers, etc. that owners of sub-modules would use to write their tests. 

Presumably, there shouldn't be much change to this module unless a new RM is added or modified for a new use-case, which I would assume our team would want to check and make sure it isn't spinning up or managing resources in an unexpected way. 

This seemed in-scope to me, but let me know if I am off-base with this line of thinking

damccorm: That's fair - I think I'd like to avoid us owning all subdirectories like https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/main/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner which are specific to sets of templates, but I guess we can add those as exclusions (like the below `.github/workflows/spanner-*` lines).

Could we add `it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner` ownership to spanner below as well? If we do that, I feel good about where we stand

Polber: Updated

.github/CODEOWNERS
```diff
@@ -21,6 +57,15 @@ v2/spanner-migrations-sdk/ @GoogleCloudPlatform/spanner-migrations-team
 v2/spanner-custom-shard/ @GoogleCloudPlatform/spanner-migrations-team
 v2/spanner-to-sourcedb/ @GoogleCloudPlatform/spanner-migrations-team
 
+# Spanner testing infra
+it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner @GoogleCloudPlatform/dataflow-templates-wg
```
damccorm, line 61: Should this point to spanner-migrations-team?

.github/CODEOWNERS
```diff
@@ -21,6 +57,15 @@ v2/spanner-migrations-sdk/ @GoogleCloudPlatform/spanner-migrations-team
...
+lt/
```
damccorm, line 62: Should this have an owner?

.github/CODEOWNERS
Polber: Oops, yes - thanks

damccorm (approved):
Thanks!

PR #1718, opened 2024-07-08
Fixing timeout handling for uniform partitioning and (temporarily) disabling splitting overhad

VardhanThigle:
## Timeout Handling
In Change for splitting a space of single/composite long/integers [PR#165](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1657/), timeout is handle via trapping `java.SQL.Timeout`. It turns out that there are situates where the DB is throwing a normal `java.sql.exception` and we will have to look into the error codes to decide whether it was a timeout.

This does not affect tables up to 1 TB size.

## Disabling the overhead of splitting process
While  counting a range completes within 5 seconds if not done in parallel, moment the dataflow job tries to count a lot of ranges in parallel, all of them timeout rendering the process of splitting and uniformizing useless. Although this PR is pushing some of the optimizations done towards better counting, for the main line code, it is also making the stages to 0, essentially removing the overhead entirely.  

The exercise to fine tune the counting process will be continued in the baground and the splits will be enabled again later on.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/JdbcIOWrapperConfig.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/JdbcIOWrapperConfig.java
@@ -114,6 +114,38 @@ public abstract class JdbcIOWrapperConfig {
   public abstract OnSignal<?> waitOn();
 
   /**
+   * If not null, maximum number of parallel queries issued to the DB during split process. Ignored
+   * if {@link JdbcIOWrapperConfig#readWithUniformPartitionsFeatureEnabled()} is false. It's best to
+   * set this to a number close to number of cores available on mySql server. Defaults to {@link
+   * JdbcIOWrapperConfig#DEFAULT_PARALLELIZATION_FOR_SLIT_PROCESS}.
+   *
+   * <p><b>Performance</b>
+   *
+   * <ul>
+   *   <li>Ensure that <a
+   *       href=https://dev.mysql.com/doc/refman/8.4/en/innodb-performance-multiple_io_threads.html>innodb_read_io_threads</a>
+   *       is set to the default value as recommended by Mysql or higher. If the partitioning is
+   *       slow due to too many queries timing out in each stage, and if the {@code SHOW ENGINE
+   *       INNODB STATUS} shows pending queries close to innodb_read_io_threads, it's an indication
+   *       to increase this setting.
+   *   <li>Ensure that <a
+   *       href=https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_buffer_pool_size>sysvar_innodb_buffer_pool_size</a>
+   *       is set to the default value as recommended by Mysql or higher.
+   * </ul>
+   */
+  @Nullable
+  public abstract Integer dbParallelizationForSplitProcess();
+
+  private static final int DEFAULT_PARALLELIZATION_FOR_SLIT_PROCESS = 100;
+
+  /**
+   * If not null, maximum number of parallel queries issued to the DB for reads. Ignored if {@link
+   * JdbcIOWrapperConfig#readWithUniformPartitionsFeatureEnabled()} is false. Defaults to null.
+   */
+  @Nullable
+  public abstract Integer dbParallelizationForReads();
+
+  /**
    * A transform that can be injected to make use of the discovered splits for additional use case
    * like creating split points on spanner before the actual read. Ignored if {@link
    * JdbcIOWrapperConfig#readWithUniformPartitionsFeatureEnabled()} is false. Defaults to null.
@@ -136,6 +168,8 @@ public abstract class JdbcIOWrapperConfig {
         .setMaxPartitions(null)
         .setWaitOn(null)
         .setMaxFetchSize(null)
+        .setDbParallelizationForReads(null)
+        .setDbParallelizationForSplitProcess(DEFAULT_PARALLELIZATION_FOR_SLIT_PROCESS)
         .setReadWithUniformPartitionsFeatureEnabled(true);
   }
 
@@ -176,6 +210,10 @@ public abstract class JdbcIOWrapperConfig {
     public abstract Builder setReadWithUniformPartitionsFeatureEnabled(Boolean value);
 
     public abstract Builder setWaitOn(@Nullable OnSignal<?> value);
+
+    public abstract Builder setDbParallelizationForSplitProcess(@Nullable Integer value);
+
+    public abstract Builder setDbParallelizationForReads(@Nullable Integer value);
 
     public abstract Builder setAdditionalOperationsOnRanges(
         @Nullable PTransform<PCollection<ImmutableList<Range>>, ?> value);
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java
@@ -222,11 +222,13 @@ public abstract class Range implements Serializable, Comparable<Range> {
     return Pair.of(
         this.toBuilder()
             .setBoundary(boundaries.getLeft())
+            .setCount(INDETERMINATE_COUNT)
             .setIsFirst(isFirst())
             .setIsLast(false)
             .build(),
         this.toBuilder()
             .setBoundary(boundaries.getRight())
+            .setCount(INDETERMINATE_COUNT)
             .setIsFirst(false)
             .setIsLast(isLast())
             .build());
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
@@ -18,8 +18,10 @@ package com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.trans
 import static org.apache.beam.sdk.util.Preconditions.checkStateNotNull;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.UniformSplitterDBAdapter;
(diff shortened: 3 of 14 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
```diff
@@ -56,6 +60,53 @@ final class RangeCountDoFn extends DoFn<Range, Range> implements Serializable {
...
+          "aborted",
+          "interrupt",
+          "interrupted",
+          "terminate",
+          "terminated",
+          "timeout",
+          "time out",
+          "time exceeded");
+
+  private static final Pattern TIMEOUT_MESSAGES_PATTERN =
```
bharadwaj-aditya, line 100: There should be a cleaner way to extract the timeout errors here.

VardhanThigle: Thanks. Fixed in the current push.
We had to exchange a list of patterns with a list of error codes, now it's mysql specific which is no doubt a cleaner approach.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/JdbcIOWrapperConfig.java
```diff
@@ -113,6 +113,38 @@ public abstract class JdbcIOWrapperConfig {
...
+   *       to increase this setting.
+   *   <li>Ensure that <a
+   *       href=https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_buffer_pool_size>sysvar_innodb_buffer_pool_size</a>
+   *       is set to the default value as recommended by Mysql or higher.
+   * </ul>
+   */
+  @Nullable
+  public abstract Integer dbParallelizationForSplitProcess();
+
+  private static final int DEFAULT_PARALLELIZATION_FOR_SLIT_PROCESS = 100;
```
bharadwaj-aditya, line 139: nit: typo - SPLIT

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
```diff
@@ -97,43 +103,49 @@ public void processElement(@Element Range input, OutputReceiver<Range> out, Proc
...
           count = Range.INDETERMINATE_COUNT;
         }
-      } catch (SQLTimeoutException e) {
+      } else {
+        logger.error(
+            "Got empty count resultSet. Range = {}, Query = {}, DataSource = {}",
+            input,
+            countQuery,
+            dataSource);
+        count = Range.INDETERMINATE_COUNT;
```
bharadwaj-aditya, line 123: if we get an empty, then this should be set to 0. Only in the case of exception it makes sense to go back to an undetermined number.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
```diff
@@ -97,43 +103,49 @@ public void processElement(@Element Range input, OutputReceiver<Range> out, Proc
...
+        logger.error(
+            "Non-timeout SQL Exception while counting Range = {}, Query = {}, DataSource = {}, timeoutMillis = {}, exception {}, sqlState {}, errorCode {}",
+            input,
+            countQuery,
+            dataSource,
+            timeoutMillis,
+            e,
+            e.getSQLState(),
+            e.getErrorCode());
+        throw e;
```
bharadwaj-aditya, line 147: could this be a temporary connection issue or something ? In which case we should not fail the splitter as this is best effort. We should fall back to undetermined count.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/ReadWithUniformPartitions.java
```diff
@@ -182,7 +197,9 @@ public PCollection<T> expand(PBegin input) {
     PCollection<Range> rangesToRead =
         peekRanges(mergedRanges).apply(ParDo.of(new UnflattenRangesDoFn()));
     return rangesToRead
-        .apply(getTransformName("ReshuffleFinal", null), Reshuffle.viaRandomKey())
+        .apply(
+            getTransformName("ReshuffleFinal", null),
+            Reshuffle.<Range>viaRandomKey().withNumBuckets(dbParallelizationForReads()))
```
bharadwaj-aditya, line 202: this number should be lower as we may have a bunch of readers on a given MySQL instance for optimal usage. But i suggest we take this optimization up as as a follow up.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/ReadWithUniformPartitions.java
```diff
@@ -339,7 +363,7 @@ private PCollection<ImmutableList<Range>> peekRanges(
   }
 
   public static long inferMaxPartitions(long count) {
-    return Math.max(1, Math.round(Math.floor(Math.sqrt(count) / 10)));
+    return Math.max(1, Math.round(Math.floor(Math.sqrt(count) / 20)));
```
bharadwaj-aditya, line 366: take this number - 20 as a constant and justify why it is set to 20 in the constant deescription.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java
```diff
@@ -222,11 +222,13 @@ public Pair<Range, Range> split(@Nullable ProcessContext processContext) {
     return Pair.of(
         this.toBuilder()
             .setBoundary(boundaries.getLeft())
+            .setCount(INDETERMINATE_COUNT)
```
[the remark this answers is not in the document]
VardhanThigle: The count `INDETERMINATE_COUNT` is an indicates that the new split range is not yet counted.
This one-liner was not affecting our production code as split ranges were always marked for counting, the reset of count is done mostly from a UT/correctness POV here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
VardhanThigle: Ideally count(*) query should never return empty, it does return 0 for no rows.
making the count `Range.INDETERMINATE_COUNT` marks the range for recounting.

This case is mostly from a code completeness (or utilize return value) sake, we need to have some handle for rs.next() returning false, since it's unexpected, making it `Range.INDETERMINATE_COUNT` will mark the range for recounting.

We have error logs for this to catch it if we hit this in a real world situation.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
VardhanThigle: we could do that, but currently all the exceptions we throw at run time are retried by beam, it's the same strategy we follow for read too.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/ReadWithUniformPartitions.java
VardhanThigle:
```
this number should be lower as we may have a bunch of readers on a given MySQL instanc
```
That's right, we need to make it more configurable and also need to get more concrete benchmarking

`suggest we take this optimization up as as a follow up.`
Sure!

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/ReadWithUniformPartitions.java
VardhanThigle: `Math.max(1, Math.round(Math.floor(Math.sqrt(count) / 10` we have taken this (verbatim) from beam's JdbcIO.
The splitter generates initial ranges in powers of 2, so thats close to double of this number, making it 20 makes the total number of inital splits similar to jdbcIO.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/JdbcIOWrapperConfig.java
VardhanThigle: Fixed in the current push.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/ReadWithUniformPartitions.java
VardhanThigle: Wrapped it in a constant in latest push.

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeCountDoFn.java
bharadwaj-aditya: Add a todo to re-evaluate this with test runs.

bharadwaj-aditya (approved):
LGTM

PR #1721, opened 2024-07-09
Spanner Load tests - Increasing job cancel timeout

darshan-sj:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner100GbLT.java
```diff
@@ -173,7 +173,7 @@ public void backfill100Gb() throws IOException, ParseException, InterruptedExcep
     // Assert Conditions
     assertThatResult(result).meetsConditions();
 
-    result = pipelineOperator.cancelJobAndFinish(createConfig(jobInfo, Duration.ofMinutes(5)));
+    result = pipelineOperator.cancelJobAndFinish(createConfig(jobInfo, Duration.ofMinutes(20)));
```
manitgupta, line 176: With the parameterization work you are planning, let's aim to make this a configuration of the test being written.

manitgupta (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner100GbLT.java
darshan-sj: Okay, I will make things configurable with parameterization. But this time duration should be same for all the tests. It is duration between calling cancel job api and job actually getting cancelled and stopping.

PR #1375, opened 2024-03-19
FLOAT32 support in datastream

shubhamswe:
[Datastream] - FLOAT32 is now supported in datastream to spanner convertors.

The changes are tested against the following production components:
1. [MySQL Database](https://pantheon.corp.google.com/sql/instances/shubhamswe-test/databases?e=13802955&hl=en&mods=monitoring_api_prod&project=span-cloud-testing) 
2. [Data Stream](https://pantheon.corp.google.com/datastream/streams/locations/us-central1/instances/shubhamswe-mysql-to-gcs-stream;tab=overview?e=13802955&hl=en&mods=monitoring_api_prod&project=span-cloud-testing)
3. [Dataflow Job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2024-04-02_05_48_28-3407817368732960406;graphView=1?project=span-cloud-testing&e=13802955&hl=en&mods=monitoring_api_prod&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))
4. [Spanner Database](https://pantheon-staging.corp.google.com/spanner/instances/shubhamswe-test/databases/float32-test/details/query;showWriteDdlOptInBanner=true?b=cloud_spanner_api_name:STAGING,cloud_spanner_api_url:STAGING,coliseum_backend:AUTOPUSH,coliseum_host_backendz:AUTOPUSH&project=span-cloud-testing)

```sh
# gcloud cli setup
$ gcloud auth login
$ gcloud auth configure-docker

# Repository setup
$ git clone git@github.com:GoogleCloudPlatform/DataflowTemplates.git
$ cd DataflowTemplates/v2

# Build Template Image 
$ export PROJECT=span-cloud-testing
$ export IMAGE_NAME=datastream-to-spanner
$ export BUCKET_NAME=gs://shubhamswe-test
$ export TARGET_GCR_IMAGE=gcr.io/${PROJECT}/templates/${IMAGE_NAME}
$ export BASE_CONTAINER_IMAGE=gcr.io/dataflow-templates-base/java11-template-launcher-base
$ export BASE_CONTAINER_IMAGE_VERSION=latest
$ export APP_ROOT=/template/${IMAGE_NAME}
$ export DATAFLOW_JAVA_COMMAND_SPEC=${APP_ROOT}/resources/${IMAGE_NAME}-command-spec.json
$ export SPANNER_HOST="https://staging-wrenchworks.sandbox.googleapis.com/"
[...]

PR #1724, opened 2024-07-09
Support float32 type in datastream

arawind:
This is a patch of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1375.

Successfully ran the dataflow template:

MySQL source, stored to GCS via datastream:
Screenshot from 2024-07-10 01-12-34

Spanner destination from datastream:
Screenshot from 2024-07-10 01-12-52

```
export PROJECT=span-cloud-testing
export IMAGE_NAME=datastream-to-spanner
export BUCKET_NAME=gs://shubhamswe-test
export TARGET_GCR_IMAGE=gcr.io/${PROJECT}/templates/${IMAGE_NAME}
export BASE_CONTAINER_IMAGE=gcr.io/dataflow-templates-base/java11-template-launcher-base
export BASE_CONTAINER_IMAGE_VERSION=latest
export APP_ROOT=/template/${IMAGE_NAME}
export DATAFLOW_JAVA_COMMAND_SPEC=${APP_ROOT}/resources/${IMAGE_NAME}-command-spec.json
#export SPANNER_HOST="https://staging-wrenchworks.sandbox.googleapis.com/"
export INSTANCE_ID="arawind-test"
export DATABASE_ID="float32-test"
export GCS_LOCATION="gs://shubhamswe-test/float32-mysql-out"
export STREAM_NAME="projects/span-cloud-testing/locations/us-central1/streams/arawind-mysql-datastream"
export TEMPLATE_IMAGE_SPEC="gs://shubhamswe-test/images/2024_07_09_01/flex/Cloud_Datastream_to_Spanner"
export JOB_NAME="${IMAGE_NAME}-`date +%Y%m%d-%H%M%S-%N`"

gcloud dataflow flex-template run ${JOB_NAME} \
        --project=${PROJECT} --region=us-central1 \
        --template-file-gcs-location=${TEMPLATE_IMAGE_SPEC} \
        --parameters instanceId=${INSTANCE_ID},databaseId=${DATABASE_ID},inputFilePattern=${GCS_LOCATION},streamName=${STREAM_NAME}
```

### DIFF
--- a/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertorTest.java
+++ b/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertorTest.java
@@ -81,6 +81,9 @@ public class ChangeEventConvertorTest {
             .column("int64_field")
             .int64()
             .endColumn()
+            .column("float32_field")
+            .float32()
+            .endColumn()
             .column("float64_field")
             .float64()
             .endColumn()
@@ -150,6 +153,9 @@ public class ChangeEventConvertorTest {
             .column("int64_field")
             .int64()
             .endColumn()
+            .column("float32_field")
+            .float32()
+            .endColumn()
             .column("float64_field")
             .float64()
             .endColumn()
@@ -213,6 +219,9 @@ public class ChangeEventConvertorTest {
             .column("int64_field")
             .int64()
             .endColumn()
+            .column("float32_field")
+            .float32()
+            .endColumn()
             .column("float64_field")
             .float64()
             .endColumn()
@@ -268,6 +277,7 @@ public class ChangeEventConvertorTest {
     changeEvent.put("bool_field", "true");
     changeEvent.put("bool_field2", true);
     changeEvent.put("int64_field", "2344");
+    changeEvent.put("float32_field", "-137.81");
     changeEvent.put("float64_field", "2344.34");
     changeEvent.put("string_field", "testtest");
     changeEvent.put("json_field", "{\"key1\": \"value1\", \"key2\": \"value2\"}");
@@ -293,6 +303,7 @@ public class ChangeEventConvertorTest {
             put("bool_field", Value.bool(true));
             put("bool_field2", Value.bool(true));
             put("int64_field", Value.int64(2344));
+            put("float32_field", Value.float32(-137.81f));
             put("float64_field", Value.float64(2344.34));
             put("string_field", Value.string("testtest"));
             put("json_field", Value.string("{\"key1\": \"value1\", \"key2\": \"value2\"}"));
@@ -420,6 +431,15 @@ public class ChangeEventConvertorTest {
     Ddl ddl = getTestDdl();
     JSONObject changeEvent = getTestChangeEvent("Users");
     changeEvent.put("int64_field", "asdfas");
+    JsonNode ce = parseChangeEvent(changeEvent.toString());
+    Mutation mutation = ChangeEventConvertor.changeEventToMutation(ddl, ce);
+  }
+
+  @Test(expected = ChangeEventConvertorException.class)
+  public void cannotConvertChangeEventWithInvalidFloat32ToMutation() throws Exception {
+    Ddl ddl = getTestDdl();
+    JSONObject changeEvent = getTestChangeEvent("Users");
+    changeEvent.put("float32_field", "asdfasdf");
     JsonNode ce = parseChangeEvent(changeEvent.toString());
     Mutation mutation = ChangeEventConvertor.changeEventToMutation(ddl, ce);
   }
--- a/v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScannerTest.java
+++ b/v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScannerTest.java
@@ -169,12 +169,15 @@ public class InformationSchemaScannerTest {
     ResultSet listColumnsResultSet = mock(ResultSet.class);
 
     when(context.executeQuery(listColumns)).thenReturn(listColumnsResultSet);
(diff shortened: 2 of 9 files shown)


v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventTypeConvertorTest.java
```diff
@@ -205,6 +205,81 @@ public void cannotConvertNonExistentRequiredFieldToLong() throws Exception {
...
+    assertEquals(
+        ChangeEventTypeConvertor.toFloat(ce, "field1", /* requiredField= */ true),
+        new Long(123457));
+  }
+
+  @Test(expected = ChangeEventConvertorException.class)
+  public void cannotConvertNonExistentRequiredFieldToFloat() throws Exception {
+    JSONObject changeEvent = new JSONObject();
+    JsonNode ce = getJsonNode(changeEvent.toString());
+    assertEquals(
```
shreyakhajanchi, line 278: nit: Maybe better to just do  `ChangeEventTypeConvertor.toFloat(ce, "field1", /* requiredField= */ true)` for exception cases instead of assert, its a bit confusing

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertorTest.java
```diff
@@ -81,6 +81,9 @@ static Ddl getTestDdl() {
             .column("int64_field")
             .int64()
             .endColumn()
+            .column("float32_field")
```
shreyakhajanchi, line 84: Shouldn't there be changes in v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertor.java?

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScannerTest.java
```diff
@@ -355,16 +367,18 @@ public void testScanGSQLDdl() {
     Ddl ddl = informationSchemaScanner.scan();
     String expectedDdl =
         "CREATE TABLE `singer` (\n"
-            + "\t`singerId`                              STRING(50) NOT NULL,\n"
+            + "\t`singerId`                              INT64 NOT NULL,\n"
```
shreyakhajanchi, line 370: why are we modifying the datatype of other columns?

arawind: This is just to increase the coverage. The types were only testing STRING, added tests for INT64. Why INT64? Felt like SingerId would be an INT64 :)

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertorTest.java
arawind: No, that file only has transformations for primary keys. FLOAT32 is not a primary keyable type, which is also why we do not add it to the list of primary keys below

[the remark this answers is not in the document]
arawind: float_colum already exists in that test. I've modified spanner-schema.sql to correct the mapping from FLOAT64 to FLOAT32

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventTypeConvertorTest.java
arawind: Done

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertorTest.java
shreyakhajanchi: yeah makes sense, i didn't notice that the file code had been refactored

manitgupta:
Can you explain what am I looking at in the screenshots above? I think the change in value of `amount1` is expected? Can you link the docs here for me to read?

Is this change "normal" and "expected" - i.e will customers be okay with this change?

1.23 in MySQL becomes 1.230000002 in Spanner.

manitgupta (commented):
The code changes themselves look good, thanks for the rigor on updating the ITs as well.

I want to better understand the implications on precision of the data while migrating a source float value to Float32 with respect to how the behaviour changes from the current mapping of Float64.

arawind:
Thanks @manitgupta!

> Can you explain what am I looking at in the screenshots above? I think the change in value of `amount1` is expected? Can you link the docs here for me to read?
> 
> Is this change "normal" and "expected" - i.e will customers be okay with this change?
> 
> 1.23 in MySQL becomes 1.230000002 in Spanner.

TL;DR: The issue is just in the Cloud Console UI, the actual value of 1.23 is still preserved. 
The underlying reason is because we're up-casting the `float` value to `double` while on the Cloud Spanner API, as that's the best we can do with the limited types that the API provides.
Client libraries do the conversions float-to-double while sending, and double-to-float while receiving, and these conversions are lossless.
Now, since Cloud Console uses a JS client, and since JS does not support a float32 type, we're doing our best to convert the float64 value in the API to float32 in JS, but it is just an approximation to the actual down casting, and we end up with these values on the UI.

The actual values are preserved across migrations, import / export etc.

Design doc is available internally at go/cloud-spanner-float32.

shreyakhajanchi (approved):
LGTM

manitgupta (approved):
Thank you @arawind!

PR #1728, opened 2024-07-12
Add table counters and separate metrics on a table level

Deep1998:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/IncrementTableCounter.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.Wait;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IncrementTableCounter extends PTransform<PBegin, PCollection<Void>> {
```
bharadwaj-aditya, line 37: please add a javadoc for what this transform is supposed to do.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/IncrementTableCounter.java
```diff
@@ -0,0 +1,87 @@
...
+    this.shardId = shardId;
+  }
+
+  @Override
+  public PCollection<Void> expand(PBegin input) {
+    List<PCollection<String>> tableNamesPCollections = new ArrayList<>();
+    for (Map.Entry<String, Wait.OnSignal<?>> entry : tableWaits.entrySet()) {
+      String tableName = entry.getKey();
+      Wait.OnSignal<?> wait = entry.getValue();
+      PCollection<String> waitedPCollection =
```
bharadwaj-aditya, line 58: This seems a little complex to populate a counter when a table completes. Could you elaborate with a few comments/docs on why this is needed here ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/MigrateForTable.java
```diff
@@ -0,0 +1,132 @@
...
+    this.schemaMapper = schemaMapper;
+    this.reader = reader;
+    this.shardId = StringUtils.isEmpty(shardId) ? "" : shardId;
+  }
+
+  @Override
+  public PCollection<Void> expand(PBegin input) {
+    ReaderTransform readerTransform = reader.getReaderTransform();
+
+    PCollectionTuple rowsAndTables = input.apply("Read_rows", readerTransform.readTransform());
```
bharadwaj-aditya, line 73: suggest we name all the stages with shard and table name or something so it is easy to relate.

Deep1998: Since we have wrapped it in a ptransform now, we set the shard/table identified at the top level pipeline. Adding additional identifiers for steps will be redundant and have overhead of passing these parameters everywhere where.

All our stages now look like:
Migrate_for_<shard_id>_<table_name>/Read_rows/...
Migrate_for_<table_name>/Write_To_Spanner/... (non-sharded cases)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/IncrementTableCounter.java
Deep1998: Added, an alternative could potentially be: Run a count transform instead of a pcreate->wait.on. The rest remains the same. But the current approach also allows us to log which table completed.

bharadwaj-aditya: So i think i understand now. The wait on logic has been moved into this this class which is where i was getting confused. 

Ideally, this should be done in the migrate table transform and this counter should just be a simple logger on completion of table. It seems like this class is doing the waiting as well as the logging which is why it is as complex.

I would suggest -
Either move this logic to the migrate Table class.
or
Rename this class to reflect that fact that it is waiting for table completion as bulk of the logic is not the counter.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/MigrateForTable.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.TupleTagList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MigrateForTable extends PTransform<PBegin, PCollection<Void>> {
```
bharadwaj-aditya, line 44: nit: Rename to MigrateTableTransform

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/IncrementTableCounter.java
Deep1998: Slight correction, the wait on is not moved here. The parent dependency wait ons are as is. This is an additional wait that needs to be added, which basically says, wait for this table to complete before incrementing counter.

We cannot bypass using this additional waitOn, this is necessary for being able to do a dummy pcreate and increment a counter on completion.

And the issue all this inside the migrateTable class dataflow ends up making a separate counter for every step then. (We end up getting per table singleton counters).

bharadwaj-aditya: Closing as discussed.

bharadwaj-aditya (approved):
LGTM

PR #1729, opened 2024-07-12
fix CdcToBigQueryChangeApplierPipelineIT debezium wait logic

Polber:
Observed flakes starting up debezium connector, and noticed the wait logic would fail on first of 3 checks. Refactored to follow intended logic.


Polber:
Test failures unrelated. Addressed in #1753

Abacn (approved):

PR #1730, opened 2024-07-12
Add IT tests for handling schema update during spanner cdc to big query pipeline running

ShuranZhang:
Add IT tests for handling schema update during a spanner cdc to bigquery pipeline running.
Previously a base case test that doesn't contain any schema change was added in https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1705 . This pull request adds two more IT tests following above workflow:

- Add a column(**disabled**): add a column in spanner source table -> add a column in corresponding bq table ->wait for 15mins(this is due to limitation of BigQuery and BQ team recommends wait >10 mins, see more [here](https://github.com/googleapis/google-cloud-python/issues/7433#issuecomment-553145337))->insert a row to source table with new column filled->verify bq table's new column is successfully populated. Due to the above bq limitation, this test requires sleeping at least 10 mins in test body to ensure no flakiness. Thus I disabled this test to avoid just hang the thread and do hardcoded sleep for such a long time.
- Add a table: add a source table in spanner -> insert a new row to new source table -> verify that the new changelog table in created and filled with correct information in bq. Different from updating an existing bq table's schema, add a completely new table does not suffer from the high latency.

I have run all IT tests together in parallel for >15 times locally and no flakiness observed. We can add more test cases later if tests in this pr are running stable.

I also added a 3 minutes dlq retry parameters to all IT tests in case there are any availability/transient errors happening on the spanner side.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigquery/BigQueryResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigquery/BigQueryResourceManager.java
@@ -145,7 +145,7 @@ public final class BigQueryResourceManager implements ResourceManager {
    * @return the table, if it exists.
    * @throws IllegalStateException if the given table name does not exist in the dataset.
    */
-  private synchronized Table getTableIfExists(String tableId) throws IllegalStateException {
+  public synchronized Table getTableIfExists(String tableId) throws IllegalStateException {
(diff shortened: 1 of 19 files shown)


Abacn:
DLPTextToBigQueryStreamingIT.testDLPTextToBigQuery failed due to #1836

Abacn (approved):

PR #1731, opened 2024-07-14
Fixing Flaky test: Increasing timeout in SpannerChangeStreamToGcsMultiShardIT

darshan-sj:


manitgupta (approved):

aksharauke (commented):
We do not need to increase the timeout - the Dataflow job completed it's work on time. This seems to be some resource contention issue in the runners. We can close this PR.

PR #1736, opened 2024-07-16
Add extra logging for the failure case of the HTTP request.

fozzie15:
Add extra logging to catch issue where bad responses return HTTP instead of valid JSON.


v2/elasticsearch-common/src/main/java/com/google/cloud/teleport/v2/elasticsearch/utils/ElasticsearchIO.java
```diff
@@ -1658,7 +1659,14 @@ static int getBackendVersion(ConnectionConfiguration connectionConfiguration) {
...
       Request request = new Request("GET", "");
       Response response = restClient.performRequest(request);
-      JsonNode jsonNode = parseResponse(response.getEntity());
+      JsonNode jsonNode;
+      try {
+        jsonNode = parseResponse(response.getEntity());
+      } catch (IOException jsonException) {
+        throw new IOException(
+            "The response is not a valid JSON: "
+                + IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8));
```
an2x, line 1668: Can you please add a unit test for this? Make sure an exception is thrown with the HTTP response body in the message?

an2x (changes requested):

PR #1677, opened 2024-06-20
Explicitly enable AutoValue annotation processor

Abacn:
Starting in JDK 23, annotation processors are no longer implicitly enabled from dependencies, this change adds configuration to make AutoValue run on the latest JDK versions.

https://bugs.openjdk.org/browse/JDK-8321314
https://mail.openjdk.org/pipermail/jdk-dev/2024-May/009028.html

PR #1744, opened 2024-07-23
Fix annotation processor

Polber:
#1677 changed the way that `maven-compiler-plugin` compiles annotation processors from being implicit to explicitly defined. The definition for `@AutoValue` annotations was added, but not `@AutoService`.

This caused the `CommonTemplateJvmInitializer` to not get compiled into the template jars with the `@AutoService` handler, which resulted in Beam not calling the necessary Jvm initialization methods for certain templates to work properly.
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0d01accc1cfa29b762a764bc226496bfde25165c/v1/src/main/java/com/google/cloud/teleport/templates/common/CommonTemplateJvmInitializer.java#L40-L44

This was namely an issue with JDBC templates for users that rely on the `--extraFilesToStage` parameter to copy files from GCS to the `extra_files` directory on the Dataflow worker.

This PR adds the definition for the `@AutoService` annotation processor to affected modules.

### DIFF
--- a/it/pom.xml
+++ b/it/pom.xml
@@ -130,6 +130,11 @@
               <artifactId>auto-value</artifactId>
               <version>${autovalue.version}</version>
             </path>
+            <path>
+              <groupId>com.google.auto.service</groupId>
+              <artifactId>auto-service</artifactId>
+              <version>${autovalue.service.version}</version>
+            </path>
           </annotationProcessorPaths>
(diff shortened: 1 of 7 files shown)


Polber:
Test failure unrelated

Abacn (approved):

PR #1754, opened 2024-07-29
feat: Adds SourceDBToSpanner PostgreSQL Adapter 

thiagotnunes:
Adds implementation for reading source PostgreSQL database.

We purposefully left out parts of the PostgreSQL adapter support, because this pull request was becoming too large. This is what still needs to be done after this PR is merged:

- Named schema support
- String primary key support (collations, charsets, charset max length, collations query)
- Add type mapping support for types other than bigint, date and varchar
- Proper character mapping (with / without length) according to [pg-mapping](https://cloud.google.com/datastream/docs/unified-types#map-psql)
- Fine tune default parameters for PostgreSQL connections


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/transformer/SourceRowToMutationDoFn.java
```diff
@@ -106,6 +106,7 @@ public void processElement(ProcessContext c, MultiOutputReceiver output) {
           .get(SourceDbToSpannerConstants.ROW_TRANSFORMATION_SUCCESS)
           .output(RowContext.builder().setRow(sourceRow).setMutation(mutation).build());
     } catch (Exception e) {
+      LOG.error("Error while processing element", e);
```
thiagotnunes, line 109: Do we want to keep this log statement?

VardhanThigle: This is in exception path. 
Have you seen a lot of these in any of your test runs?

Unlike live template, we get to order foreign key constraints in Bulk, so in the most ideal (and correctly scaled - Dataflow, spanner etc) migration there should not be a ton of exceptions.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/OptionsToConfigBuilder.java
```diff
@@ -204,5 +201,27 @@ protected static String addParamToJdbcUrl(String jdbcUrl, String paramName, Stri
...
+  }
+
+  private static JdbcIOWrapperConfig.Builder builderWithDefaultsFor(SQLDialect dialect) {
+    switch (dialect) {
+      case MYSQL:
+        return builderWithMySqlDefaults();
+      case POSTGRESQL:
+        return builderWithPostgreSQLDefaults();
+    }
+    return null;
```
VardhanThigle, line 223: [nit] From a POV of code coverage and static null checkers, would it make sense to add handle MySQL as default case and avoid the un-reachable null return?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -20,9 +20,21 @@
 
 /** Interface used by the SourcedbToSpanner pipeline to accept user input. */
 public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
+  @TemplateParameter.Enum(
```
VardhanThigle, line 23: [nit] Would adding this parameter at the end avoid a re-numbering other paramters?

I don't think the order number matters a lot when a user executes a curl or gcloud command to start the template, the parameters are always key-value pairs.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/defaults/PostgreSQLConfigDefaults.java
```diff
@@ -0,0 +1,51 @@
...
+      new PostgreSQLJdbcValueMappings();
+
+  public static final Long DEFAULT_POSTGRESQL_MAX_CONNECTIONS = 160L;
+
+  public static final FluentBackoff DEFAULT_POSTGRESQL_SCHEMA_DISCOVERY_BACKOFF =
+      FluentBackoff.DEFAULT;
+
+  /** Default Initialization Sequence for the JDBC connection. */
+  public static final ImmutableList<String> DEFAULT_POSTGRESQL_INIT_SEQ =
+      ImmutableList.of("SET TIME ZONE 'UTC'");
```
VardhanThigle, line 48: [nit/todo]:
Some of the mySQL init sequences have been added for performance considerations.
If we have not scale tested PG for tall tables yet, could we add a TODO to fine tune these (just to not loose track)?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/JdbcSourceRowMapper.java
```diff
@@ -103,7 +102,8 @@ long getCurrentTimeMicros() {
                     entry.getKey(),
                     this.mappingsProvider
                         .getMappings()
-                        .getOrDefault(entry.getValue().getName(), JdbcValueMapper.UNSUPPORTED)
```
VardhanThigle, line 106: [nit] extra new line in the diff.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceTableSchema.java
```diff
@@ -65,8 +66,14 @@ public Schema getAvroPayload() {
    * primary keys, indexing information, foreign key constraints etc.
    */
 
-  public static Builder builder() {
-    return builder(MapperType.MYSQL);
+  public static Builder builder(SQLDialect dialect) {
+    switch (dialect) {
+      case MYSQL:
```
VardhanThigle, line 71: [nit] Same as earlier, making MYSQL as default case will avoid adding un-reachable exceptions.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/provider/PostgreSQLMappingProvider.java
```diff
@@ -0,0 +1,52 @@
...
+ * @see <a href="https://cloud.google.com/datastream/docs/unified-types#map-psql">Mappings of
+ *     unified types to source and destination data types</a>
+ */
+public final class PostgreSQLMappingProvider {
+  // TODO(thiagotnunes): Add missing type mappings
+  private static final ImmutableMap<String, UnifiedTypeMapping> MAPPING =
+      ImmutableMap.<String, UnifiedMappingProvider.Type>builder()
+          .put("BIGINT", UnifiedMappingProvider.Type.LONG)
+          .put("CHARACTER VARYING", UnifiedMappingProvider.Type.STRING)
+          .put("DATE", UnifiedMappingProvider.Type.TIMESTAMP)
```
VardhanThigle, line 37: We follow Datastream's  [unified mapping](https://cloud.google.com/datastream/docs/unified-types#map-psql) to maintain compatibility and code reusability with live template.
I think PG data map's to date logical type.

For MySQL this maps to TimeStamp but I am not sure why it's done that way :) (outside of historic reasons)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/provider/PostgreSQLMappingProvider.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @see <a href="https://cloud.google.com/datastream/docs/unified-types#map-psql">Mappings of
+ *     unified types to source and destination data types</a>
+ */
+public final class PostgreSQLMappingProvider {
+  // TODO(thiagotnunes): Add missing type mappings
+  private static final ImmutableMap<String, UnifiedTypeMapping> MAPPING =
+      ImmutableMap.<String, UnifiedMappingProvider.Type>builder()
+          .put("BIGINT", UnifiedMappingProvider.Type.LONG)
+          .put("CHARACTER VARYING", UnifiedMappingProvider.Type.STRING)
```
VardhanThigle, line 36: Please have a look at [pg-mapping](https://cloud.google.com/datastream/docs/unified-types#map-psql), we might need to check the length too.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -20,9 +20,21 @@
...
+      order = 1,
+      enumOptions = {
+        @TemplateParameter.TemplateEnumOption("MYSQL"),
+        @TemplateParameter.TemplateEnumOption("POSTGRESQL")
+      },
+      description = "SQL Dialect of the source database",
+      helpText = "Possible values are `MYSQL` and `POSTGRESQL`.")
+  @Default.String("MYSQL")
+  String getSourceDbDialect();
```
VardhanThigle, line 32: Do we also want to add a parameter for PG namespace?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -46,12 +58,12 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
...
   @TemplateParameter.Text(
-      order = 3,
-      regexes = {"(^jdbc:mysql://.*|^gs://.*)"},
+      order = 4,
+      regexes = {"(^jdbc:mysql://.*|^jdbc:postgresql://.*|^gs://.*)"},
       groupName = "Source",
       description =
           "URL to connect to the source database host. It can be either of "
-              + "1. The JDBC connection URL - which must contain the host, port and source db name and can optionally contain properties like autoReconnect, maxReconnects etc. Format: `jdbc:mysql://{host}:{port}/{dbName}?{parameters}`"
+              + "1. The JDBC connection URL - which must contain the host, port and source db name and can optionally contain properties like autoReconnect, maxReconnects etc. Format: `jdbc:{mysql|postgresql}://{host}:{port}/{dbName}?{parameters}`"
```
VardhanThigle, line 66: For PG we might need to have name-space too.
We have a field for the name in our SchemaReference. I am not sure, how it's getting populated with the current changes for PG.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -0,0 +1,495 @@
...
+
+  /**
+   * Discover the indexes of tables to migrate.
+   *
+   * @param dataSource Provider for JDBC connection.
+   * @param sourceSchemaReference Source database name and (optionally namespace)
+   * @param tables Tables to migrate.
+   * @return The discovered indexes.
+   * @throws SchemaDiscoveryException - Fatal exception during Schema Discovery.
+   * @throws RetriableSchemaDiscoveryException - Retriable exception during Schema Discovery.
```
VardhanThigle, line 266: [super-nit], please feel free to ignore. For ease of review / future-reading in case of some of the MYSQL discovery queries, especially around index discover (which involves a join), we have added links to db-fiddle .

In case it's not a lot of trouble or if you have them ready from your experiments, feel free to add them for PG too.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -0,0 +1,495 @@
...
+   *
+   * @param tableName name of the table to read.
+   * @param partitionColumns if not-empty, partition columns. Set empty for first column of
+   *     partitioning.
+   */
+  @Override
+  public String getBoundaryQuery(
+      String tableName, ImmutableList<String> partitionColumns, String colName) {
+    return addWhereClause(
+        String.format("SELECT MIN(%s), MAX(%s) FROM %s", colName, colName, tableName),
```
VardhanThigle, line 423: Would it be possible to UT the queries (with mocks if it's not possible to test with in-memory db)?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -0,0 +1,495 @@
...
+      case "D":
+        return SourceColumnIndexInfo.IndexType.DATE_TIME;
+      case "S":
+        return SourceColumnIndexInfo.IndexType.STRING;
+      default:
+        return SourceColumnIndexInfo.IndexType.OTHER;
+    }
+  }
+
+  private boolean shouldPadSpace(@Nullable String pad) throws SQLException {
```
VardhanThigle, line 492: In PG, the `pad_attribute` whic is passed to this function is always "NO PAD". [Ref](https://www.postgresql.org/docs/16/infoschema-collations.html)

But

Whether trailing spaces are semantically significant or not, depends on the type of the column
[Ref](https://www.postgresql.org/docs/current/datatype-character.html)

Could we also consider the column type also in this logic?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/OptionsToConfigBuilder.java
thiagotnunes: Updated

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
thiagotnunes: I would like this to be the first choice a user has to make in the UI.

Is it not related to the numbering here?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/config/defaults/PostgreSQLConfigDefaults.java
thiagotnunes: There is already TODO a the top of the file

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/JdbcSourceRowMapper.java
thiagotnunes: This was auto-updated by `spotless`, since I added `.toUpperCase` and it no longer fits in a single line

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceTableSchema.java
thiagotnunes: Updated

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/provider/PostgreSQLMappingProvider.java
thiagotnunes: Updated

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/provider/PostgreSQLMappingProvider.java
thiagotnunes: Adding a TODO for now

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
thiagotnunes: I think you mean schema in PG?

I would prefer if we didn't deal with those for now, because the PR is quite big already, but I will add a TODO

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
thiagotnunes: That is very interesting, added and discovered that one query did not work with PG < 11. Added todo for the latter.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
thiagotnunes: Done

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
thiagotnunes: Could you point me to where we do this with MySQL, so I can base my test on that?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
VardhanThigle: That's right. Given the huge list of parameters we need to run a migration, we probably would have users with curl, gcloud or terraform.
But it's ok to reorder too.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
VardhanThigle: For specific boundary query, we test the query string
[here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapterTest.java)

For larger parts like index discovery, we test via the mocks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -0,0 +1,518 @@
...
+
+            // TODO(thiagotnunes): Collation support and string max length
+            String collation = resultSet.getString("collation");
+            if (collation != null) {
+              Integer typeLength = resultSet.getInt("type_length");
+              if (resultSet.wasNull()) {
+                typeLength = null;
+              }
+              boolean shouldPadSpace =
+                  isBlankPaddedType(resultSet.getString("type_name"), typeLength)
```
VardhanThigle, line 335: I think we care only about pad-space. And my perspective could be clouded from MYSQL, so please let me know your thoughts from PG side.

In case of MYSQL blanks are ignored at all positions, so our collation detection logic creates detects blanks separately and ignores them all together.

Space (and it's variants in unicode) though is a different case. If comparison in PAD-SPACE a space at the end is ignored from comparisons, but is very much relevent at positions other than end. For example `"a\tb" < "a b"` but `"ab\t" > "ab "`
`"a\zb" = "ab" and "ab\z" = "ab"` here "\z" stands for `ctrl-z`, which in MySql world is a blank character.

What happens in PG when a blank character is compared in the middle of the string?

thiagotnunes: A few cases in PostgreSQL, below. I think the logic is almost right, but `bpchar`, which perhaps should always be marked as `pad-space`?

```sql
-----------------------
-- Not blank padded --
-----------------------
SELECT 'a\tb' < 'a b', 'ab\t' < 'ab ';
SELECT 'a\tb'::character varying(100) < 'a b'::character varying(100), 'ab\t'::character varying(100) < 'ab '::character varying(100);
SELECT 'a\tb'::varchar(100) < 'a b'::varchar(100), 'ab\t'::varchar(100) < 'ab '::varchar(100);
SELECT 'a\tb'::text < 'a b'::text, 'ab\t'::text < 'ab '::text;
-- TRUE, TRUE

-------------------
-- Blank padded --
-------------------
SELECT 'a\tb'::character(100) < 'a b'::character(100), 'ab\t'::character(100) < 'ab '::character(100);
SELECT 'a\tb'::char(100) < 'a b'::char(100), 'ab\t'::char(100) < 'ab '::char(100);
SELECT 'a\tb'::bpchar(100) < 'a b'::bpchar(100), 'ab\t'::bpchar(100) < 'ab '::bpchar(100);
-- TRUE, FALSE (left is > than right)

---------------------------
-- Space ONLY trimmed --
---------------------------
SELECT 'ab\t'::bpchar = 'ab '::bpchar, 'ab         '::bpchar = 'ab '::bpchar;
-- FALSE, TRUE (left is equal to right)
```

VardhanThigle: Tab is not a version of space in MySQL too (characters like non-breaking space, or [quad-space](https://www.compart.com/en/unicode/U+2001) are variants of space, but that will be more relevant when you implement the actual detection query)


I thing even char ignores trailing spaces [like](https://www.db-fiddle.com/f/idsYJsV76FWX3JQa7fKysP/0)

I am not sure if PG has any character that's a MySql style blank as in a character that's insignificant in comparisons at all positions (like \z, \0 etc)

thiagotnunes: Sorry, I should had mentioned that the `\t` above is just representative in the query. You need to escape it like so:

```sql
SELECT E'a\t'::character(2)
```

There are only [these](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE) especial characters in PostgreSQL.

I tested all of them in db-fiddle [here](https://www.db-fiddle.com/f/4ko4mwKS6Es5tzvrxWhewd/0), alongside two unicodes you mentioned (quad-space and CTRL-Z), but they are not padded. The only character that is padded is space. 

AFAIU we only need to pad-space (trim space) for CHAR(n), CHARACTER(n) and BPCHAR(n). All others behave as expected.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
```diff
@@ -0,0 +1,71 @@
...
+                  Map.entry(
+                      entry.getKey(),
+                      new JdbcValueMapper<>(
+                          entry.getValue().getLeft(), entry.getValue().getRight())))
+          .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+
+  /** Get static mapping of SourceColumnType to {@link JdbcValueMapper}. */
+  @Override
+  public ImmutableMap<String, JdbcValueMapper<?>> getMappings() {
+    return SCHEMA_MAPPINGS;
```
VardhanThigle, line 64: In case we are taking up UT in a subsequent PR, please add TODOs whereever coverage is missing or the providers and dialect adapter classes.

thiagotnunes: Added TODOs

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
thiagotnunes: Added the query string test for the missing boundary query.

IT test should cover the query execution, but let me know if you want specific UTs.

VardhanThigle (approved):

PR #1769, opened 2024-08-02
Support populating Spanner source database id, instance id and a custom metadata field in Spanner CDC to Pub Sub template

ShuranZhang:
Introduce two new options for Spanner Change Streams to Pub/Sub template.

1. `includeSpannerSource`: boolean option default to false.
If this config is set to true, two new fields `spannerDatabaseId` and `spannerInstanceId` will be populated to the message data field with the associate values.

2. `outputMessageData`: string option default to empty.
If this config has non-empty string value, the string value will be populated to the `outputMessageMetadata` field in output pub sub message body. 

I verified correctness for both AVRO and JSON data formats through integration tests in `SpannerChangeStreamsToPubSubIT.java`.

### DIFF
--- a/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
+++ b/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
@@ -218,10 +218,9 @@ public interface SpannerChangeStreamsToPubSubOptions extends DataflowPipelineOpt
   @TemplateParameter.Boolean(
       order = 17,
       optional = true,
-      description =
-          "Whether or not to include the spanner database and instance to read the change stream from in the output message data",
+      description = "Include spanner database id and instance id in output message",
       helpText =
-          "Whether or not to include the spanner database and instance to read the change stream from in the output message data. Defaults to: false")
+          "Whether or not to include the spanner database id and instance id to read the change stream from in the output message data. Defaults to: false")
   @Default.Boolean(false)
   Boolean getIncludeSpannerSource();
 
--- a/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
+++ b/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
@@ -225,4 +225,18 @@ public interface SpannerChangeStreamsToPubSubOptions extends DataflowPipelineOpt
   Boolean getIncludeSpannerSource();
 
   void setIncludeSpannerSource(Boolean value);
+
+  @TemplateParameter.Text(
+      order = 18,
+      optional = true,
+      description = "Custom Value for the optional field outputMessageMetadata",
+      helpText =
+          "The string value for the custom field outputMessageMetadata in output pub/sub message. "
+              + "Defaults to empty and the field outputMessageMetadata is only populated if this "
+              + "value is non-empty. Please escape any special characters when entering the value "
+              + "here(ie: double quotes).")
+  @Default.String("")
+  String getOutputMessageMetadata();
+
+  void setOutputMessageMetadata(String value);
 }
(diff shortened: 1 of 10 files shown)


v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
```diff
@@ -214,4 +214,16 @@ public interface SpannerChangeStreamsToPubSubOptions extends DataflowPipelineOpt
   RpcPriority getRpcPriority();
 
   void setRpcPriority(RpcPriority rpcPriority);
+
+  @TemplateParameter.Boolean(
+      order = 17,
+      optional = true,
+      description =
+          "Whether or not to include the spanner database and instance to read the change stream from in the output message data",
```
tianz101, line 222: I assume the database name and intance name are from user specified names and not internal names. Can you confirm? Thanks.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/WriteDataChangeRecordsToAvro.java
```diff
@@ -46,91 +46,149 @@ public class WriteDataChangeRecordsToAvro {
...
+      }
+    }
   }
 
   private static long timestampToMicros(Timestamp ts) {
     return TimeUnit.SECONDS.toMicros(ts.getSeconds())
         + TimeUnit.NANOSECONDS.toMicros(ts.getNanos());
   }
 
-  public static com.google.cloud.teleport.v2.DataChangeRecord dataChangeRecordToAvro(
```
tianz101, line 60: It used to be public static, can you comment on why we can not keep public static any more?

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/WriteDataChangeRecordsToAvro.java
```diff
@@ -46,91 +46,149 @@ public class WriteDataChangeRecordsToAvro {
...
+    public String spannerInstanceId() {
+      return spannerInstanceId;
+    }
+
     @Override
     public com.google.cloud.teleport.v2.DataChangeRecord apply(DataChangeRecord record) {
       return dataChangeRecordToAvro(record);
     }
+
+    public DataChangeRecordToAvroFn() {}
```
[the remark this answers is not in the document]
ShuranZhang: This is needed because this file is shared with our GCS template still uses the default empty constructor.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/WriteDataChangeRecordsToAvro.java
ShuranZhang: Because we need to call class functions `spannerDatabaseId()` and `spannerInstanceId()` in this function so it is no longer stateless.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
ShuranZhang: Thanks for pointing out, I agree. I renamed every thing from `spannerDatabase` to `spannerDatabaseId`

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/FileFormatFactorySpannerChangeStreamsToPubSub.java
```diff
@@ -61,6 +62,18 @@ public static WriteToPubSubBuilder newBuilder() {
 
   protected abstract String pubsubTopicName();
 
+  @Nullable
+  protected abstract Boolean includeSpannerSource();
+
```
tianz101, line 67: Maybe 'includeSpannerSourceInfo' is a more accurate name here.

tianz101 (approved):

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/FileFormatFactorySpannerChangeStreamsToPubSub.java
ShuranZhang: To me it appears cleaner to keep the function name consistent with the new conig name `includeSpannerSource`. This also aligns with the builder pattern to have same function name with the class member variable.

Abacn (approved):
approval based on passed tests

PR #1772, opened 2024-08-06
Script to generate shard-config.json

VardhanThigle:


v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
```diff
@@ -0,0 +1,94 @@
...
+    "dataShards": []
+  }
+}'
+
+# Populate dataShards array
+data_shards=()
+for host_ip in "${host_ips[@]}"; do
+    data_shard_id=$(echo "$host_ip" | tr '.' '-')
+    databases=()
+    for database_name in "${database_names[@]}"; do
```
bharadwaj-aditya, line 53: I think this assumes the same database names in all shards. That may not be the case

bharadwaj-aditya (commented):
Please add some comments around the generation of the database names. Rest looks fine.

v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
VardhanThigle: That's right. I will add comments around the same.
It's done to keep it simple as in the most generic form, it becomes the json itself.

VardhanThigle: Fixed in the current push.

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
```diff
@@ -0,0 +1,101 @@
...
+# Replace and add the names of all the logical databases you would like to migrate.
+# Note that all logical databases must have the same schema.
+database_names=("logicalDb01" "logicalDb02")
+# Replace with the actual DB credentials.
+username="<user-name>"
+password="<password>"
+
+# The script below helps you create the shard config json.
+# Main JSON Structure.
+json='{
```
Deep1998, line 35: Maybe mention in the comment this does not need to be filled

v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
```diff
@@ -0,0 +1,101 @@
...
+      "user": "",
+      "password": "",
+      "port": "",
+      "dbName": ""
+    },
+    "dataShards": []
+  }
+}'
+
+# Code below populates the dataShards array.
```
Deep1998, line 50: Here also, it helps a new reader if we explicitly state what needs to be filled and what need not be.

v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
VardhanThigle: we have a section titled Variables above. 
This is another section script.

Does that help?

v2/sourcedb-to-spanner/src/main/scripts/create_simple_shard_config.bash
VardhanThigle: Same as above. In the steps we have told the user to update the variables and we have a section for the same.

PR #1773, opened 2024-08-06
JdbcToBigQuery: Support loading the source query from Cloud Storage.

an2x:
This was already implemented in the legacy JdbcToBigQuery templates, but not in the Flex (v2) template.

This can be used in 2 use cases: providing very large queries (> 100 KB -- this causes the job graph to be too large if provided as a constant parameter) and providing queries with unicode characters.

### DIFF
--- a/v2/jdbc-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/JdbcToBigQueryIT.java
+++ b/v2/jdbc-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/JdbcToBigQueryIT.java
@@ -192,6 +192,34 @@ public class JdbcToBigQueryIT extends JDBCBaseIT {
   }
 
   @Test
+  public void testPostgresWithUnicodeCharactersInQuery() throws IOException {
+    String tableName = "unicóde_table";
+
+    postgresResourceManager = PostgresResourceManager.builder(testName).build();
+    gcsClient.createArtifact(
+        "input/query.sql",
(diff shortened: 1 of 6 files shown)


v2/jdbc-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/JdbcToBigQueryIT.java
```diff
@@ -191,6 +191,34 @@ public void testPostgresToBigQueryFlex() throws IOException {
                     + testName));
   }
 
+  @Test
+  public void testPostgresWithUnicodeCharactersInQuery() throws IOException {
+    String tableName = "unicóde_table";
```
Polber, line 196: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/bigquery/BigQueryResourceManagerUtils.java#L95-L100

BigQuery ILLEGAL_TABLE_CHARS regex needs to be modified to either allow unicode characters according to https://cloud.google.com/bigquery/docs/tables or disable this part of the checkValidTableId entirely and allow the test to fail when it sends create table request to BQ

an2x: Fixed. Let's see if the IT succeeds now.

Polber (approved):

PR #1774, opened 2024-08-06
Upgrade Beam version to 2.58.0

jrmccluskey:
Bumps the Beam version to the newly released 2.58.0.


jrmccluskey:
Yi picked this up, closing

PR #1775, opened 2024-08-06
Migrate from the deprecated Charsets constants (in Guava) to the StandardCharsets constants (in the JDK).

damccorm:
Same change as https://github.com/apache/beam/pull/32083

guava Charset is now a delegates to java.nio Charsets, e.g.

https://github.com/google/guava/blob/8bab177c14b6c3faf82d6959a9c774b0d15e7414/guava/src/com/google/common/base/Charsets.java#L67

so we can move off the deprecated charsets


Abacn (approved):
Thank you!

PR #1779, opened 2024-08-10
Support PLACEMENT schema objects in Spanner's import/export templates

sgorse123:
The [Placement](https://cloud.google.com/spanner/docs/create-manage-data-placements#create-placement) schema object was introduced as part geo-partitioning.

This change uses [placements](https://cloud.google.com/spanner/docs/information-schema#placements) and [placement_options](https://cloud.google.com/spanner/docs/information-schema#placement-options) information schema views, which hold information about the Placements in the database.

Similar to Change Stream and Sequence schema objects, each Placement in the database is exported as a separate Avro and JSON manifest file.

This change was also manually tested by using the [export](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/README_Cloud_Spanner_to_GCS_Avro.md#building-template) and [import](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/README_GCS_Avro_to_Cloud_Spanner.md#building-template) instructions.


manitgupta (commented):
LGTM, although I do see individual ITs and UTs have been added for both export and import, can you confirm the following - 

*Have you tried exporting a non-empty spanner database with placement options to Avro and then importing the Avro files back into an empty spanner database and verifying that the old and new databases are identical?*

manitgupta (approved):
Test failure due to unrelated reason.

PR #1784, opened 2024-08-12
KafkaToBigQueryFlex Template: Expose source message offset in DLQ metadata (for Json format only)

fozzie15:
Modify the Kafka to BigQuery Flex pipeline such that topic, partition, and source offset can be written to the DLQ. Currently, this is only supported for failures converting the records into table rows.

### DIFF
--- a/v2/kafka-to-bigquery/src/test/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlexJsonIT.java
+++ b/v2/kafka-to-bigquery/src/test/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlexJsonIT.java
@@ -90,6 +90,7 @@ public final class KafkaToBigQueryFlexJsonIT extends TemplateTestBase {
   }
 
   @Test
+  @TemplateIntegrationTest(value = KafkaToBigQueryFlex.class, template = "Kafka_to_BigQuery_Flex")
   public void testKafkaToBigQueryWithExistingDLQ() throws IOException {
     TableId deadletterTableId =
         bigQueryClient.createTable(testName + "_dlq", getDeadletterSchema());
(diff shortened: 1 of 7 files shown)


v2/kafka-to-bigquery/src/test/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlexJsonIT.java
```diff
@@ -90,6 +90,7 @@ public void testKafkaToBigQuery() throws IOException {
   }
 
   @Test
+  @TemplateIntegrationTest(value = KafkaToBigQueryFlex.class, template = "Kafka_to_BigQuery_Flex")
```
an2x, line 93: It looks like this test only verifies that good records were written to the main BQ table but doesn't verify that the bad records were written to the DLQ table.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/transforms/KafkaRecordErrorConverters.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.PCollection;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KafkaRecordErrorConverters {
```
an2x, line 37: As discussed offline, please refactor this into a generic ErrorConverter in the common module + a "payload extractor" specific to KafkaRecords in kafka-common.

an2x (changes requested):

v2/kafka-to-bigquery/src/test/java/com/google/cloud/teleport/v2/templates/KafkaToBigQueryFlexJsonIT.java
fozzie15: Yes. That is the intention of that test. For the DLQ case, we have a separate test called "testKafkaToBigQueryWithExistingDLQ"

fozzie15: Added extra check for the DLQ table

an2x (approved):

PR #1789, opened 2024-08-14
Support Timestamp in proto to bigquery schema conversion

Abacn:
For example, in PubSub_Proto_to_BigQuery_Xlang, when there is a Timestamp field, it get inferred to a BigQuery RECORD field of {"seconds": "INTEGER", "nanos": "INTEGER"}, i.e., google.protobuf.Timestamp gets expanded.

### DIFF
--- a/v2/common/pom.xml
+++ b/v2/common/pom.xml
@@ -119,6 +119,10 @@
           <groupId>com.google.cloud</groupId>
           <artifactId>google-cloud-kms</artifactId>
         </dependency>
+        <dependency>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </dependency>
 
         <!-- UDFs -->
         <dependency>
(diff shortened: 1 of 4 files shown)


Abacn:
Internal bug: 355704665

Abacn:
R: @Polber

v2/common/pom.xml
```diff
@@ -119,6 +119,10 @@
           <groupId>com.google.cloud</groupId>
           <artifactId>google-cloud-kms</artifactId>
         </dependency>
+        <dependency>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </dependency>
```
Polber, line 125: Is this necessary? No new imports were added to the corresponding files

Abacn: I was following https://www.xolstice.org/protobuf-maven-plugin/usage.html explicitly asked to add these lines.

It wasn't introducing issue because protobuf-java has always been a transitive dependency of other declared dependencies, even though v2/commons also directly used protobuf without declare it.

Hoever in general one should declare a dependency if it is directly used in that module

Abacn:
Four attempts all have different tests failing, not related to the change

Polber (approved):

PR #1798, opened 2024-08-16
Add Datastream Owners

dhercher:
adding datastream as owners of the datastream templates


damccorm (approved):
Thanks!

Abacn:
this is blocking merging #1835. Should we at least add spanner-migration-team as owner of v2/datastream-common ? Also It seems once the owner added then Template team approval is not sufficient also.

PR #1809, opened 2024-08-23
Update codecov to block PRs lower than 80% coverage

Polber:
This file was moved from `.github` to root directory per instructions on codecov docs.

Future PRs will also be blocked if they are not submitted with at least 80% code coverage.


codecov.yml
```diff
@@ -9,7 +9,7 @@ coverage:
     patch:
       default:
         target: 80%
-        informational: true
+        informational: false
```
damccorm, line 12: I don't think we should do this - blocking patches on code coverage can be overly restrictive, especially for smaller changes, and can put a big burden on authors in places where good testing doesn't already exist.

We'd also need buy in from partner teams since this would impact code that our core team doesn't own.

codecov.yml
```diff
@@ -9,7 +9,7 @@ coverage:
     patch:
```
damccorm, line 9: Why are we moving the file out of `.github`?

codecov.yml
damccorm: Note that right now our repo-level coverage is ~40% and its unclear to me if that includes integration tests

codecov.yml
Polber: I think it was using the global YAML before, so we were getting checks, but it wasn't using this YAML's config

damccorm: It can go in .github too - https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml

Polber: Ah ok nice, I'll move it back

.github/codecov.yml
```diff
@@ -9,7 +9,7 @@ coverage:
     patch:
       default:
         target: 80%
-        informational: true
+        informational: false
```
damccorm, line 12: Probably we need to do this for spanner as well

damccorm:
Also, for any readers of this PR who are impacted - this 80% line is required in order for us to be compliant with some internal horizontals

.github/codecov.yml
Polber: Added

damccorm (approved):
Thanks!

PR #1815, opened 2024-08-26
Remove gcs notification tests from Datastream Spanner ITs

Polber:
Removing these tests as they are redundant as of #1674


manitgupta:
The polling approach has been deprecated in #1674 ; the GCS notification approach is still being actively used. I don't think we should remove this?

Polber:
> #1674

@manitgupta #1674 added a line to make all ITs use GCS notifications

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java#L384

So either that change should be reverted, or this PR should remove the redundant tests

PR #1816, opened 2024-08-27
[KafkaToBigQuery & KafkaToGcs] - Add OAuth Authentication Support For Schema Registries 

fozzie15:
Add support for OAuth authentication to schema registry in the Kafka to Big Query Flex and Kafka to GCS flex templates.

### DIFF
--- a/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
+++ b/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
@@ -114,10 +114,11 @@ public interface SchemaRegistryOptions extends PipelineOptions {
       enumOptions = {
         @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.NONE),
         @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.TLS),
+        @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.OAUTH),
       },
       optional = true,
       description = "Authentication Mode",
-      helpText = "Schema Registry authentication mode. Can be NONE or TLS.")
+      helpText = "Schema Registry authentication mode. Can be NONE, TLS or OAUTH.")
   @Default.String(KafkaAuthenticationMethod.NONE)
   String getSchemaRegistryAuthenticationMode();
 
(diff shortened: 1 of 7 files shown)


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
```diff
@@ -188,4 +189,52 @@ public interface SchemaRegistryOptions extends PipelineOptions {
...
+
+  @TemplateParameter.Text(
+      order = 11,
+      parentName = "schemaRegistryAuthenticationMode",
+      parentTriggerValues = {KafkaAuthenticationMethod.OAUTH},
+      optional = true,
+      description = "Client ID",
+      helpText =
+          "Client ID used to obtain client credentials grant from OAuth/OIDC identity provider.")
+  String getSchemaRegistryClientId();
```
an2x, line 201: The parameter name doesn't make it clear that this is the ClientID to be used with OAuth, it sounds like "the id of the schema registry client", which is confusing.
I'd add "OAuth" prefix to all of the new params, e.g. `schemaRegistryOAuthClientId`, `schemaRegistryOAuthTokenEndpointUrl`, etc.

an2x: Note: it's not a problem in the UI, but could be a problem when running the template via CLI and/or reading the list of parameters in the docs.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
```diff
@@ -188,4 +189,52 @@ public interface SchemaRegistryOptions extends PipelineOptions {
...
+  void setSchemaRegistryScope(String schemaRegistryScope);
+
+  @TemplateParameter.Text(
+      order = 11,
+      parentName = "schemaRegistryAuthenticationMode",
+      parentTriggerValues = {KafkaAuthenticationMethod.OAUTH},
+      optional = true,
+      description = "Client Secret ID",
+      helpText =
+          "The secret ID used in secret manager to obtain client credentials grant from OAuth/OIDC identity provider.")
```
an2x, line 236: Make this similar to the `KafkaReadOptions#getKafkaReadKeystorePasswordSecretId()` param?
1) Change description to: "The Google Cloud Secret Manager secret ID that contains the Client Secret to use to authenticate the Schema Registry client. Required if messageFormat = AVRO_CONFLUENT_WIRE_FORMAT and schemaRegistryAuthenticationMode = OAUTH".
2) Add an example (`projects/<PROJECT_ID>/secrets/<SECRET_ID>/versions/<SECRET_VERSION>`).
3) Does this have to point to a specific version or just `projects/<PROJECT_ID>/secrets/<SECRET_ID>` will work? What about `projects/<PROJECT_ID>/secrets/<SECRET_ID>/versions/latest`?

Please review other params and make names/descriptions/examples consistent between themselves and other, already existing params.

an2x (changes requested):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
fozzie15: Done.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
fozzie15: Made some changes. Leaving this comment open until another review.

an2x (approved):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/SchemaRegistryOptions.java
an2x: LGTM now, thanks!

PR #1819, opened 2024-08-28
Terraform template to launch a sharded bulk migration 

Deep1998:
This template is used to orchestrate bulk migrations jobs for a sharded setup. It accepts a shardedConfig of all the shards, and batches multiple physical shards into a job based on a batch_size parameter.


manitgupta:
Lets update the Terraform validator for the bulk directory and fix any issues it highlights - looks like some mandatory sections are missing.

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/spanner-terraform-validator.yml#L73

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
```diff
@@ -1,29 +1,129 @@
...
+provides instructions to add these roles to an existing service account.
+
+## Dataflow permissions
+
+The Dataflow service account needs to be provided the `roles/storage.objectAdmin` and `roles/spanner.databaseAdmin`
+roles. Also ensure that the workers can access the sources.
+
+## Assumptions
+
+It takes the following assumptions -
```
bharadwaj-aditya, line 78: s/takes/makes

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
```diff
@@ -1,29 +1,129 @@
-## Scenario
+## Sample Scenario: MySQL to Spanner production bulk migrations
```
bharadwaj-aditya, line 1: Sharded MySQL ?

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
```diff
@@ -1,29 +1,129 @@
...
+> **_NOTE:_**
+[SMT](https://googlecloudplatform.github.io/spanner-migration-tool/quickstart.html)
+> can be used for converting a MySQL schema to a Spanner compatible schema.
+
+1. MySQL sources can establish network connectivity with Dataflow.
 2. Appropriate permissions are added to the service account running Terraform to allow resource creation.
 3. Appropriate permissions are provided to the service account running Dataflow to write to Spanner.
-4. A GCS bucket has been provided to write the DLQ records to.
+4. A GCS directory has been provided to write the output records (DLQ/filtered events) to.
+5. A Spanner instance with database containing the data-migration compatible schema is created. If it is not
```
bharadwaj-aditya, line 88: 6. The Source Schema should be the same across all sources instances.

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
```diff
@@ -1,29 +1,129 @@
...
+   like-to-like i.e. there are modifications in the Spanner schema, provide an SMT generated session file is mandatory.
+   Check out the FAQ on [how to generate a session file](#specifying-schema-overrides).
 
-Given these assumptions, it copies data from multiple source MySQL databases to the configured Spanner database(s).
+Given these assumptions, it copies data from multiple source MySQL databases to the configured Spanner database.
```
bharadwaj-aditya, line 92: Please use the sharded terminology here to be consistent .

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
```diff
@@ -61,12 +161,260 @@ dataflow_job_urls = [
 ]
 ```
 
-**Note:** Each of the jobs will have a random suffix added to it to prevent name collisions. 
+**Note 2:** If you want to run more than 10 dataflow jobs in parallel, run terraform using the `-parallelism=n` flag.
```
bharadwaj-aditya, line 164: nit: Mention the default parallelism is 10, and then provide the example of how to change it.

Add a note that parallelism linearly increases load on Spanner. If Spanner is not sized sufficiently,all the jobs will slow down.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,431 @@
+## Sample Scenario: Sharded MySQL to Spanner production bulk migration
```
bharadwaj-aditya, line 1: How is this different from multiple sources ?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/terraform.tfvars
```diff
@@ -0,0 +1,33 @@
...
+  jdbcDriverJars           = "gs://path/to/driver/jars"
+  jdbcDriverClassName      = "com.mysql.jdbc.driver"
+  num_partitions           = 4000
+  max_connections          = 320
+  instanceId               = "my-spanner-instance"
+  databaseId               = "my-spanner-database"
+  projectId                = "my-spanner-project"
+  spannerHost              = "https://batch-spanner.googleapis.com"
+  local_session_file_path  = "/local/path/to/smt/session/file"
+  local_sharding_config    = "/Users/deepchowdhury/Downloads/128shardconfig.json"
```
bharadwaj-aditya, line 15: Please remove the user id's.

replace with ~/Downloads

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/terraform_simple.tfvars
```diff
@@ -0,0 +1,18 @@
...
+common_params = {
+  project                  = "project-name"
+  region                   = "us-central1"               # Or your desired region
+  working_directory_bucket = "bucket-name"               # example "test-bucket"
+  working_directory_prefix = "path/to/working/directory" # should not start or end with a '/'
+  instanceId               = "my-spanner-instance"
+  databaseId               = "my-spanner-database"
+  projectId                = "my-spanner-project"
+  local_session_file_path  = "/local/path/to/smt/session/file"
+  local_sharding_config    = "/Users/deepchowdhury/Downloads/128shardconfig.json"
```
bharadwaj-aditya, line 13: replace user details with ~

v2/sourcedb-to-spanner/terraform/samples/multiple-jobs/README.md
Deep1998: This multiple-jobs readme is not for sharded but launching n jobs for any n mysql instances, sharded or not.

The sharded readme is separate. Might make sense to scrap the multiple jobs readme altogether?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: multiple sources runs any random n sources, like mentioned above might make sense to remove it since its confusing and not a strong use case for terraform.

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya (dismissed):
LGTM

manitgupta:
You should still provide a sample for the simplest use-case of just launching a single database/instance bulk migration.
Deleting the `multiple-jobs` one makes sense, but please add a simpler, newer one which - 

1. Configures any required permissions
2. Runs the Dataflow job.

Deep1998:
> You should still provide a sample for the simplest use-case of just launching a single database/instance bulk migration. Deleting the `multiple-jobs` one makes sense, but please add a simpler, newer one which -
> 
> 1. Configures any required permissions
> 2. Runs the Dataflow job.

I was thinking of doing this but launching a single job via terraform seems like an overkill as launching bulk is pretty straightforward. Customers would likely use this only for sharded use cases.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,435 @@
...
+- compute.globalOperations.get
+- compute.networks.addPeering
+- compute.networks.get
+- compute.networks.listPeeringRoutes
+- compute.networks.removePeering
+- compute.networks.use
+- compute.routes.get
+- compute.routes.list
+- compute.subnetworks.get
+- compute.subnetworks.list
```
manitgupta, line 37: Why are these needed?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,435 @@
...
+It makes the following assumptions -
+
+> **_NOTE:_**
+[SMT](https://googlecloudplatform.github.io/spanner-migration-tool/quickstart.html)
+> can be used for converting a MySQL schema to a Spanner compatible schema.
+
+1. Dataflow can establish network connectivity with the MySQL source shards.
+2. Appropriate permissions are added to the service account running Terraform to allow resource creation.
+3. Appropriate permissions are provided to the service account running Dataflow to write to Spanner.
+4. A GCS directory has been provided to write the output records (DLQ/filtered events) to.
```
manitgupta, line 86: Why is this an assumption, is the DLQ not mandatory?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,435 @@
...
+5. A Spanner instance with database containing the data-migration compatible schema is created. This includes a
+   migration_shard_id column being present in the Spanner schema.
```
manitgupta, line 88: Is this mandatory? What if the user does not want this column at all?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,435 @@
...
+6. An SMT generated session file is provided containing the schema mapping information is created locally. This is
+   mandatory for sharded migrations. Check out the FAQ
+   on [how to generate a session file](#specifying-schema-overrides).
+7. A sharding config containing source connection information of all the physical and logical shards is created locally.
```
manitgupta, line 92: Why does this need to be defined in the JSON? This should be natively defined inside Terraform, using Terraform constructs. Something like the `shard_list` defined in the live templates. 

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/terraform/samples/mysql-sharded-end-to-end/variables.tf#L76

The Terraform configuration should be internally converted into a JSON file (or whatever representation) that the Dataflow job understands via functions like `jsonencode`.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/main.tf
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  ]
+
+  working_directory_gcs = "gs://${var.common_params.working_directory_bucket}/${var.common_params.working_directory_prefix}"
+}
+
+resource "random_pet" "job_names" {
+  count = length(local.source_configs)
+}
```
manitgupta, line 20: Add the `smt-` prefix.
Nit: Rename `job_names` to `prefix` or something similar.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/main.tf
```diff
@@ -0,0 +1,78 @@
...
+        dataShards : slice(local.dataShards, batch_start, min(batch_start + var.common_params.batch_size, length(local.dataShards)))
+      }
+    }
+  ]
+
+  working_directory_gcs = "gs://${var.common_params.working_directory_bucket}/${var.common_params.working_directory_prefix}"
+}
+
+resource "random_pet" "job_names" {
+  count = length(local.source_configs)
```
manitgupta, line 19: Add this as a resource tag wherever possible e.g Dataflow job

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/shardingConfig.json
```diff
@@ -0,0 +1,51 @@
...
+          {
+            "dbName": "db3",
+            "databaseId": "logicaldb3",
+            "refDataShardId": "data-shard2"
+          }
+        ]
+      }
+    ]
+  }
+}
```
manitgupta, line 50: This should be native TF configuration and the JSON should be generated from that.

The biggest advantage is that in TF you define the objects declaratively so the user has all the information needed to construct an input from the TF definition itself.

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/terraform/samples/mysql-sharded-end-to-end/variables.tf#L76

manitgupta (commented):
I will recommend running the TF on a fresh GCP project to verify the permissions required, if not done already.

manitgupta:
> I was thinking of doing this but launching a single job via terraform seems like an overkill as launching bulk is pretty straightforward. Customers would likely use this only for sharded use cases.

I don't think its overkill. On the contrary, TF gives a bunch of advantages around state management that most other ways don't give. Moreover, during a migration if the user is already using TF in other places (for e.g live migration), they will likely use it here as well.

I think adding a small rudimentary example will be quick and checks our boxes in terms of providing examples. Should we go ahead and add it?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/shardingConfig.json
Deep1998: Our official documentation refers to launching it via gcloud which already uses this file. They will likely have this config already created. This was done to avoid recreation of this and avoid the extra step. Also a single config file that gcloud/terraform anything can use, and not generate new config based on tool.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/main.tf
Deep1998: With terraform in place, do we want to add smt prefixes? I think we shoud even remove it from the live terraform templates.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: This is a mandatory column for sharded migrations AFAIK. They can rename it but not avoid it for sharded flows. And a session file is also mandatory for sharded flows.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: It is mandatory, the assumption means the GCS working directory is already created by the user and not one we are creating. INside that we create all the output directories for all jobs.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: May not be, I added them because they were also present on the live template.

manitgupta: Okay - let's assess the permissions required for bulk migration specifically and list the minimal required permission set here. This is what we did for live template as well.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
manitgupta: Got it, I think its a bit confusing. Consider rephrasing to something like `GCS bucket with a directory to store DLQ events is already created...` or something similar.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/main.tf
manitgupta: Terraform is also in under the ambit of "SMT" from a customer perspective. You are thinking `SMT = Harbourbridge`, I am thinking of `SMT = anything that the migration team provides`. The current job names are random and there is no way to filter or aggregate resources created by our tooling in the customer's account - so having a prefix is very helpful.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/shardingConfig.json
manitgupta: > They will likely have this config already created. This was done to avoid recreation of this and avoid the extra step.

I don't agree. I think you are confusing both of these things. Customer would likely use **one of** gCloud or Terraform to run the job - so there is no notion of an extra step or recreation here IMO. 

> Also a single config file that gcloud/terraform anything can use, and not generate new config based on tool.

This is **not** in fact a single global configuration. There are a lot of other configurations that are added in the gCloud command in addition to this config file. The **true** configuration file - which gives the complete snapshot of what a user configured, is the user's `tfvars` file.

Due to its many benefits around state management, configuration isolation and debuggability - we should in general encourage users towards using Terraform wherever possible, just like we used to encourage users to run via SMT instead of running via gCloud etc earlier. 

The gCloud way was strictly a way to get early users onto the product while we did the Terraform work. This is also the reason why I am insisting on having a simple, single-job template in place as well. 

Once the Terraform work is merged, you should also consider changing the official documentation/recommendation.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,418 @@
...
+  shard. Please check your network configuration and credentials.
+- If you are still getting a Timeout after this, this means the job graph construction is taking too long. You could
+  potentially have too many tables allocated in a single Dataflow job. The recommendation is to set the `batch_size`
+  such that the total number of tables in a single job does not exceed `150`. This means if a logical shard has 15
+  tables, and a physical shard has 2 logical shards, do not set the batch_size more than 5.
+
+### Job graph is not loading/Custom counters not visible on Dataflow panel
+
+For very large graphs, this can happen. The graph section would be empty and the counters won't load. But worry not, the
+migration should progress nonetheless.
```
manitgupta, line 194: This will be a blind migration and may not be acceptable to users...we should give the instructions to observe the counters directly on cloud monitoring.

CC: @bharadwaj-aditya

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -0,0 +1,418 @@
...
+- **Check Dataflow metrics:** Are memory/CPU limits being hit? Dataflow should autoscale to required number of
+  nodes.
+    - CPU/memory limits being hit means the `max-workers` parameter might be too low. It is recommended to use smaller
+      machines (`Ex: n1-highmem-4`) for most workloads. However, if you have tall tables (singe table larger than 1tb),
+      we recommend using larger machine sizes (`n1-highmem-32`).
+    - For tables with `string` primary keys, the throughput is expected to be lower than tables with `int` primary keys.
+    - Consider bumping up `maxConnection` and `numPartition` parameters when launching the jobs. We recommend `320` for
+      maxConnections and `10000` for numPartitions.
+
+### Configuring to run using a VPC
```
manitgupta, line 211: Are there any firewall rules that need to be configured in a VPC to get Dataflow VMs talking to each and the source database VMs?

manitgupta (commented):
Have we tried running this terraform template against MySQL VMs that are in a VPC and needs to be connected to via the Dataflow job that are - 

1. In the same project and in the same VPC.
2. In a different project (service), with a shared VPC being used to establish the connection?

I think there will likely be additional configuration in terms of firewalls rules or permissions that will be required to run these cases which should be present in the README.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: They are incorrect on cloud monitoring itself, on the console they do not show up at all. There is no source of truth in these cases other than manually running count watchers on spanner.

manitgupta: Are they delayed or are they incorrect? I'm confused - there must be many types of counter metrics that we publish, right?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: Updated docs

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -213,6 +213,37 @@ There can be multiple reasons for this.
...
+
+- Configuring MySQL to allow connections from the Dataflow IPs.
+- Configuring the firewall to allow ingress TCP connection on the MySQL port
+  from the Dataflow IPs.
+
+#### Configuring connectivity on Cloud SQL
+
+Cloud SQL allows multiple ways to establish connectivity. With google
+private access enabled, Dataflow should automatically be able to [access
+the instance](https://cloud.google.com/sql/docs/mysql/authorize-networks).
```
manitgupta, line 228: This is not true IMO. Have you validated this?

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
```diff
@@ -213,6 +213,37 @@ There can be multiple reasons for this.
...
+the firewall is configured to allow connections between all private IPs.
+If the target does not allow the dataflow IPs, we recommend the following configurations
+for allowing connectivity:
+
+- Using network tags to configure MySQL VM. Apply the `databases` network
+  tag on the GCE VM.
+- Add firewall rule that allows tcp ingress on 3306 from resources
+  with the `dataflow` [network tag](https://cloud.google.com/vpc/docs/add-remove-network-tags) on targets
+  with the `databases` tag.
+- Specify the Dataflow ips in source ranges to be allowlisted in the firewall rule.
```
manitgupta, line 243: What will this be? Dataflow runs a large number of individual GCP VMs and how would the customer know which one will these be?
I think you are confusing how network tags and source ranges work - please check once..

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: THis should be within the same VPC. Rephrased.

v2/sourcedb-to-spanner/terraform/samples/sharded-bulk-migration/README.md
Deep1998: Removed the Ip line. Its not needed

manitgupta (approved):

PR #1820, opened 2024-08-28
BigTable to Parquet: only set numShards for >0

damondouglas:
This PR fixes a recent bug discovered by a user that when numShards is null, there's an error in Dataflow that the numShards cannot be 0.


ron-gal (approved):

damondouglas:
Failures in Integration Tests have nothing to do with the code in this PR.

PR #1826, opened 2024-08-29
Do not run spanner-pr tests in Java PR

Abacn:
spanner-pr tests have been flaky, mostly due to different kind of resource exhausted. This PR moves them out of. the common java pr GHA.

### DIFF
--- a/cicd/internal/flags/common-flags_test.go
+++ b/cicd/internal/flags/common-flags_test.go
@@ -18,6 +18,7 @@ package flags
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 )
 
@@ -50,5 +51,25 @@ func TestModulesToBuild(t *testing.T) {
 		if !reflect.DeepEqual(actual, test.expected) {
 			t.Errorf("Returned modules are not equal. Expected %v. Got %v.", test.expected, actual)
 		}
+	}
+}
+
+func TestDefaultExcludedSubModules(t *testing.T) {
+	// common modules won't excluded
+	modulesToBuild = "DEFAULT"
+	defaults := ModulesToBuild()
+	mods := []string{"SPANNER"}
+	var s []string
+	for _, m := range mods {
+		modulesToBuild = m
+		ms := ModulesToBuild()
+		for _, n := range ms {
+			if !strings.HasPrefix(n, "plugins/") {
+				s = append(s, "!"+n)
+			}
+		}
+	}
+	if !reflect.DeepEqual(defaults, s) {
+		t.Errorf("Returned modules are not equal. Expected %v. Got %v.", s, defaults)
 	}
 }
--- a/.github/workflows/java-pr.yml
+++ b/.github/workflows/java-pr.yml
@@ -165,6 +165,7 @@ jobs:
       - name: Run Integration Tests
         run: | 
           ./cicd/run-it-tests \
+          --modules-to-build="DEFAULT" \
           --it-region="us-central1" \
           --it-project="cloud-teleport-testing" \
           --it-artifact-bucket="cloud-teleport-testing-it-gitactions" \
--- a/.github/workflows/spanner-pr.yml
+++ b/.github/workflows/spanner-pr.yml
@@ -21,14 +21,20 @@ on:
     branches:
     - 'main'
     paths:
-      # Include spanner paths only
-      - '.github/workflows/spanner-pr.yml'
-      - 'v2/datastream-to-spanner/**'
+      # Template wide common modules
+      - 'v2/common/**'
+      - 'v2/datastream-common/**'
+      - 'it/google-cloud-platform/**'
+      - 'it/conditions/**'
+      # Component common modules
       - 'v2/spanner-common/**'
-      - 'v2/spanner-change-streams-to-sharded-file-sink/**'
-      - 'v2/gcs-to-sourcedb/**'
       - 'v2/spanner-migrations-sdk/**'
       - 'v2/spanner-custom-shard/**'
+      # Include spanner paths
+      - '.github/workflows/spanner-pr.yml'
+      - 'v2/datastream-to-spanner/**'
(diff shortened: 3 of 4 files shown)


Abacn:
java pr instead run on dispatch_workflow trigger: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/10617855255/job/29432175100 because pr trigger won't pick up changes in github action yaml

Abacn:
Verified there are 6 modules no longer build on Java PR workflow. Previously,

```
Building plaintext-logging 1.0-SNAPSHOT                           [5/74]
```

now

```
Building spanner-common 1.0-SNAPSHOT                             [26/68]
```

Abacn:
R: @Polber @damccorm 

cc: @darshan-sj @aksharauke

cicd/internal/flags/common-flags_test.go
```diff
@@ -52,3 +53,23 @@ func TestModulesToBuild(t *testing.T) {
 		}
 	}
 }
+
+func TestDefaultExcludedSubModules(t *testing.T) {
```
Abacn, line 57: the refactor and tests makes it convenient for future workflow split (just need to add new moduleMap entry)

Polber (commented):
My concern is that changes that would normally affect all templates won't be tested against all templates.

For example, SPANNER includes v2/sourcedb-to-spanner/. This module depends on v2/common. Say someone modifies the CommonTemplateJvmInitializer in v2/common to fix something in one of the JDBC templates. The Sourcedb_to_Spanner_Flex template also uses this class. So, even though there is no direct dependency on common, spanner is still affected when there are changes in it.

Or another simple example - changes to it/ will affect all templates, so it should probably be tested against all templates

Abacn:
Changes affects all templates won't necessary to test against all templates. If it is in common module, it should also affects non-spanner module. If a change only affects spanner module, the code involved shouldn't be in the common module at the first place.

If there are code path sounds necessary to exercise spanner tests, we can add those to trigger path of spanner-pr.yml

Currently spanner integration tests are running twice if a PR touches trigger files or both java-pr and spanner-pr. This is highly ineffective. The current situation is a big workflow running all integration test is not sustainable. It's taking increasingly long, and spanner tests consistently becomes trouble maker. It's blocking major template maintenance work. Please consider split the tests as necessary movement as of review.

Abacn:
This is more like a outage fix change, to fix the already permared java-pr workflow. The spanner integration tests has various issues, which will be considered later by corresponding code owners.

Polber:
> Changes affects all templates won't necessary to test against all templates. If it is in common module, it should also affects non-spanner module. If a change only affects spanner module, the code involved shouldn't be in the common module at the first place.
> 
> If there are code path sounds necessary to exercise spanner tests, we can add those to trigger path of spanner-pr.yml
> 
> Currently spanner integration tests are running twice if a PR touches trigger files or both java-pr and spanner-pr. This is highly ineffective. The current situation is a big workflow running all integration test is not sustainable. It's taking increasingly long, and spanner tests consistently becomes trouble maker. It's blocking major template maintenance work. Please consider split the tests as necessary movement as of review.

v2/datastream-common
v2/common
v2/spanner-common
v2/spanner-migrations-sdk
v2/spanner-custom-shard
it/igoogle-cloud-platform
it/conditions

Polber:
> Changes affects all templates won't necessary to test against all templates. If it is in common module, it should also affects non-spanner module. If a change only affects spanner module, the code involved shouldn't be in the common module at the first place.
> 
> If there are code path sounds necessary to exercise spanner tests, we can add those to trigger path of spanner-pr.yml
> 
> Currently spanner integration tests are running twice if a PR touches trigger files or both java-pr and spanner-pr. This is highly ineffective. The current situation is a big workflow running all integration test is not sustainable. It's taking increasingly long, and spanner tests consistently becomes trouble maker. It's blocking major template maintenance work. Please consider split the tests as necessary movement as of review.

@Abacn I think splitting makes sense, I was just pointing out it could lead to unexpected build failures for non-default modules (i.e. spanner, and soon kafka)

For spanner at least, maybe we should just inform them of the change and point out the following modules are transitive dependencies of the already listed modules:

v2/common
v2/datastream-common

v2/spanner-common
v2/spanner-migrations-sdk
v2/spanner-custom-shard

it/google-cloud-platform
it/conditions

Adding these to the spanner module list could help prevent unexpected failures, and I don't think there are any IT's that would add slowdown to the existing workflow (just some unit tests)

Polber (dismissed):

Abacn:
one integration test failing, not related to this PR:

SpannerChangeStreamsToBigQueryIT.testSpannerChangeStreamsToBigQueryFloatColumns

```
org.apache.beam.sdk.extensions.gcp.util.RetryHttpRequestInitializer - Request failed with code 412, performed 0 retries due to IOExceptions, performed 0 retries due to unsuccessful status codes, HTTP framework says request can be retried, (caller responsible for retrying): https://storage.googleapis.com/upload/storage/v1/b/dataflow-staging-us-central1-269744978479/o?ifGenerationMatch=0&name=staging/common-1.0-SNAPSHOT-i81N2xsQTnm2tfk1XSXEZrxZwuvedBIdmffiqGZdnco.jar&uploadType=resumable&upload_id=AD-8ljt6Dxbk175_A5TsvTt6BOYXTMSxG7deYbHhQUcPu8eV9MYYhsqQAIqqm4PXbstfTPWXdTE7WKIjBvgFYBfWm3InVP_MoE-itu47IyVvWo8T. 

...

Exception: java.lang.NoSuchMethodError thrown from the UncaughtExceptionHandler in thread "main"
```

same test on spanner-pr passed. Likely some flakiness then failing the test, and immediately triggering an unknown dependency version issue when template launching is failing

Abacn:
opened #1836 for the flaky test seen

shunping (approved):
LGTM

PR #1829, opened 2024-08-30
Rollback YamlTemplate to 2.57

Polber:
ReadFromPubSub using JSON format is broken in Beam YAML 2.58.0

Since the formatter for JSON imports a function (part of DoFn) that is in the yaml module (json_utils.py), and the init.py defines from apache_beam.yaml.yaml_transform import *, all the deps of yaml_transform get imported when the Dataflow worker executes the ReadFromPubSub formatter DoFn.

This is a problem because as of beam 2.58, jinja is a dependency of yaml_transform: [go/beam-pr/31684](http://goto.google.com/beam-pr/31684)

However, the worker base image requirements were not updated to include jinja until 2.59: [go/beam-pr/32132](http://goto.google.com/beam-pr/32132)

This means that the workers are trying to import a package they do not have installed.


robertwb:
Could you document the root cause and reason for the rollback in the description?

liferoad (approved):

PR #1837, opened 2024-09-04
feat: optimising writes to spanner

asthamohta:
Writing to only CDC events to Shadow Table. For all Backfill events, if row is present in Shadow table, row is skip, else it is directly written to Spanner.


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
```diff
@@ -184,7 +185,11 @@ public void processElement(ProcessContext c) {
                     }
 
                     // Apply shadow and data table mutations.
-                    transaction.buffer(changeEventContext.getMutations());
+                    if (isBackfill) {
+                      transaction.buffer(changeEventContext.getDataTableMutations());
+                    } else {
+                      transaction.buffer(changeEventContext.getMutations());
+                    }
```
manitgupta, line 192: Add unit test. 

Also check if the ITs need to be updated or a new one needs to be added.

manitgupta (commented):
I think the change itself looks good - let's do some scale & cost testing, and some validations on the correctness to ensure we are not missing any corner cases.

PR #1839, opened 2024-09-05
Load test fixes and multiple PK mutation in same trx handling

aksharauke:
The PR has following changes:
1. Connection pool is added - a singleton at Dataflow worker level with pool size dependent on the worker harness thread number 
2. If there are multiple changes to the same PK in a given transaction- then to check if the source already has latest data - there is logic added to also consider the request sequence number in addition to the commit timestamp

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
@@ -511,7 +511,10 @@ public class SpannerToSourceDb {
                         options.getShardingCustomJarPath(),
                         options.getShardingCustomClassName(),
                         options.getShardingCustomParameters(),
-                        options.getMaxShardConnections())))
+                        options.getMaxShardConnections()
+                            * shards
+                                .size()))) // currently assuming that all shards accept the same
+                                           // number of max connections
             .setCoder(
                 KvCoder.of(
                     VarLongCoder.of(), SerializableCoder.of(TrimmedShardedDataChangeRecord.class)))
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
@@ -48,6 +48,7 @@ import com.google.cloud.teleport.v2.values.FailsafeElement;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.apache.beam.runners.dataflow.options.DataflowPipelineDebugOptions;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineWorkerPoolOptions;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineWorkerPoolOptions.AutoscalingAlgorithmType;
@@ -490,7 +491,7 @@ public class SpannerToSourceDb {
     } else {
       mergedRecords = dlqRecords;
     }
-
+    DataflowPipelineDebugOptions debugOptions = options.as(DataflowPipelineDebugOptions.class);
     SourceWriterTransform.Result sourceWriterOutput =
         mergedRecords
             .apply(
@@ -524,7 +525,10 @@ public class SpannerToSourceDb {
                     options.getSourceDbTimezoneOffset(),
                     ddl,
                     options.getShadowTablePrefix(),
-                    options.getSkipDirectoryName()));
+                    options.getSkipDirectoryName(),
+                    debugOptions.getNumberOfWorkerHarnessThreads() > 0
+                        ? debugOptions.getNumberOfWorkerHarnessThreads()
+                        : Constants.DEFAULT_WORKER_HARNESS_THREAD_COUNT));
 
     PCollection<FailsafeElement<String, String>> dlqPermErrorRecords =
         reconsumedElements
@@ -535,6 +539,7 @@ public class SpannerToSourceDb {
         sourceWriterOutput
             .permanentErrors()
             .setCoder(StringUtf8Coder.of())
+            .apply("Reshuffle3", Reshuffle.viaRandomKey())
             .apply(
                 "Convert permanent errors from source writer to DLQ format",
                 ParDo.of(new ConvertChangeStreamErrorRecordToFailsafeElementFn()))
@@ -564,6 +569,7 @@ public class SpannerToSourceDb {
         sourceWriterOutput
             .retryableErrors()
             .setCoder(StringUtf8Coder.of())
+            .apply("Reshuffle4", Reshuffle.viaRandomKey())
             .apply(
                 "Convert retryable errors from source writer to DLQ format",
                 ParDo.of(new ConvertChangeStreamErrorRecordToFailsafeElementFn()))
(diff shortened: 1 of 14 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -231,7 +231,10 @@ public void processElement(ProcessContext c) throws Exception {
 
       record.setShard(qualifiedShard);
       String finalKeyString = tableName + "_" + keysJsonStr + "_" + qualifiedShard;
-      Long finalKey = finalKeyString.hashCode() % maxConnectionsPerShard;
+      Long finalKey =
+          finalKeyString.hashCode()
+              % maxConnectionsAcrossAllShards; // The total parallelism is
+                                               // maxConnectionsAcrossAllShards
```
Deep1998, line 237: Question: I remember discussing this guarantees ordering per key since Dataflow processes a thread in the same order. But is it a guarantee by dataflow? Example lets say a thread restarts when it had (key1, write1) followed by (key1, write2), will this order be retained after the thread restarts?

Otherwise in these cases, it could happen the MySQL write went through but the spanner write was not executed. This could lead to inconsistencies.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -196,20 +212,28 @@ public void processElement(ProcessContext c) {
...
       } catch (Exception ex) {
         LOG.error("Failed to write to source", ex);
         // we are only interested in the retryable errors
         // https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
-        // Error 1452 and 1451
+        // Error 1452 and 1451 and transient communication errors
+
+        if (ex.getMessage().contains("a foreign key constraint fails")
+            || ex.getMessage().contains("No operations allowed after statement closed")
+            || ex.getMessage().contains("Got timeout reading communication packets")) {
```
Deep1998, line 235: Why can't we use error codes here for matching?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/ConnectionHelper.java
```diff
@@ -0,0 +1,83 @@
...
+        try (StringReader reader = new StringReader(properties)) {
+          jdbcProperties.load(reader);
+        } catch (IOException e) {
+          System.err.println("Error converting string to properties: " + e.getMessage());
+        }
+      }
+
+      for (String key : jdbcProperties.stringPropertyNames()) {
+        String value = jdbcProperties.getProperty(key);
+        config.addDataSourceProperty(key, value);
```
Deep1998, line 67: We seem to be modifying a config here, but it is not being used anywhere after this. It is last used to initalise `ds` which is put in the map as is.  Is this expected?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -524,7 +528,10 @@ public static PipelineResult run(Options options) {
                     options.getSourceDbTimezoneOffset(),
                     ddl,
                     options.getShadowTablePrefix(),
-                    options.getSkipDirectoryName()));
+                    options.getSkipDirectoryName(),
+                    debugOptions.getNumberOfWorkerHarnessThreads() > 0
+                        ? debugOptions.getNumberOfWorkerHarnessThreads()
```
Deep1998, line 533: I had a general question on how jdbcIO sets the pool size per worker, since it does not seem to be using getNumberOfWorkerHarnessThreads in the code that I see? We can maybe replicate whatever they do.

If this is giving improvements as per testing, we can take that as a followup.

aksharauke: JDBCIO gives abstractions to provide the DataSourceProvider - the caller can create it based on their choice of the connection pool, size etc. Meaning there is no set way.  There is a default DBCP2 pool (PoolingDataSource ) - pool size is default to -1 not sure what DBCP2 interprets it as. I could not find in-house implementation where connection pool size is provided ( for instance CdcJdbcIO and DynamicJdbcIO,v2/jdbc-common/src/main/java/com/google/cloud/teleport/v2/auto/blocks/ReadFromJdbc.java)

I have changed the connection pool logic a bit: I have taken max number of workers configured and the max connections allowed per shard - took their division as the connection pool size for the worker. This is more intuitive - since it guarantees that it will never cross the max allowed connections.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/ConnectionHelper.java
aksharauke: Thanks! Fixed

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/ConnectionHelper.java
```diff
@@ -0,0 +1,83 @@
...
+      config.setMaximumPoolSize(maxConnections);
+      config.setConnectionInitSql(
+          "SET SESSION net_read_timeout=1200"); // to avoid timeouts at network level layer
+      HikariDataSource ds = new HikariDataSource(config);
+      Properties jdbcProperties = new Properties();
+      if (properties != null && !properties.isEmpty()) {
+        try (StringReader reader = new StringReader(properties)) {
+          jdbcProperties.load(reader);
+        } catch (IOException e) {
+          System.err.println("Error converting string to properties: " + e.getMessage());
```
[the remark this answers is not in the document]
aksharauke: Done

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
aksharauke: Thanks! Done

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
aksharauke: Yeah, as per this: https://g3doc.corp.google.com/company/gfw/support/cloud/playbooks/dataflow/streaming_pipeline_slow_stuck.md?cl=head#keys-in-dataflow - there is a guarantee by Dataflow. For your scenario about worker restart - yes this is possible that for a given element - in mid-processing, the worker restarted. Then Dataflow will give the same item to different worker thread - but it will not happen that two threads are processing the same element in parallel. So the half-processed element will begin it's processing again - since there was no update done to Spanner shadow table - it will do upsert to MySQL and this time the Spanner write will also happen. 

There is a second scenario - let's suppose there is only one Dataflow key in the system, for simplicity. It can happen that MySQL write succeeded and the Spanner shadow table write failed - some Spanner issue - then this record gets written to DLQ for reprocessing. Meanwhile, it can happen that data with an earlier commit timestamp arrives - we will end up writing this to MySQL and update the commit timestamp in Spanner. After DLQ record is processed,MySQL will have new data again. So in essence: the source is eventually consistent (by design)

Deep1998 (dismissed):

Deep1998 (approved):

PR #1840, opened 2024-09-05
feat: Modularization: Schema overrides via string and file

manitgupta:


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
```diff
@@ -756,4 +810,30 @@ private static DeadLetterQueueManager buildDlqManager(Options options) {
...
+
+  private static ISchemaOverridesParser configureSchemaOverrides(Options options) {
+    // incorrect configuration
+    if (!options.getSchemaOverridesFilePath().isEmpty()
+        && (!options.getTableOverrides().isEmpty() || !options.getColumnOverrides().isEmpty())) {
+      throw new IllegalArgumentException(
+          "Only one of file based or string based overrides must be configured! Please correct the configuration and re-run the job");
+    }
+    // string based overrides
+    if (!options.getTableOverrides().isEmpty() || !options.getColumnOverrides().isEmpty()) {
```
bharadwaj-aditya, line 822: how does this handle if both table and column overrides are non-empty ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.migrations.schema;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+public interface ISchemaOverridesParser {
```
bharadwaj-aditya, line 20: Can you re-use this interface ? 

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java

bharadwaj-aditya (changes requested):

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
manitgupta: That is covered in the parser. If they conflict and the parsing fails, an exception is thrown. Otherwise, both are honored.

manitgupta:
@bharadwaj-aditya 
I took a 2nd look at making changes to the session file based mapping logic inside bulk and live migration template.

Code ref -

1. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/30370f0460b2a29d4266d0ef95f6ec1e06a20d9e/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java#L64

2. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/30370f0460b2a29d4266d0ef95f6ec1e06a20d9e/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java#L88

Observations - 

1. Post integration of `SessionFileMapper` in live migration template, it would only be used to call two utility methods `getSpannerTableName` and `getSpannerColumnName` instead of accessing them directly via the `schema` object, which live migration template does today. This is not much benefit.
2. `SessionFileMapper` requires a `Ddl` object as input to construct it, which is not required at the place where overrides are implemented currently. 
3. A large part of the mapping logic for bulk migration is actually written inside the `PipelineController`, similar is the case for live migrations, where it is written in `ChangeEventTransformerDoFn`. In the discussion that we had on greater encapsulation of the mapping logic, it seems like we will need to refactor **both** bulk and live migration to achieve it. 

I will have to dig a bit deeper into the bulk code, but it looks like actually some of the methods like `getSpannerColumnName` in `SessionFileMapper` are actually only used in UTs. It's not immediately clear to me why.

Screenshot 2024-09-25 at 3 03 40 PM

I'll think a bit more on how we can go about this, but we should treat it as a separate exercise IMO.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
```diff
@@ -169,6 +177,35 @@ JsonNode removeDroppedColumns(JsonNode changeEvent, String tableId) {
...
+    List<String> sourceFieldNames = ChangeEventUtils.getEventColumnKeys(changeEvent);
+    sourceFieldNames.forEach(
+        sourceFieldName -> {
+          Pair<String, String> spannerTableColumn =
+              schemaOverridesParser.getColumnOverrideOrDefault(sourceTableName, sourceFieldName);
+          // a valid column override for the table in this changeEvent exist
+          // 1.  the table name of the source should match the one specified in the override
+          // 2. the column name override should be a different value than the current source field
+          // name.
+          if (sourceTableName.equals(spannerTableColumn.getLeft())
```
bharadwaj-aditya, line 199: is this check needed ? the source table is being passed to the override parser in any case

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Gets the spanner table name given the source table name, or source table name if no override is
+   * configured.
+   *
+   * @param sourceTableName The source table name
+   * @return The overridden spanner table name
+   */
+  String getTableOverrideOrDefault(String sourceTableName);
```
bharadwaj-aditya, line 29: nit: the purpose of this method is to return the name to be used in spanner. can we just call this getSpannerTableName ?

In top level interfaces it is generally better to have a deterministic function name rather than have or clauses to abstract out the logic.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Gets the spanner column name given the source table name, or the source column name if override
+   * is configured.
+   *
+   * @param sourceTableName the source table name for which column name is overridden
+   * @param sourceColumnName the source column name being overridden
+   * @return A pair of spannerTableName and spannerColumnName
+   */
+  Pair<String, String> getColumnOverrideOrDefault(String sourceTableName, String sourceColumnName);
```
bharadwaj-aditya, line 39: Similarly, this would just be getSpannerColumn name - it will return the Spanner column name only if the table name and column name are available in the overrrides. If the tuple is not available, then just return the original column name.

An alternative would be to rename this method to getColumnOverride.
If no override is configured for this tuple, just return null. I think a check for this is being done at the caller for this in any case, so we're not really reducing any complexity by having the or clause here.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaFileOverride.java
```diff
@@ -0,0 +1,40 @@
...
+
+  private final Map<String, Map<String, String>> renamedColumns;
+
+  public SchemaFileOverride(
+      Map<String, String> renamedTables, Map<String, Map<String, String>> renamedColumns) {
+    this.renamedTables = renamedTables;
+    this.renamedColumns = renamedColumns;
+  }
+
+  public Map<String, Map<String, String>> getRenamedColumns() {
```
bharadwaj-aditya, line 33: nit - calling this renamed column tuple or such would help distinguish the purpose of this better.

bharadwaj-aditya:
> @bharadwaj-aditya I took a 2nd look at making changes to the session file based mapping logic inside bulk and live migration template.
> 
> Code ref -
> 
> 1. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/30370f0460b2a29d4266d0ef95f6ec1e06a20d9e/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java#L64
> 2. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/30370f0460b2a29d4266d0ef95f6ec1e06a20d9e/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java#L88
> 
> Observations -
> 
> 1. Post integration of `SessionFileMapper` in live migration template, it would only be used to call two utility methods `getSpannerTableName` and `getSpannerColumnName` instead of accessing them directly via the `schema` object, which live migration template does today. This is not much benefit.
> 2. `SessionFileMapper` requires a `Ddl` object as input to construct it, which is not required at the place where overrides are implemented currently.
> 3. A large part of the mapping logic for bulk migration is actually written inside the `PipelineController`, similar is the case for live migrations, where it is written in `ChangeEventTransformerDoFn`. In the discussion that we had on greater encapsulation of the mapping logic, it seems like we will need to refactor **both** bulk and live migration to achieve it.
> 
> I will have to dig a bit deeper into the bulk code, but it looks like actually some of the methods like `getSpannerColumnName` in `SessionFileMapper` are actually only used in UTs. It's not immediately clear to me why.
> 
> Screenshot 2024-09-25 at 3 03 40 PM
> 
> I'll think a bit more on how we can go about this, but we should treat it as a separate exercise IMO.

Regarding 1 - Agreed the ISchemaMapper does some stuff which is not required here. 

Regarding 2 - Agreed - it is not required currently, but the logic to resolve it etc exists in the template in any case, any concern in using that in this flow ? It comes as the ddlView which is loaded as a part of the pipeline as opposed to launcher, but i'm not sure that is reason enough to not use it.

Regarding 3 - 
In the bulk migration flow, the approach is to be SpannerSchema first, so the calls being made are getSourceColumn instead of getSpannerColumn. The mapper is being used for these calls in both identity and session.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
manitgupta: Agreed that this check is not needed, I just wanted to be more explicit and implement the two conditions written above in the docs.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
manitgupta: These type of methods are quite common in Java. My implementation here is just an extension of an already existing concept - 

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#getOrDefault(java.lang.Object,V)

I am not particular about the naming here though, so if you think it is confusion, I can change it.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
manitgupta: These type of methods are quite common in Java. My implementation here is just an extension of an already existing concept - 

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#getOrDefault(java.lang.Object,V)

I am not particular about the naming here though, so if you think it is confusion, I can change it.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaFileOverride.java
manitgupta: Renamed to `getRenamedColumnTupleMap` since it returns a column.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
manitgupta: Updated the name.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
manitgupta: Updated the name.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
bharadwaj-aditya: This seems quite confusing which stems from the return type pair of the interface being confusing.

Case in point. The check here seems to verify sourceTableName against the SpannerTableColumn.getLeft() which seems odd - what is being read the source or spanner ?

It took me a while to figure out which table is being returned. even the interface javadocs don't mention this. I suggest just keep the interface simple and return the column name for given source table. Let the implementation of the interface ensure that the table is retreived correctly or default value is handled. This seems like a leak in responsibility.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Gets the spanner column name given the source table name, or the source column name if override
+   * is configured.
+   *
+   * @param sourceTableName the source table name for which column name is overridden
+   * @param sourceColumnName the source column name being overridden
+   * @return A pair of spannerTableName and spannerColumnName
+   */
+  Pair<String, String> getColumnOverride(String sourceTableName, String sourceColumnName);
```
bharadwaj-aditya, line 39: This interface is quite confusing suggest this just return the overriden column name. Refer to comment above.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
```diff
@@ -810,4 +865,30 @@ private static DeadLetterQueueManager buildDlqManager(Options options) {
       return DeadLetterQueueManager.create(dlqDirectory, retryDlqUri, 0);
     }
   }
+
+  private static ISchemaOverridesParser configureSchemaOverrides(Options options) {
+    // incorrect configuration
+    if (!options.getSchemaOverridesFilePath().isEmpty()
```
bharadwaj-aditya, line 871: can you add UT coverage for this ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaFileOverridesParser.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class SchemaFileOverridesParser implements ISchemaOverridesParser, Serializable {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SchemaFileOverridesParser.class);
+  private static Gson gson = new Gson();
+
+  final SchemaFileOverride schemaFileOverride;
+
+  public SchemaFileOverridesParser(String overridesFilePath) {
+    try (InputStream stream =
```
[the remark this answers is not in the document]
manitgupta: It is covered - https://screenshot.googleplex.com/9Tjd6SNe75VLCgy

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
manitgupta: This is part of `DatastreamToSpanner` class which initializes the beam pipeline. 
We don't seem to have UTs for this type of code - 

https://app.codecov.io/gh/GoogleCloudPlatform/DataflowTemplates/blob/main/v2%2Fdatastream-to-spanner%2Fsrc%2Fmain%2Fjava%2Fcom%2Fgoogle%2Fcloud%2Fteleport%2Fv2%2Ftemplates%2FDataStreamToSpanner.java

https://app.codecov.io/gh/GoogleCloudPlatform/DataflowTemplates/blob/main/v2%2Fsourcedb-to-spanner%2Fsrc%2Fmain%2Fjava%2Fcom%2Fgoogle%2Fcloud%2Fteleport%2Fv2%2Ftemplates%2FSourceDbToSpanner.java

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaOverridesParser.java
manitgupta: Done

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
manitgupta: Done

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaStringOverridesParser.java
```diff
@@ -0,0 +1,173 @@
...
+                  String.format("%s.%s", sourceTableColumn[0], sourceTableColumn[1]),
+                  new ImmutablePair<>(destTableColumn[0], destTableColumn[1]));
+            } else {
+              throw new IllegalArgumentException(
+                  String.format(
+                      "Malformed pair encountered: %s, please fix the columnOverrides and rerun the job.",
+                      pair));
+            }
+          } else {
+            throw new IllegalArgumentException(
```
bharadwaj-aditya, line 133: we should test for invalid scenarios as well, more important considering these are user inputs.

bharadwaj-aditya (approved):
LGTM

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/SchemaStringOverridesParser.java
manitgupta: User input is gated via regex. If any malformed string which slips through the regex will be caught in this exception which has been unit tested.

PR #1842, opened 2024-09-05
add spanner-migrations-team as CODEOWNERS for v2/datastream-common

Abacn:

PR #1849, opened 2024-09-09
Revert "adding datastream as owners of our templates (#1798)"

Abacn:
Also Revert "add spanner-migrations-team as CODEOWNERS for v2/datastream-common (#1842)"

#1798 added a "GoogleCloudPlatform/datastream" to some code path owners, however, this team currently only has one member, and real datastream team is prevented from making code changes to datastream templates. Revert it for now until the GitHub team settings are correctly configured


Abacn:
R: @damccorm @dhercher @yair-harel

damccorm:
I think we should leave this - the right thing to do here is to continue to add folks to the datastream GitHub team and for them to accept invites

PR #1859, opened 2024-09-11
Added support for WHERE ... IS NOT NULL for indexes

atask-g:
Included the INFORMATION_SCHEMA.INDEXES.FILTER column in the listIndexes query to enable support for WHERE ... IS NOT NULL.

I ran manual testing by exporting a database with an index that has the WHERE ... IS NOT NULL clause and was able to successfully import it with the correct resulting schema and all rows restored.

@darshan-sj

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Index.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Index.java
@@ -47,7 +47,6 @@ public abstract class Index implements Serializable {
   // restricted for gsql
   abstract boolean nullFiltered();
 
-  // restricted for pg
   @Nullable
   abstract String filter();
 
@@ -179,6 +178,11 @@ public abstract class Index implements Serializable {
     if (interleaveIn() != null) {
       appendable.append(", INTERLEAVE IN ").append(quoteIdentifier(interleaveIn(), dialect()));
     }
+
+    if (!nullFiltered() && filter() != null && !filter().isEmpty()) {
+      appendable.append(" WHERE ").append(filter());
+    }
+
     if (options() != null) {
       String optionsString = String.join(",", options());
       if (!optionsString.isEmpty()) {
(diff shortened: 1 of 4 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Index.java
```diff
@@ -179,6 +178,11 @@ private void prettyPrintGsql(Appendable appendable) throws IOException {
     if (interleaveIn() != null) {
       appendable.append(", INTERLEAVE IN ").append(quoteIdentifier(interleaveIn(), dialect()));
     }
+
+    if (!nullFiltered() && filter() != null && !filter().isEmpty()) {
+      appendable.append(" WHERE ").append(filter());
```
darshan-sj, line 183: This enables where clause for all type of indexes. Not just Search index. Correct?

darshan-sj (approved):
LGTM.

Please run an Export and Import manually and document the testing done in the PR description.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Index.java
atask-g: Yes, this enables it for all index types and is intentional.

atask-g:
I ran manual testing by exporting a database with an index that has the WHERE ... IS NOT NULL clause and was able to successfully import it with the correct resulting schema and all rows restored.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -445,7 +442,7 @@ Statement listIndexesSQL() {
       case GOOGLE_STANDARD_SQL:
         return Statement.of(
             "SELECT t.table_schema, t.table_name, t.index_name, t.parent_table_name, t.is_unique,"
-                + " t.is_null_filtered, t.index_type, t.search_partition_by, t.search_order_by"
+                + " t.is_null_filtered, t.filter, t.index_type, t.search_partition_by, t.search_order_by"
```
damccorm, line 445: I think this may be causing test failures - e.g. https://pantheon.corp.google.com/dataflow/jobs/us-central1/2024-09-24_01_47_34-4743833279596043378;bottomTab=JOB_LOGS;expandBottomPanel=true;logsSeverity=ERROR;graphView=0?project=cloud-teleport-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=13802955&mods=-ai_platform_fake_service,ai_platform_staging_service

PR #1862, opened 2024-09-11
Build PRs using artifact registry packages

Polber:
This PR configures maven on GitHub Actions workers (self-hosted) to pull dependencies from a Google-hosted artifact registry. This ensures that PR's are able to stage templates the same way that templates are staged during the release process to detect possible package discrepancies earlier.

### DIFF
--- a/cicd/cmd/run-build/main.go
+++ b/cicd/cmd/run-build/main.go
@@ -35,7 +35,8 @@ func main() {
 		mvnFlags.SkipShade(),
 		mvnFlags.SkipSpotlessCheck(),
 		mvnFlags.SkipTests(),
-		mvnFlags.ThreadCount(3))
+		mvnFlags.ThreadCount(3),
+		flags.InternalMaven())
 	if err != nil {
 		log.Fatalf("%v\n", err)
 	}
--- a/cicd/cmd/run-build/main.go
+++ b/cicd/cmd/run-build/main.go
@@ -36,7 +36,7 @@ func main() {
 		mvnFlags.SkipSpotlessCheck(),
 		mvnFlags.SkipTests(),
 		mvnFlags.ThreadCount(3),
-		flags.InternalMaven())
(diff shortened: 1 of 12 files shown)


damccorm:
Could you add a description of what/why this is doing now that we pulled out the urgent pieces? It does also look like checks might be persistently failing (not sure if that's a flake or not)

Polber:
@damccorm It's also hard to test because new caches get created by other workflows which affect this one being able to use AR packages instead of maven central. Once it is merged, I'll be able to clear caches and all new ones will use AR packages.

I'm pretty confident this change works, but let's wait until next week to merge so I can keep an eye on actions and see if anything fails.

cicd/cmd/run-build/main.go
```diff
@@ -35,7 +35,8 @@ func main() {
 		mvnFlags.SkipShade(),
 		mvnFlags.SkipSpotlessCheck(),
 		mvnFlags.SkipTests(),
-		mvnFlags.ThreadCount(3))
+		mvnFlags.ThreadCount(3),
+		mvnFlags.InternalMaven())
```
damccorm, line 39: Is this accessible to everyone? If not, should we disable it for non-CI/release builds by default?

Polber: @damccorm All builds on GitHub actions have access to the Airlock repo, and the `cicd/` folder only runs on GitHub actions, so there shouldn't be a case where these don't run, unless I'm missing something

damccorm: > and the cicd/ folder only runs on GitHub actions

You're right, this is the part I wasn't considering - I forgot people don't use these for their normal local development. Honestly, that workflow seems wrong to me (ideally there would be little divergence from remote/local to make it easy to repro issues), but that's out of scope here

damccorm (approved):
LGTM - feel free to merge at a time when you're able to monitor/rollback if needed

PR #1863, opened 2024-09-12
feat: update pg index discovery query for 9.4+

thiagotnunes:
Update the SourceDbToSpanner index discovery query for PostgreSQL as a source to work in PG versions older than 11 (9.4+).

Before version 11, PG did not support the INCLUDE clause in indexes, so the pg_catalog.pg_index table did not contain the column 'indnkeyatts' (number of key columns in an index). It only contained the column 'indnatts' (number of columns in an index).

In this change, we first try to read from the 'indnkeyatts' and fallback to 'indnatts'.


VardhanThigle (dismissed):
Overall look good to me.

For our future reference by wider team (folks working on mysql, pg or other sources), could you please add perma-links to dbfiddle for these queries?

thiagotnunes:
Updated db-fiddle query

PR #1864, opened 2024-09-12
Fixing isue of discovering unsigned types

VardhanThigle:
## Root cause
1. `integer unsigned` is discovered as `integer` instead of `integer unsigned`.
2. This is because the code was referring to `information_schema.column.DATA_TYPE` instead of `information_schema.column.COLUMN_TYPE`.
## Why was this not caught in UT?
While we have UT for unsigned types, we don't have the real `information_schema` in UT, it's the mocked version that we use to unit test unsigned types. So it's possibly impossible to catch differences in the behavior of `DATA_TYPE` and `COLUMN_TYPE` columns of the `information_schema` in the UT.
## Why was this not caught in IT?
Needed implementation for the test case for `int unsigned`
## Fix
1. Use `information_schema.column.COLUMN_TYPE` instead of `information_schema.column.DATA_TYPE`
2. Enhance the exception log to have table and schema name.
3. Add IT coverage for unsigned integer
## Testing
1. End to End migration on failing schema with `unsignedint_max` as well as failing value.
2. IT that covers `integer unsigned` with max boundary.


bharadwaj-aditya (dismissed):
LGTM - can you also add tests for other unsigned types to be future proof.

VardhanThigle:
> LGTM - can you also add tests for other unsigned types to be future proof.

Done. Added in the current push.

Deep1998 (approved):

PR #1866, opened 2024-09-13
Refactor Dockerfile-based images plugin

Polber:
This PR refactors and cleans up plugin code for compiling Dockerfile-based templates (YAML, PYTHON, XLANG)


metadata/src/main/java/com/google/cloud/teleport/metadata/HelperImage.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that marks a helper image for a Dataflow Template. The helper images will use the same
+ * underlying image as the template they are annotating.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface HelperImage {
```
damccorm, line 29: Could we split this PR into a few different PRs? It is rather large and is trying to do several things at once (refactor/cleanup, helperImage, job-builder-server(?)), and I'm having a hard time figuring out how to review it as a result.

In general, it is probably better to separate large diff refactors from other prs when possible

Polber: Sure, I pulled out everything besides the refactoring of the Dockerfile generator. The refactor basically just consolidates all the code from the 3 separate generators into a single generator that is constructed using a builder pattern with defaults. This is also another step in the direction of allowing a single Dockerfile for all the image types

Polber: It also adds a couple configurable options to be compatible with the HelperImage annotation in the next PR

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
```diff
@@ -70,18 +70,26 @@ public class TemplatesReleaseMojo extends TemplatesBaseMojo {
   @Parameter(defaultValue = "${gcpTempLocation}", readonly = true, required = false)
   protected String gcpTempLocation;
 
-  @Parameter(
-      name = "baseContainerImage",
-      defaultValue =
-          "gcr.io/dataflow-templates-base/java11-template-launcher-base-distroless:latest",
-      required = false)
+  @Parameter(defaultValue = "${baseContainerImage}", readonly = true, required = false)
```
damccorm, line 73: I don't follow this piece - why are we changing the default value here?

damccorm: (same question applies to defaults across the board)

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -886,6 +926,65 @@ private void stageXlangUsingDockerfile(String imagePath, String dockerfile)
...
+                }
+              });
+    } catch (Exception e) {
+      LOG.warn("unable to copy jar files");
+      throw e;
+    }
+  }
+
+  private void setDockerfileDefaults() {
+    if (baseContainerImage == null) {
```
damccorm, line 968: Will this ever be null? Doesn't it have a default value?

damccorm (commented):
Thanks - splitting it up helped. The generator changes looked good, I didn't follow everything happening in the plugins though and asked a few questions

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
Polber: Other plugins (i.e. Run and Release) call this plugin and can set those values in theory (though this never happens in practice)

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
Polber: It can be null if it is not set on the command line when running the maven command. In which case we want to use a default value

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
Polber: Existing example of a parameter not being set and therefore being set to a default value: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java#L272-L276

damccorm: Right - my question was why we're changing the existing defaults though

damccorm:
Looks like this is breaking checks also

damccorm (approved):
Sorry - missed that this was ready for re-review. This looks good to me if we can get a green signal. Looks like `googlecloud-to-neo4j ` is broken for this PR - might just be a flake, but lets retry/pull in the latest `main` and make sure its not this PR breaking it

damccorm (approved):

PR #1868, opened 2024-09-13
Auto-generate fully specified requirements

damccorm:
Uses the same mechanisms as https://github.com/apache/beam/blob/master/sdks/python/container/run_generate_requirements.sh to auto-generate full requirements files with all dependencies specified.

Without this change, we were being flagged by internal tooling which wants us to use fully specified requirements. Once we have this, we can enhance it and use it to do `pip installs` with `--require-hashes` to enforce security requirements.


damccorm:
(internal bug id - 360236369)

Polber (approved):

python/generate_dependencies.sh
```diff
@@ -0,0 +1,75 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
```
Polber, line 1: nit
```suggestion
# Copyright 2024 Google Inc. All Rights Reserved.
```
and below

damccorm:
Force merging since approval came before a header string only change

PR #1871, opened 2024-09-16
Require hashes for all pip installs

damccorm:
Continuation of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1912 to make our pip installs more secure. We should require hashes to make sure the underlying dependency doesn't move out from under us.


plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -16,16 +14,15 @@ ${filesToCopy}
...
 
 # Install dependencies to launch the pipeline and download to reduce startup time
 # Remove Jinja2 dependency once YAML templatization support is added to Beam
 # TODO - remove `pip uninstall apache-beam` line when repo is upgraded to Beam 2.59.0
 RUN python -m venv /venv \
     && /venv/bin/pip uninstall apache-beam -y \
-    && /venv/bin/pip install --no-cache-dir --upgrade pip setuptools \
-    && /venv/bin/pip install --no-cache-dir -U -r $REQUIREMENTS_FILE \
+    && /venv/bin/pip install --require-hashes --no-cache-dir -U -r $REQUIREMENTS_FILE \
     && /venv/bin/pip install --no-cache-dir -U Jinja2 \
```
Polber, line 26: I'm pretty sure this is not needed as of Beam 2.59, so you could remove to avoid installing without hash
```suggestion
```

plugins/core-plugin/src/main/resources/Dockerfile-template-python
```diff
@@ -8,11 +8,9 @@ WORKDIR $WORKDIR
...
 ENV FLEX_TEMPLATE_PYTHON_PY_FILE=main.py
 
-ARG BEAM_VERSION=${beamVersion}
-ARG BEAM_PACKAGE=apache-beam[gcp]==$BEAM_VERSION
+RUN if ! [ -f requirements.txt ] ; then >&2 echo "error: no requirements.txt file found" && exit 1 ; fi
 
-RUN pip install $BEAM_PACKAGE
-RUN pip install -U -r $FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE
+RUN pip install -U -r --require-hashes $FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE
 RUN pip download --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE
```
Polber, line 14: Should `pip download` also use `--require-hashes`?

Same for the other Dockerfiles

plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
```diff
@@ -16,13 +14,13 @@ ${filesToCopy}
...
-# Create requirements.txt file if not provided
-RUN if ! [ -f requirements.txt ] ; then echo "$BEAM_PACKAGE" > requirements.txt ; fi
+# Throw if requirements.txt file not provided
+RUN if ! [ -f requirements.txt ] ; then >&2 echo "error: no requirements.txt file found" && exit 1 ; fi
 
 # Install dependencies to launch the pipeline and download to reduce startup time
 RUN python -m venv /venv \
-    && /venv/bin/pip install --no-cache-dir --upgrade pip setuptools \
-    && /venv/bin/pip install --no-cache-dir -U -r $REQUIREMENTS_FILE \
+    && python -m ensurepip --upgrade \
```
Polber, line 22: Does this actually ensure pip for the venv?
```suggestion
    && /venv/bin/python -m ensurepip --upgrade \
```

Same for the others

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -699,10 +710,20 @@ private void stageFlexPythonTemplate(
...
     if (!dockerfile.exists()) {
+      // Copy in requirements.txt if present
+      File sourceRequirements = new File(outputClassesDirectory.getPath() + "/requirements.txt");
+      File destRequirements = new File(dockerfileContainer + "/requirements.txt");
+      if (sourceRequirements.exists()) {
+        Files.copy(
+            sourceRequirements.toPath(),
+            destRequirements.toPath(),
+            StandardCopyOption.REPLACE_EXISTING);
+      }
```
Polber, line 726: nit: Why is this needed?

Cloud Build runs in the `outputClassesDirectory`, so it shouldn't even be looking at the `dockerfilePath`

Polber (changes requested):

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
damccorm: `dockerfilePath` is in `outputClassesDirectory`, so these are related. I found that as it stands right now (without my change), the requirements.txt files were not being found even when present, but colocating them with the dockerfile allowed docker to find them without issue.

damccorm: But it definitely does fail without this change

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
damccorm: Thanks - done

plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
damccorm: I added the suggestion, not sure I understand the question though.

plugins/core-plugin/src/main/resources/Dockerfile-template-python
damccorm: Yeah, done

plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
Polber: "The ensurepip package provides support for bootstrapping the pip installer into an existing Python installation or virtual environment."

I just wasn't sure if calling `python -m ensurepip` would perform the intended function for the venv since it was running on the system python rather than the venv python (which I think is just a pointer to the system python, but again I was just unsure)

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
Polber: Ok interesting because we used to use requirements files and they were stored in the resources directory of the template, so not sure what changed. But ok SGTM

Polber (approved):

damccorm:
postgressql failures are unrelated to this change, so I'm going to merge

PR #1873, opened 2024-09-17
Improve bulk terraform template to work seamlessly

Deep1998:
Added permission and firewall rules along with other changes to make shared VPC bulk orchestration seamless.
Tests:
Orchestrated a sharded bulk migration in a fresh shared VPC setup done via spanner-common samples.

Follow-up PR: Move allow-datastream firewall rule from infra-setup script to live migration terraform template.


bharadwaj-aditya (approved):
LGTM

PR #1876, opened 2024-09-18
fixes for multi shard load test

aksharauke:
1. The Dao map was incorrectly being emptied, corrected the initialization
2. The MySQL driver name is corrected
3. Excessive logs are removed - which were being printed per record


darshan-sj (approved):
LGTM

PR #1877, opened 2024-09-18
Adding support for bigInt Unsigned as Primary Key for partitioning and connectionProperties in shard-config

VardhanThigle:
## Overview
This PR covers a few cases keeping in view upcoming use-cases of bulk migration template.

1. Adding Support for `BigInt Unsigned` as primary key for partitioning. This is the next type we are addeding as primary key to the earlier list of everything that's subset of `long` and `string`.
2. Adding support for `connectionProperties` in Shard-Config
3. Allowing `portNumber` to be an integer in shard-config.

## Testing
1. UT coverage for the changes.
2. IT coverage for `BigInt Unsigned` and `String` as primary keys. `Long` primary key is already covered by existing integration tests.
3. End to End migration on 10 physical shard migration with `BigInt Unsigned` as primary key.
    The shard-config used for the end-to-end migration, also adds connectionProperties and integer Port number for test
    ```json
      "dataShardId": "X-X-X-X",
        "host": "X.X.X.X",
        "user": "root",
        "password": "XXXX",
        "port": 3306,
        "connectionProperties": "tcpKeepAlive=false",
        "dbName": "",
     ```

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
@@ -18,6 +18,7 @@ package com.google.cloud.teleport.v2.spanner.migrations.utils;
 import com.google.cloud.teleport.v2.spanner.migrations.shard.Shard;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.GsonBuilder;
+import com.google.gson.ToNumberPolicy;
 import com.google.gson.reflect.TypeToken;
 import java.io.IOException;
 import java.io.InputStream;
@@ -184,6 +185,7 @@ public class ShardFileReader {
     Map shardConfigMap =
         new GsonBuilder()
             .setFieldNamingPolicy(FieldNamingPolicy.IDENTITY)
+            .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
             .create()
             .fromJson(jsonString, shardConfiguration);
 
@@ -225,7 +227,8 @@ public class ShardFileReader {
               (String) (dataShard.get("user")),
               password,
               "",
-              (String) (dataShard.get("secretManagerUri")));
+              (String) (dataShard.get("secretManagerUri")),
+              dataShard.getOrDefault("connectionProperties", "").toString());
 
       for (Map database : databases) {
         shard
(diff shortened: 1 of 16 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
```diff
@@ -216,16 +216,22 @@ public List<Shard> readForwardMigrationShardingConfig(String sourceShardsFilePat
...
+      // inadvertently accept
+      // incorrect inputs like "3306.1".
+      // So we trim the decimal point followed by a string of zeros.
+      String regexForTrailingDecimal = "\\.0+$";
       Shard shard =
           new Shard(
               "",
               host,
-              dataShard.getOrDefault("port", 0).toString(),
+              dataShard.getOrDefault("port", 0).toString().replaceAll(regexForTrailingDecimal, ""),
```
bharadwaj-aditya, line 229: we should fail if this sort of config is passed. Unless the string is empty or can be parsed as a number, we should just fail this bit. If we go down the path of cleaning up the input, there may be a whole bunch of cases which we don't handle.

Is there any real world case you can think of where we should be handling this case ?

bharadwaj-aditya (changes requested):

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
Deep1998: We can alternatively take a string, convert it to number and cast it to int. 
However, I think we should enforce a singe format, that port should either be inputted as an int or a string.

Deep1998 (commented):
LGTM

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
VardhanThigle: It's always a string in json.
when the user adds 3306 w/o quotes in the file, it gets interpreted as "3306.0" (A string with `.0` appended)
We can round this down and convert it to an integer but then we will need to also make sure we don't accept "3306.1"

So I just trimmed the ".0+" if it's present.

bharadwaj-aditya: A cleaner way to do this would be one of the following 
1. check the type after parsing - if it is a number, just do number.tointeger
2. use the tonumber polizy instead -https://javadoc.io/doc/com.google.code.gson/gson/2.9.0/com.google.gson/com/google/gson/ToNumberPolicy.html

VardhanThigle: Thanks a ton! The toNubmerPolicy makes it lot more idomatic and less hacky.
I have updated the code to use the same.

Deep1998 (approved):

bharadwaj-aditya (approved):
LGTM

PR #1882, opened 2024-09-20
Add Job Builder server image

Polber:
~~Depends on #1866~~

~~Introduce `@HelperImage` tag to build a clone of the Dockerfile for a given template with overrides for a different entrypoint, files copied over, etc. This will allow an image that is traditionally based on the template image to be built and versioned using the same base image - without the need for child Dockerfile.~~

UPDATE: This PR now stages the Job Builder server image as its own "Template" without staging the spec file to GCS. This way it can reuse the existing `@Template` parameter and not be as tightly bound to the Yaml Template should we want to to a standalone release.


metadata/src/main/java/com/google/cloud/teleport/metadata/HelperImage.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that marks a helper image for a Dataflow Template. The helper images will use the same
+ * underlying image as the template they are annotating.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface HelperImage {
```
damccorm, line 29: Reading through this PR, I'm -1 on this change at this point in time.

I think that this is probably a premature optimization which is just as likely to cost us time down the line as save it. Specific things I'm worried about:

1) What if we want to release job builder separately from the rest of the templates (which seems reasonable)
2) What if we want job builder to have a different set of dependencies than any other template? I could see this being the case if, for example, we wanted it to contain all ML dependencies (and basically be a fat jar)
3) I don't really see the case for this being generically useful - I think it is unlikely that we'll want to continue to attach other helper images.
4) I already found the container publish flow confusing - this adds an extra layer which makes it harder to understand what we're doing.

I would instead vote for having a separate command (or arg to a top level command) to stage the job server. It could use its own requirements file to install dependencies, it could still use the yaml image (though it doesn't really have to in the future), and it can be run independently.

Basically, I don't think we need to totally couple job builder server release to the yaml template release, and I don't see a ton of advantages to doing so

Polber: Refactored to use existing `@Template` annotation. Let me know what you think

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
```diff
@@ -176,15 +172,12 @@ public Builder setEntryPoint(String[] entryPoint) {
...
-        if (!filesToCopy.get(file).isEmpty()) {
-          files.append(String.format(" %s", String.join(" ", filesToCopy.get(file))));
-        }
-        files.append(" /$WORKDIR/\n");
+    public Builder setFilesToCopy(List<String> filesToCopy) {
+      StringBuilder files = new StringBuilder("COPY");
+      for (String file : filesToCopy) {
+        files.append(String.format(" %s", file));
       }
+      files.append(" $WORKDIR/\n");
```
damccorm, line 180: Why do we need this change? Doesn't this make this function less robust since it will now leave behind `COPY $WORKDIR` if there are no filesToCopy?

damccorm: (to be clear, I think switching to a list input is a good idea, I just don't know why we're changing the core ordering of the logic.

damccorm (dismissed):
Just had minor feedback, otherwise LGTM

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
Polber: I made an assumption the list would never be empty. I added preconditions checks (to this method and others) to reinforce the assumptions.

I also added Java docs to all the methods to make the usage clearer

damccorm: I'm ok with this and am not going to block on it, but I do actually like the original version more; in general it is better to be permissive with your inputs if you can (https://en.wikipedia.org/wiki/Robustness_principle)

damccorm (approved):

PR #1887, opened 2024-09-23
feat: add several types to PGAdapter

thiagotnunes:
Adds several type mappings for a PostgreSQL database.

* bigint
* bigserial
* bit
* bit varying
* bool
* boolean
* bytea
* char
* character
* character varying
* date
* decimal
* double precision
* float4
* float8
* int
* integer
* int2
* int4
* int8
* json
* jsonb
* money
* numeric
* oid
* real
* serial
* serial2
* serial4
* serial8
* smallint
* smallserial
* text
* timestamp
* timestamptz
* timestamp with time zone
* timestamp without time zone
* uuid
* varbit
* varchar


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
```diff
@@ -19,34 +19,216 @@
...
+
   private static final ImmutableMap<String, JdbcValueMapper<?>> SCHEMA_MAPPINGS =
       ImmutableMap.<String, Pair<ResultSetValueExtractor<?>, ResultSetValueMapper<?>>>builder()
           .put("BIGINT", Pair.of(ResultSet::getLong, valuePassThrough))
+          .put("BIGSERIAL", Pair.of(ResultSet::getLong, valuePassThrough))
+          .put("BIT", Pair.of(bytesExtractor, valuePassThrough))
+          .put("BIT VARYING", Pair.of(bytesExtractor, valuePassThrough))
+          .put("BOOL", Pair.of(ResultSet::getBoolean, valuePassThrough))
+          .put("BOOLEAN", Pair.of(ResultSet::getBoolean, valuePassThrough))
+          .put("BOX", Pair.of(ResultSet::getString, valuePassThrough))
```
VardhanThigle, line 174: For MySQL, types that are not supported by SMT for schema migration are marked unsupported to fail fast. 
If we have made a conscious choice to support the Geometric types with strings, It's fine to have this change (for schema migration other things come into consideration, especially around, if the user has an index mounted on these special types, would converting them to a string be ok?)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
```diff
@@ -19,34 +19,216 @@
...
+          .put("BYTEA", Pair.of(bytesExtractor, valuePassThrough))
+          .put("CHAR", Pair.of(ResultSet::getString, valuePassThrough))
+          .put("CHARACTER", Pair.of(ResultSet::getString, valuePassThrough))
           .put("CHARACTER VARYING", Pair.of(ResultSet::getString, valuePassThrough))
-          .put("DATE", Pair.of(utcDateExtractor, sqlDateToAvro))
+          .put("CIDR", Pair.of(ResultSet::getString, valuePassThrough))
+          .put("CIRCLE", Pair.of(ResultSet::getString, valuePassThrough))
+          .put("CITEXT", Pair.of(ResultSet::getString, valuePassThrough))
+          .put("DATE", Pair.of(dateExtractor, dateToAvro))
+          .put("DECIMAL", Pair.of(ResultSet::getObject, numericToAvro))
```
VardhanThigle, line 183: For PG decimal, we don't have a mapping [here](https://cloud.google.com/datastream/docs/unified-types), would it make sense to default to Mysql mapping then?

MySQL mapping is [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/MysqlJdbcValueMappings.java#L202)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
```diff
@@ -19,34 +19,216 @@
...
+        String timestampTz = rs.getString(fieldName);
+        if (timestampTz == null) {
+          return null;
+        }
+        return OffsetDateTime.parse(timestampTz, TIMESTAMPTZ_FORMAT);
+      };
+
+  // Value might be a Double.NaN or a valid BigDecimal
+  private static final ResultSetValueMapper<Number> numericToAvro =
+      (value, schema) -> value.toString();
```
VardhanThigle, line 135: should we consider the precision and scale from the schema argument as required by the [unified-types](https://cloud.google.com/datastream/docs/unified-types) documentation?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: I tried following this https://cloud.google.com/datastream/docs/unified-types#map-psql for type mapping, so you will see that the `PostgreSQLMappingProvider.java` has some types marked as unsupported. Indexes on geometric types using Gist which is not supported by Spanner.

What do you suggest?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: Yes, I would like to do that in a future PR, since this approach works with and without precision for now. I also have to do that for VARCHAR

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: DECIMAL is an alias for NUMERIC.

I cannot do the same mapping as MySQL as PostgreSQL allows for NaN, which is not representable in a BigDecimal.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
```diff
@@ -19,34 +19,216 @@
...
+          .put("SMALLSERIAL", Pair.of(ResultSet::getInt, valuePassThrough))
           .put("TEXT", Pair.of(ResultSet::getString, valuePassThrough))
+          .put("TIME", Pair.of(timeExtractor, timeToAvro))
+          .put("TIMETZ", Pair.of(timetzExtractor, timetzToAvro))
+          .put("TIMESTAMP", Pair.of(timestampExtractor, timestampToAvro))
+          .put("TIMESTAMPTZ", Pair.of(timestamptzExtractor, timestamptzToAvro))
+          .put("TIME WITH TIME ZONE", Pair.of(timetzExtractor, timetzToAvro))
+          .put("TIME WITHOUT TIME ZONE", Pair.of(timeExtractor, timeToAvro))
+          .put("TIMESTAMP WITH TIME ZONE", Pair.of(timestamptzExtractor, timestamptzToAvro))
+          .put("TIMESTAMP WITHOUT TIME ZONE", Pair.of(timestampExtractor, timestampToAvro))
```
VardhanThigle, line 224: [nit and can be taken up later], should we normalize `TIMESTAMP WITHOUT TIME ZONE` to `TIMESTAMP` and similarly for `TIME WITHOUT TIME ZONE` to `TIME` in schema discovery?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
VardhanThigle: Got it.
I think to have parity with live, we should map unsupported types to unsupported.
In which case, the schema will always be a NULL, and so we might not need a `value mapping` for those types until we support them.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: Good point we can reduce the testing then.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: Done

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/rowmapper/provider/PostgreSQLJdbcValueMappings.java
thiagotnunes: Added TODO

VardhanThigle (approved):

PR #1889, opened 2024-09-23
Update live terraform templates to support end to end migration

Deep1998:
This does the following changes -

- Move allow datastream firewall rule to the live template from the earlier setup test-infra template.
- Add a enable_backfill and skip_dataflow flag 
- Update the readme to launch a live+ bulk combined migration


v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/main.tf
```diff
@@ -6,6 +6,23 @@ locals {
...
+  description = "Allow traffic from private connectivity endpoint of Datastream"
+
+  allow {
+    protocol = "tcp"
+    ports    = ["3306"]
+  }
+  source_ranges = [var.datastream_params.private_connectivity.range]
+  target_tags   = ["databases"]
+}
+
```
manitgupta, line 25: 1. For the firewall rule, how does the user configure what targets the rule should apply to? 
Depending on the existing user infra setup - this can mean different things - some may have an IP range to configure, others may have a tag etc. You should provide the user the ability to make this configuration.

2. Adding a firewall rule is a very sensitive operation from a security POV. Customers may not want to have this automated. Just like other existing optional options in the template around security sensitive things (refer to README e.g *adding permissions to service account*), you should also make this optional, so the user can safely skip it.

v2/datastream-to-spanner/terraform/samples/postgresql-end-to-end/outputs.tf
```diff
@@ -8,7 +8,7 @@ output "resource_ids" {
     gcs_bucket                           = google_storage_bucket.datastream_bucket.name
     pubsub_topic                         = google_pubsub_topic.datastream_topic.name
     pubsub_subscription                  = google_pubsub_subscription.datastream_subscription.name
-    dataflow_job                         = google_dataflow_flex_template_job.live_migration_job.job_id
+    dataflow_job                         = var.dataflow_params.skip_dataflow ? "Dataflow job creation skipped" : google_dataflow_flex_template_job.live_migration_job[0].job_id
```
manitgupta, line 11: Nit: Can't we just leave it empty?

v2/datastream-to-spanner/terraform/samples/mysql-sharded-end-to-end/README.md
```diff
@@ -296,6 +303,40 @@ can exclude it from the state file using `terraform state rm` command.
...
+3. **Live Migration (with Dataflow)**
+    - Once the bulk migration is complete, run the live Terraform template again, this time with the following
+      variables:
+        - `skip_dataflow` = false (This enables the creation of the Dataflow job for CDC data)
+        - `enable_backfill` = false (Keep backfill disabled as it was handled in the bulk migration)
+    - This step creates a Dataflow job that migrates the CDC data captured by Datastream.
+
+Once Spanner is up to speed with the source instance, turn off the source instance and wait for Spanner to fully
+catch up. Once caught up, cut over the application to use Spanner.
+
```
manitgupta, line 339: I wonder if we should have this as a dedicated README in `spanner-common` instead of as an FAQ in `datastream-to-spanner` template - seems too important. WDYT? It covers multiple templates, so it makes sense. 

We should have a diagram/flow chart to describe the overall process and leave room for enhancing it in the future via additional sections (completing the migration, validation etc). Looks to me that it deserves its known README. In future, it can be converted to Cloud docs.

In these FAQ, you can maybe explain the specific params and refer to the other README?

v2/datastream-to-spanner/terraform/samples/mysql-sharded-end-to-end/README.md
```diff
@@ -371,7 +415,7 @@ and configures it in the source profile created for the Datastream stream.
 
 > **ALERT:** Private connectivity resource destruction is currently not
 > supported in Terraform due to the ability to delete nested
->resources: [#17920](https://github.com/hashicorp/terraform-provider-google/issues/17290),
+> resources: [#17920](https://github.com/hashicorp/terraform-provider-google/issues/17290),
 > [#13054](https://github.com/hashicorp/terraform-provider-google/issues/13054).
 > Until this is supported, the private connectivity resource will need to be
 > manually deleted via the console or the gcloud CLI before running
```
manitgupta, line 421: Could you remove this alert - this bug is now fixed.

v2/datastream-to-spanner/terraform/samples/mysql-end-to-end/README.md
```diff
@@ -67,27 +70,26 @@ Following permissions are required -
...
 
-> **_Note on IAM:_** 
+> **_Note on IAM:_**
 >
 > 1. For ease of use, this sample automatically adds the
 > required
 > roles to the service account used for running the migration. In order to
-> do this, we need the `resourcemanager.projects.setIamPolicy` permission. If
-> granting
+> do this, we need the `roles/iam.securityAdmin` role. If granting
```
manitgupta, line 78: This section is at a permission revel, you should talk about the specific permission and not the role here.

v2/datastream-to-spanner/terraform/samples/mysql-sharded-end-to-end/README.md
Deep1998: Agreed, this will require a much more detailed guide, likely in cloud docs. I can move it to spanner-common but will someone ever open a spanner-common readme by themselves? Maybe we should have 2 duplicates, both in live and bulk. For the non-wanderers, we will anyways point to proper documentation.

manitgupta: > Agreed, this will require a much more detailed guide, likely in cloud docs.

Agreed, this should be a first-class citizen in cloud docs.

> I can move it to spanner-common but will someone ever open a spanner-common readme by themselves? Maybe we should have 2 duplicates under a major heading in the readme itself, both in live and bulk. For the non-wanderers, we will anyways point to proper documentation.

My opinion on this lies in the first part itself, I do not think customers will self-discover this on their own until this documentation is part of the cloud docs. With that said, I would recommend placing this in a guide of its own (which I think we agree on). Between duplicating it between bulk and live v/s placing in spanner-common, I think the added discoverability is incremental compared to cloud docs, so I would recommend placing in spanner-common, but I am okay with duplication too, as long as it is it's own dedicated `README` file.

We should also add the diagram/flow pieces as well.

manitgupta (approved):

PR #1891, opened 2024-09-24
load test automation

aksharauke:
Current load test is limited by the mysql instance. In future will add logic to have a dedicated mysql instance.


darshan-sj (approved):
LGTM

PR #1893, opened 2024-09-25
feat: maps AVRO BYTES ByteBuffer column

thiagotnunes:
AVRO GenericData uses java.nio.ByteBuffer as a representation of BYTES data (see
https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L1093).

This PR maps a ByteBuffer record into a byte array so it can be written into Spanner.


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java
```diff
@@ -281,6 +282,12 @@ static ByteArray avroFieldToByteArray(Object recordValue, Schema fieldSchema) {
         }
         return ByteArray.copyFrom(Hex.decodeHex(s));
       }
+      if (fieldSchema.getType().equals(Schema.Type.BYTES)) {
+        // For a bytes avro type, expect a java.nio.ByteBuffer.
+        if (recordValue instanceof ByteBuffer) {
+          return ByteArray.copyFrom(((ByteBuffer) recordValue).array());
+        }
+      }
       return ByteArray.copyFrom((byte[]) recordValue);
```
Deep1998, line 291: This was added as a catchall for byte types but currently for mysql none of them are using this flow AFAIK. We should make the default catchall as ByteBuffer if that is the avro standard.

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertorTest.java
Deep1998, line 1: We should just update the bytes_col with byte buffer and remove the byte_buffer_col

thiagotnunes: Done

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java
thiagotnunes: Done, I was worried that some path was using it so I did not default to it.

Deep1998 (dismissed):

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroToValueMapper.java
```diff
@@ -281,12 +282,12 @@ static ByteArray avroFieldToByteArray(Object recordValue, Schema fieldSchema) {
         }
         return ByteArray.copyFrom(Hex.decodeHex(s));
       }
-      return ByteArray.copyFrom((byte[]) recordValue);
+      return ByteArray.copyFrom(((ByteBuffer) recordValue).array());
     } catch (Exception e) {
       throw new AvroTypeConvertorException(
           "Unable to convert "
               + fieldSchema.getType()
-              + " to byte array, with value: "
```
Deep1998, line 289: This should be array since the output type is array and not a buffer.

thiagotnunes: Done

VardhanThigle (approved):

Issue #1807, opened 2024-08-22
[Bug]: Spanner staging tests are breaking for ImportPipelineIT and ExportPipelineIT


Polber:
### Related Template(s)

Cloud_Spanner_to_GCS_Avro, GCS_Avro_to_Cloud_Spanner

### Template Version

HEAD

### What happened?

The ImportPipelineIT and ExportPipelineIT tests have been failing for Spanner Staging cases. There were a series of errors like the following:
```
Caused by: java.lang.IllegalStateException: Column 1 is not of correct type: expected BOOL but was STRING
	com.google.common.base.Preconditions.checkState(Preconditions.java:857)
	com.google.cloud.spanner.AbstractStructReader.checkNonNullOfType(AbstractStructReader.java:616)
	com.google.cloud.spanner.AbstractStructReader.getBoolean(AbstractStructReader.java:155)
	com.google.cloud.teleport.spanner.ddl.InformationSchemaScanner.listPlacements(InformationSchemaScanner.java:1450)
	com.google.cloud.teleport.spanner.ddl.InformationSchemaScanner.scan(InformationSchemaScanner.java:105)
	com.google.cloud.teleport.spanner.ReadInformationSchema$ReadInformationSchemaFn.processElement(ReadInformationSchema.java:96)
```
that seem to have been resolved as of #1792, however the below log output shows the current output.

Since this appears to only occur on staging spanner tests, it may be due to an incompatible change on the staged spanner instance, in which case disabling these tests will be the temporary fix until a patch is released, or the ITs are updated.

### Relevant log output

```shell
org.apache.beam.it.gcp.spanner.SpannerResourceManagerException: Failed to execute statement.
	at org.apache.beam.it.gcp.spanner.SpannerResourceManager.executeDdlStatements(SpannerResourceManager.java:307)
	at org.apache.beam.it.gcp.spanner.SpannerResourceManager.executeDdlStatement(SpannerResourceManager.java:283)
	at com.google.cloud.teleport.spanner.ExportPipelineIT.testSpannerToGCSAvroBase(ExportPipelineIT.java:210)
	at com.google.cloud.teleport.spanner.ExportPipelineIT.testSpannerToGCSAvroStaging(ExportPipelineIT.java:168)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1895, opened 2024-09-25
Disable testSpannerToGCSAvroStaging

Abacn:
test signal long been red could impose risk that other breakages slipped into the main branch, which is just happened

mitigate #1807


Abacn:
R: @VardhanThigle @manitgupta

manitgupta:
What is the gap here, can we pick this up as an oncall item?

@Deep1998 @bharadwaj-aditya

Deep1998:
@darshan-sj this test should work now with this [PR merged](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1902), right?

PR #1898, opened 2024-09-27
Fix readStringFromKafka and readAvroFromKafka

Surabhiad:


v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
```diff
@@ -97,10 +95,8 @@ public static PTransform<PBegin, PCollection<KV<byte[], GenericRecord>>> readAvr
             .withKeyDeserializerAndCoder(
                 ByteArrayDeserializer.class, NullableCoder.of(ByteArrayCoder.of()))
             .withValueDeserializer(new KafkaSchemaDeserializerProvider(avroSchema))
-            .withConsumerConfigUpdates(config);
-    if (sslConfig != null) {
-      kafkaRecords = kafkaRecords.withConsumerFactoryFn(new FileAwareConsumerFactoryFn());
-    }
+            .withConsumerConfigUpdates(config)
+            .withConsumerFactoryFn(new FileAwareConsumerFactoryFn());
```
sjvanrossum, line 99: Would you mind fixing the other inconsistency in `readAvroFromKafka` as well? That method does not apply the effect of setting `--enableCommitOffsets`, see the suggestion below.

A clean up nit: The `enableCommitOffsets` parameter is typed as a boxed primitive (`Boolean`), but none of the method bodies for `KafkaTransform.read*` require any property of that type (e.g. nullness) so the parameter type can be replaced with its primitive counterpart (`boolean`).

```
  /**
   * Configures Kafka consumer that reads Avro GenericRecord.
   *
   * @param bootstrapServers Kafka servers to read from
   * @param topicsList Kafka topics to read from
   * @param config configuration for the Kafka consumer
   * @param enableCommitOffsets applies {@link KafkaIO.Read#commitOffsetsInFinalize()} if set
   * @param avroSchema schema for the Kafka record's value
   * @return PCollection of Kafka Key & Value Pair deserialized in string format
   */
  public static PTransform<PBegin, PCollection<KV<byte[], GenericRecord>>> readAvroFromKafka(
      String bootstrapServers,
      List<String> topicsList,
      Map<String, Object> config,
      Boolean enableCommitOffsets,
      String avroSchema) {
    KafkaIO.Read<byte[], GenericRecord> kafkaRecords =
        KafkaIO.<byte[], GenericRecord>read()
            .withBootstrapServers(bootstrapServers)
            .withTopics(topicsList)
            .withKeyDeserializerAndCoder(
                ByteArrayDeserializer.class, NullableCoder.of(ByteArrayCoder.of()))
            .withValueDeserializer(new KafkaSchemaDeserializerProvider(avroSchema))
            .withConsumerConfigUpdates(config)
            .withConsumerFactoryFn(new FileAwareConsumerFactoryFn());

    if (enableCommitOffsets) {
      kafkaRecords = kafkaRecords.commitOffsetsInFinalize();
    }
    return kafkaRecords.withoutMetadata();
  }
```

Surabhiad: the method `readAvroFromKafka` isn't used in the `KafkaToBigQueryFlex` template, hence I didn't make any changes to the method.

Sent the cleanup on a different [pr](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1901)

an2x (approved):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
an2x: commitOffsetsInFinalize changes have been added to this PR.

v2/kafka-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/KafkaToBigQuery.java
```diff
@@ -483,7 +483,7 @@ public static PipelineResult runAvroPipeline(
             .apply(
                 "ReadFromKafka",
                 KafkaTransform.readAvroFromKafka(
-                    bootstrapServers, topicsList, kafkaConfig, avroSchema, null))
+                    bootstrapServers, topicsList, kafkaConfig, avroSchema, null, false))
```
sjvanrossum: Not sure how much this template is used still, but it would seem to me like this pipeline should provide `KafkaReadOptions.getEnableCommitOffsets()` instead of `false`.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
```diff
@@ -124,7 +128,7 @@ public static KafkaIO.Read<byte[], byte[]> readBytesFromKafka(
             .withConsumerConfigUpdates(config)
             .withConsumerFactoryFn(new FileAwareConsumerFactoryFn());
 
-    if (enableCommitOffsets) {
+    if (enableCommitOffsets != null && enableCommitOffsets) {
```
sjvanrossum, line 128: As mentioned in an earlier comment, it's ok to change the parameter type from `Boolean` to `boolean` since `null` has no differentiated meaning compared to false for these methods. The caller should ensure that the provided argument is non-null. That's the case for all callers in this repository, since `KafkaReadOptions` defines a default value for the pipeline option `enableCommitOffsets` and the caller in the legacy template currently sets this to `false` unconditionally.

sjvanrossum (commented):
I'm not aware of the support cycle for the legacy template or its day-to-day active use, but unconditionally setting `enableCommitOffsets` to `false` for all Kafka reads leaves out a fairly important setting for users that require more accurate tracking of committed offsets to avoid data loss or duplicates. Changing that behavior, if it's still relevant for this template, can be done separate from this PR.

@an2x any thoughts?

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/transforms/KafkaTransform.java
sjvanrossum: Ah, I see #1901 addresses this.

an2x:
> I'm not aware of the support cycle for the legacy template or its day-to-day active use, but unconditionally setting `enableCommitOffsets` to `false` for all Kafka reads leaves out a fairly important setting for users that require more accurate tracking of committed offsets to avoid data loss or duplicates. Changing that behavior, if it's still relevant for this template, can be done separate from this PR.
> 
> @an2x any thoughts?

This is the existing behavior in the legacy template. `KafkaTransform.readAvroFromKafka` used to not enable offset commits at all, so now when this was added as an option, we pass `enableCommitOffsets = false` from the KafkaToBigQuery templates to maintain the same behavior.

While I agree it's not the right thing to do in general, enabling commits now when they have been disabled for a long time would be a breaking change. Some users of the legacy template may depend on the fact that it doesn't commit offsets. I'd err on the side of not making breaking changes. If someone wants to use a better and a more robust template they should migrate to the new one.

PR #1903, opened 2024-09-29
Fixed the broken mysql link

liferoad:
https://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html should be the correct one.


damccorm (approved):
Thanks

PR #1906, opened 2024-09-30
Add retry settings for notification creation in GCSResourceManager

Deep1998:
Spanner ITs often fail with rate limit exceeded or reached notification limit when running multiple ITs in parallel. Adding a retry mechanism for notification creation.


Abacn (approved):
thanks!

PR #1912, opened 2024-10-03
Improve generation of requirements

damccorm:
This collects several changes to the tooling used to automatically generate requirements:

1) It adds different base requirements for python and yaml/x-lang jobs since these tend to require different base dependencies. This just matches what is installed by default in the dockerfile for now.
2) It adds setuptools to all the base requirements, which is needed for the tooling we use in our dockerfiles.
3) It generates requirements for a bunch of different templates which need a requirements file.
4) It switches to python 3.11 for generating dependencies since this is what we use in our dockerfiles
5) It aliases pip install in our dependency generation so that it doesn't get flagged by tooling. This is ok since we're only ever manually running these tools (which is allowed)

I split the manual changes (37 lines of changes) and the generated changes (1000s of lines) into different commits to make reviewing the PR easier. I'd suggest focusing on the first commit with the manual changes since that is driving all the generated changes.

This is a continuation of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1868 to make our pip installs more secure


python/generate_dependencies.sh
```diff
@@ -15,27 +15,32 @@
...
+  echo "Your python3.11 installation does not have a required venv module. See s.apache.org/beam-python-dev-wiki for Python installation tips."
   exit 1
 fi
 
 set -ex
 
 ENV_PATH="$PWD/__build__/python${PY_VERSION/./}_requirements_gen"
 rm -rf "$ENV_PATH" 2>/dev/null || true
-python3 -m venv "$ENV_PATH"
+# These python versions need to be kept in sync with our dockerfile python versions (https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/0ac92513838ca525adb3f616c9e1f65237334d1e/plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java#L46)
```
Polber, line 32: Perhaps a similar comment in the DockerfileGenerator file would help ensure synchronization

python/default_base_python_requirements.txt
```diff
@@ -1 +1,2 @@
 apache-beam[gcp]
```
Polber, line 1: If this is base for all future python templates, perhaps at least including dataframe is a good idea
```suggestion
apache-beam[gcp,dataframe]
```

v1/src/main/python/base_requirements.txt
Polber, line 1: I didn't even realize there was a `v1/src/main/python` - I'm pretty confident we do not use this directory at all. The only file that mentions these files is https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/src/main/java/com/google/cloud/teleport/templates/common/SubprocessTextTransformer.java

Which I'm pretty sure was an attempt at a Python UDF transform years ago that is not actually used anywhere (and shouldn't be used now that we have the working one that leverages PythonCallableSource).

I think we can remove `v1/src/main/python` and `v1/src/main/java/com/google/cloud/teleport/templates/common/SubprocessTextTransformer.java` entirely

python/default_base_yaml_requirements.txt
Polber, line 1: I don't think it hurts to use the same base requirements file for all the templates (using this file as the default and removing default_base_python_requirements.txt).

python/generate_all_dependencies.sh
Polber, line 1: Nit: this works for now, but I think creating a small maven plugin mojo would allow use to parse all templates with `TemplateType.YAML`, `TemplateType.PYTHON` and `TemplateType.XLANG`, and generate all of the requirements files that way (similar to how we generate docs), rather than having to keep this list up-to-date

python/default_base_python_requirements.txt
damccorm: I don't think dataframes are used by most python transforms, so I'd rather not add this unless we find we need it

python/default_base_yaml_requirements.txt
damccorm: I originally had it this way, but split it because then you end up installing a bunch of extras you don't need for the python template. It may be that we don't need these extras for most x-lang images as well, but I'd rather not scope that into this PR if so.

python/generate_all_dependencies.sh
damccorm: I agree, and am interested in doing that later if maintaining this list ends up being burdensome - I'm just trying to keep changes small/self-contained and avoid jumping to optimizations before we need them. We'll need to shell out to a script like this regardless, even in that plugin.

v1/src/main/python/base_requirements.txt
damccorm: SGTM - I tried removing it, we'll see if any checks blow up :)

python/default_base_python_requirements.txt
Polber: Are they not used by most if not all FIleIO's? Just a suggestion, definitely not a blocker

Polber (approved):
Thanks!

python/default_base_python_requirements.txt
damccorm: No, just a couple - https://github.com/apache/beam/blob/707d204595e1b338daffacee2703fa46f038e858/sdks/python/apache_beam/io/textio.py#L1061

PR #1918, opened 2024-10-07
Export resource manager metrics and custom counters for Spanner load tests

darshan-sj:


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/monitoring/MonitoringClient.java
```diff
@@ -410,6 +410,53 @@ public List<Long> listTimeSeriesAsLong(ListTimeSeriesRequest request) {
...
+   * Gets aggregated metric for a given filter string for the given interval. The aggregation will
+   * be applied over the entire interval and only one value will be returned.
+   *
+   * @param projectId the project id
+   * @param filter filter string
+   * @param interval time interval
+   * @param aggregationFunction the aggregation function to apply on the time series for the whole
+   *     interval
+   */
+  public @Nullable Double getAggregatedMetric(
```
manitgupta, line 423: Does this package have unit tests? If yes, can we add for these as well?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
```diff
@@ -394,6 +400,50 @@ private static <T> RetryPolicy<T> retryOnDeadlineExceeded() {
...
   }
 
+  /**
+   * Collects the performance metrics for the pubsub resources.
+   *
+   * @param monitoringClient Monitoring client
+   * @param metrics The pubsub metrics will be populated in this map
+   */
+  public void getMetrics(
+      @NonNull MonitoringClient monitoringClient, @NonNull Map<String, Double> metrics) {
```
manitgupta, line 410: Why should the consumer of a `Pubsub` resource care or know about the `monitoringClient`? 

As a user, I am using a library to manage by pubsub resources and shouldn't have to care about how it does the things that are internal to it. 

It should be initialized internally (you can do a lazy initialization if you want).

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
```diff
@@ -394,6 +400,50 @@ private static <T> RetryPolicy<T> retryOnDeadlineExceeded() {
         .build();
   }
 
+  /**
+   * Collects the performance metrics for the pubsub resources.
+   *
```
manitgupta, line 405: Don't see any UTs updated. Is it possible to unit test any of this? 

We have a `Map` of metrics, it should be possible to assert on it?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/monitoring/MonitoringClient.java
darshan-sj: MonitoringClient doesn't have unit tests. I'll check if I can add.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
darshan-sj: I have added UTs for Pubsub and Spanner RMs.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
darshan-sj: Changed the implementation to accept the monitoring client when initializing the PubsubResourceManager. 

The idea is, the Load test (or integation test) would have an ensemble of RMs which are required to execute the test. If a resourceManager like Pubsub requires another RM like MonitoringClient to perform some functions, then the test can pass in the MonitoringClient to Pubsub RM during its creation.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
```diff
@@ -85,6 +91,8 @@ public final class PubsubResourceManager implements ResourceManager {
 
   private final SchemaServiceClient schemaServiceClient;
 
+  private final MonitoringClient monitoringClient;
```
manitgupta, line 94: Nit: Let's call the variable `monitoringResourceManager` so that the idea of dependent resource manager creation is clearer?

manitgupta (approved):
One minor comment on a variable name.

manitgupta (approved):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
```diff
@@ -394,12 +405,65 @@ private static <T> RetryPolicy<T> retryOnDeadlineExceeded() {
         .build();
   }
 
+  /**
+   * Collects the performance metrics for the pubsub resources.
+   *
+   * @param metrics The pubsub metrics will be populated in this map
+   */
+  public void getMetrics(@NonNull Map<String, Double> metrics) {
```
bharadwaj-aditya, line 413: it is odd to have a get method which populates a map and then returns a void. Can you rename this to collectPubSubMetrics or something similar ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
```diff
@@ -469,6 +480,53 @@ public synchronized void cleanupAll() {
     LOG.info("Manager successfully cleaned up.");
   }
 
+  /**
+   * Collects the performance metrics for the spanner database resource like Average CPU
+   * utilization.
+   *
+   * @param metrics The spanner metrics will be populated in this map
+   */
+  public void getMetrics(@NonNull Map<String, Double> metrics) {
```
bharadwaj-aditya, line 489: same as above. Please rename.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
darshan-sj: Done.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
darshan-sj: Done.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/pubsub/PubsubResourceManager.java
darshan-sj: Monitoring client is different from resource managers in the sense that there is no underlying "resource" it manages. It doesn't have create and cleanup life cycles and it is just like a wrapper around Monitoring API library. That is why it is not called Monitoring Resource manager.

PR #1923, opened 2024-10-09
Organize it module and use system properties for it/jdbc

Polber:


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/LoadTestBase.java
```diff
@@ -215,19 +219,23 @@ private void computeDataflowMetrics(
     // cost info
     LOG.info("Calculating approximate cost for {} under {}", launchInfo.jobId(), project);
     TimeInterval workerTimeInterval = getWorkerTimeInterval(launchInfo);
-    metrics.put(
-        "RunTime",
-        (double)
-            Timestamps.between(workerTimeInterval.getStartTime(), workerTimeInterval.getEndTime())
-                .getSeconds());
+    if (workerTimeInterval.hasStartTime() && workerTimeInterval.hasEndTime()) {
```
damccorm, line 222: Why do we need this change? Is it just copying Beam? Or something else? If we're copying Beam, can we just use Beam's implementation?

damccorm: Same question below

it/splunk/src/main/java/com/google/cloud/teleport/it/splunk/SplunkClientFactory.java
```diff
@@ -0,0 +1,47 @@
...
+  SplunkClientFactory() {}
+
+  /**
+   * Returns an HTTP client that is used to send HTTP messages to a Splunk server instance with HEC.
+   *
+   * @return An HTTP client for sending HTTP messages to Splunk HEC.
+   */
+  CloseableHttpClient getHttpClient() {
+    return HttpClientBuilder.create().build();
+  }
```
damccorm, line 36: Where are these splunk files coming from?

it/common/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
damccorm, line 2: proactive ask - please don't rebase, it will be very helpful to be able to see layered commits in the UI for a PR of this size

it/common/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
damccorm: Ok, having reviewed this it isn't as messy as I expected. I still think splitting the JDBC tests out into their own pr would've been helpful. Splitting further would've still been helpful just to keep volume per review lower (which normally helps my time to review). Generally, my principles here are:

1) Keep PRs as small/atomic as possible
2) (especially this) If you have a large but very simple change, don't layer on unrelated changes (even if they're not very complicated). It makes things much harder to review (and rollback if needed)

If I find myself writing PR descriptions like: `Does X. Also does Y.`, that's often a hint that I should split things up (that was the first thing I noticed here).

ron-gal (approved):
Approved for Bigtable changes (that are mostly moving `TemplateTestBase` around and updating imports

PR #1926, opened 2024-10-10
It

aksharauke:
Corrected the object name prefix to create the PubSub notifications

### DIFF
--- a/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbCustomShardIT.java
+++ b/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbCustomShardIT.java
@@ -112,7 +112,7 @@ public class SpannerToSourceDbCustomShardIT extends SpannerToSourceDbITBase {
             createPubsubResources(
                 getClass().getSimpleName(),
                 pubsubResourceManager,
-                getGcsPath("dlq", gcsResourceManager));
+                getGcsPath("dlq", gcsResourceManager).replace("gs://" + artifactBucketName, ""));
         jobInfo =
             launchDataflowJob(
                 gcsResourceManager,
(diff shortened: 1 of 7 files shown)


damccorm:
It looks like tests are still failing, I think this may be due to a dependency cache issue. @Polber does that look right? If so, what do we need to do to fix this?

aksharauke:
> It looks like tests are still failing, I think this may be due to a dependency cache issue. @Polber does that look right? If so, what do we need to do to fix this?

The problematic integration test is fixed though. The spanner-to-sourcedb has passed
[INFO] spanner-to-sourcedb ................................ SUCCESS [17:56 min]

But yeah, tests in other templates are not passing.

damccorm:
Rerunning checks after clearing the cache

damccorm:
Actually @aksharauke could you try pulling in the latest changes from master - looks like caching was actually just hiding the issue before, but https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1927 should fix the problem generally

aksharauke:
> Actually @aksharauke could you try pulling in the latest changes from master - looks like caching was actually just hiding the issue before, but #1927 should fix the problem generally

This PR does not change the file affected by #1927, so just retriggering the integration tests should help I suppose?

damccorm:
> This PR does not change the file affected by https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1927, so just retriggering the integration tests should help I suppose?

I don't think so - when you rerun IIRC GitHub will still run the rerun at the same merge commit, which won't include the change. So I think we need to pull in the change directly (or push any other change which will cause the merge commit to be recomputed)

aksharauke:
> > This PR does not change the file affected by #1927, so just retriggering the integration tests should help I suppose?
> 
> I don't think so - when you rerun IIRC GitHub will still run the rerun at the same merge commit, which won't include the change. So I think we need to pull in the change directly (or push any other change which will cause the merge commit to be recomputed)

I see, thanks. Done, updated my branch.

shreyakhajanchi (approved):

PR #1927, opened 2024-10-10
upgrade postgres version in JDBCBaseIT

Polber:


damccorm (approved):
Thanks! Optimistically merging to get other checks green (and since this is pretty obviously correct given https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/9519892aed0fab320212113b9f3755e2b5d0a569/pom.xml#L94)

PR #1930, opened 2024-10-10
Increase timeout for DataTypesIT#allTypesTest

Polber:
DataTypesIT#allTypesTest has been flaky lately, causing PR failures. Last couple runs took ~25 minutes, so I increased the timeout to at least temporarily fix the failures.


thiagotnunes (approved):
Thanks for the fix

VardhanThigle (approved):

PR #1931, opened 2024-10-11
Add cutover time to help with flaky IT test due to eventual consistency of CDC propogation

VardhanThigle:
## Overview of Fix
At the outset the flaky test seems like a problem due to cdc propagation delay.
With this PR we are adding a 30 seconds of cut-over time, while this can't eliminate the possibility of flakiness, given the high throughput of migration and low number of test entries in both shards, a 30 seconds cut over time seems to be reasonable wait for all CDCs to propagate without increasing the test's run time by a lot.

## Alternative
Checking for pub/sub to get completely acked might be a lot less flakier. But not doing it at the moment in the view that it's best to not add too much complexity to test code unless absolutely needed (to avoid the need to debug or test  the tests themselves)

### DIFF
--- a/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
+++ b/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
@@ -74,6 +74,7 @@ public class DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT
 
   private static final String SPANNER_DDL_RESOURCE =
       "DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT/spanner-schema.sql";
+  public static final int CUTOVER_MILLIS = 30 * 1000;
 
   private static HashSet<DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT>
       testInstances = new HashSet<>();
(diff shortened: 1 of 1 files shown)


shreyakhajanchi (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
```diff
@@ -247,6 +254,14 @@ public void multiShardMigration() {
             .waitForCondition(createConfig(jobInfo2, Duration.ofMinutes(10)), rowsConditionCheck);
```
Deep1998, line 254: Isnt this wait for condition being used to specify how long to wait till conditions match? We could probably just bump up these timeouts instead of adding a thread.sleep() for consistency. 


However, seems like we are already giving plenty of wait time.

VardhanThigle: That's only for row count match.
The flakyness is due to CDC propagation delay, for which there's no easy check.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerShardedMigrationWithMigrationShardIdColumnIT.java
```diff
@@ -203,7 +204,13 @@ public void multiShardMigration() {
     PipelineOperator.Result result =
         pipelineOperator()
             .waitForCondition(createConfig(jobInfo1, Duration.ofMinutes(8)), conditionCheck);
-
+    // Sleep for cutover time to wait till all CDCs propagate.
+    // This will reduce the chance of flakiness.
+    // A real world customer also has a small cut over time to reach consistency.
+    try {
+      Thread.sleep(CUTOVER_MILLIS);
```
darshan-sj, line 211: This sleep is not required actually. The assertResult below actually checks if all the avro files are uploaded. So, only the second sleep in this PR is required. WDYT?

VardhanThigle: Thanks. I will remove the sleep here.
Could we please clean up the 2 waves of upload into 1 to avoid confusion?

PR #1932, opened 2024-10-11
Remove unused/invalid beam version reference

damccorm:
This is causing staging to fail (but was somehow not caught by presubmits, not really sure how)


Polber (changes requested):
The comment about 2.57 is wrong.  This was assuming a change to Beam's python bootstrap script that was never maded.

This line is needed to trick the bootstrap script into using the python libraries that were already downloaded by the Docker script

damccorm:
Thanks - this is a good catch. We should find a way to eventually clean this up, but definitely the right call to leave it for now

PR #1938, opened 2024-10-14
Validate Beam 2.60.0rc2

Abacn:


Abacn:
YAMLTemplateIT test failed with

```
value of : listArtifacts(...).size() expected : 1 but was : 0 iterable was: []
```

however, confirmed that the Dataflow job succeeded and write data (job id `2024-10-14_14_03_35-9637013313796080446`) likely some test infra need changes to handle RC validation use case after MOSS changes.

Other than that, only PythonUDF test failed which is expected (currently rc validation does not support)

PR #1943, opened 2024-10-16
instructions to cut-back

aksharauke:


v2/spanner-to-sourcedb/README.md
```diff
@@ -315,6 +315,47 @@ Reverse transformation can not be supported for following scenarios out of the b
...
+4. Due to a current Beam bug, ensure that the change stream metadata table is deleted from the metadata database before re-run. This table will be of naming format ```Metadata_<metadata database name>_<uuid>```
+5. The template does not Drain due to change stream constant polling, hence it will need to be forcefully cancelled. However, upon re-run, since the startTimestamp is the same, it would re-read the change streams and there will not be data-loss.
+
+#### What to do when job has been running for duration longer than change stream retention
+If the job has been running for longer than the change stream retention period, then:
+1. Set the startTimestamp to current date - change stream retention period.
+2. To avoid data-loss, ensure that the DataFreshness for the Write to SourceDb stage has not crossed the change stream retention period, meaning it should not cross 7 days - this would be an unlikey event and if alerts are set-up, it should be caught early on.
+3. If there are any reocrds on the retry directory, do the below:
```
manitgupta, line 350: typo

manitgupta (dismissed):

v2/spanner-to-sourcedb/README.md
aksharauke: Thanks Manit, fixed.

shreyakhajanchi (approved):
LGTM

PR #1944, opened 2024-10-16
Added support for vector indexes

atask-g:
- Modified the Index builder to also handle vector indexes. 
- Added tests to validate vector index support.
- Refactored the index code to unify the logic for search and vector indexes.

@darshan-sj


darshan-sj (dismissed):
LGTM

darshan-sj (approved):

PR #1951, opened 2024-10-21
Tests: Adding Reverse Replication Tests

asthamohta:


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbNewTablesIT.java
```diff
@@ -0,0 +1,160 @@
...
+    Mutation m1 =
+        Mutation.newInsertOrUpdateBuilder("Users").set("id").to(1).set("name").to("A").build();
+    spannerResourceManager.write(m1);
+
+    Mutation m2 =
+        Mutation.newInsertOrUpdateBuilder("Users2").set("id").to(2).set("name").to("B").build();
+    spannerResourceManager.write(m2);
+  }
+
+  private void assertRowInMySQL() throws InterruptedException {
```
aksharauke, line 152: Skipped tables get written to 'skip' sub-directory in DLQ directory. Can we add a check for that there exists a file there?

aksharauke (dismissed):
Just one small comment

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbNewTablesIT.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    ResourceManagerUtils.cleanResources(
+        spannerResourceManager,
+        jdbcResourceManager,
+        spannerMetadataResourceManager,
+        gcsResourceManager,
+        pubsubResourceManager);
+  }
+
+  @Test
```
shreyakhajanchi, line 133: Please add a description on what CUJ is being tested

shreyakhajanchi (changes requested):
I would recommend merging these tests into the SpannerToSourceDbIT.class that we already have, we should minimize on creating resources like dataflow job, pub/sub notification, etc since it leads to quota limits and merge tests wherever possible as is currently there in live , following the rule that if job parameter changes then only add a separate IT class for a test.

v2/spanner-to-sourcedb/pom.xml
```diff
@@ -89,6 +89,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+      <dependency>
+        <groupId>io.swagger.core.v3</groupId>
```
aksharauke, line 93: Why do we need swagger dependency?

aksharauke (approved):

v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbIT/spanner-schema.sql
```diff
@@ -1,9 +1,14 @@
...
-    name STRING(25),
+    full_name STRING(25),
     `from` STRING(25)
 ) PRIMARY KEY(id);
 
+CREATE TABLE IF NOT EXISTS Users2 (
+    id INT64 NOT NULL,
+    name STRING(25),
+    ) PRIMARY KEY(id);
+
```
shreyakhajanchi, line 11: where is this table being used? I don't think its added in session file either

shreyakhajanchi (approved):

PR #1952, opened 2024-10-21
Fix for job-builder image release

Polber:
The new Job Builder image will fail during release when trying to write metadata since this metadata file is never generated due to skipping the `gcloud dataflow flex-template build...` step as this is not needed for Job Builder.


damccorm:
Thanks!

PR #1954, opened 2024-10-22
Fix flaky tests

fbiville:
Relax assertions around key and property order when they are specified with the legacy object syntax, since order is not guaranteed.


shunping:
LGTM. The failed tests seem to related to resource issue.

shunping (approved):

PR #1958, opened 2024-10-24
Added the queryTempDataset option for the BQ read options

### LINKED ISSUE #1907: [Feature Request]: Allow specifying withQueryTempDataset for the BQ query
### Related Template(s)

bigquery-to-bigtable

### What feature(s) are you requesting?

BigQueryIO fromQuery tries to create a temporary dataset to materialize query results into. This means the executing service accounts must have the `bigquery.datasets.create` permission. Alternatively, there is a `withQueryTempDataset` method that allows specifying a dataset to use so that the creation permission is not required ([BigQueryIO docs](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.TypedRead.html#withQueryTempDataset-java.lang.String-)). 

It would be great if this template / [BigQueryReadOptions](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ba9af646f5255a58c81f84c3a3dad0642bf24a06/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryConverters.java#L164) supported a queryTempDataset parameter to avoid granting excess permissions to the service account.

liferoad:
Fixes #1907

### DIFF
--- a/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryConverters.java
+++ b/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryConverters.java
@@ -229,6 +229,18 @@ public class BigQueryConverters {
     String getQueryLocation();
 
     void setQueryLocation(String query);
+
+    @TemplateParameter.Text(
+            order = 6,
+            optional = true,
+            description = "BigQuery temporary dataset reference when reading from the query.",
+            helpText =
+                    "With this option, you can set an existing dataset to create the temporary table "
+                        + "to store the results of the query.",
+            example = "temp_dataset")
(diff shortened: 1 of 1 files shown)


liferoad:
```
Caused by: java.util.concurrent.ExecutionException: com.google.cloud.spanner.SpannerException: RESOURCE_EXHAUSTED: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: CPU overload detected on server hosting z114_3042243 for Isolation Key [/AS/u/]. Failing-fast.
retry_delay {
  nanos: 100000000
}
```

Polber (approved):

PR #1961, opened 2024-10-25
Fix node target reference lookup and node match mode

fbiville:
Make node match mode mandatory.

Look up all active node targets when resolving start/end node
target references, not just the active node targets sharing the
relationship target's source.

---

~Note: for the e2e test to pass, https://github.com/neo4j/import-spec/pull/153 needs to be merged and released.~

EDIT: merge & release done


ali-ince (approved):
LGTM

Polber (approved):
LGTM from core templates infra perspective

liferoad (approved):

PR #1962, opened 2024-10-26
Add cutover time to help with flaky IT test due to eventual consistency of CDC propogation

darshan-sj:
**Overview of Fix**
At the outset the flaky test seems like a problem due to cdc propagation delay.
With this PR we are adding a 30 seconds of cut-over time, while this can't eliminate the possibility of flakiness, given the high throughput of migration and low number of test entries, a 30 seconds cut over time seems to be reasonable wait for all CDCs to propagate without increasing the test's run time by a lot.

### DIFF
--- a/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java
+++ b/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java
@@ -176,6 +176,12 @@ public class DataStreamToSpannerDDLIT extends DataStreamToSpannerITBase {
     // Assert Conditions
     assertThatResult(result).meetsConditions();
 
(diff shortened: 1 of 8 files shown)


bharadwaj-aditya:
2 questions here
1. Is 30seconds enough ? The integration tests are still failing after introducing a 30second delay
2. This will introduce an additional runtime of 30 seconds per test which will add up to overall runtime quite a bit. Can we be a little smarter about this ? Maybe put in a poll at a lesser interval or so ?

darshan-sj:
1. 30 seconds delay seems enough for now, since we are already waiting for all the rows to arrive in a condition check and the delay was not introduced in all the tests so, introducing them here.
2. There are 17 tests, 30 seconds in each running parallel in 3 threads shouldn't introduce much delay.

darshan-sj:
This PR is stale - I have created a new PR https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1963 

closing this.

PR #1963, opened 2024-10-27
Fixing Flaky tests

darshan-sj:
**Overview of Fix**
1. Introducing 30 seconds of cut-over time delay for all the tests which have cdc events.
2. Introducing retry in createSpannerDdl in SpannerResourceManager.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
@@ -54,7 +54,6 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import org.apache.beam.it.common.ResourceManager;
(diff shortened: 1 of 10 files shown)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
```diff
@@ -310,11 +309,16 @@ public synchronized void executeDdlStatements(List<String> statements)
 
     LOG.info("Executing DDL statements '{}' on database {}.", statements, databaseId);
     try {
-      databaseAdminClient
-          .updateDatabaseDdl(instanceId, databaseId, statements, /* operationId= */ null)
-          .get();
+      // executeDdlStatments can fail for spanner staging because of failfast.
+      Failsafe.with(retryOnQuotaException())
```
VardhanThigle, line 313: Would this need us to update the beam version/commit id in our templates?

VardhanThigle: Resolving, as discussed, this is not part of the `apache-beam` (though the folder structure is naming it that way)

VardhanThigle (approved):

VardhanThigle (approved):

PR #1972, opened 2024-10-29
Add requirements.txt to job builder server

Polber:
Also, remove wildcard from default `filesToCopy` as requirements.txt (with hashes) is now required for all templates


damccorm (approved):
Thanks!

PR #1978, opened 2024-10-30
Fix for reserved words in schema: Reverse replication

shreyakhajanchi:


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/DMLGenerator.java
```diff
@@ -140,10 +140,10 @@ private static String getUpsertStatement(
     for (Map.Entry<String, String> entry : columnNameValues.entrySet()) {
       String colName = entry.getKey();
       String colValue = entry.getValue();
-      allColumns += colName;
+      allColumns += "`" + colName + "`";
```
aksharauke, line 143: Also add the backticks to the table table. If someone has already modified the session file manually and added backticks, we must filter them.

aksharauke (approved):

PR #1979, opened 2024-10-30
Escaping column names to handle reserved keywords

VardhanThigle:
#Overview
1. Escaping column names to handle reserved keywords #1979
2. Addition of UT for the bigInt fix.

TODO:
Integration test for these changes will be taken up as a follow on.


bharadwaj-aditya (approved):
LGTM

PR #1980, opened 2024-10-30
handling for empty tables with bigint keys

bharadwaj-aditya:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -96,10 +97,12 @@ private static Boundary<java.math.BigInteger> fromBigIntegers(
       throws SQLException {
     Preconditions.checkArgument(partitionColumn.columnClass().equals(BigInteger.class));
     resultSet.next();
+    BigDecimal start = resultSet.getBigDecimal(1);
+    BigDecimal end = resultSet.getBigDecimal(2);
     return Boundary.<java.math.BigInteger>builder()
         .setPartitionColumn(partitionColumn)
-        .setStart(resultSet.getBigDecimal(1).toBigInteger())
```
VardhanThigle, line 101: Sorry for the duplication :)
I had incorporated this diff in [pull/1979](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1979/files)
I am running an end to end test.

We also need to make sure that the range is correctly marked as unsplitable, which is covered in pr#1979

VardhanThigle: Resolving as discussed over chat.

VardhanThigle (approved):

PR #1981, opened 2024-10-30
use Airlock for PyPi packages

Polber:
This PR will force Java PR and Release workflows to use Airlock's PyPi repo instead of the default PyPi repo for python-related templates.


liferoad (dismissed):

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -17,18 +17,24 @@ WORKDIR $WORKDIR
...
     && mv /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
 
 # Cache provider environments for faster startup and expansion time
 RUN python -m apache_beam.yaml.cache_provider_artifacts
 
+<#if saSecretName?? && airlockJavaRepo??>
+COPY cache_provider_jars.py /template
+RUN mkdir -p $HOME/.apache_beam/cache/jars
+RUN python cache_provider_jars.py --cache_dir="/root/.apache_beam/cache/jars/" --maven_central_repository_url="https://_json_key_base64:$(gcloud secrets versions access latest --secret=${saSecretName} | base64 -w 0)@us-maven.pkg.dev/artifact-foundry-prod/${airlockJavaRepo}"
+</#if>
```
damccorm, line 37: What is our goal with this block? Would be good to have a comment explaining what we are doing here

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -61,6 +67,11 @@ COPY --from=python-base /usr/local/lib/libpython$PY_VERSION* /usr/local/lib/
 COPY --from=python-base /bin/dash /bin/sh
 COPY --from=python-base /usr/bin/which.debianutils /usr/bin/which
 
+# Copy cached beam library cache from python-base
+COPY --from=python-base /bin/mkdir /bin/mkdir
+RUN mkdir -p /root/.apache_beam/cache
+COPY --from=python-base /root/.apache_beam/cache /root/.apache_beam/cache
```
damccorm, line 73: Same question - why do we need this?

python/src/main/python/yaml-template/cache_provider_jars.py
```diff
@@ -0,0 +1,57 @@
...
+  for providers in yaml_provider.standard_providers().values():
+    for provider in providers:
+      providers_by_id[id(provider)] = provider
+  for provider in providers_by_id.values():
+    t = time.time()
+    artifacts = provider.cache_artifacts()
+    if artifacts:
+      for artifact in artifacts:
+        if artifact[-4:] == '.jar':
+          command = ['curl', '-L', known_args.maven_central_repository_url + artifact.replace(JavaJarServer.MAVEN_CENTRAL_REPOSITORY, ''), '-o', known_args.cache_dir + str(artifact).split('/')[-1]]
```
damccorm, line 49: Oh, is this why we're doing this? Basically, so we can override how this caching is done? It seems like this might be better suited as a parameter that we can pass into the beam provider function

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
Polber: This copies the cached Beam venv used for cloning new venv's during template launch. Basically, any time a user specifies a Python provider, the YAML framework will spin up a venv to hold Beam and any other packages defined by the provider. This new venv is cloned from the one that was cached into the image. Since the cache was created in the builder step, it needs to be copied over to the actual image.

I can expand the comment

damccorm: Ok, I think this is fine

damccorm (approved):
Thanks, LGTM

PR #1983, opened 2024-10-31
added level wise table processing

bharadwaj-aditya:
All the tables are sorted into "levels" - with the definition of a level being that only after the tables in the level above the current tables level is processed, it is safe to process the current table. The levels are decided based on interleaving and foreign keys similar to the previous implementation.

A repercussion of this is 
1. that all the tables of a given level need to be processed before the next level can be picked up. This may lead to a drop in overall completion time of each table if a few tables at the earlier level take longer, however it is unlikely  to in total job execution time.
2. As there is a limit on the number of graphs that can be used from a dataflow job graph generation perspective, this approach will reduce the number of graphs and improve startup time and allow jobs to process arbitrarily high number of tables.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
@@ -139,8 +139,15 @@ public class PipelineController {
               .map(t -> tableSelector.getSchemaMapper().getSourceTableName("", t))
               .collect(Collectors.toList());
       LOG.info("level: {} source tables: {}", currentLevel, spannerTables);
+      PCollection<Void> previousLevelPCollection = levelVsOutputMap.get((currentLevel - 1) + "");
+      if (currentLevel > 0 && previousLevelPCollection == null) {
+        LOG.warn(
+            "proceeding without waiting for parent. current level: {}  tables: {}",
+            currentLevel,
+            spannerTables);
+      }
       OnSignal<@UnknownKeyFor @Nullable @Initialized Object> waitOnSignal =
-          currentLevel > 0 ? Wait.on(levelVsOutputMap.get(currentLevel - 1)) : null;
+          previousLevelPCollection != null ? Wait.on(previousLevelPCollection) : null;
       JdbcIoWrapper jdbcIoWrapper =
           JdbcIoWrapper.of(configContainer.getJDBCIOWrapperConfig(sourceTables, waitOnSignal));
       if (jdbcIoWrapper.getTableReaders().isEmpty()) {
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
@@ -206,12 +206,12 @@ public class PipelineController {
     return schemaMapper;
   }
 
-  static interface DbConfigContainer {
+  interface DbConfigContainer {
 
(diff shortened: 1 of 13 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
```diff
@@ -220,88 +102,94 @@ static PipelineResult executeShardedMigration(
...
+      List<String> spannerTables = levelToSpannerTableList.get(currentLevel);
+      LOG.info("processing level: {} spanner tables: {}", currentLevel, spannerTables);
+      List<String> sourceTables =
+          spannerTables.stream()
+              .map(t -> tableSelector.getSchemaMapper().getSourceTableName("", t))
+              .collect(Collectors.toList());
+      LOG.info("level: {} source tables: {}", currentLevel, spannerTables);
+      PCollection<Void> previousLevelPCollection = levelVsOutputMap.get(currentLevel - 1);
+      if (currentLevel > 0 && previousLevelPCollection == null) {
+        LOG.warn(
```
Deep1998, line 143: In what scenarios will `previousLevelPCollection` be null despite having parents dependency?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
```diff
@@ -220,88 +102,94 @@ static PipelineResult executeShardedMigration(
...
+          JdbcIoWrapper.of(configContainer.getJDBCIOWrapperConfig(sourceTables, waitOnSignal));
+      if (jdbcIoWrapper.getTableReaders().isEmpty()) {
+        LOG.info("not creating reader as tables are not found at source: {}", sourceTables);
+        // If tables of 1 level are ignored in middle, then the subsequent level will not wait to
+        // begin processing.
+        continue;
+      }
+      ReaderImpl reader = ReaderImpl.of(jdbcIoWrapper);
+      String suffix = generateSuffix(configContainer.getShardId(), currentLevel + "");
+      String shardIdColumn = ""; // TODO- Check if this is needed
```
Deep1998, line 160: We seem to be passing 2 shard id columns now in MigrateTableTransform: shardId and shardIdColumn. THe usage of the latter seems like former should suffice? @VardhanThigle could you verify?

Deep1998 (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
bharadwaj-aditya: If a list of tables does not include parent, then that will come up in the table selector. So i don't think this will come up, but i've added a log just in case some edge case we've not thought of crops up.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
bharadwaj-aditya: I don't think the shardId column is needed here. Hence the TODO - if it is confirmed we can remove it.

VardhanThigle: The `shardIdColumn` is needed for DLQ retrials.

bharadwaj-aditya: As discussed, will take this up as a follow on PR.

PR #1988, opened 2024-11-04
corrected dlq shard id handling

bharadwaj-aditya:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -83,11 +85,14 @@ public PTransform<PCollection<String>, PDone> getDlqTransform() {
...
-      String dlqDirectory, Ddl ddl, String shardIdColumn, SQLDialect sqlDialect) {
+      String dlqDirectory,
+      Ddl ddl,
+      Map<String, String> srcTableToShardIdColumnMap,
+      SQLDialect sqlDialect) {
     this.dlqDirectory = dlqDirectory;
     this.dlqTransform = createDLQTransform(dlqDirectory);
     this.ddl = ddl;
-    this.shardIdColumn = shardIdColumn;
+    this.srcTableToShardIdColumnMap = srcTableToShardIdColumnMap;
```
Deep1998, line 95: nit: Should we create this new map and pass here rather than sticking to our common struct like `schemaMapper`  being pass around everywhere, and extend the mapper to support both source and spanner table names?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/MigrateTableTransform.java
```diff
@@ -51,25 +52,26 @@ public class MigrateTableTransform extends PTransform<PBegin, PCollection<Void>>
...
+      Map<String, String> srcTableToShardIdColumnMap) {
     this.options = options;
     this.spannerConfig = spannerConfig;
     this.ddl = ddl;
     this.schemaMapper = schemaMapper;
     this.reader = reader;
     this.shardId = StringUtils.isEmpty(shardId) ? "" : shardId;
-    this.shardIdColumn = shardIdColumn;
     this.sqlDialect = SQLDialect.valueOf(options.getSourceDbDialect());
+    this.srcTableToShardIdColumnMap = srcTableToShardIdColumnMap;
```
VardhanThigle, line 74: `migrateTableTransform` is a table level.
Do we need to have the map here? Can we just use `ISchemaMapper schemaMapper` which is already available here to map it to the `shardIdColumn` directly and just pass it to the DLQ create?

Deep1998 (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/MigrateTableTransform.java
bharadwaj-aditya: So the input comes across the tables in the current level. So this transform runs for a list of tables and the DLQ needs to identify based on which table the row is coming for.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
bharadwaj-aditya: The stuff passed to the transform needs to be serializable, so a complex object may not work.

VardhanThigle (approved):

PR #1990, opened 2024-11-05
Enable HIDDEN for all column types

atask-g:
- Modified DdlToAvroSchemaConverter to process the HIDDEN column attribute for all column types. This was originally only restricted to generated columns.
- Added unit and integration tests to cover all supported variations.


v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
```diff
@@ -156,6 +156,7 @@ public void tableWithAllTypes() throws Exception {
             + " `arr_proto_field_2`     ARRAY<`com.google.cloud.teleport.spanner.tests.Order`>,"
             + " `arr_nested_enum`       ARRAY<`com.google.cloud.teleport.spanner.tests.Order.PaymentMode`>,"
             + " `arr_enum_field`        ARRAY<`com.google.cloud.teleport.spanner.tests.TestEnum`>,"
+            + " `hidden_column`         STRING(MAX) HIDDEN,"
```
darshan-sj, line 159: Have you tested this DDL in spanner prod? I hope this feature is available in prod.

darshan-sj (dismissed):

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
atask-g: The feature is not new in Spanner, I originally added support for HIDDEN with search indexes and restricted it at the time to only generated columns. This change is extending HIDDEN support to non-generated columns as well as is supported by GoogleSQL SDL. I just retested in prod and found a missing change and also added a test to cover this.

darshan-sj (approved):

PR #1991, opened 2024-11-05
test: test reserved keywords in tables

thiagotnunes:
Adds test for migrating a PostgreSQL table with reserved keywords.


v2/sourcedb-to-spanner/src/test/resources/ReservedKeywordsIT/postgresql-schema.sql
```diff
@@ -0,0 +1,142 @@
+CREATE TABLE reserved_keywords (
+  id INT8 PRIMARY KEY,
```
bharadwaj-aditya, line 2: is it possible to have a reserved keyword in the primary key ?

i would suspect that most queries would use it in the where clause and it might break. Can you add a table with that test case?

bharadwaj-aditya (commented):
One case to add. Rest looks fine.
thanks.

v2/sourcedb-to-spanner/src/test/resources/ReservedKeywordsIT/postgresql-schema.sql
thiagotnunes: Done

bharadwaj-aditya (approved):
LGTM

VardhanThigle (approved):

v2/sourcedb-to-spanner/src/test/resources/ReservedKeywordsIT/postgresql-schema.sql
VardhanThigle: A note: As discussed over chat, we will need columns not named `id` as the first primary key and this might need a bit of deeper dive in the current  IT code to allow such a possibility.
Will take it up in MySQL side of the changes.

PR #1995, opened 2024-11-07
Handle extra columns present on Spanner 

Deep1998:
### Description
This PR supports Spanner having extra columns not present at source. Earlier, an additional column was a TODO, so apart from the shard id, extra cols would cause the transformation to error out. Additional columns on Spanner can be populated via 
1. Custom transformations
2. Autogeneration like generated columns, default columns etc. (These are now omitted from the mutation)
3. Synthetic PK (These are populated with a UUID like the live template)
4. ShardId (This was already handled previously)

### Testing
Unit Tests Added
 
Pending
- [x] Manual migration using extra columns on Spanner.

### DIFF
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/transformer/SourceRowToMutationDoFnTest.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/transformer/SourceRowToMutationDoFnTest.java
@@ -87,6 +87,8 @@ public class SourceRowToMutationDoFnTest {
         .thenReturn(Type.string());
     when(mockIschemaMapper.getSpannerColumns(anyString(), anyString()))
         .thenReturn(List.of("spFirstName", "spLastName"));
+    when(mockIschemaMapper.colExistsAtSource(anyString(), anyString(), anyString()))
+        .thenReturn(true);
 
     PCollection<Mutation> mutations =
         transform(sourceRows, SourceRowToMutationDoFn.create(mockIschemaMapper, null));
@@ -131,7 +133,8 @@ public class SourceRowToMutationDoFnTest {
         .thenReturn(Type.string());
     when(mockIschemaMapper.getSpannerColumns(anyString(), anyString()))
         .thenReturn(List.of("spFirstName", "spLastName"));
-
+    when(mockIschemaMapper.colExistsAtSource(anyString(), anyString(), anyString()))
+        .thenReturn(true);
     // Create custom transformer.
     Map<String, Object> spannerRecord = Map.of("age", 10);
     MigrationTransformationResponse migrationTransformationResponse =
@@ -179,7 +182,7 @@ public class SourceRowToMutationDoFnTest {
   }
 
   @Test
-  public void testCustomTranformationFilteredEvents() throws InvalidTransformationException {
+  public void testCustomTransformationFilteredEvents() throws InvalidTransformationException {
     final String testTable = "srcTable";
     var schemaRef = SchemaTestUtils.generateSchemaReference("public", "mydb");
     var schema = SchemaTestUtils.generateTestTableSchema(testTable);


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
```diff
@@ -95,21 +98,46 @@ public GenericRecordTypeConvertor(
...
+        }
         // If current column is migration shard id, populate value.
         if (spannerColName.equals(shardIdCol)) {
           result = populateShardId(result, shardIdCol);
           continue;
         }
+
+        // For session based mapper, populate synthetic primary key with UUID. For identity mapper,
+        // the schemaMapper returns null.
+        if (spannerColName.equals(
```
VardhanThigle, line 121: note, I might be missing entire context here, so please feel free to resolve this in case that's not the case.
if `schemaMapper.getSyntheticPrimaryKeyColName`  could return null, we might need to use [Object.equal](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Objects.html#equal-java.lang.Object-java.lang.Object-) to avoid NPA.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
```diff
@@ -95,21 +98,49 @@ public GenericRecordTypeConvertor(
   public Map<String, Value> transformChangeEvent(GenericRecord record, String srcTableName)
       throws InvalidTransformationException {
     Map<String, Value> result = new HashMap<>();
+    result = populateCustomTransformations(result, record, srcTableName);
+    // If the row needs to be filtered.
+    if (result == null) {
```
bharadwaj-aditya, line 103: please add a log here.

Deep1998: ADded debug log.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
```diff
@@ -136,10 +167,30 @@ public Map<String, Value> transformChangeEvent(GenericRecord record, String srcT
...
     return result;
   }
 
+  private String getUUID() {
+    return UUID.randomUUID().toString();
+  }
+
+  // Returns true if the column exists at source. For identity mapper, it checks if the column
+  // exists in the GenericRecord.
+  private boolean colExistsAtSource(
```
bharadwaj-aditya, line 179: this seems like a leaky abstraction. We should not have switch cases based on types of mappers here.

Can we add a method in the schema mapper like hasSourceColumn or so skip if not present ?

bharadwaj-aditya (commented):
Can you also check if this works for the tokenList use case ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
Deep1998: That was my original intent but we cannot do that as the IdentityMapper is not aware of any source specific schema details. For this, we rely on the generic record to give us the source column names which is not ideal.

In the long term, once we develop capabilities to read the source schema in the job itself, we can use that in the identity mapper and refactor this.

VardhanThigle (approved):

Deep1998:
TOKENLIST is a new datatype in Spanner that is not handled here. That will be taken up in a follow up PR.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
Deep1998: Added a comment in the code.

bharadwaj-aditya: I wouldn't completely agree on this. 

I think we should do both these checks for all mappers in any case.
Get the sourceColumn, in the identity mapper it would just return the same name.

And then you check if the record has the sourceColumn - regardless of the mapper.

VardhanThigle (approved):

bharadwaj-aditya (approved):
LGTM

PR #1996, opened 2024-11-07
feat: treat empty strings as null values from csv imports

Emrehzl94:
This PR enables empty strings in CSV files to be treated as null values during import and adds an integration test to verify this behavior.


ali-ince (approved):
LGTM

PR #1997, opened 2024-11-07
Add TOKENLIST spanner type

Deep1998:
TOKENLIST is a spanner type used only for Full-Text Search and is not supported in PG Spanner. Added the type so the ddl object can parse it.

Testing:
Tested migration on spanner table with additional FTS column, additional generated column, default column and customer transformation column.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
@@ -94,8 +94,7 @@ public interface ISchemaMapper extends Serializable {
   String getSyntheticPrimaryKeyColName(String namespace, String spannerTableName);
 
   /**
-   * Returns true if the Spanner column exists at the source. For identity mappers, since there is
-   * no source information, it always returns true.
+   * Returns true if a corresponding source column exists for the provided Spanner column.
    *
    * @param namespace is currently not operational.
    */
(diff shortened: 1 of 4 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/schema/ISchemaMapper.java
```diff
@@ -94,7 +94,8 @@ List<String> getSpannerColumns(String namespace, String spannerTable)
   String getSyntheticPrimaryKeyColName(String namespace, String spannerTableName);
 
   /**
-   * Returns true if the Spanner column exists at the source.
+   * Returns true if the Spanner column exists at the source. For identity mappers, since there is
```
bharadwaj-aditya, line 97: nit: this is specific to the identity mapper, we probably should not put it here.

bharadwaj-aditya (approved):
LGTM - but can we figure out a way of not having to add to DDL everytime something is added to Spanner ?

Deep1998:
> LGTM - but can we figure out a way of not having to add to DDL everytime something is added to Spanner ?

Hmm that would be an interesting problem. For new datatypes added, making it agnostic likely wont be possible as the templates will require updates to handle them.

For types like these, we can maybe update the ddl class to ignore hidden columns and not construct their DDL,  extendible to future hidden types, but how often do we really get a new hidden type?

VardhanThigle (dismissed):
If easily possible, please add todos for UTs.

bharadwaj-aditya:
> > LGTM - but can we figure out a way of not having to add to DDL everytime something is added to Spanner ?
> 
> Hmm that would be an interesting problem. For new datatypes added, making it agnostic likely wont be possible as the templates will require updates to handle them.
> 
> For types like these, we can maybe update the ddl class to ignore hidden columns and not construct their DDL, extendible to future hidden types, but how often do we really get a new hidden type?

It doesnt happen often. But it is good to not have stuff break when it does. Because it is a rare change, the possibility of someone missing it is even more.

Deep1998:
Correction to what I said earlier, simply ignoring HIDDEN columns also wont work since they can be for enabled on non-tokelist columns as well. I dont see a good way to handle this currently.

VardhanThigle (approved):

PR #1999, opened 2024-11-08
fix: add yaml support for specification parser

Emrehzl94:
This PR updates the job specification parser to support both JSON and YAML formats. If the JSON parsing fails, the content is treated as YAML and converted to a JSON object for consistent handling


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapper.java
```diff
@@ -38,14 +39,25 @@ public class JobSpecMapper {
...
 
   public static ImportSpecification parse(String jobSpecUri, OptionsParams options) {
-    var json = fetchContent(jobSpecUri);
-    var spec = new JSONObject(json);
+    String content = fetchContent(jobSpecUri);
+    JSONObject spec;
+
+    try {
+      spec = new JSONObject(content);
+    } catch (Exception jsonException) {
```
venikkin, line 47: issue(minor): If the user specify malformed json, they won't be able to see the parser error. Instead, they will receive an error about malformed yaml. However, I don't see how this could be avoided without explicit configuration option, which is probably excessive.

venikkin (approved):

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapper.java
Emrehzl94: yes, you're right and it seems only way to avoid is also catch yaml exception and show both errors for json and yaml, I mean something like this; 

```
    try {
      spec = new JSONObject(content);
    } catch (Exception jsonException) {
      Yaml yaml = new Yaml();
      try {
        Map<String, Object> yamlMap = yaml.load(content);
        spec = new JSONObject(yamlMap);

      } catch (Exception yamlException) {
        throw new IllegalArgumentException("Parsing failed: content is neither valid JSON nor valid YAML." +
                "\nJSON parse error: " + jsonException.getMessage() +
                "\nYAML parse error: " + yamlException.getMessage(), yamlException);
      }
    }
```

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapper.java
```diff
@@ -38,19 +39,45 @@ public class JobSpecMapper {
...
     } catch (SpecificationException e) {
       throw validationFailure(e);
     }
   }
 
+  private static JSONObject getJsonObject(String content) {
+    JSONObject spec;
+
+    try {
+      spec = new JSONObject(content);
```
fbiville, line 61: nit: I'd rather use `return` directly and in the `catch` block than rely on a variable

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapper.java
```diff
@@ -38,19 +39,45 @@ public class JobSpecMapper {
...
+    } catch (Exception jsonException) {
```
fbiville, line 62: issue: you should use `org.json.JSONException` instead

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapper.java
```diff
@@ -38,19 +39,45 @@ public class JobSpecMapper {
...
+      Yaml yaml = new Yaml();
+      try {
+        Map<String, Object> yamlMap = yaml.load(content);
+        spec = new JSONObject(yamlMap);
+
+      } catch (Exception yamlException) {
```
fbiville, line 68: issue: same with `org.yaml.snakeyaml.error.YAMLException`

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/database/NewSpecCypherGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Parameterized.Parameter public String fileFormat;
+
+  @Parameterized.Parameters(name = "{0}")
+  public static List<String> testParameters() {
+    return Arrays.asList("json", "yaml");
+  }
+
+  @Test
+  public void merges_nodes_creates_relationship() {
```
fbiville, line 43: issue: I'm not sure I follow why there is a new test here?
I'd expect new tests in `JobSpecMapperTest` covering cases like:

 - valid JSON, legacy spec
 - valid JSON, import spec
 - valid Yaml, import spec
 - invalid JSON and Yaml
 - valid JSON, wrong format (neither legacy nor import spec)
 - valid Yaml, wrong format

fbiville (changes requested):

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/database/NewSpecCypherGeneratorTest.java
Emrehzl94: we don't have explicitly JobSpecMapperTest, you mean I should add this tests for CypherGeneratorTest or create a new test file for this cases?

fbiville: let's create `JobSpecMapperTest` for this

Emrehzl94: already did, commit is gonna be pushed soon!

v2/googlecloud-to-neo4j/src/test/resources/testing-specs/job-spec-mapper-test/invalid-json.json
```diff
@@ -0,0 +1,103 @@
+{
+    "version": "1,
```
fbiville, line 2: nit: you could actually stop there and make the document much shorter :)

v2/googlecloud-to-neo4j/src/test/java/com/google/cloud/teleport/v2/neo4j/model/helpers/JobSpecMapperTest.java
```diff
@@ -0,0 +1,305 @@
...
+import com.google.cloud.teleport.v2.neo4j.model.job.OptionsParams;
+import java.util.List;
+import org.junit.Test;
+
+public class JobSpecMapperTest {
+
+  private static final String SPEC_PATH = "src/test/resources/testing-specs/job-spec-mapper-test";
+
+  @Test
+  public void valid_json_legacy_spec() {
```
fbiville, line 30:
```suggestion
  public void parses_valid_json_legacy_spec() {
```

fbiville: applies to other tests too

v2/googlecloud-to-neo4j/src/test/resources/testing-specs/job-spec-mapper-test/invalid-yaml.yaml
```diff
@@ -0,0 +1,61 @@
+version: "1
```
fbiville, line 1: same as above

liferoad (approved):

PR #2001, opened 2024-11-11
tests: Adding Forward Migration Tests

asthamohta:
Tests added:
1. Sequence
2. Dropped table
3. Mixed IT: 1. Foreign Keys. 2. Table Dropped. 3. Column Rename. 4. DLQ Retry. 5. Missing PK 6. Indexes


manitgupta (commented):
What is the `EndToEndIT` supposed to do and how is different from all the other tests that are already present?

asthamohta:
The End-to-End test checks for multiple conditions at the same time in the same table unlike other tests. It checks for, for examples FK,Column Rename+,tables dropped,Indexes etc, all on the same table

manitgupta:
` multiple conditions at the same time in the same table unlike other tests. It checks for, for examples FK,Column Rename+,tables dropped,Indexes etc, all on the same table`

How does that add any additional value v/s checking these operations on a different sets of tables? 

It's not clear to me what is the definition of end-to-end..

Can we define this please?

darshan-sj:
In our existing ITs we test for column rename and DLQ handling for FK/Interleaving separately. We had a bug in our template where the DLQ handling was not happening properly when the columns are renamed, and it dint get caught through ITs.
So, we wanted to add one IT which covers all the cases simultaneously.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerEndToEndIT.java
```diff
@@ -0,0 +1,192 @@
...
+import org.junit.runners.JUnit4;
+
+/**
+ * An integration test for {@link DataStreamToSpanner} Flex template which tests use-cases where a
+ * session file is required.
+ */
+@Category({TemplateIntegrationTest.class, SkipDirectRunnerTest.class})
+@TemplateIntegrationTest(DataStreamToSpanner.class)
+@RunWith(JUnit4.class)
+public class DataStreamToSpannerEndToEndIT extends DataStreamToSpannerITBase {
```
darshan-sj, line 51: We should not call this EndToEnd it. Maybe call it DLQ retry test along with all other cases.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java
```diff
@@ -411,6 +412,31 @@ public void migrationTestWithCharsetConversion() {
...
+  @Test
+  public void migrationTestWithSequenceColumns() {
+    // Construct a ChainedConditionCheck with 2 stages.
+    // 1. Send initial wave of events
+    // 2. Wait on Spanner to have events
+    ChainedConditionCheck conditionCheck =
+        ChainedConditionCheck.builder(
+                List.of(
+                    uploadDataStreamFile(
+                        jobInfo, TABLE8, "sequence.avro", "DataStreamToSpannerDDLIT/Singers.avro"),
```
darshan-sj, line 424: please create a sql file in corresponding resource directory with SQL statements used to create this avro file.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerEndToEndIT.java
```diff
@@ -0,0 +1,192 @@
...
+  /** Cleanup dataflow job and all the resources and resource managers. */
+  @AfterClass
+  public static void cleanUp() throws IOException {
+    for (DataStreamToSpannerEndToEndIT instance : testInstances) {
+      instance.tearDownBase();
+    }
+    ResourceManagerUtils.cleanResources(spannerResourceManager, pubsubResourceManager);
+  }
+
+  @Test
```
darshan-sj, line 106: Add a comment here detailing all test cases this test covers simultaneously.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerEndToEndIT.java
```diff
@@ -0,0 +1,192 @@
...
+  public void endToEndMigrationTest() {
+    // Construct a ChainedConditionCheck with 2 stages.
+    // 1. Send initial wave of events
+    // 2. Wait on Spanner to have events
+    ChainedConditionCheck conditionCheck =
```
darshan-sj, line 111: We want to simulate DLQ retry processing by inserting child row first and then insert corresponding parent later, hence the mysql-schema should not contain FK relationship between Authors and Books.

The test case should look like this:
1. upload Books avro file containing few rows referring author 1 and few rows referring author 2.
2. upload genre avro file
3. upload author avro file containing author 1.
4. check for Author table to contain 1 row. This ensures that current wave of avro files are almost processed by the template.

Add a small sleep after Line No. 135

1. Then upload avro file containing Author number 2 (and others)
2. Wait for all rows to appear in author table, books table and genre table.
3. Then assert table contents.

Maybe this is what you are doing, it will be clear if you add the SQL statements used to create the avro files in the resources folder. Also, add comments here in this test to help others understand the test case.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerEndToEndIT.java
```diff
@@ -0,0 +1,192 @@
...
+            .build();
+
+    result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(8)), conditionCheck);
+
+    // Assert Conditions
+    assertThatResult(result).meetsConditions();
+
+    assertAuthorsTableContents();
```
darshan-sj, line 162: You should also assert contents of Books table.

v2/datastream-to-spanner/src/test/resources/DataStreamToSpannerEndToEndIT/mysql-schema.sql
```diff
@@ -0,0 +1,19 @@
...
+    `name` varchar(25),             -- To: full_name STRING(25) Column name renamed
+    `last_update` timestamp,        -- To: Column dropped in spanner
+    PRIMARY KEY (`author_id`)
+);
+
+CREATE TABLE `Books` (
+    `id` int NOT NULL,
+    `title` varchar(200),
+    `author_id` int,
+    FOREIGN KEY (author_id) REFERENCES Author(author_id)
```
darshan-sj, line 12: As mentioned earlier, mysql should not contain the FK constraint.

v2/datastream-to-spanner/src/test/resources/DataStreamToSpannerIT/mysql-session.json
```diff
@@ -415,6 +415,129 @@
       "ForeignKeys": null,
       "Indexes": null,
       "Id": "t2"
+    },
```
darshan-sj, line 418: As discussed, move these changes to DataStreamToSpannerSessionIT and add a test case method there (droppedTableTest).

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java
```diff
@@ -411,6 +412,31 @@ public void migrationTestWithCharsetConversion() {
...
+                        .build()))
+            .build();
+
+    // Wait for conditions
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(8)), conditionCheck);
+
+    // Assert Conditions
+    assertThatResult(result).meetsConditions();
```
darshan-sj, line 437: We don't need any asserts here is it? If the autogenerated sequence is predictable, we should assert the contents of the spanner table here.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerEndToEndIT.java
asthamohta: Hi Darshan. So this is what I have done. Uploaded child rows(Books), one parent row(Authors) on which none of the child rows depends and the genre table. Now since only one of these will get migrated, we check if 1 row is migrated to Authors. In this duration the Child rows(Books) have gone to dlq for retry. Then we upload the three parents rows(Author) on which the previous 3 child rows depend. We now check that both child(books) and parents(author) rows are migrated

asthamohta: I will add the SQLs

v2/datastream-to-spanner/src/test/resources/DataStreamToSpannerEndToEndIT/mysql-schema.sql
asthamohta: Even if MySql has the FK constraint and we migrate with child rows migrated first the dlq does get created(I checked manually as well, it does). i don't understand why parent should not have FK constraint. What difference does this make?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java
asthamohta: We can't do that because everytime a different autogenerated key might get generated.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerMixedIT.java
```diff
@@ -0,0 +1,218 @@
...
+            .build();
+
+    // Wait for conditions
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(8)), conditionCheck);
+
+    // Assert Conditions
+    assertThatResult(result).meetsConditions();
+
```
darshan-sj, line 137: We should add a sleep here in order to make sure the child rows go to DLQ.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerMixedIT.java
```diff
@@ -0,0 +1,218 @@
...
+                        "authors_2.avro",
+                        "DataStreamToSpannerMixedIT/Authors_2.avro"),
+                    SpannerRowsCheck.builder(spannerResourceManager, TABLE1)
+                        .setMinRows(4)
+                        .setMaxRows(4)
+                        .build(),
+                    SpannerRowsCheck.builder(spannerResourceManager, TABLE2)
+                        .setMinRows(3)
+                        .setMaxRows(3)
+                        .build()))
```
darshan-sj, line 153: We should also assert Genre table?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerMixedIT.java
```diff
@@ -0,0 +1,218 @@
...
+
+  private void assertBooksTableContents() {
+    List<Map<String, Object>> bookEvents = new ArrayList<>();
+
+    Map<String, Object> bookRow1 = new HashMap<>();
+    bookRow1.put("id", 1);
+    bookRow1.put("title", "Pride and Prejudice");
+
+    Map<String, Object> bookRow2 = new HashMap<>();
+    bookRow2.put("id", 1);
```
darshan-sj, line 204: typo? asserting the same row 3 times.

darshan-sj (approved):
LGTM

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerMixedIT.java
asthamohta: We can't as that is a dropped table

darshan-sj (approved):
LGTM

PR #2010, opened 2024-11-14
Add IT to verify extra cols

manitgupta:
Add an IT to verify that live migration works if the corresponding Spanner table has some extra columns which do not exist in the source tables.

### DIFF
--- a/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerSessionIT.java
+++ b/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerSessionIT.java
@@ -171,7 +171,7 @@ public class DataStreamToSpannerSessionIT extends DataStreamToSpannerITBase {
   }
 
   @Test
-  public void migrationTestWithSyntheticPK() {
+  public void migrationTestWithSyntheticPKAndExtraColumn() {
     // Construct a ChainedConditionCheck with 2 stages.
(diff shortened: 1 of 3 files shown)


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerSessionIT.java
```diff
@@ -171,7 +171,7 @@ public void migrationTestWithRenameAndDropColumn() {
   }
 
   @Test
-  public void migrationTestWithSyntheticPK() {
+  public void migrationTestWithSyntheticPKAndExtraColumn() {
     // Construct a ChainedConditionCheck with 2 stages.
     // 1. Send initial wave of events
     // 2. Wait on Spanner to have events
```
darshan-sj, line 177: We should also assert the contents of extra column in assertBooksBackfillContents method.

darshan-sj (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerSessionIT.java
manitgupta: As discussed offline, the `extraCol1` is null and asserting it seems to break `hasRecordsUnorderedCaseInsensitiveColumns` method. Going ahead with the current approach.

PR #2013, opened 2024-11-14
Favor built-in transform over custom one

fbiville:
This replaces our custom KV transform, since Beam already includes one.


ali-ince (approved):
LGTM

liferoad (approved):

PR #2016, opened 2024-11-15
Add DLQ support to JdbcToBigQuery using BQ Storage Write API

Polber:
This PR adds proper error handling to the JdbcToBigQueryFlex template when using Storage Write API.

Before this change, the WriteToBigQuery step would succeed even when there were write failures. These errors were never captured or handled, thus resulting in dropped rows. This PR will now fail the pipeline on write errors by default when using Storage Write API. If a deadletter table is specified, then errors will instead be written the the DLQ table (created if needed).

### DIFF
--- a/v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java
+++ b/v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java
@@ -15,28 +15,45 @@
  */
 package com.google.cloud.teleport.v2.templates;
 
+import static com.google.cloud.teleport.v2.transforms.BigQueryConverters.wrapBigQueryInsertError;
 import static com.google.cloud.teleport.v2.utils.GCSUtils.getGcsFileAsString;
 import static com.google.cloud.teleport.v2.utils.KMSUtils.maybeDecrypt;
 
+import com.google.api.services.bigquery.model.ErrorProto;
 import com.google.api.services.bigquery.model.TableRow;
 import com.google.cloud.teleport.metadata.Template;
 import com.google.cloud.teleport.metadata.TemplateCategory;
+import com.google.cloud.teleport.v2.coders.FailsafeElementCoder;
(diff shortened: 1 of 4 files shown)


damccorm (dismissed):
Thanks

damccorm (approved):

v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java
```diff
@@ -175,12 +202,71 @@ static PipelineResult run(JdbcToBigQueryOptions options, Write<TableRow> writeTo
...
+                .setErrorRecordsTable(options.getOutputDeadletterTable())
+                .setErrorRecordsTableSchema(ResourceUtils.getDeadletterTableSchemaJson())
+                .setUseWindowedTimestamp(false)
+                .build());
+      } else {
+        /*
+         * Step 3b.
+         * Fail pipeline upon write errors if no DLQ was specified
+         */
+        insertErrors.apply(ParDo.of(new ThrowWriteErrorsDoFn()));
```
Abacn, line 246: iiuc now error will be thrown if there is a failed write record and dlq table is not specified (the default), which means some pipeline succeeded before now would fail. This is a breaking change. However I understand previous behavior (silently drop record) was bad either. Shall we call this out somewhere, while still give user an option preserve the old behavior if they notices breakage, that is to drop bad records like the previous behavior (e.g. set deadletter table to a magic name like "ignored"

damccorm: I don't think we should give an option to ignore this - it is silent data loss, so I think we should force users to apply it. Agreed on calling it out somewhere - the PR description will get auto-added to the release notes, so maybe we could make that more descriptive? Alternately, we could coordinate with the next release manager to make sure it gets mentioned explicitly

Abacn: Thanks, the current form sounds good to me

Abacn (approved):

PR #2019, opened 2024-11-18
Cassandra Schema Discovery

VardhanThigle:
# Implementing SchemaDiscovery Interface for Cassandra Source.
## This includes
1. [X] Discovering tables on a given Cassandra Instance
2. [X] Discovering schema of tables on a given Cassandra Instance.
## Will be implemented in later PRs (not included in current PR):
1. [ ] Avro Mapping for primitive types
2. [ ] End to End flow with pipeline options for migrating primitive types
3. [ ] Integration tests (once end to end flow is in place)
4. [ ] Support for advanced types.

### DIFF
--- a/v2/sourcedb-to-spanner/pom.xml
+++ b/v2/sourcedb-to-spanner/pom.xml
@@ -29,6 +29,7 @@
   <artifactId>sourcedb-to-spanner</artifactId>
 
   <dependencies>
+
     <dependency>
       <groupId>com.google.cloud.teleport.v2</groupId>
       <artifactId>common</artifactId>
@@ -68,6 +69,13 @@
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
       <version>${postgresql.version}</version>
+    </dependency>
+
+    <!-- https://mvnrepository.com/artifact/com.datastax.oss/java-driver-core -->
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <version>4.17.0</version>
     </dependency>
 
     <!-- Test Dependencies -->
@@ -128,5 +136,11 @@
       <version>1.0-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
+      <dependency>
+        <groupId>com.github.nosan</groupId>
+        <artifactId>embedded-cassandra</artifactId>
+        <version>5.0.0</version>
+        <scope>test</scope>
+      </dependency>
   </dependencies>
 </project>
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraConnector.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraConnector.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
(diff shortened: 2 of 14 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraConnector.java
```diff
@@ -0,0 +1,96 @@
...
+ * Optionally allow driver config from GCS file.
+ */
+public final class CassandraConnector implements AutoCloseable {
+  private final CqlSession session;
+  private static final Logger LOG = LoggerFactory.getLogger(CassandraConnector.class);
+
+  public CassandraConnector(
+      CassandraDataSource dataSource, CassandraSchemaReference schemaReference) {
+    Log.info(
+        "Connecting to Cassandra Source dataSource = {}, schemaReference = {}",
```
Deep1998, line 44: Just verifying, but this prints the actual object and not the pointer right?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraDataSource.java
```diff
@@ -15,40 +15,77 @@
...
  */
 @AutoValue
 public abstract class CassandraDataSource implements Serializable {
 
   /** Name of the Cassandra Cluster. */
   public abstract String clusterName();
 
+  /** Name of local Datacenter. Must be specified if contactPoints are not empty */
+  @Nullable
+  public abstract String localDataCenter();
```
Deep1998, line 41: Don't we also need to add auth details here? Like username password etc?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/testutils/BasicTestSchema.java
Deep1998, line 1: Should this schema related file be in utils? Since different tests could use different schemas, it should reside besides(or within) the unit test file itself?

Basically the unit test decides what cql to use what schema to create?

VardhanThigle: Different tests can sure use different schemas. But given that we are spawnning embedded cassandra, it will be great if they share the same cluster.
The basicTest and related cqlsh are meant to servive for most of the tests (with minor additions as needed).

As we forsee that many uts will share the basic_schema.cqlsh, we have a utility class that provides paths needed to spwan that schema in the UT.

The test utils in any way don't limit from other tests using their own cluster if they have a very different use case.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraConnector.java
VardhanThigle: That was the case for `java.sqlx.DataSOurce` as it was a very opaque object.  `CassandraDataSource` is built using `AutoValue` so it comes by default for string pretty printing.

The jdbc issue got fixed when we introduced an extension to basic java.sqlx.DataSource for Jdbc.

v2/sourcedb-to-spanner/pom.xml
```diff
@@ -128,5 +136,11 @@
       <version>1.0-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
+      <dependency>
+        <groupId>com.github.nosan</groupId>
+        <artifactId>embedded-cassandra</artifactId>
+        <version>5.0.0</version>
```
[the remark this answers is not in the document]
VardhanThigle: Thanks.
This is the test  embeded cassandra framework dependency.
I got some issues in using 4.0 test dependency (mostly around the it's compatibility with non-JVM 8,  JVM versions).  
Regardless of the embedded cassandra version, embedded cassandra let's us set the version for running the actual cassandra process which is used by UT.
I tried to play with that, but setting that to 4.0 is needed some tweaks to work with our UT JVM.
I have added a TODO for the same.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/iowrapper/CassandraDataSource.java
VardhanThigle: Thanks. Added in the current push.

PR #2020, opened 2024-11-18
Parameterizing spanner Host in spanner tests

darshan-sj:
- If "spannerHost" parameter is passed then the spanner tests will be run only on that spannerHost. Else, all of them will be run twice: once with "staging" as spanner host and once with "default" as spanner host. The test names will appear as testSpannerToGCSAvroBase[staging] and testSpannerToGCSAvroBase[default]


v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
```diff
@@ -44,14 +46,35 @@
...
+  private String spannerHost;
+  private String spannerHostName;
+
+  public ExportPipelineIT(String spannerHost, String spannerHostName) {
+    this.spannerHost = spannerHost;
+    this.spannerHostName = spannerHostName;
+  }
+
+  @Parameters(name = "{1}")
+  public static Collection primeNumbers() {
```
manitgupta, line 67: Why is this method called `primeNumbers` o.O?

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
```diff
@@ -44,14 +46,35 @@
...
 import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
 
 /** Integration test for {@link ExportPipeline Spanner to GCS Avro} template. */
 @Category(TemplateIntegrationTest.class)
 @TemplateIntegrationTest(ExportPipeline.class)
-@RunWith(JUnit4.class)
+@RunWith(Parameterized.class)
```
manitgupta, line 55: Is it possible to create a `SpannerTestBase` and parameterize that so that we don't have to do this every time we right a new test? 

https://stackoverflow.com/questions/67477549/junit4-parameterized-test-extends-parameterized-base-class-need-cartesian-pr

If not , it is also possible that the test author just forgets writing the new test in this way. How will be ensure that this style of writing tests is uniformly followed in future.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
```diff
@@ -90,11 +90,13 @@ public abstract class TemplateTestBase {
...
       new TestWatcher() {
         protected void starting(Description description) {
-          LOG.info(
-              "Starting integration test {}.{}",
-              description.getClassName(),
-              description.getMethodName());
           testName = description.getMethodName();
+          // In case of parameterization the testName can contain subscript like testName[paramName]
+          // Converting testName from testName[paramName] to testNameParamName since it is used to
+          // create many resources and it cannot contain special characters.
```
manitgupta, line 96: Why did this not fail earlier? What has changed?

darshan-sj: We never had parameterized tests in ITs in this repo. So, the testName never contained subscripts earlier.

darshan-sj: As mentioned in the comment the subscript is actually parameter name.

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
darshan-sj: Corrected. copy paste mistake :)

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
darshan-sj: Moved the parameterization to base class. This simplifies the test structure. Thank you!

Polber (commented):
LGTM - much cleaner than the existing solution. I will let someone from Spanner give the official approval

Also, there is other cleanup that will be required either in this PR, or I can submit the following changes after this PR is merged: https://github.com/GoogleCloudPlatform/DataflowTemplates/compare/main...Polber:DataflowTemplates:jkinard/spanner-cleanup

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerTemplateITBase.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.beam.it.gcp.TemplateTestBase;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+/**
+ * Base class for all Spanner template integration tests. This class parameterizes the spannerHost
+ * for all subclasses. If the "spannerHost" system property is not set, the value of spannerHost
+ * will be set to STAGING_SPANNER_HOST and DEFAULT_SPANNER_HOST (as defined in {@link
+ * SpannerResourceManager}). All tests in the base class will be run twice: once with spannerHost
```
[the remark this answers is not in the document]
Polber: I agree with this. Staging tests tend to be flaky and can block PRs - perhaps defaulting to DEFAULT_SPANNER_HOST except for the release workflow would be safer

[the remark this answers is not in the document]
darshan-sj: This is the current behaviour. All the tests run twice now already. I don't want to change that behavior as part of this PR. We can discuss and take up differentiating PR builds and release builds as a followup.

manitgupta (approved):

PR #2024, opened 2024-11-19
add KMS key support to BigQueryConverters

Polber:
This PR adds support for encrypting temporary table created during BigQueryIO read for any templates that use the BigQueryConverters class (v2) to read from BigQuery.

This only applies when using `query` to read from the BigQuery source


v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/BigQueryConverters.java
```diff
@@ -241,6 +242,18 @@ public interface BigQueryReadOptions extends PipelineOptions {
     String getQueryTempDataset();
 
     void setQueryTempDataset(String queryTempDataset);
+
+    @TemplateParameter.KmsEncryptionKey(
+        order = 8,
```
liferoad, line 247: why do we skip 7 here?

Polber: Because I messed up 😓 Thanks

PR #2028, opened 2024-11-21
[Draft: Not For review] Allowing Jdbc FetchSize Configuration for handling large rows

VardhanThigle:
[TODO]


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -117,8 +117,19 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
 
   void setNumPartitions(Integer value);
 
-  @TemplateParameter.Text(
+  @TemplateParameter.Integer(
       order = 9,
+      optional = true,
+      description = "The number of rows to fetch per page read for JDBC source.",
+      helpText =
+          "The number of rows to fetch per page read for JDBC source."
```
bharadwaj-aditya, line 125: nit: Please mention the default here - it will help folks identify what a good size should be

bharadwaj-aditya (commented):
Comments around the help text, rest looks fine.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -117,8 +117,19 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
...
+              + " Please refer to: Statement.setFetchSize(int) It should ONLY be used if the default value throws memory errors.")
```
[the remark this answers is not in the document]
bharadwaj-aditya: It would also be good to set the default value here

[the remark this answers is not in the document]
VardhanThigle: The default value is implicitly null and that can't be set explicitly (it gives a compile time error, but optional parameters are null unless a default is set). We cover this in unit tests.

`Statement.setFetchSize(int)` Got it. It came from javadocs but help text wont have the @link. Will fix this in the next patch.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
VardhanThigle: Thanks. Fixed in the current push.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -463,7 +467,9 @@ private static PTransform<PBegin, PCollection<SourceRow>> getReadWithUniformPart
       readWithUniformPartitionsBuilder =
           readWithUniformPartitionsBuilder.setMaxPartitionsHint((long) tableConfig.maxPartitions());
     }
-    return readWithUniformPartitionsBuilder.build();
+    ReadWithUniformPartitions readWithUniformPartitions = readWithUniformPartitionsBuilder.build();
+    LOG.info("Configured ReadWithUniformPartitions {} for {}", readWithUniformPartitions, config);
```
bharadwaj-aditya, line 471: Make sure this does not log any PII

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: It does log names of column and index details, but most of our logs do have that information for debugging purpose.
It does not log any rows, if that's what we wanted to confirm here.

manitgupta (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -117,8 +117,20 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
 
   void setNumPartitions(Integer value);
 
-  @TemplateParameter.Text(
+  @TemplateParameter.Integer(
       order = 9,
+      optional = true,
+      description = "The number of rows to fetch per page read for JDBC source.",
+      helpText =
+          "The number of rows to fetch per page read for JDBC source. If not set, the default of JdbcIO of 50_000 rows gets used."
```
bharadwaj-aditya, line 125: I don't think having an additional flag for mysql makes sense. Can we just check if this is passed and use this value ?

bharadwaj-aditya (approved):
minor suggestion on config. Rest looks fine.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
VardhanThigle: The flag `fetcSize` is meant for all of the jdbc sources - mySQl, PG, oracle etc.
We have not added any flag for setting the cursor mode, as users can set it via connection properties (fetchsize needs to be set while preparing the statement and hence not a connection property) either in shard config or URL, so for that part we don't have any flag, it's just a recommendation.

PR #2031, opened 2024-11-24
Avro format Import/Export support for Spanner NOT ENFORCED Foreign Key

jennyjiang98:
Adds Avro format Import/Export support for NOT ENFORCED Spanner Foreign Key (GoogleSQL dialect)

- Foreign keys are `ENFORCED` by default. When exporting unenforced foreign keys, the `NOT ENFORCED` keyword is attached. The `ENFORCED` is omitted when exporting enforced foreign keys.
 
I ran manual testing in staging & prod, Postgresql & Googlesql, by exporting a database with NOT ENFORCED foreign keys and was able to successfully import it with the correct resulting schema and all rows restored.

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/ForeignKey.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/ForeignKey.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.teleport.spanner.ddl;
 
+import static com.google.cloud.spanner.Dialect.GOOGLE_STANDARD_SQL;
 import static com.google.cloud.teleport.spanner.common.NameUtils.quoteIdentifier;
 import static com.google.common.base.Strings.isNullOrEmpty;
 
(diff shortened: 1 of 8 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/ddl/ForeignKey.java
```diff
@@ -133,6 +137,9 @@ private void prettyPrint(Appendable appendable) throws IOException {
               "Foreign Key action not supported: " + action.get().getSqlString());
       }
     }
+    if (isEnforced() != null && !isEnforced()) {
+      appendable.append(" NOT ENFORCED");
```
darshan-sj, line 141: I don't see dialect being used here or anywhere in this file. Is the SQL statements to create FKs same for both GSQL and postgres?

Is this feature going to be released for both GSQL and PG together?

If this feature is not supported for Postgres, can you add a check here to append " NOT ENFORCED" only for GSQL and put a TODO comment to remove the dialect check once it is supported for PG.

darshan-sj (dismissed):
Looks good.

darshan-sj (approved):

VardhanThigle (approved):

PR #2034, opened 2024-11-26
feat: add temp project and dataset id config properties

Emrehzl94:
This PR introduces tempProjectId and tempDatasetId configuration properties, allowing temporary data to be written to a specified project and dataset ID for the BigQuery source.


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/sources/BigQuerySource.java
```diff
@@ -22,11 +22,19 @@ public class BigQuerySource implements Source {
...
 
   public BigQuerySource(String name, String query) {
+    this(name, query, null, null);
+  }
 
+  public BigQuerySource(
+      String name, String query, String tempDataBeamProject, String tempDataBeamDataset) {
     this.name = name;
     this.query = query;
+    this.tempDataBeamProject = tempDataBeamProject;
```
fbiville, line 36: suggestion: I'd remove `Beam` from both of these names (applies everywhere, but I wanted to point this out at the source [no pun intended])

Emrehzl94: I didn't want it to be like `temp_data_dataset` or `tempDataDataset`, two consecutive data word in the properties and field names. Maybe `tempProject` and `tempDataset` could be, but this time the meaning is not explicit I think.

PR #2035, opened 2024-11-26
Add extra validation to SpannerToBigQuery Template

jrmccluskey:
The SpannerToBigQuery template wasn't failing with a clear error if the create disposition was not CREATE_NEVER but no BigQuery schema was provided, instead just throwing a vague NullPointerException in the write step. This replaces that exception with an IllegalArgumentException and a clear message before runtime.

### DIFF
--- a/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerToBigQueryOptions.java
+++ b/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerToBigQueryOptions.java
@@ -96,7 +96,7 @@ public interface SpannerToBigQueryOptions
       optional = true,
       description = "Cloud Storage path to BigQuery JSON schema",
       helpText =
-          "The Cloud Storage path (gs://) to the JSON file that defines your BigQuery schema.",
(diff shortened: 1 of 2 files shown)


Abacn:
Sorry for late reply.

>  just throwing a vague NullPointerException in the write step

Would you mind pointing out the code path for this Exception?

I'm wondering by any chance the original setting could work in some use case, e.g. CREATE_IF_NEEDED + not provide schema + the table actually existed, and with this change it now fails.

Abacn:
> The failure pops up here:
> 
> https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/22f4c0e7e02fa9cffdeab63d2ea8b5a038bd36d4/v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToBigQuery.java#L110

I see, NPE raised in getGcsFileAsString. Currently the error will happen regardless of CREATE_DISPOSITION. We should move `.withJsonSchema(getGcsFileAsString(..)` into a branch  as here 

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/a77d886b41573b8cbfa6988c094956052cbf734c/v2/jdbc-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/JdbcToBigQuery.java#L290

then CREATE_NEVER won't cause NPE

Abacn (approved):
thank you!

PR #2040, opened 2024-12-01
custom transformation implementation

shreyakhajanchi:
1. Added support for custom transformations in the `spanner-to-sourcedb` reverse replication template
2. Added support for filtered events, they will be written to a new directory `FilterEventsDirectoryName`
3. Added metrics for calculating transformation latency and counting invalid transformation exceptions
4. Updated unit tests

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
@@ -82,7 +82,8 @@ public class MySQLDMLGenerator implements IDMLGenerator {
             sourceTable,
             dmlGeneratorRequest.getNewValuesJson(),
             dmlGeneratorRequest.getKeyValuesJson(),
-            dmlGeneratorRequest.getSourceDbTimezoneOffset());
+            dmlGeneratorRequest.getSourceDbTimezoneOffset(),
+            dmlGeneratorRequest.getCustomTransformationResponse());
     if (pkcolumnNameValues == null) {
       LOG.warn(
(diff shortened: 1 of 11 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -224,13 +227,21 @@ private static Map<String, String> getColumnValues(
...
+    }
     for (Map.Entry<String, SourceColumnDefinition> entry : sourceTable.getColDefs().entrySet()) {
       SourceColumnDefinition sourceColDef = entry.getValue();
 
       String colName = sourceColDef.getName();
       if (sourcePKs.contains(colName)) {
         continue; // we only need non-primary keys
       }
+      if (customTransformColumns != null && customTransformColumns.contains(colName)) {
+        response.put(colName, (String) customTransformationResponse.get(colName));
```
Deep1998, line 242: nit: We should use toString() as its more robust and handles more types than casting.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
Deep1998, line 1: Is the custom transformation error counter same as the generic error counter (like other templates) and unchanged so we dont modify it in this PR? Just verifying since I didn't seen any error counters in this one.

Deep1998:
We also need some sort of exhaustive units tests covering all supported response java types and ensure the right DML gets generated, besides  end-to-end ITs.

Deep1998 (approved):

PR #1460, opened 2024-04-22
Type mappings for Bulk Reader

VardhanThigle:
Implementing Schema type mappings for Bulk reader.
The Bulk Reader will map various types of source schemas into [Unified types](https://cloud.google.com/datastream/docs/unified-types) .Support for mapping MySQL types to unified types is added in this PR.

The unified type mapping, in general has 2 steps:
Step-1 (inner step) - Generate Avro schemas  for the unified types. For most of the types, this is a pain building of Avro Schema. For a few types like Decimal or Varchar, extra fields like length, precision and scale are needed.
Step-2 (outer step) - Map the source column type name (strings as returned by the scan of information schema table) into the avro schema generators for step-1.

PR #2048, opened 2024-12-04
Cassandra Schema and Value Mapping

VardhanThigle:
# Cassandra Schema and Data mapping for primitive types.
This PR covers mapping of primitive types to Avro for Cassandra.
As of this PR, all the non-primitive types are mapped to `null`.

TODOS:
- end to end integration with pipeline options and `CassandraIO`.
- integration and scale testing 
- support for non-primitive types.


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/rowmapper/FieldMapper.java
Deep1998, line 1: These new fieldmapper, rowmapper constructs seem different from the jdbc rowmapper classes. Is there a way we can keep these constructs same and reusable? or will they not work for Cassandra?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappingsProvider.java
```diff
@@ -0,0 +1,135 @@
...
+              .set(IntervalNano.MONTHS_FIELD_NAME, value.getMonths())
+              .set(IntervalNano.DAYS_FIELD_NAME, value.getDays())
+              .set(IntervalNano.NANOS_FIELD_NAME, value.getNanoseconds())
+              .build();
+
+  /**
+   * Cassandra represents `Time` field as 64 bit singed integer representing number of nanoseconds
+   * since midnight. See <a
+   * href=https://cassandra.apache.org/doc/stable/cassandra/cql/types.html>types documentation</a>
+   * for further details.
```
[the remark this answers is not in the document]
VardhanThigle: Thanks for this suggestion.
For now, I have added comments in the datastream code.
The new type is covered in Cassandra Bulk Design doc.

Will add a readme for all the mappings to avro across MySql, PG and Cassandra mappings as a followup.

Deep1998:
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/rowmapper/FieldMapper.java
VardhanThigle: The driver api and object type is the subtle difference here:
Although semantically `getInt` on `Jdbc` looks similar to `getInt` for Cassandra Row, JDBC's result set does not implement Cassandra's [GettableByName](https://docs.datastax.com/en/latest-java-driver-api/com/datastax/oss/driver/api/core/data/GettableByName.html) interface and vice-versa.

If we want to call the same function on two objects of different types, they need to implement same interface.

Also note that, Cassandra does have a resultSet type, which on iteratation gives a Cassandra specific `Row`. This `Row` implements the `GettableByName` interface which looks some what similar to JDBC. ResultSet masks this iteration under the `next` call until null, where as in Cassandra world we map the `ResultSet` to an `Iterator<Row>`.

This begs the next question, should we create one more cushion interface to wrap the two?
We won't get any advantage by that, as the interface will have to route each type to it's actual functional (which would proably add to the code we need to maintain than help us reduce it)

There's more layer of difference here:
Jdbc result Set is more liberal in types, for example you can read prettymuch any field via getString, or you can always do `getInt` on a tiny int type, Cassandra driver is more strongly typed in this aspects.

For more complex operations, Cassandra driver supports registering codecs to read and write types, which JDBC does not.

So all in all, it makes sense to not tightly couple Cassandra and Jdbc although the interfaces driver interfaces look similar  they are not exactly same.  For sources that have exactly same connector interface - for example MySql, PG (and may be later Oracle) we have brought them under the same interface. For connectors that don't have the same interface it gives us better maintainability by not coupling them tightly.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/UnifiedTypeMapper.java
```diff
@@ -96,6 +100,7 @@ public enum MapperType {
     MYSQL,
     POSTGRESQL,
     ORACLE,
-    SQLSERVER
+    SQLSERVER,
```
bharadwaj-aditya, line 103: It was not introduced in this PR, but why is there a mention of SQL server in this flow ?

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroTestingHelper.java
```diff
@@ -91,4 +118,17 @@ public static GenericRecord createIntervalRecord(Integer months, Integer hours,
     genericRecord.put("micros", micros);
     return genericRecord;
   }
+
+  public static GenericRecord createIntervalNanosRecord(
+      Long years, Long months, Long days, Long hours, Long minutes, Long seconds, Long nanos) {
+    GenericRecord genericRecord = new GenericData.Record(INTERVAL_NANOS_SCHEMA);
+    genericRecord.put("years", years);
```
bharadwaj-aditya, line 125: it would be good to move all these strings into constants. They seem to be all over the code

bharadwaj-aditya (commented):
Looks fine overall. Please address the comments before merging.

v2/datastream-common/src/main/java/com/google/cloud/teleport/v2/datastream/transforms/FormatDatastreamRecordToJson.java
```diff
@@ -563,6 +565,48 @@ static void handleDatastreamRecordType(
...
+          /*
+           * The `intervalNano` maps to nano second precision interval type used by Cassandra Interval.
+           * On spanner this will map to `string` or `Interval` type.
+           * This is added here for DQL retrials for sourcedb-to-spanner.
+           *
+           * TODO(b/383689307):
+           * There's a lot of community in handling avro types between {@link FormatDatastreamRecordToJson} and {@link com.google.cloud.teleport.v2.spanner.migrations.avro.GenericRecordTypeConvertor}.
+           * Adding inter-package dependency might not be the best route, and we might eventually want to build a common package for handling common logic between the two.
+           */
+        case "intervalNano":
```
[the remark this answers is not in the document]
VardhanThigle: This is done for DLQ handling.
The bug we have linked in the todo is to migrate entirely to the common code that was developed for bulk migration avro handling for live template.

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/avro/AvroTestingHelper.java
VardhanThigle: The strings are specific to the schema. 
We will take up moving all the avro schema and logical type definitions to common code as a part of 
b/383689307

Currently the datastream package also uses string literals directly and that seems to have flown into the live changes.

v2/datastream-common/src/main/java/com/google/cloud/teleport/v2/datastream/transforms/FormatDatastreamRecordToJson.java
```diff
@@ -563,6 +565,48 @@ static void handleDatastreamRecordType(
                   .withZoneSameInstant(ZoneId.of("UTC"))
                   .format(DEFAULT_TIMESTAMP_WITH_TZ_FORMATTER));
           break;
+          /*
+           * The `intervalNano` maps to nano second precision interval type used by Cassandra Interval.
+           * On spanner this will map to `string` or `Interval` type.
+           * This is added here for DQL retrials for sourcedb-to-spanner.
+           *
+           * TODO(b/383689307):
+           * There's a lot of community in handling avro types between {@link FormatDatastreamRecordToJson} and {@link com.google.cloud.teleport.v2.spanner.migrations.avro.GenericRecordTypeConvertor}.
```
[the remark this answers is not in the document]
VardhanThigle: Thanks. That was an autocorrect effect. I changed it to commonality.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappingsProvider.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private static final CassandraMappings CASSANDRA_MAPPINGS =
+      CassandraMappings.builder()
+          .put("ASCII", UnifiedMappingProvider.Type.STRING, Row::getString, valuePassThrough)
+          .put("BIGINT", UnifiedMappingProvider.Type.LONG, Row::getLong, valuePassThrough)
+          .put("BLOB", UnifiedMappingProvider.Type.STRING, Row::getBytes, ByteBufferToHexString)
+          .put("BOOLEAN", UnifiedMappingProvider.Type.BOOLEAN, Row::getBool, valuePassThrough)
+          .put("COUNTER", UnifiedMappingProvider.Type.LONG, Row::getLong, valuePassThrough)
+          .put("DATE", UnifiedMappingProvider.Type.DATE, Row::getDate, localDateToAvroLogicalDate)
+          // Unlike MySql, the Cassandra decimal does not have precision and scale fixed in the
```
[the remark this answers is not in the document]
VardhanThigle: For primitive types, most of our mapping is quite like MySql. We have covered all primitive types in the PR here.
Decimal stands out a bit, so added a common as to why we map it to a string here.

Decimal on MySql is has precision and scale decided at schema time which maps to Avro's Decimal schema, which needs the scale and precision while constructing schema.

Cassandra however is more flexible in this regard. So we map the type to a string.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/rowmapper/RowMapper.java
```diff
@@ -0,0 +1,85 @@
...
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
+import org.apache.avro.Schema;
+import org.apache.commons.collections4.Transformer;
+import org.checkerframework.checker.initialization.qual.Initialized;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.UnknownKeyFor;
+
+@AutoValue
+public abstract class RowMapper implements Transformer<Row, SourceRow>, Serializable {
```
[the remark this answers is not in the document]
VardhanThigle: `CassandraMappingsProvider` is accessed outside Cassandra package as a Unified Mapping Provider.

RowMapper will be using by `io.cassandra` package - It needs to be public as it will be accessed outside it's immediate package, but still it will be used only within the `io.cassandra` umbrella.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/typemapping/UnifiedTypeMapper.java
VardhanThigle: It was cover edge cases  to test checks on unmapped source type in the initial mapping [pull/1460](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/1460)

When we get to supporting `SQLSERVER` will add an artificial type like `UNSUPPORTED` here.

Deep1998 (approved):

VardhanThigle (commented):
IT tests are failing, due to 
```json
{"errors":[{"code":"TOOMANYREQUESTS","message":"Quota exceeded for quota metric 'Requests per project in the US multi-region' and limit 'Requests per project in the US multi-region per minute' of service 'artifactregistry.googleapis.com' for consumer 'project_number:167633336322'."}]}
```
The same [workflow](https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/spanner-pr.yml) had passed on this PR 13 hours ago.

Even the [scheduled workflow](https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/12325107750) on main seems to be flaky.

Issue #2038, opened 2024-11-27
[Bug]: Comma separated list of topics is not WAI in kafka-to-bigquery


atognolag:
### Related Template(s)

kafka_to_bigquery_flex

### Template Version

2024-11-06-01_rc00

### What happened?

The string of comma separated Kafka topics is not correctly parsed into a list and is instead provided as a string to the KafkaIO where a List<String> is expected

### Relevant log output

```shell
{"message":"org.apache.beam.sdk.extensions.gcp.options.GcpOptions$DefaultProjectFactory - Inferred default GCP project \u0027sdg-af-goog\u0027 from gcloud. If this is the incorrect project, please cancel this Pipeline and specify the command-line argument --project.","severity":"INFO"}
[...]

PR #2060, opened 2024-12-12
Allow defining multiple source kafka topics 

Fadelis:
Supporting possibility listening to multiple Kafka topics in dataflows would be very valuable. It was supported in non Flex templates and seems like the base options example hints to it as well, but somehow it wasn't not implemented (and there's a bug for it #2038) https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ff60a5f32174b364f8b8ece66fe789a55aef5009/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java#L40
With the possibility of listening to multiple topics, it would be also very beneficial to be able to enrich the stored record with the topic of the message, so added such functionality in a similar fashion as the `persistKafkaKey` option.

### DIFF
--- a/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/utils/KafkaTopicUtils.java
+++ b/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/utils/KafkaTopicUtils.java
@@ -18,6 +18,8 @@ package com.google.cloud.teleport.v2.kafka.utils;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class KafkaTopicUtils {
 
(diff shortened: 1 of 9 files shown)


Fadelis:
Bump @GoogleCloudPlatform/dataflow-templates-wg

an2x:
The new Flex templates support reading from a single topic only (the example is incorrect, I'll fix the example), because they have a lot of extra functionality compared to the legacy templates (e.g. reading messages with different schema from the same topic, writing to different BigQuery tables based on schema, etc.), and supporting this while also reading from multiple topics is not straightforward and will require more changes than just parsing the input topic as a list.

At the minimum, we would need better support for the `schemaFormat = SINGLE_SCHEMA_FILE` case (because different topics typically have messages with different schema, so the user would somehow have to provide a schema file for every topic, which is hard to configure through the current UI), and better support for the `DYNAMIC_TABLE_NAMS` write strategy in the Kafka to BigQuery table (because right now the target table names are auto-generated and we don't account for messages potentially coming from different topics, there could be conflicts).

All in all, IMO all this would make the template way too complicated. It's supposed to cover only some basic use cases. If someone has a complex pipeline that requires all this functionality while also reading from multiple topics, I'd recommend writing a custom pipeline for this use case instead (or a custom template).

PR #2063, opened 2024-12-16
[Sourcedb-to-spanner] Bulk migration Mysql to spanner 1tb Load test

darshan-sj:


v2/sourcedb-to-spanner/src/test/resources/SourceDbToSpannerLT/mysql-spanner-schema.sql
```diff
@@ -0,0 +1,41 @@
+CREATE TABLE IF NOT EXISTS table1 (
```
VardhanThigle, line 1: If we are testing for wide rows, please set the fetchsize as per [user-guide](https://docs.google.com/document/d/1P2uZLPmw_0av1KN4vt7voaxDw4r119SIgKfxSn5vAmE/edit?tab=t.0#bookmark=id.vgljjjz9hidk)

darshan-sj: Done.

it/google-cloud-platform/src/main/java/org/apache/beam/it/jdbc/StaticJDBCResource.java
```diff
@@ -0,0 +1,89 @@
...
+  private final String username;
+  private final String password;
+  private final int port;
+
+  private final String database;
+
+  enum SourceType {
+    ORACLE,
+    MYSQL,
+    POSTGRESQL,
```
VardhanThigle, line 32: [Super-nit, can be a todo]
In case the `it` folder is not covered under UT, could we exempt it for the Codecov tests?

it/google-cloud-platform/src/main/java/org/apache/beam/it/jdbc/StaticJDBCResource.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public abstract static class Builder<T extends org.apache.beam.it.jdbc.StaticJDBCResource> {
+    private final String hostname;
+    private final String username;
+    private final String password;
+    private final int port;
+
+    private final String database;
+
+    public Builder(String hostname, String username, String password, int port, String database) {
```
VardhanThigle, line 79: [nit]
If the builder isn't a fluent builder, would plain constructor on the `StaticJDBCResource` work too?

In case you would like to have a fluent builder pattern, you could use AutoValue and AutoBuilder , but note that it will make this class final, and would need a slightly different way to structure the code here.

VardhanThigle (approved):

it/google-cloud-platform/src/main/java/org/apache/beam/it/jdbc/StaticJDBCResource.java
darshan-sj: Added a todo comment

it/google-cloud-platform/src/main/java/org/apache/beam/it/jdbc/StaticJDBCResource.java
darshan-sj: This pattern was followed in other ResourceManager classes like - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datastream/JDBCSource.java  - I did not want to deviate from that.

VardhanThigle (approved):

PR #2077, opened 2024-12-19
Cassandra Driver Config Loader for Cassandra Driver Config in GCS

VardhanThigle:
# Overview
1. A common static utility to load [Cassandra Driver config ](https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/configuration/index.html)from GCS.
2. The implementation uses the Cassandra Driver APIs to parse the file, so all the options that the driver supports (now and with upgrades) are supported for parsing out of the box.
3. The implementation also provides utility to serialize and deserialize the Config to [OptionMap](https://docs.datastax.com/en/drivers/java/4.9/com/datastax/oss/driver/api/core/config/OptionsMap.html)
## Testing
1. All lines covered in UT.
2. UT to parse the [reference config](https://docs.datastax.com/en/developer/java-driver/4.3/manual/core/configuration/reference/index.html) as provided by the Cassandra community.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (C) 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.migrations.utils;
+
+import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
+import com.datastax.oss.driver.api.core.config.OptionsMap;
+import com.datastax.oss.driver.api.core.config.TypedDriverOption;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableMap;
+import com.typesafe.config.ConfigException;
+import java.io.FileNotFoundException;
+import java.net.URL;
+import java.util.Map.Entry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A common static utility class that allows the spanner migration pipelines to ingest Cassandra
+ * Driver config file from GCS. Cassandra has a structured config file to accept all the driver
+ * parameters, be it list of host ip addresses, credentials, retry policy and many more. Most of
+ * these parameters are very specific to the Cassandra Database. Refer to the <a
+ * href=>https://docs.datastax.com/en/developer/java-driver/4.3/manual/core/configuration/reference/index.html>reference
+ * configuration</a> for the file format.
+ */
+public final class CassandraDriverConfigLoader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CassandraDriverConfigLoader.class);
+  private static final ImmutableMap<String, TypedDriverOption> OPTIONS_SUPPORTED_BY_DRIVER =
+      getOptionsSupportedByDriver();
+
+  /**
+   * Load the Cassandra Config from a file as a {@link DriverConfigLoader}.
+   *
+   * @param path A complete gcs path to the config file of the form "gs://path/to/file".
+   * @return DriverConfigLoader.
+   * @throws FileNotFoundException - If file is not found at specified path.
+   */
+  public static DriverConfigLoader loadFile(String path) throws FileNotFoundException {
+    URL url = loadSingleFile(path);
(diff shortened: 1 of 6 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
```diff
@@ -0,0 +1,147 @@
...
+      throw (new IllegalArgumentException(
+          String.format(
+              "Need to provide a single Cassandra driver config file in the specified path %s. Found %d files",
+              path, urls.length)));
+    }
+    return urls[0];
+  }
+
+  @VisibleForTesting
+  protected static void putOptionToMap(
```
bharadwaj-aditya, line 123: nit: Rename to putInOptionsMap or something. The current method rename sounds like the options map is created/added completely.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
```diff
@@ -0,0 +1,147 @@
...
+      getOptionsSupportedByDriver();
+
+  /**
+   * Load the Cassandra Config from a file as a {@link DriverConfigLoader}.
+   *
+   * @param path A complete gcs path to the config file of the form "gs://path/to/file".
+   * @return DriverConfigLoader.
+   * @throws FileNotFoundException - If file is not found at specified path.
+   */
+  public static DriverConfigLoader load(String path) throws FileNotFoundException {
```
bharadwaj-aditya, line 52: nit: suggest renaming to loadFile / loadFromPath or something which is more obvious, especially considering this is a public method.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
```diff
@@ -0,0 +1,147 @@
...
+   * java.io.Serializable}. At the time of opening a conneciton to Cassandra, it can be deserialized
+   * by {@link CassandraDriverConfigLoader#fromOptionsMap(OptionsMap)}. Note: Implementation Detail,
+   * Cassandra Driver does not provide a direct method to convert a link {@link DriverConfigLoader}
+   * into an {@link OptionsMap}, or build an {@link OptionsMap} from a file.
+   *
+   * @param path A complete gcs path to the config file of the form "gs://path/to/file".
+   * @return DriverConfigLoader.
+   * @throws FileNotFoundException - If file is not found at specified path.
+   */
+  public static OptionsMap loadToSerializableOptionsMap(String path) throws FileNotFoundException {
```
bharadwaj-aditya, line 77: nit: rename to be clearer - something like getOptionsMapFromFile or something.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
bharadwaj-aditya: does this method need to be public ? I think only 1 method will probably need to be public from this class - which is load/getOptionsMap

bharadwaj-aditya (commented):
A few minor comments. Looks fine overall.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
```diff
@@ -0,0 +1,147 @@
...
+      LOG.error(
+          "Parsing error while parsing Cassandra Driver config from path {}", path, parseException);
+      throw parseException;
+    }
+  }
+
+  /**
+   * Load the Cassandra Config from a file as a {@link java.io.Serializable} {@link OptionsMap}.
+   * This {@link OptionsMap} can be stored in any object that needs to implement {@link
+   * java.io.Serializable}. At the time of opening a conneciton to Cassandra, it can be deserialized
```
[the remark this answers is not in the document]
VardhanThigle: Thanks.
Fixed in the current push.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
VardhanThigle: Thanks.
This method actually is `putOptionToOptionMap`, which sort of becomes redundant and hence had used `putOptionToMap`.
Changed this to `putInOptionsMap` as suggested. That works too.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
VardhanThigle: Thanks. Changed to `loadFile`.

The method is public to allow any pipeline to use loading to `DriverConfigLoader` directly if needed in their architecture at the time of pipeline construction.
I

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
VardhanThigle: Thanks. Fixed in current push.

bharadwaj-aditya (approved):
LGTM

PR #2080, opened 2024-12-20
Use [self-hosted, it] for prepare java cache workflow

liferoad:


liferoad:
This fixes the no-space left error caused by the GitHub workers.

damccorm (approved):

PR #2106, opened 2025-01-03
Fix precision error for datetime

Deep1998:
Screenshot 2025-01-03 at 4 48 04 PM


VardhanThigle (approved):

bharadwaj-aditya (approved):

Deep1998:
bulk templates ITs have succeeded, merging since this change is not touched by any other template and other failures are due to quota issues.

PR #2127, opened 2025-01-10
Changing project id for spanner-staging-tests and run on spanner runners

darshan-sj:


cicd/cmd/run-spanner-staging-it-tests/main.go
```diff
@@ -56,8 +56,7 @@ func main() {
 		mvnFlags.SkipJib(),
 		mvnFlags.SkipShade(),
 		mvnFlags.RunSpannerStagingIntegrationTests(),
-		mvnFlags.ThreadCount(4),
-		mvnFlags.IntegrationTestParallelism(3),
+		mvnFlags.IntegrationTestParallelism(1),
```
manitgupta, line 59: Why is it 1? What issues does more parallelism cause?

darshan-sj: The dataflow job link was not logged. I was experimenting to see if the issue was due to parallelism. I have reverted this change now.

.github/workflows/spanner-staging-tests.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     name: Spanner Dataflow Templates Integration Tests
     timeout-minutes: 180
     # Run on any runner that matches all the specified runs-on values.
-    runs-on: [ self-hosted, it ]
+    runs-on: [ self-hosted, spanner ]
```
bharadwaj-aditya, line 38: earlier the parameters were it-project etc - corresponding to it.

Now it says Spanner. Where does the name get used ?

darshan-sj: This is the tag of github actions runner on which needs to pick up this workflow. I have created a new pool of github actions runners and tagged it `spanner` . These are in new project and is done to provide isolation for the staging tests

manitgupta:
I see that no Spanner tests have been run on this PR. 

1. Should we include the workflow definition files (`.github`) in the file list that triggers the Spanner tests?
2. Without that, how do we verify if the `spanner` tag is correctly set up and will work?

darshan-sj:
> I see that no Spanner tests have been run on this PR.
> 
> 1. Should we include the workflow definition files (`.github`) in the file list that triggers the Spanner tests?
> 2. Without that, how do we verify if the `spanner` tag is correctly set up and will work?

I have run the spanner staging workflow on this branch manually and verified that they are running on spanner instances - https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/spanner-staging-tests.yml

I have also configured this workflow to run if there are any changes to the config file itself. It did not get triggered for this PR.

manitgupta (approved):

PR #2128, opened 2025-01-12
Spanner Import/Export INTERLEAVE IN

jjfox15:
Add support for Import/Export of Spanner `INTERLEAVE IN` tables. This new feature allows for omitting the `PARENT` keyword in `CREATE/ALTER <table> INTERLEAVE IN [PARENT] <parent_table>` clauses. `INTERLEAVE IN` tables are similar to `INTERLEAVE IN PARENT`, but do not enforce parent/child referential integrity. `INTERLEAVE IN` tables do not support `ON DELETE` actions.

I manually verified this works by creating a table with two children, one `INTERLEAVE IN` and the other `INTERLEAVE IN PARENT`. All tables had a few rows - and the `INTERLEAVE IN` table contained a row for which the parent row was missing.

I exported this db to avro using [these instructuons](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/cf5b98b007d04b8851c22b953e708443aa428fd8/v1/README_Cloud_Spanner_to_GCS_Avro.md), and imported to an empty db following [these](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/cf5b98b007d04b8851c22b953e708443aa428fd8/v1/README_GCS_Avro_to_Cloud_Spanner.md). I manually checked the target db contained the correct tables and rows, and also verified the entries in the `INFORMATION_SCHEMA.Tables` were identical across both tables (specifically the `INTERLEAVE_TYPE`, `PARENT_TABLE_NAME`, and `ON_DELETE_ACTION` columns)


hengfengli (approved):
LGTM.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -272,13 +273,26 @@ private void listTables(Ddl.Builder builder) {
...
+      Table.InterleaveType interleaveType = null;
+      if (!Strings.isNullOrEmpty(interleaveTypeStr)) {
+        interleaveType =
+            interleaveTypeStr.equals("IN PARENT") ? InterleaveType.IN_PARENT : InterleaveType.IN;
+      }
+      String onDeleteAction = resultSet.isNull(4) ? null : resultSet.getString(4);
 
       // Error out when the parent table or on delete action are set incorrectly.
-      if (Strings.isNullOrEmpty(parentTableName) != Strings.isNullOrEmpty(onDeleteAction)) {
+      if (Strings.isNullOrEmpty(parentTableName) != Strings.isNullOrEmpty(interleaveTypeStr)) {
```
darshan-sj, line 285: Can you explain why this is not a valid combination?

parentTableName is not null && interleaveTypeStr is null  -> This would be the current situation in prod right? Or is the interleaveType column in info schema is fully backfilled with `IN PARENT` ?

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -272,13 +273,26 @@ private void listTables(Ddl.Builder builder) {
...
+        throw new IllegalStateException(
+            String.format(
+                "Invalid combination of parentTableName %s and interleaveType %s",
+                parentTableName, interleaveType));
+      }
+
+      if ((interleaveType == InterleaveType.IN_PARENT) == Strings.isNullOrEmpty(onDeleteAction)) {
```
darshan-sj, line 292: Why is this an invalid combination? If interleaveType is IN PARENT, onDeleteAction cannot be null ?

darshan-sj: Actually the if condition expressions in both the above if blocks are super confusing. Either add comments explaining scenarios or avoid using == and != to compare boolean operands and expand the expr with && ||

darshan-sj (commented):
Please check If there are Unit tests or Integration tests for the files being modified in this PR. If there are can you add test cases for INTERLEAVE IN ?

v1/src/main/java/com/google/cloud/teleport/spanner/DdlToAvroSchemaConverter.java
```diff
@@ -119,6 +121,9 @@ public Collection<Schema> convert(Ddl ddl) {
       recordBuilder.prop(GOOGLE_STORAGE, "CloudSpanner");
       if (table.interleaveInParent() != null) {
         recordBuilder.prop(SPANNER_PARENT, table.interleaveInParent());
+        recordBuilder.prop(
+            SPANNER_INTERLEAVE_TYPE,
+            table.interleaveType() == InterleaveType.IN ? "IN" : "IN PARENT");
         recordBuilder.prop(
             SPANNER_ON_DELETE_ACTION, table.onDeleteCascade() ? "cascade" : "no action");
```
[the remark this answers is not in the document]
jjfox15: Table.PrettyPrint only emits the ON DELETE clause when the on delete action is CASCADE. By omitting the clause, no error for INTERLEAVE IN.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -241,7 +242,7 @@ private void listTables(Ddl.Builder builder) {
       case POSTGRESQL:
         queryBuilder =
             Statement.newBuilder(
-                "SELECT t.table_schema, t.table_name, t.parent_table_name, t.on_delete_action FROM"
+                "SELECT t.table_schema, t.table_name, t.parent_table_name, t.interleave_type, t.on_delete_action FROM"
```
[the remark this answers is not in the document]
jjfox15: ah, important context. Pg does not yet support INTERLEAVE IN, but, the interleave_type column exists in information_schema for both dialects. This column predates my feature and work in this area by quite some time.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -272,13 +273,26 @@ private void listTables(Ddl.Builder builder) {
       // Parent table and child table has to be in same schema.
       String parentTableName =
           resultSet.isNull(2) ? null : getQualifiedName(tableSchema, resultSet.getString(2));
-      String onDeleteAction = resultSet.isNull(3) ? null : resultSet.getString(3);
+      String interleaveTypeStr = resultSet.isNull(3) ? null : resultSet.getString(3);
+      Table.InterleaveType interleaveType = null;
+      if (!Strings.isNullOrEmpty(interleaveTypeStr)) {
+        interleaveType =
+            interleaveTypeStr.equals("IN PARENT") ? InterleaveType.IN_PARENT : InterleaveType.IN;
```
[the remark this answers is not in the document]
jjfox15: Yep, I double checked

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
jjfox15: See my above comment, but no, if parentTableName is not null interleaveTypeStr will never be null, even before I released my feature.

As I understand it, the only place backwards compatibility is relevant is the avro -> ddl converter, in case we try to import from an export performed before this change. There, it is important that we default to NO ACTION in case the property doesnt exist.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
jjfox15: "If interleaveType is IN PARENT, onDeleteAction cannot be null ?" -> yes that is correct.

Sure thing though, I'll add comments

darshan-sj (dismissed):
Please add tests.

Please test Export + import for a PG database with your feature.

darshan-sj (commented):
Please add a test case in ExportPipelineIT

darshan-sj (approved):
Looks good

PR #2130, opened 2025-01-13
Enable SSL mode in embedded Cassandra used by unit tests.

VardhanThigle:
# Overview
This is a child of #2114.
This has UT only changes to enable SSL in embedded Cassandra used by Bulk UT.

### DIFF
--- a/v2/sourcedb-to-spanner/pom.xml
+++ b/v2/sourcedb-to-spanner/pom.xml
@@ -158,5 +158,11 @@
       <version>4.1</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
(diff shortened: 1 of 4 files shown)


Deep1998:
Could we get a summary in the comments of why all this is needed to enable SSL as readers may not be fully aware of the nuances of enabling SSL?

VardhanThigle:
> Could we get a summary as a code comment of why all this is needed to enable SSL as readers may not be fully aware of the nuances of enabling SSL?
Thanks.
Added Java Docs. Although this is a UT only code, in this case JavaDocs are needed.

Let's syncp over a call if I could help with more details around the SSL enablements.

Deep1998 (approved):

shreyakhajanchi (approved):

PR #2132, opened 2025-01-13
MongoDbToBigQuery - ISO standard for date and time

OreOreDa:
We have noticed that the format of dates and timestamp are broken, compared to the mongodb fields, e.g. a date appears as "Mar 12, 2024, 3:05:34 AM", instead of the original ISO format

This PR fixes the issue, thanks !


v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/templates/MongoDbUtils.java
```diff
@@ -119,7 +141,8 @@ public static Document getMongoDbDocument(String uri, String dbName, String coll
     return doc;
   }
 
-  public static TableRow getTableSchema(Document document, String userOption) {
+  public static TableRow getTableSchema(
```
Abacn, line 144: Could you please add a unit test for legacy / iso time format?

v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/mongodb/options/MongoDbToBigQueryOptions.java
```diff
@@ -158,5 +158,15 @@ public interface JavascriptDocumentTransformerOptions extends PipelineOptions {
     String getJavascriptDocumentTransformFunctionName();
 
     void setJavascriptDocumentTransformFunctionName(String javascriptDocumentTransformFunctionName);
+
+    @TemplateParameter.Boolean(
+        order = 2,
+        optional = true,
+        description = "Use legacy time format.",
+        helpText =
```
Abacn, line 166: The description is "use legacy time format" but the option name is "useIsoTimeFormat". The helpText (default is true) is also not true. Please fix.

Abacn (commented):
Thank you for the fix! Left a few comment.

PR #2153, opened 2025-01-27
Handle Frozen Case and Error Categorization

taherkl:
Few Enhancement and Bug Fixes

1. Handle Frozen Case for Collection data Types
2. Handle Error Categorization for Moving in the proper bucket
3. Handle Float Conversion AutoBoxing
4. Handle Bytes Datatype

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
@@ -532,11 +532,8 @@ public class CassandraTypeHandler {
   private static Object parseFloatingPoint(String columnType, Object colValue) {
     if (columnType.equals("double")) {
       return Double.parseDouble((String) colValue);
-    } else if (columnType.equals("float")) {
-      return Float.parseFloat((String) colValue);
     }
-    throw new IllegalArgumentException(
-        String.format("Unable to parse %s : value %s", columnType, colValue));
+    return Float.parseFloat((String) colValue);
   }
 
   private static LocalDate parseDate(Object colValue) {
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
@@ -118,8 +118,11 @@ public class CassandraTypeHandler {
    * @throws IllegalArgumentException If the value is neither a valid number string, byte array, nor
    *     a valid {@link ByteBuffer} for varint representation.
    */
-  private static BigInteger handleCassandraVarintType(String value) {
-    return new BigInteger(value);
+  private static BigInteger handleCassandraVarintType(Object value) {
+    if (value instanceof byte[]) {
+      return new BigInteger((byte[]) value);
+    }
+    return new BigInteger(value.toString());
   }
 
   /**
@@ -353,6 +356,10 @@ public class CassandraTypeHandler {
   private static PreparedStatementValueObject<?> parseAndCastToCassandraType(
       String columnType, Object colValue) {
 
+    if (columnType.startsWith("frozen<")) {
+      return parseAndCastToCassandraType(extractInnerType(columnType), colValue);
+    }
+
     // Handle collection types
     if (columnType.startsWith("list<")) {
       return safeHandle(
@@ -443,7 +450,7 @@ public class CassandraTypeHandler {
 
       case "varint":
         return PreparedStatementValueObject.create(
-            columnType, safeHandle(() -> handleCassandraVarintType(colValue.toString())));
+            columnType, safeHandle(() -> handleCassandraVarintType(colValue)));
 
       case "duration":
         return PreparedStatementValueObject.create(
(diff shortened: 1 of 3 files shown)


taherkl:
Hi @shreyakhajanchi Can you please run the workflows as well as provide your review? Thanks

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -240,6 +242,8 @@ public void processElement(ProcessContext c) {
           | com.mysql.cj.jdbc.exceptions.CommunicationsException
           | java.sql.SQLIntegrityConstraintViolationException
           | java.sql.SQLTransientConnectionException
+          | CodecNotFoundException
+          | QueryExecutionException
```
shreyakhajanchi, line 246: What parts of the code are throwing this exception?

pawankashyapollion: **QueryExecutionException**  : Whenever  we are going to perform DDL Statement Execution, Sometime we are getting  QueryExecutionException in case of Cassandra node when there is an issue with the execution of the query 

**CodecNotFoundException**: We are getting this issue we are trying to map invalid mapping to Cassandra node
for Example: Double to float etc.

shreyakhajanchi: I had edited the earlier comment. Can you please take a look at the updated comment?

pawankashyapollion: Yes @shreyakhajanchi 
_Will a query trying to insert null value in a non-nullable column fail with QueryExecutionException?_ -> will throw QueryExecutionException

Why is CodecNotFoundException retryable , isn't that caused by incorrect datatype mapping? -> Yes, If in case our code missed some corner case which leads to incorrect datatype mapping, it is not that in terms of parsing it could be the place of binding

shreyakhajanchi: `Yes, If in case our code missed some corner case which leads to incorrect datatype mapping, it is not that in terms of parsing it could be the place of binding` - It should be a severe error in that case

`Will a query trying to insert null value in a non-nullable column fail with QueryExecutionException?` - this is also a case of severe error right, how will retry solve this issue?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
```diff
@@ -523,9 +530,13 @@ private static BigDecimal parseDecimal(Object colValue) {
...
   private static Object parseFloatingPoint(String columnType, Object colValue) {
-    return columnType.equals("double")
-        ? Double.parseDouble((String) colValue)
-        : Float.parseFloat((String) colValue);
+    if (columnType.equals("double")) {
+      return Double.parseDouble((String) colValue);
+    } else if (columnType.equals("float")) {
+      return Float.parseFloat((String) colValue);
+    }
+    throw new IllegalArgumentException(
```
shreyakhajanchi, line 538: Add a test case covering this branch

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
pawankashyapollion: @shreyakhajanchi Yes I have moved both to severe error

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
pawankashyapollion: @shreyakhajanchi I guess this should resolved now

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
shreyakhajanchi: In case of query timeouts what will be the type of exception?

pawankashyapollion: The **_QueryExecutionException_** usually occurs when there’s something wrong with the query itself, like a mismatch between the database structure and the query, or problems with the data types. It can also happen if you violate certain rules in your database, like trying to insert duplicate values, or the system can’t meet the required consistency levels or Constraints Volilation Exception etc. 

It is the Parent Class for other exceptions like WriteFailureException etc which can happen due to timeout or invalid query

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
pawankashyapollion: @shreyakhajanchi we just simplified the logic

shreyakhajanchi (approved):

PR #2154, opened 2025-01-28
Support avro arrays for postgres insertion.

claudevdm:


v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToPostgresDML.java
```diff
@@ -102,6 +107,56 @@ public String cleanDataTypeValueSql(
...
+              elements.add(formatArrayElement(elementValue));
+            } else {
+              elements.add(getNullValueSql());
+            }
+          } else if (!element.isNull()) {
+            elements.add(formatArrayElement(element));
+          }
+        }
+      }
+      return "ARRAY[" + String.join(",", elements) + "]";
```
dhercher, line 149: why are we adding ARRAY here?  shouldn't it just be the value?

dhercher (changes requested):

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToPostgresDML.java
claudevdm: AFAIK Postgres arrays either require ARRAY[value1,value2] or array literals as curly braces like {value1,value2}.

claudevdm: https://www.postgresql.org/docs/current/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS

dhercher: The CL is formatting this data directly into Postgres syntax but it should be either string json or an object at this point.  In the Postgres writer you'll want to wrap thi value in the array syntax (the file i sent earlier)
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToPostgresDML.java#L103

dhercher: apologies, nvm, i had thought we were doing this in the avro reader ... ignore

dhercher (approved):

PR #2155, opened 2025-01-28
Cassandra  Reverse Replication - Integration Tests

taherkl:
1. Handle Integration test for different Table from Spanner to one Table in cassandra Insert Query
2. Handle Integration test for One table in spanner which support all of the data type conversion to cassandra


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * Inserts the given Document into a table.
+   *
+   * <p>A database will be created here, if one does not already exist.
+   *
+   * @param tableName The name of the table to insert the document into.
+   * @param document The document to insert into the table.
+   * @return A boolean indicating whether the Document was inserted successfully.
+   */
+  public synchronized boolean insertDocument(String tableName, Map<String, Object> document) {
```
shreyakhajanchi, line 188: why are rows referred to as documents? Can we use `rows` instead?

v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceDbDatatypeIT/cassandra-schema.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE AllDatatypeColumns (
```
shreyakhajanchi, line 1: add uuid and inet types as well

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraDbITBase.java
```diff
@@ -0,0 +1,245 @@
+/*
```
shreyakhajanchi, line 1: why are we creating a seperate SpannerToCassandraDbITBase class, we should reuse `SpannerToSourceDbITBase` since most of the methods are common

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
```diff
@@ -0,0 +1,422 @@
...
+        createCassandraSchema(cassandraResourceManager, CASSANDRA_SCHEMA_FILE_RESOURCE);
+        createAndUploadCassandraConfigToGcs(
+            gcsResourceManager, cassandraResourceManager, CASSANDRA_CONFIG_FILE_RESOURCE);
+        pubsubResourceManager = setUpPubSubResourceManager();
+        subscriptionName =
+            createPubsubResources(
+                getClass().getSimpleName(),
+                pubsubResourceManager,
+                getGcsPath("dlq", gcsResourceManager).replace("gs://" + artifactBucketName, ""));
+        jobInfo =
```
shreyakhajanchi, line 109: why do we need two seperate class i.e two seperate dataflow jobs for these 2 ITs , it should be merged into 1 test as long as the dataflow job parameters are same. Each new IT creates a seperate set of resources, we should aim to minimize this wherever possible

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
```diff
@@ -0,0 +1,422 @@
...
+    ResourceManagerUtils.cleanResources(
+        spannerResourceManager,
+        cassandraResourceManager,
+        spannerMetadataResourceManager,
+        gcsResourceManager,
+        pubsubResourceManager);
+  }
+
+  @Test
+  public void spannerToCassandraSourceDataTypeConversionTest()
```
shreyakhajanchi, line 143: This test method is for default type mapping , add a test for cross type mapping similar to : https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2631a1d683033bec88833de4cdc3094bbb1a9aa3/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerDDLIT.java#L248 . You can add a new table in the same database and test on it

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
```diff
@@ -0,0 +1,422 @@
...
+            .set("longtext_column")
+            .to("Very long text data that exceeds the medium text column length for long text.")
+            .set("enum_column")
+            .to("OptionA")
+            .set("bool_column")
+            .to(Value.bool(Boolean.TRUE))
+            .set("other_bool_column")
+            .to(Value.bool(Boolean.FALSE))
+            .set("bytes_column")
+            .to(Value.bytes(ByteArray.copyFrom("SGVsbG8gd29ybGQ=".getBytes())))
```
shreyakhajanchi, line 209: I would suggest use readable values in the format: `Value.bytes(ByteArray.copyFrom("bytes_column_value"))`. Same comment for other places

v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceDbDatatypeIT/cassandra-schema.sql
pawankashyapollion: @shreyakhajanchi we are adding the same in our custom transformation

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraDbITBase.java
pawankashyapollion: @shreyakhajanchi we have created separately because of resources we have differences in handling launch job and also for the Schema Population

I will try to accommodate the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
pawankashyapollion: Sure @shreyakhajanchi

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
pawankashyapollion: @shreyakhajanchi we are already handling the same in our customTransformationIT

shreyakhajanchi: the customTransformationIT is different , it will be similar to https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbCustomTransformationIT.java , the types test should be included in this same class

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
```diff
@@ -0,0 +1,366 @@
...
+ *
+ * <p>The class supports one database and multiple collections per database object. A database is
+ * created when the first collection is created if one has not been created already.
+ *
+ * <p>The database name is formed using testId. The database name will be "{testId}-{ISO8601 time,
+ * microsecond precision}", with additional formatting.
+ *
+ * <p>The class is thread-safe.
+ */
+public class CassandraSharedResourceManager
```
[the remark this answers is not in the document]
pawankashyapollion: @shreyakhajanchi addressed the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
pawankashyapollion: @shreyakhajanchi addressed the same and removed unused code

v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceDbDatatypeIT/cassandra-schema.sql
pawankashyapollion: @shreyakhajanchi addressed the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraDbITBase.java
pawankashyapollion: @shreyakhajanchi addressed the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
pawankashyapollion: @shreyakhajanchi addressed the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
pawankashyapollion: @shreyakhajanchi can we have a sync on it

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbDatatypeIT.java
pawankashyapollion: @shreyakhajanchi addressed the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
```diff
@@ -138,13 +221,18 @@ public PipelineLauncher.LaunchInfo launchDataflowJob(
...
+                getGcsPath(
+                    sourceType != null ? "input/cassandra-config.conf" : "input/shard.json",
+                    gcsResourceManager));
             put("changeStreamName", "allstream");
             put("dlqGcsPubSubSubscription", subscriptionName);
             put("deadLetterQueueDirectory", getGcsPath("dlq", gcsResourceManager));
             put("maxShardConnections", "5");
             put("maxNumWorkers", "1");
             put("numWorkers", "1");
+            put("sourceType", sourceType);
```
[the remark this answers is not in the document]
taherkl: @shreyakhajanchi The fix has been pushed. Can you please rerun the pipeline?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
```diff
@@ -121,6 +121,11 @@ private interface HandlerSupplier<T> {
   private static BigInteger handleCassandraVarintType(Object value) {
     if (value instanceof byte[]) {
       return new BigInteger((byte[]) value);
+    } else if (value instanceof ByteBuffer) {
+      ByteBuffer byteBuffer = (ByteBuffer) value;
+      byte[] byteArray = new byte[byteBuffer.remaining()];
+      byteBuffer.get(byteArray);
+      return new BigInteger(byteArray);
```
VardhanThigle, line 128: [nit],  Consider if  it's possible to use [ByteBuffer.html#array](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#array--) here.

Something on the lines of
```java
return new BigInteger(((ByteBuffer)value).array())
```

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
```diff
@@ -331,7 +331,14 @@ private static Object handleSpannerColumnType(
           return null;
         }
         String hexEncodedString = valuesJson.optString(columnName);
-        return convertBinaryEncodedStringToByteArray(hexEncodedString);
+        return safeHandle(
+            () -> {
+              try {
+                return safeHandle(() -> convertBinaryEncodedStringToByteArray(hexEncodedString));
```
VardhanThigle, line 337: Could you please elaborate a bit on this frangment of change?

At the high level, it seems like we are trying to parse a hex encoded string, in case of any exception, we error log it and then rethrow it as `IllegalArgumentExeption`, we then catch that IllegalArgumentException and then parse it as blob, if this fails too, we again throw an illegal arument exception.

A few questions come to mind for someone who is taking a slightly new look at this:

Is there a value dependent decision taken here? Can we avoid that by looking at the source and destination types? 

Could we add more comments around this?

Can we not log errors where we don't fail? Debug is good, since it will be a toil to filter these errors coming from each rows to get to the real errors that the oncall has to deubg.

But before goind down that path, can we avoid the parsing exception by first principles if we know what are the source and desitnation types and encodings in a more deterministic way?

pawankashyapollion: we are doing it because of below reason
1. safeHandle(...) – Supplier Wrapper
Purpose: Ensures that the operation is executed safely, likely handling exceptions gracefully throw generic exception IllegalArgumentException.
Reason: Prevents unexpected crashes and centralizes error handling.
2. convertBinaryEncodedStringToByteArray(hexEncodedString)
Purpose: Converts a binary-encoded hexadecimal string into a byte array.
Reason: Some data formats require conversion binary string to bytes Array before processing.
Potential Failure: May throw IllegalArgumentException if the input is invalid.
3. Catching IllegalArgumentException Instead of Generic Exception
Purpose: Provides an alternative parsing strategy if the binary Encoded String conversion fails like data are in BASE64 Decoded String.
Reason: Instead of failing outright, it attempts another method (parseBlobType) to process the string.
Why IllegalArgumentException?: This exception is commonly thrown when input format is incorrect, ensuring robustness.
4. parseBlobType(hexEncodedString) – Fallback Handling
Purpose: Attempts to parse the input as a different type (likely a BLOB, Byte Array and String).
Reason: Ensures that even if the first conversion fails, an alternative method is attempted before giving up.
Why Needed?: The input may be in different formats, so multiple parsing strategies increase reliability.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
pawankashyapollion: Addressed @VardhanThigle

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
pawankashyapollion: @VardhanThigle there is not value decision taken here

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
shreyakhajanchi: There are still many places where we are using the term document instead of rows

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,929 @@
...
+      throw new RuntimeException(
+          "Failed to read from Cassandra table: " + ALL_DATA_TYPES_CUSTOM_CONVERSION_TABLE, e);
+    }
+
+    assertThat(rows).hasSize(2);
+    Row row = rows.iterator().next();
+    assertAll(
+        () -> assertThat(row.getString("varchar_column")).isEqualTo("SampleVarchar2"),
+        () -> assertThat(row.getByte("tinyint_column")).isEqualTo((byte) 127));
+  }
```
shreyakhajanchi, line 928: assert for all datatypes

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,929 @@
...
+  /**
+   * Writes a row containing all supported data types into the Spanner database.
+   *
+   * <p>This method creates and inserts a row into the {@code ALL_DATA_TYPES_TABLE} with various
+   * data types, including text, numerical, date/time, boolean, byte arrays, lists, sets, and maps.
+   * The values are set explicitly to ensure compatibility with Spanner's schema.
+   *
+   * <p>The following data types are included:
+   *
+   * <ul>
```
shreyakhajanchi, line 340: you can remove these verbose comments, the methods are self explainatory. Same comment for other places

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,929 @@
...
+                .isEqualTo("This is some sample text data for the text column."),
+        () ->
+            assertThat(row.getLocalTime("time_column"))
+                .isEqualTo(java.time.LocalTime.parse("12:30:00.000000000")),
+        () ->
+            assertThat(row.getInstant("timestamp_column"))
+                .isEqualTo(java.time.Instant.parse("2025-01-27T10:30:00.000Z")),
+        () ->
+            assertThat(row.getBigInteger("varint_column"))
+                .isEqualTo(java.math.BigInteger.valueOf(123456789L)));
```
shreyakhajanchi, line 687: I don't see an assertion for byte_column , can you please add assertion for all columns

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,929 @@
...
+            .set("bigint_column")
+            .to(9223372036854775807L)
+            .set("float_column")
+            .to(3.14159)
+            .set("double_column")
+            .to(2.718281828459045)
+            .set("decimal_column")
+            .to(new BigDecimal("12345.6789"))
+            .set("datetime_column")
+            .to(Value.timestamp(Timestamp.parseTimestamp("2025-01-27T10:30:00Z")))
```
shreyakhajanchi, line 413: can you modify this value to have micro second precision and assert against cassandra

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraResourceManager.java
```diff
@@ -0,0 +1,352 @@
...
+    this(
+        null,
+        new CassandraContainer<>(
+            DockerImageName.parse(builder.containerImageName).withTag(builder.containerImageTag)),
+        builder);
+  }
+
+  @VisibleForTesting
+  @SuppressWarnings("nullness")
+  CassandraResourceManager(
```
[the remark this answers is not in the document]
pawankashyapollion: @shreyakhajanchi let me check and update here

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: @shreyakhajanchi as both table is containing same record and we have already assert in other test so we are validating here for necessary changes to asset record is added properly

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,929 @@
...
+   *   <li>Inserts or updates a row in the "users2" table with an ID of 2.
+   *   <li>Executes a transactionally buffered insert/update operation in the "users" table with an
+   *       ID of 3, using a transaction tag for tracking.
+   * </ul>
+   *
+   * The transaction uses a Spanner client with a specific transaction tag
+   * ("txBy=forwardMigration").
+   */
+  private void writeBasicRowInSpanner() {
+    Mutation m1 =
```
shreyakhajanchi, line 239: can you please add an update and delete mutation also for the test

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraResourceManager.java
```diff
@@ -0,0 +1,352 @@
...
+   * Inserts the given Documents into a collection.
+   *
+   * <p>Note: Implementations may do collection creation here, if one does not already exist.
+   *
+   * @param tableName The name of the collection to insert the documents into.
+   * @param documents A list of documents to insert into the collection.
+   * @return A boolean indicating whether the Documents were inserted successfully.
+   * @throws IllegalArgumentException if there is an error inserting the documents.
+   */
+  public synchronized boolean insertDocuments(String tableName, List<Map<String, Object>> documents)
```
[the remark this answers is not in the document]
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraResourceManager.java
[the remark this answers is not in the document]
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
pawankashyapollion: @shreyakhajanchi address the same

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
shreyakhajanchi: i don't see an update mutation, can you add it as a part of same test

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
shreyakhajanchi: but the datatype conversion is from string type so it would be better to add assertions on all columns for atleast 1 row

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
shreyakhajanchi: you have modified inserts in different table and asserts in different table

shreyakhajanchi (commented):
The CassandraIT is failing in github action with error `Caused by: com.google.cloud.spanner.SpannerException: FAILED_PRECONDITION: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Wrong number of key parts for users. Expected: 1. Got: ["1","3"]` please check and fix the IT

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
shreyakhajanchi: the term is still being used in logs in read method

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: Sure @shreyakhajanchi

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
pawankashyapollion: @shreyakhajanchi we have fixed this issue

pawankashyapollion:
> The CassandraIT is failing in github action with error `Caused by: com.google.cloud.spanner.SpannerException: FAILED_PRECONDITION: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Wrong number of key parts for users. Expected: 1. Got: ["1","3"]` please check and fix the IT

@shreyakhajanchi we have identified and fixed the issue

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/CassandraSharedResourceManager.java
pawankashyapollion: @shreyakhajanchi address the same

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToPostgresDML.java
```diff
@@ -114,6 +114,10 @@ public String cleanDataTypeValueSql(
     if (dataType.startsWith("_")) {
       return convertJsonToPostgresArray(columnValue, dataType.toUpperCase(), columnName);
     }
+    // Arrays in Postgres are prefixed with underscore e.g. _INT4 for integer array.
```
shreyakhajanchi, line 117: Rebasing issue?

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -0,0 +1,1138 @@
...
+      throw new RuntimeException(
+          "Failed to read from Cassandra table: " + ALL_DATA_TYPES_CUSTOM_CONVERSION_TABLE, e);
+    }
+
+    assertThat(rows).hasSize(2);
+    Row row = rows.iterator().next();
+    LOG.info("[AssertStringToActualRowsInCassandraDB] row: {}", row.getFormattedContents());
+    assertAll(
+        () -> assertThat(row.getString("varchar_column")).isEqualTo("SampleVarchar2"),
+        () -> assertThat(row.getByte("tinyint_column")).isEqualTo((byte) 127),
```
shreyakhajanchi, line 1013: why are we doing getbytes for tinyint column?

pawankashyapollion: @shreyakhajanchi for fetching Tiny int it is more suitable way instead of int

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToPostgresDML.java
pawankashyapollion: @shreyakhajanchi Address the Rebase issue

shreyakhajanchi (approved):

PR #2167, opened 2025-02-06
Added VIRTUAL column support to PG dataflow

atask-g:
- added support for VIRTUAL keyword for PG non-stored columns
- added unit and integration tests
- ran manual tests in staging and prod

R:@n-d-joshi
R:@darshan-sj

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Column.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Column.java
@@ -140,7 +140,7 @@ public abstract class Column implements Serializable {
       if (isStored()) {
         appendable.append(" STORED");
       } else if (dialect() == Dialect.POSTGRESQL) {
-          appendable.append(" VIRTUAL");
+        appendable.append(" VIRTUAL");
       }
     }
     if (isPlacementKey()) {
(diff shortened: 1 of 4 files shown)


n-d-joshi:
LGTM on these changes. We will also need to have some tests for FTS - will follow up with you offline on those.

n-d-joshi (approved):

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Column.java
```diff
@@ -139,6 +139,8 @@ public void prettyPrint(Appendable appendable) throws IOException {
       appendable.append(" AS (").append(generationExpression()).append(")");
       if (isStored()) {
         appendable.append(" STORED");
+      } else if (dialect() == Dialect.POSTGRESQL) {
+        appendable.append(" VIRTUAL");
```
darshan-sj, line 143: VIRTUAL keyword is only supported for PG?

darshan-sj (commented):
No changes are required in InformationSchemaScanner?
Is the feature released in production?
Also, no keyword at the end of the expression currently is defaulted to VIRTUAL ? If so, I'm trying to understand why this change is needed.

atask-g:
Replying to previous comment inline:

* No changes are required in InformationSchemaScanner?
  * No, this is only needed in the pretty printing of the column to indicate a non-stored generated column. In INFORMATION_SCHEMA it's represented as isStored == false.
* Is the feature released in production?
  * Yes, I also confirmed by manually testing in prod.
* Also, no keyword at the end of the expression currently is defaulted to VIRTUAL ? If so, I'm trying to understand why this change is needed.
  * This change is only needed for PG, in GSQL the absence of STORED indicates non-stored. In PG, VIRTUAL indicates non-stored. So, if the STORED keyword is not specified, then the VIRTUAL keyword must be provided.

darshan-sj (approved):
LGTM

PR #2171, opened 2025-02-10
Fix issues for sequences and identity columns

hengfengli:
Fix a few issues related to sequences and identity columns: 
- `null` strings appear in identity columns. If a parameter is null, it should not appear in the output file. 
- A conversion is required from `BIT_REVERSED_POSITIVE_SEQUENCE` to `bit_reversed_positive` because the `identity_kind` in the information_schema table has a value `BIT_REVERSED_POSITIVE_SEQUENCE` which is different than `bit_reversed_positive`.
- `sequence_kind="default"` is not shown properly for the statement `CREATE SEQUENCE myseq`.


manitgupta (approved):
@darshan-sj Do we also test this functionality in an IT? Do we need to update a corresponding IT as well?

darshan-sj (commented):
Can you check ExportPipelineIT and ImportPipelineIT and please add a test case if it is possible to add for this edgecase.

darshan-sj (approved):
LGTM

PR #2181, opened 2025-02-12
Added SCRAM-SHA-512 support to dataflow kafka-to-bigquery

vgnanasekaran:
Added SCRAM-SHA-512 support to dataflow kafka-to-bigquery

### DIFF
--- a/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
+++ b/v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
@@ -98,14 +98,16 @@ public interface KafkaReadOptions extends PipelineOptions {
             KafkaAuthenticationMethod.APPLICATION_DEFAULT_CREDENTIALS),
         @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.SASL_PLAIN),
         @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.TLS),
+        @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.SASL_SCRAM_512),
         @TemplateParameter.TemplateEnumOption(KafkaAuthenticationMethod.NONE),
       },
       description = "Kafka Source Authentication Mode",
       helpText =
           "The mode of authentication to use with the Kafka cluster. "
               + "Use `KafkaAuthenticationMethod.NONE` for no authentication, `KafkaAuthenticationMethod.SASL_PLAIN` for SASL/PLAIN username and password, "
-              + "and `KafkaAuthenticationMethod.TLS` for certificate-based authentication. `KafkaAuthenticationMethod.APPLICATION_DEFAULT_CREDENTIALS` "
-              + "should be used only for Google Cloud Apache Kafka for BigQuery cluster, it allows to authenticate using application default credentials.")
+              + "`KafkaAuthenticationMethod.SASL_SCRAM_512` for SASL_SCRAM_512 authentication and `KafkaAuthenticationMethod.TLS` for certificate-based "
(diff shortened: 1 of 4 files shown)


vgnanasekaran:
@an2x As discussed, I have created the PR to support scram-sha-512 in Dataflow Kaka2BQ. Request you to review and share your comments.

Thank you,
Venkat

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
```diff
@@ -215,4 +217,62 @@ final class Offset {
   String getKafkaReadKeyPasswordSecretId();
 
   void setKafkaReadKeyPasswordSecretId(String sourceKeyPasswordSecretId);
+
```
an2x, line 220: Looks good, but can you please rename the new parameters from `getKafkaReadSASLScram512*`  to `getKafkaReadSaslScram*` (e.g. `getKafkaReadSASLScram512UsernameSecretId` to `getKafkaReadSaslScramUsernameSecretId`), and rename the corresponding variables in `KafkaConfig` accordingly?

The reason for "SASL" -> "Sasl" change is that by convention, in all the Java identifiers in this project only the first latter in each word is capitalized, even if this is an abbreviation.

The reason for removing "512" is to allow extending this to support SASL SCRAM 256 auth as well in the future (if needed) without the need to introduce another set of ``getKafkaReadSaslScram256*` parameters (we'll just re-use these).

an2x (changes requested):

an2x:
Note: the checks are failing because of some recent changes in GitHub actions. Please also update the PR to pick up the most recent changes from the main branch, this should've been fixed in https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/023b53e626ab1519aeda2fe9d8ee7a8326e53b0f.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
vgnanasekaran: @an2x thank you for reviewing the PR and sharing your feedback. My understanding is the parameter prompts in the UI is based on the parent name  and parentTriggerValues. For example if the chosen authentication mode is KafkaAuthenticationMethod.SASL_PLAIN then only two input fields will be displayed in the UI (Username and password). To address any differences in inputs, between SASL SCRAM 256  & SASL SCRAM 512, I thought of keeping them separate. If you think that is not required, I will rename the parameters as recommended.

vgnanasekaran:
@an2x per your feedback, I have updated the parameter names and also have merged the recent changes.

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
an2x: @vgnanasekaran It looks like one change is still missing: to rename "SASL" -> "Sasl" in the parameter names, could you please change that as well?
Sorry for nitpicking, but the parameter names are case-sensitive and this is inconsistent with pretty much every other parameter and may be confusing for the users.

an2x (changes requested):

v2/kafka-common/src/main/java/com/google/cloud/teleport/v2/kafka/options/KafkaReadOptions.java
vgnanasekaran: @an2x not a problem at all. I want to stick to the standards followed by your team.

vgnanasekaran:
@an2x thank you for your approval.

PR #2186, opened 2025-02-13
Change default shadow table prefix of reverse replication template

manitgupta:


manitgupta:
Test failures are unrelated to PR. Tests related to reverse replication have passed.

PR #2187, opened 2025-02-13
Remove sdk_container_image example from terraform.tfvars

manitgupta:


asthamohta (approved):

manitgupta:
This change is unrelated to the template code and does not impact how ITs will fail or succeed.

PR #2198, opened 2025-02-17
bug-fix: Use jdbc connection properties for reverse migration

asthamohta:


manitgupta:
Could you please update the docs as well with the example of how to use `connectionProperties`? I had raised https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2188/files earlier when we thought that this was a documentation bug to fix it but haven't merged it yet.

manitgupta (approved):

PR #2203, opened 2025-02-18
Support inf and -inf values

dhercher:
support inf and -inf datetime values


claudevdm:
Can you please clarify a couple of things:

1. The Datastream docs mention that Timestamps are stored as ISO-8601 strings , and I assume that means the dates are stored in the GCS avro file as an ISO-8601 string? https://cloud.google.com/datastream/docs/unified-types#unified-types
2. This PR looks like it is expecting the incoming timestamp to be of type Long. Am I missing something? At what point is the ISO-8601 string converted to a long? Or am I incorrect and the timestamps are not actually stored in ISO string format?

ravijaingit (approved):

PR #2205, opened 2025-02-19
Report Lineage for CsvToBigQuery template

Abacn:
R: @rohitsinha54

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
+++ b/v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
@@ -26,6 +26,7 @@ import java.nio.charset.Charset;
 import org.apache.beam.sdk.io.Compression;
 import org.apache.beam.sdk.io.FileIO;
 import org.apache.beam.sdk.io.FileIO.ReadableFile;
+import org.apache.beam.sdk.io.FileSystems;
 import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.ValueProvider;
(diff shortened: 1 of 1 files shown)


v1/src/main/java/com/google/cloud/teleport/templates/common/CsvConverters.java
```diff
@@ -248,6 +249,8 @@ record ->
         LOG.error("Headers do not match, consistency cannot be guaranteed");
         throw new RuntimeException("Could not read Csv headers: " + ioe.getMessage());
       }
+      // Report source Lineage
+      FileSystems.reportSourceLineage(filePath.getMetadata().resourceId());
```
rohitsinha54, line 253: Thank you for the change. 
I see above this support file pattern matching too so this can be used to read many csv files *.csv. 

I am hoping Filesystems limit of reporting directory for >100 trigger for this custom IO Thank you for the change. 
I see above this support file pattern matching too so this can be used to read many csv files *.csv. 

I am hoping Filesystems limit of reporting directory for >100 trigger for this custom IO. Is it possible to test it quickly by simulating it.

Abacn: that's true. Currently there is a guard on StringSetData's limit (1 MB) to limit number of records. Since this is known to be a batch Template this guard should be sufficient.

Of course one can do same thing as https://github.com/apache/beam/blob/06454e57745611eda4cda1f858daa58544c36b73/sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSourceTransform.java#L117

I tend to think it's unnecessary for this use case

rohitsinha54: I do agree with you that given the limit now on beam side and also on service side the need for limiting is less. 

This IO seem to be directly reading the files matched. I think if it is just a matter of putting that logic here then it will be good to have it here for consistency among IOs else it can be confusing to customers as "when i use csv io I see many more files" vs "when i use file IO i see less or directory".

Thank you very much for the change.

Abacn: Done, PTAL

PR #2208, opened 2025-02-20
fix: distribute load on multiple telport spanner instances

asthamohta:


cicd/cmd/run-it-tests/main.go
```diff
@@ -59,7 +65,7 @@ func main() {
 		mvnFlags.ThreadCount(4),
 		mvnFlags.IntegrationTestParallelism(3),
 		mvnFlags.StaticBigtableInstance("teleport"),
-		mvnFlags.StaticSpannerInstance("teleport"),
+		mvnFlags.StaticSpannerInstance(randomTeleportSpannerInstance),
```
manitgupta, line 68: I think this will mean **ALL** ITs for a given run will run on the same spanner instance. Is that what we want?
I thought we wanted to a random spanner instance per IT, similar to I think what we did with GCS buckets?

@darshan-sj / @shreyakhajanchi Please confirm.

shreyakhajanchi: Yes, we had finalized on randomly allocating a bucket out of 10 buckets for each of the IT. I think @asthamohta was  OOO at that time and might be unaware of the discussion, [PR link](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2223)

PR #2209, opened 2025-02-20
Adding support for Cassandra map

VardhanThigle:
Adding support for  Handling Cassandra `MAP<Primitive, Primitive>` types.
`Map<Primitive, Primitive>` is mapped to `Json{"String":"String"}`

TODO: Integration tests.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappings.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappings.java
@@ -19,11 +19,14 @@ import com.google.auto.value.AutoValue;
 import com.google.cloud.teleport.v2.source.reader.io.cassandra.rowmapper.CassandraFieldMapper;
 import com.google.cloud.teleport.v2.source.reader.io.cassandra.rowmapper.CassandraRowValueArrayMapper;
 import com.google.cloud.teleport.v2.source.reader.io.cassandra.rowmapper.CassandraRowValueExtractor;
+import com.google.cloud.teleport.v2.source.reader.io.cassandra.rowmapper.CassandraRowValueMapMapper;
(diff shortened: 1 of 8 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappings.java
```diff
@@ -108,6 +119,46 @@ private <T> void putSet(
...
+      ImmutableMap<String, Class> classMappings = typeToClassMappingBuilder().build();
+      ImmutableMap<String, CassandraFieldMapper<?>> fieldMappers = fieldMappingBuilder().build();
+      ImmutableMap<String, UnifiedTypeMapping> typeMappings = typeMappingBuilder().build();
+      ImmutableSet<String> primitiveTypes = classMappings.keySet();
+
+      for (String key : primitiveTypes) {
+        Class keyClass = classMappings.get(key);
+        for (String value : primitiveTypes) {
+          // The Map type as represented in Cassandra Schema
+          String mapType = "MAP<" + key.toUpperCase() + "," + value.toUpperCase() + ">";
```
Deep1998, line 141: Does cassandra support all primitive types as key and values? like are decimal/timestamp/date types valid key types?

Deep1998 (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/cassandra/mappings/CassandraMappings.java
VardhanThigle: Except for `Duration` all are valid. `Duration` is a newly added type.
Even for` Duration`  if there's a support in near future, our mapping is going to work, basically `getMap(class, class)` interface from `Cassandra` will be same . 
In our UT, we have all types covered as a key once and value once.

PR #2211, opened 2025-02-20
Set multi release manifest for classic template shadow jar

Abacn:
This fixes classic template running on Java 21, which causes same error as https://github.com/apache/beam/issues/33471


Abacn:
R: @shunping

liferoad (approved):

shunping (approved):
LGTM

PR #2219, opened 2025-02-26
Add logic to skip runnerV2 for the ITs

Rudra-Gujarathi:
This code adds the logic for skipping runner V2 for the integration test methods that are annotated with SkipRunnerV2Test in the JUnit categories or using system property in the profile.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
@@ -135,6 +135,7 @@ public abstract class TemplateTestBase {
   protected GcsResourceManager artifactClient;
 
   private boolean usingDirectRunner;
+  private boolean skipRunnerV2;
   protected PipelineLauncher pipelineLauncher;
   protected boolean skipBaseCleanup;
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -677,6 +677,10 @@
             <version>${surefire.version}</version>
             <configuration combine.self="override">
               <systemProperties>
+                <property>
+                  <name>skipRunnerV2Test</name>
+                  <value>true</value>
+                </property>
```
damccorm, line 683: As discussed offline, I don't think we need this property, we can just key off of the `SkipRunnerV2Test` category for all suites

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
```diff
@@ -493,8 +498,9 @@ protected LaunchInfo launchTemplate(
     String unifiedWorkerHarnessContainerImage =
         System.getProperty("unifiedWorkerHarnessContainerImage");
     if (System.getProperty("unifiedWorker") != null || unifiedWorkerHarnessContainerImage != null) {
```
damccorm, line 500: We won't want to append any of the experiments in this block if we're skipping runner v2 (these are all v2 adjacent features), so we should just be able to do the if condition at the top level:

`if (!skipRunnerV2 && (System.getProperty("unifiedWorker") != null || unifiedWorkerHarnessContainerImage != null))`

damccorm (commented):
Thanks for looking at this!

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
Rudra-Gujarathi: Resolved. Thank you!

damccorm (approved):
Thanks, LGTM!

damccorm (approved):
Thanks!

PR #2222, opened 2025-02-27
Update Dockerfile-template-yaml

svetakvsundhar:
This change adds JARs to the cache in the Dockerfile, so that a Dataflow pipeline doesn't need to access public maven during pipeline runtime.

Verified that the JARs are in the cache:

Screenshot 2025-03-04 at 9 04 08 PM

Pipeline expanding `ReadFromKafka` from cache:

Screenshot 2025-03-04 at 9 10 13 PM

### DIFF
--- a/plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
+++ b/plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
@@ -28,7 +28,10 @@ RUN python -m venv /venv \
     && mv /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
 
 # Cache provider environments for faster startup and expansion time
-RUN python -m apache_beam.utils.subprocess_server
+RUN mkdir -p /root/.apache_beam/cache/jars/
+RUN wget -O /root/.apache_beam/cache/jars/beam-sdks-java-io-expansion-service-${beamVersion}.jar https://repo1.maven.org/maven2/org/apache/beam/beam-sdks-java-io-expansion-service/${beamVersion}/beam-sdks-java-io-expansion-service-${beamVersion}.jar
+RUN wget -O /root/.apache_beam/cache/jars/beam-sdks-java-extensions-schemaio-expansion-service-${beamVersion}.jar https://repo1.maven.org/maven2/org/apache/beam/beam-sdks-java-extensions-schemaio-expansion-service/${beamVersion}/beam-sdks-java-extensions-schemaio-expansion-service-${beamVersion}.jar
+RUN wget -O /root/.apache_beam/cache/jars/beam-sdks-java-io-google-cloud-platform-expansion-service-${beamVersion}.jar https://repo1.maven.org/maven2/org/apache/beam/beam-sdks-java-io-google-cloud-platform-expansion-service/${beamVersion}/beam-sdks-java-io-google-cloud-platform-expansion-service-${beamVersion}.jar
 RUN python -m apache_beam.yaml.cache_provider_artifacts
 
 #============================================================#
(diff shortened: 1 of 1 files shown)


svetakvsundhar:
R: @damccorm 
R: https://github.com/derrickaw

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -28,6 +28,7 @@ RUN python -m venv /venv \
     && mv /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
 
 # Cache provider environments for faster startup and expansion time
+RUN python -m apache_beam.utils.subprocess_server
```
damccorm, line 31: How does this actually help? Unlike [cache_provider_artifacts](https://github.com/apache/beam/blob/c8172fbaadb4c11373baed9d2d48fc8175ed0699/sdks/python/apache_beam/yaml/cache_provider_artifacts.py#L46) this doesn't seem to execute anything on `__main__` - https://github.com/apache/beam/blob/master/sdks/python/apache_beam/utils/subprocess_server.py

damccorm: If the goal is to start/cache the things needed to run a yaml pipeline, should we instead just run a super simple yaml pipeline as part of this startup? That might be more robust anyways

svetakvsundhar: This was specific to the `beam-sdks-java-io-expansion-service` jar, that was being pulled specifically when executing a Beam YAML pipeline while reading from kafka. happy to sync offline if itd help

damccorm: Yeah, I think I understand the goal, my question is more how does this accomplish the goal? I don't think this will actually do anything, though I could be missing something.

I'm also wondering if a more robust solution would be to just run a very simple yaml pipeline which calls into java. That does a few things:

1) Should cache `beam-sdks-java-io-expansion-service`
2) Should do the same with any other artifacts that we might introduce
3) Provides a sanity check that this dockerfile does actually correctly support yaml

This could be a really small pipeline, e.g.

```
python -m apache_beam.yaml.main --yaml_pipeline="
pipeline:
  type: chain
  transforms:
    - type: Create
      name: Source
      config:
        elements:
        - {b: 1, c: 1.0}
    - type: Filter
      name: Transform
      config:
        keep:
          expression: c > 0
        language: java
    - type: LogForTesting
"
```

liferoad: `python -m apache_beam.utils.subprocess_server` does not download the jars. I suggest we just manually download jars (https://github.com/apache/beam/blob/master/sdks/java/expansion-service/container/expansion_service_config.yml) to the docker image we build here.

plugins/core-plugin/src/main/resources/Dockerfile-template-yaml
```diff
@@ -28,6 +28,10 @@ RUN python -m venv /venv \
     && mv /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
 
 # Cache provider environments for faster startup and expansion time
+RUN mkdir -p ~/.apache_beam/cache/jars
+RUN wget -O ~/.apache_beam/cache/jars/beam-sdks-java-io-expansion-service-2.63.0.jar https://repo1.maven.org/maven2/org/apache/beam/beam-sdks-java-io-expansion-service/2.63.0/beam-sdks-java-io-expansion-service-2.63.0.jar
```
liferoad, line 32: can we switch back to beamVersion?

svetakvsundhar: Done

derrickaw (approved):

liferoad (approved):

damccorm:
This LGTM - rerunning test suites to try to get a green signal

Abacn:
This breaks Beam release candidate validation as the artifact does not exist in maven central yet

PR #2227, opened 2025-03-03
Add promote artifact method in release plugin

Abacn:
Tested with

```
mvn verify -PtemplatesRelease \
  -DprojectId=... \
  -DbucketName=.... \
  -DlibrariesBucketName=.... \
  -DstagePrefix=moss-test-1 \
  -Dmaven.test.skip -T8 -e \
  -DartifactRegistry=<final registry> \
  -DstagingArtifactRegistry=<staging registry> \
  --settings=.mvn/settings.xml \
  -pl v2/streaming-data-generator -am
```

added a **stagingArtifactRegistry** parameter, if set, the image will be built under **stagingArtifactRegistry** before promote to **artifactRegistry** (note: if artifactRegistry is empty, default to gcr.io/<projectId>)

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
@@ -136,16 +136,31 @@ class PromoteHelper {
   /** Add "latest" tag after promotion. */
   private void addTag() throws IOException, InterruptedException {
     // TODO: remove this once copy tag is supported by promote API
-    String[] command =
-        new String[] {
-          "gcloud",
-          "artifacts",
-          "docker",
-          "tags",
-          "add",
-          String.format("%s@%s", targetPath, sourceDigest),
-          String.format("%s:latest", targetPath)
-        };
+    String[] command;
+    if (targetSpec.repository.endsWith("gcr.io")) {
+      // gcr.io repository needs to use `gcloud container` to add tag
+      command =
+          new String[] {
+            "gcloud",
+            "container",
+            "images",
+            "add-tag",
+            "-q",
+            String.format("%s@%s", targetPath, sourceDigest),
+            String.format("%s:latest", targetPath)
+          };
+    } else {
+      command =
+          new String[] {
+            "gcloud",
+            "artifacts",
+            "docker",
+            "tags",
+            "add",
+            String.format("%s@%s", targetPath, sourceDigest),
+            String.format("%s:latest", targetPath)
+          };
+    }
     TemplatesStageMojo.runCommandCapturesOutput(command, null);
   }
 
@@ -200,6 +215,11 @@ class PromoteHelper {
 
     public final @Nullable String imageName;
 
+    /**
+     * Construct an {@code ArtifactRegImageSpec} from an image url. Supported image urls include
+     * [region.]gcr.io/projectId[/...] and region-docker.pkg.dev/projectId[/...]. See {@code
+     * PromoteHelperTest#testArtifactRegImageSpec} for example image urls.
+     */
     public ArtifactRegImageSpec(String imagePath) {
       String[] segments = imagePath.split("/", 3);
       if (segments.length < 3) {
@@ -216,7 +236,7 @@ class PromoteHelper {
       }
       this.project = segments[1];
       if (segments[0].endsWith("gcr.io")) {
-        this.repository = "gcr.io";
+        this.repository = segments[0];
         this.imageName = segments[2];
         if ("gcr.io".equals(segments[0])) {
           this.location = "us";
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
@@ -1,0 +1,266 @@
+/*
+ * Copyright (C) 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


Abacn:
ImportPipelineIT.testPostgresImportPipeline[Staging] failed, not related to the change

Abacn:
R: @liferoad @damccorm

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1027,7 +1100,15 @@ static String generateFlexTemplateImagePath(
...
-                value + "/" + projectIdUrl + "/" + stagePrefix.toLowerCase() + "/" + containerName)
+                value.contains("gcr.io")
+                    ? value
+                        + "/"
+                        + projectIdUrl
+                        + "/"
+                        + stagePrefix.toLowerCase()
+                        + "/"
+                        + containerName
+                    : value + "/" + stagePrefix.toLowerCase() + "/" + containerName)
```
damccorm, line 1111: Can we rename `generateStagingFlexTemplateImagePath` to `generateImagePath` and then use it to construct these?

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1038,6 +1119,11 @@ static String generateFlexTemplateImagePath(
                 + containerName);
   }
 
+  static String generateStagingFlexTemplateImagePath(
+      String containerName, String stagingArtifactRegistry, String stagePrefix) {
```
damccorm, line 1123: Nit: this would probably read clearer if we took parameters in the same order that we use them below.

damccorm (commented):
Thanks!

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
```diff
@@ -0,0 +1,266 @@
...
+          "-nv",
+          String.format("--header=Authorization: Bearer %s", token),
+          "https://artifactregistry.googleapis.com/v1/" + operation
+        };
+
+    RetryPolicy<?> retry =
+        RetryPolicy.builder()
+            .handleIf(throwable -> throwable instanceof QueryOperationRunnable.RetryableException)
+            .withBackoff(Duration.ofSeconds(5), Duration.ofSeconds(30))
+            .withMaxRetries(5)
```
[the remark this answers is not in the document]
Abacn: From my testing it runs within seconds, and retry seems not necessary, but as this is technically a LRO, wrapped inside a retry is preferred in general

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
Abacn: This method mirrors `generateFlexTemplateImagePath` which is used generate the final image path so a "Staging" infix is added

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
Abacn: that's fair. The current order was due to `generateFlexTemplateImagePath` had same parameter order (after removing parameters that are not used in this method)

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
```diff
@@ -0,0 +1,266 @@
...
+
+  /** Artifact registry image spec. */
+  static class ArtifactRegImageSpec {
+    public final String project;
+    public final String repository;
+    public final String location;
+
+    public final @Nullable String imageName;
+
+    public ArtifactRegImageSpec(String imagePath) {
```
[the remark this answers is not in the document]
Abacn: added. linked to a test that had different kinds of imagePath

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -108,9 +112,27 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
...
 
+  /**
+   * artifact registry.
+   *
+   * <p>If not set, images will be built to [artifactRegion.]gcr.io/[projectId].
+   *
+   * <p>If set to "xxx.gcr.io", image will be built to xxx.gcr.io/[projectId].
+   *
+   * <p>Otherwise, image will be built to artifactRegion.
+   */
```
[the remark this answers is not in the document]
Abacn: added

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
damccorm: My point is that this is doing the same thing that `generateStagingFlexTemplateImagePath` is doing (combining a registry, path, and containerName), so it seems like it is something we should share.

Though I guess that leads to the question, why do we need this extra logic for the final path, but not the staging path? Aren't they really doing the same thing, just with a different artifact registry instance?

damccorm: In other words, instead of calling `generateStagingFlexTemplateImagePath(containerName, stagingArtifactRegistry, stagePrefix)`, wouldn't we get an equal/slightly more robust outcome by calling `generateFlexTemplateImagePath(containerName, stagingArtifactRegistry, stagePrefix)`?

Abacn: Thanks, consolidated two methods, and ran a mvn verify again confirmed it still work

PR #2230, opened 2025-03-04
Adding All Datatypes IT for Cassandra Migration

VardhanThigle:
# Adding All Datatypes IT for Cassandra Migration

Adds automated integration test for all Cassandra datatypes with min and max values for each primitive type (atleast once individually and once as a part of collections).

## Note
While integration testing, it was realized that we would need to stringify the types like date correctly when converting them to a map - this will be taken care as a part of TODO [b/401116246](https://b.corp.google.com/issues/401116246)


v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/SourceDbToSpannerITBase.java
```diff
@@ -56,6 +62,17 @@ public PostgresResourceManager setUpPostgreSQLResourceManager() {
     return PostgresResourceManager.builder(testName).build();
   }
 
+  public CassandraResourceManager setupCassandraResourceManager() {
+    /* The default is Cassandra 4.1 image. TODO: Explore testing with non 4.1 tags. */
+
+    /* Max Cassandra Keyspace is 48 characters. Base Resource Manager adds 24 characters of date-time at the end.
+     * That's why we need to take a smaller subsequence of the testId.
+     */
+    String uniqueId = testId.substring(0, Math.min(20, testId.length()));
```
darshan-sj, line 71: Ideally, the CassandraResourceManager should be aware of this and should take care of this stripping logic. 

I'm okay with this change as of now, can you also make this change in Cassandra RM in Apache Beam. That is strip the id if it is greater than 20?

darshan-sj (approved):
LGTM

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/SourceDbToSpannerITBase.java
VardhanThigle: I think as of now RR does not use the resource manager from beam directly. RR team is aware of this issue as well as the fix I had proposed them (in a call, as well as this PR)
As per my sync with the RR folks, they were looking for an example to get the beam's resource manager working - which is this one -:)

PR #2237, opened 2025-03-05
Add IF NOT EXISTS clause for spanner ddls used in ITs

Deep1998:
Add IF NOT EXISTS clause and enable DatastreamToSpannerIT.

SpannerResourceManager uses a retry policy when setting up IT resources. We have faced issues in the past where Spanner ddl execution returns resource exhausted followed by an alreadyExists error, indicating background completion of the ddl can happen.

We add if not exists clause to avoid ITs from failing due to this.


Deep1998:
Ran ITs 4 times, intermittent failures on unrelated issue.b/400992122

PR #2238, opened 2025-03-05
Reverse Replication - IT for Boundary values and extended map datatypes

taherkl:
This PR introduces integration tests to validate the conversion of boundary values and map data types from Spanner to Cassandra. The tests ensure that extreme values for various data types, including int, bigint, decimal, float, double, date, timestamp, UUID, and inet, are accurately transferred and preserved.

- Boundary Value Testing: Inserts maximum and edge-case values of various data types into Spanner and verifies correctness in Cassandra.
- Map Data Type Validations: Ensures correct transformation and mapping of complex data structures between the two databases.
- String-Based Representations: Confirms that string representations used in Spanner translate accurately into native Cassandra types.
- Data Integrity Checks: Compares stored and retrieved values across systems to maintain precision.
- Robust Assertions: Uses comprehensive assertions to detect discrepancies and prevent potential conversion issues.


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/codec/DurationCodec.java
```diff
@@ -0,0 +1,97 @@
+/*
```
shreyakhajanchi, line 1: What is the issue in using CqlDuration[https://docs.datastax.com/en/drivers/java/4.1/com/datastax/oss/driver/api/core/data/CqlDuration.html] instead of Java Duration ? Based on the [driver](https://docs.datastax.com/en/drivers/java/4.1/com/datastax/oss/driver/api/core/data/SettableByName.html) documentation this should be used.

pawankashyapollion: @shreyakhajanchi the Cassandra 4.X version does not support Duration Codec Directly so we need to attach Codec for that

shreyakhajanchi: It doesn't support Java Duration but it supports CqlDuration right?

taherkl: This is resolved.

shreyakhajanchi (commented):
Overall LGTM. Just a minor comment

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/connection/CassandraConnectionHelperTest.java
```diff
@@ -77,7 +82,11 @@ public void testInit_ShouldInitializeConnectionPool() {
         mockedCqlSession.when(CqlSession::builder).thenReturn(cqlSessionBuilder);
         when(cqlSessionBuilder.withConfigLoader(driverConfigLoader)).thenReturn(cqlSessionBuilder);
         when(cqlSessionBuilder.build()).thenReturn(cqlSession);
-
+        MutableCodecRegistry mockRegistry = mock(MutableCodecRegistry.class);
+        doNothing().when(mockRegistry).register((TypeCodec<?>) any());
+        DriverContext mockContext = mock(DriverContext.class);
+        when(mockContext.getCodecRegistry()).thenReturn(mockRegistry);
+        when(cqlSession.getContext()).thenReturn(mockContext);
```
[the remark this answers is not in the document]
taherkl: Will remove it

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -1162,4 +1197,340 @@ private void assertStringToActualRowsInCassandraDB() throws MultipleFailureExcep
             assertThat(row.getBigInteger("varint_column"))
                 .isEqualTo(java.math.BigInteger.valueOf(123456789L)));
   }
+
+  private void insertBoundaryValuesIntoSpanner() {
+    Mutation mutation =
```
shreyakhajanchi, line 1202: Can you cover MIN values also and add missing columns like max duration, etc. You can refer to the below link for comparision: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/CassandraAllDataTypesIT.java

taherkl: Hi @shreyakhajanchi Have added min value columns as well as missing max value columns. Please review

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -1162,4 +1197,383 @@ private void assertStringToActualRowsInCassandraDB() throws MultipleFailureExcep
...
+                .isEqualTo(java.time.LocalTime.parse("23:59:59.999999")),
+        () ->
+            assertThat(row.getLocalTime("min_time_column"))
+                .isEqualTo(java.time.LocalTime.parse("00:00:00.000000")),
+        () ->
+            assertThat(row.getInstant("timestamp_column"))
+                .isEqualTo(java.time.Instant.parse("9999-12-31T23:59:59.999Z")),
+        () ->
+            assertThat(row.getInstant("min_timestamp_column"))
+                .isEqualTo(java.time.Instant.parse("0001-01-01T00:00:00.000Z")),
```
shreyakhajanchi, line 1408: this behaviour of truncating precision is incorrect, if the value is outside the cassandra limits for any datatype,we should move the record to DLQ

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -1162,4 +1197,383 @@ private void assertStringToActualRowsInCassandraDB() throws MultipleFailureExcep
...
+          "Failed to read from Cassandra table: " + BOUNDARY_CONVERSION_TABLE, e);
+    }
+    assertThat(rows).hasSize(1);
+    Row row = rows.iterator().next();
+    assertAll(
+        // Basic Data Types
+        () -> assertThat(row.getString("varchar_column")).isEqualTo("SampleVarchar"),
+        () -> assertThat(row.getByte("tinyint_column")).isEqualTo(Byte.MAX_VALUE),
+        () -> assertThat(row.getByte("min_tinyint_column")).isEqualTo(Byte.MIN_VALUE),
+        () -> assertThat(row.getShort("smallint_column")).isEqualTo(Short.MAX_VALUE),
```
shreyakhajanchi, line 1345: instead of increasing the number of columns, insert two seperate records - one with min values and one with max as in example shared earlier

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
taherkl: spanner supports nanosecond precision, whereas Cassandra doesn't and stores it in milliseconds without throwing any error. In this case we will need to explicitly check the precision and handle for nanoseconds...will that be okay?

shreyakhajanchi: Let me check this and get back

shreyakhajanchi (approved):

PR #2239, opened 2025-03-06
Bump timeouts for tests involving FKs/interleaved dependenceis

Deep1998:
Our current streaming templates rely on retries for handling referential dependencies. This can often be unpredictable and we have faced multiple timeout issues in the past. Bumping up the timeouts for such tests should help us make the ITs more reliable.

### DIFF
--- a/v2/gcs-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSourceDbInterleaveIT.java
+++ b/v2/gcs-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSourceDbInterleaveIT.java
@@ -169,7 +169,7 @@ public class GCSToSourceDbInterleaveIT extends TemplateTestBase {
     PipelineOperator.Result result =
         pipelineOperator()
             .waitForCondition(
-                createConfig(writerJobInfo, Duration.ofMinutes(10)),
+                createConfig(writerJobInfo, Duration.ofMinutes(30)),
                 () -> jdbcResourceManager.getRowCount("child21") == 1);
     assertThatResult(result).meetsConditions();
     // child21 is the last row to be inserted. Hence by this time all other records must be updated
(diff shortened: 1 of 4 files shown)


v2/gcs-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSourceDbInterleaveIT.java
```diff
@@ -169,7 +169,7 @@ private void assertInsertedRowsInMySQL() throws InterruptedException {
     PipelineOperator.Result result =
         pipelineOperator()
             .waitForCondition(
-                createConfig(writerJobInfo, Duration.ofMinutes(10)),
+                createConfig(writerJobInfo, Duration.ofMinutes(30)),
```
darshan-sj, line 172: Should we pass the dlqRetryInterval minutes to something less than 10 mins, instead of increasing the timeout by so much?

darshan-sj (commented):
Looks good

v2/gcs-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/GCSToSourceDbInterleaveIT.java
Deep1998: After digging deeper, I found the dlq retry minutes are only used when running in retry mode, where it polls directly on gcs. In normal mode, there is no retry configured. The moment a notification is available in DLQ topic, it consumes.

darshan-sj (dismissed):
LGTM

VardhanThigle (approved):
Please keep a tab and revert if this increases the test times by a lot.

Deep1998:
Merging as failure is in an unrelated module

Abacn:
We cannot keep bumping timeout. Currently Java PR test takes 4h+ and timing out. Please fix the underlying race if it exists. Also, each template should only exercises handful of integration test, which is expensive. Please consider unit testing coverage in the case of testing a matrix of feature flags

PR #2241, opened 2025-03-07
Fixed default DLQ path for Cloud Spanner Change Streams to Big Query

dedocibula:
Prior implementation would result in every pipeline picking the same DLQ bucket within one project which would mean a past failed pipeline could be picked up by newly created pipeline for error change replay. Given that's dangerous we will make sure each pipeline has its own unique DLQ. Note that users can also supply their own pipeline using `deadLetterQueueDirectory` pipeline option - https://cloud.google.com/dataflow/docs/guides/templates/provided/cloud-spanner-change-streams-to-bigquery#optional-parameters


Abacn (approved):
Thank you!

Abacn:
there is failing test in unrelated module "datastream-to-sql", merging for now

```
2025-03-07T22:31:42.0648323Z [ERROR] Errors: 
2025-03-07T22:31:42.0650196Z [ERROR]   DataStreamToSQLIT.testDataStreamOracleToMySqlJsonGCSNotifications:146->simpleOracleToJdbcTest:222 » DatastreamResourceManager
```

PR #2254, opened 2025-03-18
Fix MOSS promotion for Python templates

Abacn:
- Use source project to build template for Python and Yaml template mirrors the change for Java

- Fix imageSpecFile override for non-Java flex image

Tested with testing project, job-builder-server and yaml template successfully pushed


damccorm (approved):
Thanks

PR #2255, opened 2025-03-19
Add the support of default_time_zone database option.

hengfengli:
The `default_time_zone` database option allows to specify a time zone in date/timestamp functions in queries when a time zone is unspecified. Currently, the default time zone is always `America/Los_Angeles`. With this support, we allow users to set other time zones, e.g., UTC.


hengfengli:
I have run `ExportPipelineIT` and `ImportPipelineIT` integration tests which work fine.

darshan-sj (approved):
Looks good

PR #2259, opened 2025-03-19
Force timeout for DataStreamToSQLIT

Abacn:
Also move pubsub-to-kafka to correct PR workflow

It is found datastream-to-sql is the module that currently timing out in Dataflow integration test


Abacn:
Workflow no longer timing out indefinetely

Abacn:
R: @tvalentyn

Rudra-Gujarathi (approved):

Issue #2191, opened 2025-02-13
[Bug]: Flaky Spanner Integration test due to "Could not initialize class com.google.spanner.v1.Session$LabelsDefault"


Abacn:
### Related Template(s)

Spanner templates

### Template Version

N/A

### What happened?

Spanner PR is flaky, integration test job failed launch. Example error:

```
2025-02-13 13:52:51.974 EST
java.lang.NoClassDefFoundError: Could not initialize class com.google.spanner.v1.Session$LabelsDefaultEntryHolder
2025-02-13 13:52:51.974 EST
at com.google.spanner.v1.Session.internalGetLabels(Session.java:147)
2025-02-13 13:52:51.974 EST
at com.google.spanner.v1.Session.getSerializedSize(Session.java:490)
2025-02-13 13:52:51.974 EST
at com.google.protobuf.CodedOutputStream.computeMessageSizeNoTag(CodedOutputStream.java:861)
2025-02-13 13:52:51.974 EST
at com.google.protobuf.CodedOutputStream.computeMessageSize(CodedOutputStream.java:641)
2025-02-13 13:52:51.974 EST
at com.google.spanner.v1.BatchCreateSessionsRequest.getSerializedSize(BatchCreateSessionsRequest.java:232)
2025-02-13 13:52:51.974 EST
at io.grpc.protobuf.lite.ProtoInputStream.available(ProtoInputStream.java:108)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.MessageFramer.getKnownLength(MessageFramer.java:204)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.MessageFramer.writePayload(MessageFramer.java:139)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.AbstractStream.writeMessage(AbstractStream.java:66)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.ForwardingClientStream.writeMessage(ForwardingClientStream.java:37)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.DelayedStream$6.run(DelayedStream.java:282)
2025-02-13 13:52:51.974 EST
at io.grpc.internal.DelayedStream.drainPendingCalls(DelayedStream.java:182)
2025-02-13 13:52:51.975 EST
at io.grpc.internal.DelayedStream.access$100(DelayedStream.java:44)
2025-02-13 13:52:51.975 EST
[...]

PR #2261, opened 2025-03-20
Update protobuf version

damccorm:
Try to mitigate https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/2191

### DIFF
--- a/v1/pom.xml
+++ b/v1/pom.xml
@@ -535,6 +535,11 @@
       <version>${datastax.query.builder.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>4.29.4</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


sakthivelmanii:
```
<dependency>
      <groupId>com.google.protobuf</groupId>
      <artifactId>protobuf-java</artifactId>
      <version>4.29.4</version>
</dependency>
```

Can you please add the protobuf runtime to the latest version in the dependencies in both pom.xml file?

damccorm:
> ```
> <dependency>
>       <groupId>com.google.protobuf</groupId>
>       <artifactId>protobuf-java</artifactId>
>       <version>4.29.4</version>
> </dependency>
> ```
> 
> Can you please add the protobuf runtime to the latest version in the dependencies in both pom.xml file?

I tried this - I'm a bit worried about diverging from Beam (https://github.com/apache/beam/blob/9e1cf5aaee4abc9c36c7425bcdc39cf6a9fdc75c/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L633), which is driven by the google bom, but as long as we're passing tests I think it is ok. We can remove this in the future.

damccorm:
Test suite failures are `datastream-to-sql` and `datastream-to-bigquery` both of which are failing on master as well - https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/java-pr.yml?query=branch%3Amain

PR #2262, opened 2025-03-20
[PropertyGraph] Explicitly output node reference key columns

smukil:
In the general case, ommitted node key columns are filled in via the element table primary keys.
However, it is always better to explicitly call out the key columns to avoid ambiguity in reference key selection in edge cases.


smukil:
@darshan-sj This is a tiny fix. Would it be possible to get a quick review for this? Ty!

darshan-sj (dismissed):
LGTM

PR #2268, opened 2025-03-22
Bug fixes of handling null values and added missing IT

akashthawaitcc:
1. Handled null values on custom transformation.
2. Fix: empty values are migrated as null.
3. Added IT of the above scenarios.


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
```diff
@@ -531,7 +531,7 @@ private static LocalDate parseDate(Object colValue) {
    * @return a {@link List} containing parsed values, or an empty list if {@code colValue} is null
    */
   private static List<?> parseCassandraList(String columnType, JSONArray colValue) {
-    if (colValue == null) {
+    if (colValue == null || colValue.isEmpty()) {
```
VardhanThigle, line 534: Shouldn't there be a difference in a list column having null value and an empty list?

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
```diff
@@ -176,6 +177,262 @@ public void spannerToCasandraSourceDbBasic() throws InterruptedException, IOExce
...
+      throws InterruptedException, IOException, MultipleFailureException {
+    assertThatPipeline(jobInfo).isRunning();
+    writeJSONEmptyInSpanner();
+    assertJSONEmptyRowInCassandraDB();
+  }
+
+  /** Insert Empty rows in Spanner. */
+  private void writeJSONEmptyInSpanner() {
+    Mutation.WriteBuilder mutationBuilder =
+        Mutation.newInsertOrUpdateBuilder(EMPTY_STRING_JSON_TABLE)
```
VardhanThigle, line 200: should we also test null ?

taherkl: @VardhanThigle Null has already been tested in a seperate test case spannerToCassandraSourceAllDataTypeConversionTest

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
pawankashyapollion: @VardhanThigle Maintaining a clear distinction between NULL values and empty collections (List, Set, Map) ensures data consistency and integrity.

For instance, in a nested structure, such as a Map<String, List<T>>, if one of the values is NULL, the key itself may be removed, leading to potential data loss. By explicitly storing empty collections, we preserve the key structure while ensuring that the absence of elements is intentional rather than an accidental omission.

This approach improves data readability, simplifies query handling, and prevents unintended inconsistencies in database operations.

VardhanThigle: I guess I am not fully understanding the code flow here, so please explain more.
Let's take a concrete example

I have a  column named `list_col` : `list<bigint>` on cassndra mapped as an `array<int64>` on spanner. 


My understanding of this code that that if the `list_col` itself is null, we will return an empty list here.

The QQ is that shouldn't we return a null in that case?

I am not talking about contents of the list being null at this moment.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDbIT.java
VardhanThigle: Thanks.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraTypeHandler.java
taherkl: Hi @VardhanThigle If you check the IT "writeAllDataTypeRowsInSpanner", we are inserting list_of_sets_column as null and also asserting as null on cassandra. Doesn't the code work as expected as per this IT?

VardhanThigle: Does `writeAllDataTypeRowsInSpanner` cover for spanner array to list/set mapping?

taherkl: Hi @VardhanThigle In the CassandraDao.java, we handle for null.

VardhanThigle: Thanks.

In that case would this part of code need to handle a null for colValue atall? or should we just add a `@NotNull` imported from  `javax.validation.constraints.NotNull` to the parameter?

pawankashyapollion: @VardhanThigle we can add **NotNull** Constraints but It does not create any impact as in It is always return NULL INSTANCE for those col Values which is null as in AutoValue does not support Null Value 

Here we are checking Instance of NULL to identified and assigned null for those keys which we received as null from Changes stream

VardhanThigle: If this function does not expect a null, then let's add a `checkNotNull` from  `com.google.common.base.Preconditions` on the argument at the beginning of the funciton.

Here is the intention:
In case of a (introduced in future) bug, if this functions ends up getting a null and we silently convert it to an empty list, that's data corruption.
In such a situation, the pipeline should fail with an assert but never cause silent data corruptions.

Making an empty `Array<int>` as `null` or null `Array<int>` as an empty Cassandra list would be data corruption.

taherkl: @VardhanThigle Addressed the changes as recommended. Thanks for detailed explaination

VardhanThigle (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DataStreamToSpannerIT.java
```diff
@@ -63,7 +63,6 @@
 import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
```
VardhanThigle, line 66: Why have we removed this line from `DataStreamToSpannerIT` ? 
It's luckily causing the unit tests to fail. Please revert it back.

taherkl: @VardhanThigle Spotless removed this.

VardhanThigle: Please add it back.

taherkl: @VardhanThigle Done. Please rerun the pipeline

VardhanThigle:
@taherkl 
The delta after fixing the review comments is not covered in UT,
Please cover the lines in the UT.

VardhanThigle (approved):

PR #2271, opened 2025-03-25
Don't assert on datetime row because this could vary by timezone

damccorm:
Right now, we're seeing failures in internal tests (internal cl/732015551) because of issues with timezones. I'm not actually sure where the disconnect is coming from, but it seems like we're assuming a certain timezone in these tests and are running into issues.

There's not really a reason for us to care about the datetime column in the output, so we can easily get rid of this to unblock these tests.


damccorm:
Test failures are from known broken datastream tests

Abacn (approved):

PR #2256, opened 2025-03-19
Add cleanup scripts for spanner tests

manitgupta:
Added cron jobs which periodically cleanup orphaned test infrastructure used by Spanner ITs.
This is expected to improve the stability of these tests and reduce the flakiness. AI from - b/issues/397641023/

### Testing 
Sample runs of this workflow - 

https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/cleanup-spanner-test-infra.yml

PR #2274, opened 2025-03-26
fix: distribute load on multiple telport spanner instances

asthamohta:


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
```diff
@@ -649,7 +651,15 @@ public Builder useStaticInstance() {
      */
     @SuppressWarnings("nullness")
     public Builder maybeUseStaticInstance() {
```
darshan-sj, line 653: If there is a UT for maybeUseStaticInstance. Please add a test case.

darshan-sj (commented):
Can you also update the instance list in the cleanup script added in --https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2256 ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
asthamohta: There is no test

asthamohta:
@darshan-sj also increasing the time for cleanup since the current 5hrs may clean up resources needed for LT which run for 14 hrs

PR #2275, opened 2025-03-26
Configuring numPartitions for Cassandra and allowing Spanner Mutation BatchSize to be configurable

VardhanThigle:
# Making Number of Partitions on CassandraIo configurable
## Overview
CassandraIO defaults to number of partitions equal to number of hosts on Cassandra cluster. This default behavior leads to many timeout errors when run at scale.
With this PR we are percolating the existing option of numPartitions to CassandraIO.
## TODOS
1. Post Cloud Next (Auto-Infer Cassandra numPartitions from size estimate tables).

### DIFF
--- a/v2/sourcedb-to-spanner/README_Sourcedb_to_Spanner.md
+++ b/v2/sourcedb-to-spanner/README_Sourcedb_to_Spanner.md
@@ -241,6 +241,8 @@ Please ensure that the Cassandra Cluster you are migrating from is healthy and t
 #### Prerequisite-4: Spanner
 You will need to provision a spanner database where you would like to migrate the data. The database would need to have tables with a schema that maps to the schema on the source.
 The tables which are present both on Spanner and Cassandra would be the ones that are migrated.
+#### Prerequisite-5: GCS
+You would need a GCS bucket to stage your build, driver configuration file, and provide an output directory for DLQs.
 ### Run Migration
 
 **Using the staged template**:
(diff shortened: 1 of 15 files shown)


v2/sourcedb-to-spanner/README_Sourcedb_to_Spanner.md
```diff
@@ -291,6 +300,9 @@ gcloud dataflow flex-template run "sourcedb-to-spanner-flex-job" \
   --project "$PROJECT" \
   --region "$REGION" \
   --network "$NETWORK"
+  --max-workers "$MAX_WORKERS"
+  --num-workers "$NUM_WORKERS"
```
Deep1998, line 304: Dont we need ot add the getBatchSizeForSpannerMutations flag here? Our recommendaiton should be 0?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -263,4 +263,15 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
...
   void setInsertOnlyModeForSpannerMutations(Boolean value);
+
+  @TemplateParameter.Text(
+      order = 22,
+      optional = true,
+      description = "BatchSize for Spanner Mutation.",
+      helpText =
+          "BatchSize in bytes for Spanner Mutations. if set less than 0, default of Apache Beam's SpannerIO is used, which is 1MB. Set this to 0 or 10, to disable batching mutations.")
+  @Default.Long(-1)
+  Long getBatchSizeForSpannerMutations();
```
Deep1998, line 274: We should keep an internal TODO to track the repercussions of switching to unbatched. If the throughput is same as Cassandra MAx rread throughput, maybe its worth making it a default config for all sources.

Deep1998 (approved):

v2/sourcedb-to-spanner/README_Sourcedb_to_Spanner.md
VardhanThigle: We will need it in the parameters.
I have added this to trouble shooting as if you see spiky throughput use this flag.
Default is `-1` which would mean we wont artificially set anything for beam and let beam's defaults take care.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
VardhanThigle: Added and shared over chat.

PR #2276, opened 2025-03-26
Remove duplicated cdc-agg module

Abacn:
checking template image staging it is found cdc-agg is built twice, staged twice, e.g.

gcr.io/dataflow-templates/2025-03-25-00_rc00/cdc-agg

checking https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/cdc-parent/cdc-agg/pom.xml

it just puts cdc-common, cdc-embedded-connector, cdc-change-applier together. However, where the template resides, cdc-change-applier module, also has cdc-common, cdc-embedded-connector

This module was introduced back in https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/d3aecd5dea08a6519a14298015396c83533e3ab6. From the description it intends to build a uber jar. However cdc-embedded-connector currently is already a uber jar


Abacn:
R: @damccorm

damccorm (approved):
Thanks!

PR #2278, opened 2025-03-27
[DO NOT MERGE] Validate Beam 2.64.0rc2

Amar3tto:


Abacn:
We need to bump slf4j to 2.x here:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/1df2af45a575461f5f8cb2aee16b23b3308f7703/pom.xml#L79

otherwise logs are missing for tests

Amar3tto:
Failing tests:
1. `PubSubToBigQueryYamlIT.testPubSubToBigQuery`
2. `YAMLTemplateIT.testSimpleCompositeSpec`
3. `YAMLTemplateIT.testSimpleCompositeSpecFile`

Also, made a fix for `TextImportPipelineIT` and `SpannerChangeStreamsToBigQueryIT`, should be successful after that.

v1/src/test/java/com/google/cloud/teleport/spanner/spannerio/LineageUtils.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Constructs an iterable of string parts representing a fully qualified name (FQN), based on the
+   * provided system, optional subtype, hierarchical segments, and an optional separator to be
+   * appended after the final segment. See {@link Lineage} class for details.
+   */
+  static Iterable<String> getFQNParts(
```
damccorm, line 45: Noting that previously we were relying on an internal method, so it is fine that it is no longer visible from a validation perspective. When we convert this to the actual upgrade PR, I think we should either:

1) Just get rid of this file entirely and hardcode the strings. If we're not actually relying on the lineage libraries to generate this, then it doesn't really get us much.
2) Continue to use `getFQNParts` (still available at [master](https://github.com/apache/beam/blob/ffd7d021b2e8dc48d1c6be0d3f88bc26d47a3416/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/Lineage.java#L68)) and just define `getFqName` using that.

I would probably recommend (1) since it doesn't rely on internal beam methods which are prone to change and we mostly want to check that the metric name itself doesn't change. @Abacn thoughts?

v1/src/test/java/com/google/cloud/teleport/templates/TextImportPipelineIT.java
```diff
@@ -130,6 +130,7 @@ public void testImportCsv() throws IOException {
             .addParameter("instanceId", googleSqlResourceManager.getInstanceId())
             .addParameter("databaseId", googleSqlResourceManager.getDatabaseId())
             .addParameter("spannerProjectId", PROJECT)
+            .addParameter("spannerHost", spannerHost)
```
damccorm, line 133: Hm, why did we need this? Did something change which now requires the host to be specified?

v2/cdc-parent/cdc-embedded-connector/src/main/java/com/google/cloud/dataflow/cdc/connector/App.java
```diff
@@ -104,9 +103,6 @@ public static DebeziumToPubSubDataSender setup(String[] args) throws Configurati
 
     // Printing the information about the bindings for SLF4J:
     Configuration config = getConnectorConfiguration(args);
-    final StaticLoggerBinder binder = StaticLoggerBinder.getSingleton();
-    System.out.println("Logger Binding: " + binder.getLoggerFactory());
-    System.out.println(binder.getLoggerFactoryClassStr());
```
damccorm, line 109: Similarly, why did we need this - was this causing errors?

v1/src/test/java/com/google/cloud/teleport/spanner/spannerio/LineageUtils.java
damccorm: I don't think this impacts release validation, so we can defer for the moment. It would be good to answer the other 2 questions before we finalize the release though

v2/cdc-parent/cdc-embedded-connector/src/main/java/com/google/cloud/dataflow/cdc/connector/App.java
damccorm: Sounds good, thanks! That makes sense and should be fine for validation at least.

v1/src/test/java/com/google/cloud/teleport/templates/TextImportPipelineIT.java
Amar3tto: The `google_cloud_spanner_version` has changed from 6.88.0 to 6.89.0.
In the older version there is a `SpannerOptions.CLOUD_SPANNER_HOST_PATTERN.matcher(host).matches()` which causes NPE if host is empty.
I think instead of this change I should just update version to the 6.89.0 for the DataflowTemplates as well.

v1/src/test/java/com/google/cloud/teleport/spanner/spannerio/LineageUtils.java
Abacn: This is due to https://github.com/apache/beam/pull/33891 changed Lineage method name, and in forked SpannerIO we could sync change with that PR also (or fork a LineageUtils like here)

It's technically a breaking change but because Lineage is marked as internal so its fine in terms of release validation

v1/src/test/java/com/google/cloud/teleport/templates/TextImportPipelineIT.java
Amar3tto: @damccorm I set 6.89.0, 
Could you please approve workflows

damccorm: Sure - I think the problem is actually https://github.com/apache/beam/pull/34489 though

Amar3tto:
There were Spanner IT test failures because of slf4j / logback conflicts. Tried to exclude slf4j dependencies, seems didn't help, need to investigate further

liferoad:
Validation is done. Close this.

PR #2281, opened 2025-03-28
fix: Sharded DB behavior should be the same regardless of shard count

asthamohta:


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/Shard.java
```diff
@@ -133,6 +136,14 @@ public Map<String, String> getDbNameToLogicalShardIdMap() {
     return dbNameToLogicalShardIdMap;
   }
 
+  public Boolean getIsShardedMigration() {
```
VardhanThigle, line 139: Context:
We are changing the `Shard` class which is shared between tempaltes.
What are the implications of setting this to true or false in the context of bulk?
If this is flag is entirely ignored by any template, could we document this in the readme as well as the shard.java file as a commendtr?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/ShardFileReader.java
```diff
@@ -53,21 +60,124 @@ public ShardFileReader(ISecretManagerAccessor secretManagerAccessor) {
     this.secretManagerAccessor = secretManagerAccessor;
   }
 
-  public List<Shard> getOrderedShardDetails(String sourceShardsFilePath) {
+  public static class ShardConfig {
```
manitgupta, line 63: I am confused, why did this file need to change? There is nothing that should change with respect to accepting the shard configuration from the user? 
I thought the only thing we were changing was adding a template level flag?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -526,19 +527,22 @@ public static PipelineResult run(Options options) {
...
 
     } else {
       CassandraConfigFileReader cassandraConfigFileReader = new CassandraConfigFileReader();
       shards = cassandraConfigFileReader.getCassandraShard(options.getSourceShardsFilePath());
+      shardConfig = new ShardConfig(shards, false);
       LOG.info("Cassandra config is: {}", shards.get(0));
       shardingMode = Constants.SHARDING_MODE_SINGLE_SHARD;
     }
-    if (shards.size() == 1) {
+    if (shards.size() == 1 && !shardConfig.isShardedMigration()) {
```
manitgupta, line 545: I think this needs to be a template level flag and not a shard level configuration..

PR #2284, opened 2025-03-28
Surface error on cloud build

Abacn:
These errors were silenced to make release workflow pass. However, it also ignores real build errors and caused confusion in, e.g. #2278

The reason release workflow nonzero retval was the service account used on the production project has only limited permission, and error shown below:

```
[INFO] ERROR: (gcloud.builds.submit) 
[INFO] The build is running, and logs are being written to the default logs bucket.
[INFO] This tool can only stream logs if you are Viewer/Owner of the project and, if applicable, allowed by your VPC-SC security policy.
[INFO] 
[INFO] The default logs bucket is always outside any VPC-SC security perimeter.
[INFO] If you want your logs saved inside your VPC-SC perimeter, use your own bucket.
[INFO] See https://cloud.google.com/build/docs/securing-builds/store-manage-build-logs.
```

otherwise, the return value is zero here, for test workflow (ran in different gcp project and permission are granted)

Solution: https://github.com/google-github-actions/setup-gcloud/issues/281#issuecomment-1369620825


damccorm (approved):
Thanks!

PR #2286, opened 2025-03-31
Do not set null Spanner Host in SpannerChangeStreamsToBigQuery Templates

Abacn:
See https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2278#discussion_r2021054220

only this template is affected because SpannerChangeStreamsToBigQueryOptions getSpannerHost defaults to null, while other templates they default to "https://spanner.googleapis.com"

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/spannerio/SpannerAccessor.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/spannerio/SpannerAccessor.java
@@ -223,7 +223,10 @@ public class SpannerAccessor implements AutoCloseable {
     }
     ValueProvider<String> host = spannerConfig.getHost();
     if (host != null) {
-      builder.setHost(host.get());
+      String hostValue = host.get();
+      if (hostValue != null && !hostValue.trim().isEmpty()) {
+        builder.setHost(hostValue);
+      }
     }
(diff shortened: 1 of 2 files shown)


Abacn:
R: @damccorm

damccorm (changes requested):
Actually, could we update https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/086332d0a9915efee9963a484c8450434be0371a/v1/src/main/java/com/google/cloud/teleport/spanner/spannerio/SpannerAccessor.java#L226 as well?

v1/src/main/java/com/google/cloud/teleport/spanner/spannerio/SpannerAccessor.java
```diff
@@ -223,7 +223,10 @@ static SpannerOptions buildSpannerOptions(SpannerConfig spannerConfig) {
     }
     ValueProvider<String> host = spannerConfig.getHost();
     if (host != null) {
-      builder.setHost(host.get());
+      String hostValue = host.get();
+      if (hostValue != null && !hostValue.trim().isEmpty()) {
+        builder.setHost(hostValue);
```
damccorm, line 228: Lets get a decision on https://github.com/apache/beam/pull/34489#discussion_r2021280649 before moving forward here

cc/ @nielm

damccorm (approved):
LGTM

PR #2287, opened 2025-03-31
Test adjustments

Abacn:
Disable test failures due to #2288 #2289 . Otherwise lack of green signal causes more test failures spawning 

* spanner and kafka pr will run on main pom.xml change

* ~Bump DatastreamToSQL test timeout and consolidate test scenario~

hopefully this could fix tests currently failing on main without introducing more overhead

when there are n feature flags, we cannot afford to setup O(2**n) integration tests. Do O(n) tests with each covers one feature flag instead

### DIFF
--- a/.github/workflows/kafka-pr.yml
+++ b/.github/workflows/kafka-pr.yml
@@ -22,6 +22,7 @@ on:
     - 'main'
     paths:
       # Template wide common modules
+      - 'pom.xml'
       - 'v2/common/**'
       - 'v2/datastream-common/**'
       - 'it/google-cloud-platform/**'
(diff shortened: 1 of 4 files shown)


Abacn:
still failing with increased timeout, suggests test stuck wasn't due to test timeout

Abacn:
R: @tvalentyn

tvalentyn (approved):

Abacn:
actually checked cloud console, oracletobigquery took ~25 minutes to run. which implies we probably should bump time out to 30 min.

will follow up in next PR

Abacn:
Going to merge this as this does not change any code logic

Issue #2279, opened 2025-03-27
[Task]: Consolidate v2 submodules by source&sink type


Abacn:
### Related Template(s)

Flex

### What feature(s) are you requesting?

As number of templates keep growing, build and release workflow takes longer and longer to run. After #2157, only modules contains templates will build shaded jar. This saves build time, however, in release time there is not only build jar involved, but vulnerability scan and SBOM generation, etc for each artifact released. As a result the internal release workflow is hitting the timeout limit.

Currently the v2/submodules are named after source-to-sink. Further consolidation is possible for modules that contain templates of common dependencies. For example, kafka-to-bigquery, kafka-to-gcs, kafka-to-pubsub all involves beam-io-kafka and beam-io-gcp. pubsub-to-kafka and kafka-to-pubsub are essentially same dependencies. In their images, only entry point differs.

In the short term, we should consolidates submodules by source&sink type, for example, all modules aforementioned can be solidated into a single "kafka-and-googlecloud" module. And so on.

In the long term, we may want to make multiple templates pin to single image container image, to reduce the number of container images release that is linear to the time cost of each release workflow

Abacn:
Current time for each module:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Google Cloud Dataflow Templates 1.0-SNAPSHOT:
[INFO] 
[INFO] Google Cloud Dataflow Templates .................... SUCCESS [ 26.110 s]
[INFO] Templates Metadata Module .......................... SUCCESS [ 57.807 s]
[INFO] Templates Plugins Project .......................... SUCCESS [  7.022 s]
[INFO] Templates Core Plugin .............................. SUCCESS [  9.892 s]
[INFO] Templates Maven Plugin ............................. SUCCESS [02:39 min]
[INFO] integration-testing-lib ............................ SUCCESS [01:59 min]
[INFO] it-google-cloud-platform ........................... SUCCESS [07:31 min]
[INFO] it-datadog ......................................... SUCCESS [02:39 min]
[INFO] it-conditions ...................................... SUCCESS [  3.701 s]
[INFO] it-jdbc ............................................ SUCCESS [ 41.001 s]
[INFO] it-splunk .......................................... SUCCESS [ 42.357 s]
[INFO] lt ................................................. SUCCESS [  2.318 s]
[INFO] observability ...................................... SUCCESS [  1.108 s]
[...]

PR #692, opened 2023-05-11
Migrate JdbcToBigQuery and JdbcToPubsub templates from DynamicJdbcIO to JdbcIO.

pranavbhandari24:

PR #2293, opened 2025-04-01
Move jdbc templates out of googlecloud-to-googlecloud to renamed jdbc-and-googlecloud

Abacn:
Part of #2279

googlecloud-to-googlecloud currently has 20+ template and is the bottleneck of release workflow

pubsub-to-jdbc doesn't belong to googlecloud-to-googlecloud

Also deleted DynamicJdbcIO.DynamicDataSource as JdbcIO already supports load driverJar at runtime (and used by other JdbcIO templates)

* Eliminate JdbcIO related dependencies in googlecloud-to-googlecloud Templates


Abacn:
There is a single test failure, not related to this change: PubSubToBigQueryAutoIT time out

Abacn:
R: @ahmedabu98  could you please take a look? Thanks! Also R: @an2x

ahmedabu98 (approved):
I'm not familiar with the differences between DynamicDataSource and JDBC's data source, but assuming they're interchangeable, this LGTM

Abacn:
Thanks. DynamicJdbcIO was forked from Beam JdbcIO. There was efforts to unfork it: #692 and done for read transforms. The write transform remains because JdbcIO does not support DLQ at that time. A future clean up would be move it to use Beam builtin DLQ functionality

PR #2295, opened 2025-04-02
feat: add interleave in feature to optimise bulk migration

asthamohta:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
```diff
@@ -75,32 +75,6 @@ public static void cleanUp() throws IOException {
     ResourceManagerUtils.cleanResources(spannerResourceManager);
   }
 
-  @Test
-  public void testInterleaveInsertChildBeforeParent() {
```
bharadwaj-aditya, line 79: This test should still exist for scenarios where people use interleave in with parent.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
```diff
@@ -125,30 +99,6 @@ public void testFKInsertChildBeforeParent() {
     assertSpannerExceptionClassification(exception, RETRYABLE_ERROR, actualTag);
   }
 
-  @Test
-  public void testUniqueIndexError() {
```
bharadwaj-aditya, line 129: why is this removed ?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
```diff
@@ -228,24 +178,6 @@ public void testColumnNotFound() {
     assertSpannerExceptionClassification(exception, PERMANENT_ERROR, actualTag);
   }
 
-  @Test
-  public void testInterleavingDeleteParentWhenChildExists() {
```
bharadwaj-aditya, line 232: same comment as above. Interleave with parent should still be supported.

v2/datastream-to-spanner/src/test/resources/DataStreamToSpanner100GbLT/spanner-schema-interleaving.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS person2 (
      FK_ID2 INT64,
      ID INT64 NOT NULL,
 ) PRIMARY KEY(ID, FK_ID2),
-  INTERLEAVE IN PARENT person1 ON DELETE CASCADE;
```
bharadwaj-aditya, line 13: Please add tests for interleave in (without parent).

The existing tests for interleave in with parent should continue to exist

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Ddl.java
```diff
@@ -98,7 +99,8 @@ public List<String> tablesReferenced(String tableName) {
       throw new IllegalStateException(
           "attempting to fetch table which does not exist in spanner:" + tableName);
     }
-    if (table.interleaveInParent() != null) {
+    if (table.interleaveInParent() != null
+        && !Objects.equals(table.interleaveType(), new String("IN"))) {
```
bharadwaj-aditya, line 103: we need to support both types of interleaving here.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -145,9 +145,12 @@ private void listTables(Ddl.Builder builder) {
       String tableName = resultSet.getString(0);
       String parentTableName = resultSet.isNull(1) ? null : resultSet.getString(1);
       String onDeleteAction = resultSet.isNull(2) ? null : resultSet.getString(2);
+      String interleaveType = resultSet.isNull(3) ? null : resultSet.getString(3);
 
       // Error out when the parent table or on delete action are set incorrectly.
-      if (Strings.isNullOrEmpty(parentTableName) != Strings.isNullOrEmpty(onDeleteAction)) {
+      if (interleaveType != null
```
bharadwaj-aditya, line 151: this class is used in the import export flow as well. We can not assume only 1 type of interleaving here.

bharadwaj-aditya (changes requested):
The changes need to add support for INTERLEAVE IN - not change INTERLEAVE IN PARENT to INTERLEAVE IN.

Both need to be supported as they are possible cases in the migration flow.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Ddl.java
asthamohta: This ensures that if it is INTERLEAVE IN then dont a dd a parent, else add a parent. Support for both types will work

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/InformationSchemaScanner.java
asthamohta: Again as stated above, the code flow will continue to be the same for INTERLEAVE IN PARENT

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
```diff
@@ -101,6 +101,28 @@ public void testInterleaveInsertChildBeforeParent() {
...
+            .to("Child")
+            .build();
+    try {
+      spannerResourceManager.writeInTransaction(List.of(mutation));
+    } catch (SpannerException e) {
+      exception = e;
+      actualTag = DatastreamToSpannerExceptionClassifier.classify(SpannerExceptionParser.parse(e));
+    }
+    Assert.assertNull(exception);
+  }
```
manitgupta, line 124: `actualTag` should also be asserted right?

v2/datastream-to-spanner/src/test/resources/DataStreamToSpanner100GbLT/spanner-schema-interleaving.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS person2 (
...
+  INTERLEAVE IN person1;
```
manitgupta, line 13: Is there a separate test which preserves the FKA - `ON DELETE CASCADE`? Can you point me to it?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Table.java
```diff
@@ -117,14 +121,19 @@ private void prettyPrintPg(
...
           .append(" \nINTERLEAVE IN PARENT " + identifierQuote)
           .append(interleaveInParent())
           .append(identifierQuote);
       if (onDeleteCascade()) {
         appendable.append(" ON DELETE CASCADE");
       }
+    } else if (interleaveInParent() != null && Objects.equals(interleaveType(), "IN")) {
+      appendable
+          .append(" \nINTERLEAVE IN " + identifierQuote)
+          .append(interleaveInParent())
```
manitgupta, line 135: This seems a bit off...this case is for `INTERLEAVE IN` branch right? Why does it contain a reference to `interleaveInParent()`? 

I think you might be reusing stuff, but this is making it confusing. Can we remedy this? For example, can we simplify to have -

`public abstract String interleaveIn();` instead of the `interleaveType` bit? Alternatively, we can keep `interleaveType()` but then we need to rename `interleaveInParent` I think.

v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbInterleaveMultiShardIT/spanner-schema.sql
```diff
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS child21 (
   in_ts TIMESTAMP,
   migration_shard_id STRING(50),
 )  PRIMARY KEY(id, child_id),
- INTERLEAVE IN PARENT parent2 ON DELETE CASCADE;
+ INTERLEAVE IN parent2;
```
manitgupta, line 32: Same as earlier, should this schema be modified in place, or does there need to be another table added which contains `INTERLEAVE IN` while preserving the earlier `INTERLEAVE IN PARENT parent2 ON DELETE CASCADE` ?

v2/datastream-to-spanner/src/test/resources/DataStreamToSpanner100GbLT/spanner-schema-interleaving.sql
asthamohta: We will only do LTs for Interleave IN as discussed. There is a FK LT but not with ON DELETE CASCADE: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/datastream-to-spanner/src/test/resources/DataStreamToSpanner100GbLT/spanner-schema-foreign-key.sql

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Table.java
asthamohta: Renaming this to reduce confusing

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
asthamohta: actualTag is set in the exception flow. There is not exception

v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbInterleaveMultiShardIT/spanner-schema.sql
asthamohta: Added

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierIT.java
manitgupta: That does not matter I think, we should be doing something like this -

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/e0d274c3bd7f2eb4fb343daeb64775ddc668d6ee/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/spanner/DatastreamToSpannerExceptionClassifierTest.java#L195


WDYT? This example seems to assert the output of the error classifier?

manitgupta (commented):
Some minor updates to ITs are needed; discussed offline

manitgupta:
The IT suite seems to be failing - https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2295/checks?check_run_id=41661962671

Can you please take a look?

PR #2298, opened 2025-04-02
Skip staging part of AR path for job-builder

damccorm:
When we're only staging the image (job builder), having the postfix doesn't get us as much; right now this is all config driven from the UI side. This gives us the option of just pinning to latest for the image rather than needing to manually update every time, and we can still handle rollout safety by pinning to image shas if needed.


Abacn (approved):
Thanks, this LGTM

> we can pin SHAs

Just wondering if each image also published with "latest" and a version tag, then user can pin a tag rather than need to look up SHAs. This is a separate improvement can be put to backlog

damccorm:
> Thanks, this LGTM
> 
> > we can pin SHAs
> 
> Just wondering if each image also published with "latest" and a version tag, then user can pin a tag rather than need to look up SHAs. This is a separate improvement can be put to backlog

Yep, agreed, though users here just means the Dataflow service. We could do something like this eventually if we want an easier slow rollout

PR #2310, opened 2025-04-10
Reverse replication low priority integration tests

akashthawaitcc:
This PR include

- Table Name Length - 1 to 63 characters (MySQL)
- Column Name Length - 1 to 63 characters (MySQL)
- Table Name Length - 1–42 characters (Cassandra)
- Column Name Length - 1–37 characters (Cassandra)


akashthawaitcc:
> Why are we creating new ITs ? we should reuse the existing ITs and add these tables to it. Please reuse `SpannerToCassandraSourceDbIT` and `SpannerToSourceDbIT` for the same

@shreyakhajanchi We kept it separate because each test has a different schema. For Mysql, we can't because of the different session files; if we try to do so, it might break others.

shreyakhajanchi:
> > Why are we creating new ITs ? we should reuse the existing ITs and add these tables to it. Please reuse `SpannerToCassandraSourceDbIT` and `SpannerToSourceDbIT` for the same
> 
> @shreyakhajanchi We kept it separate because each test has a different schema. For Mysql, we can't because of the different session files; if we try to do so, it might break others.

You can regenerate the updated session file with new and old table in SMT right?

VardhanThigle (dismissed):

VardhanThigle:
Please debug the IT failures before merge.

taherkl:
> Please debug the IT failures before merge.

IT failures has been fixed. Please rerun the pipeline.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbIT.java
```diff
@@ -206,4 +217,36 @@ private void assertRowInMySQL() throws InterruptedException {
     assertThat(rows.get(0).get("name")).isEqualTo("FF");
     assertThat(rows.get(0).get("from")).isEqualTo("AA");
   }
+
+  @Test
+  public void spannerToMySQLSourceDbMaxColAndTableNameTest()
```
VardhanThigle, line 222: Having multiple Tests share the same dataflow pipeline is inherently flaky.
Just in case they don't schedule in parallel, the pipeline would be terminated causing flakiness.

VardhanThigle (approved):

PR #2315, opened 2025-04-11
Fixing flaky Spanner import export tests

darshan-sj:
Added DROP IF EXISTS clauses for all create statements in the Spanner Import export integration tests.


v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
```diff
@@ -198,70 +201,23 @@ private void testSpannerToGCSAvroBase(
       Function<PipelineLauncher.LaunchConfig.Builder, PipelineLauncher.LaunchConfig.Builder>
           paramsAdder)
       throws IOException {
-    // Arrange
     String setDefaultTimeZoneStatement =
         "ALTER DATABASE db SET OPTIONS (default_time_zone = 'UTC')";
```
shreyakhajanchi, line 205: nit: We should move this statement as well to `spanner-gsql-ddl.sql`

v1/src/test/java/com/google/cloud/teleport/templates/SpannerToTextIT.java
```diff
@@ -76,15 +76,17 @@ private void testSpannerToGCSTextBase(
           paramsAdder)
       throws IOException {
     // Arrange
-    String createTableStatement =
+    List<String> statements = new ArrayList<>();
```
shreyakhajanchi, line 79: why not create sql files for these as well to keep the tests consistent? Also, why is there a difference in naming convention for table names in the templates test vs spanner tests given that we are using fresh database every time?

shreyakhajanchi (approved):
LGTM. Left a couple of nits.

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
darshan-sj: UpdateDDLStatements call can fail in the middle, while processing many ddl statements and it can happen that some of the statements would have been applied. Now, when we get the RESOURCE_EXHAUSTED, we will re-apply all these ddl statements. And if then if this set default time zone statement is run on a non-empty db (containing some tables) this statement it will fail. Hence we have to apply it separately.

v1/src/test/java/com/google/cloud/teleport/templates/SpannerToTextIT.java
darshan-sj: The CSV import export ITs are simple and had only 1 or 2 DDL statements. Creating a sql file and loading it seemed like an overkill. We can move it to resources folder if it gets big.

darshan-sj:
The Integration test failures are not related to this PR.

PR #2321, opened 2025-04-15
fix: ensure that default spanner API is batch-spanner.googleapis.com

### LINKED ISSUE #2304: [Bug]: Spanner-related templates use incorrect host endpoints
### Related Template(s)

All spanner related templates

### Template Version

all

### What happened?

When a template that uses spanner does not have a hostname specified, it uses `spanner.googleapis.com`

This should be `batch-spanner.googleapis.com` 

see: https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerConfig.java#L45C1-L46C1

### Relevant log output

```shell

```

nielm:
Some of the dataflow templates use the incorrect endpoint for batch spanner actions.

This PR renames the defaults from spanner.googleapis.com to batch-spanner.googleapis.com to match the default host in Beam SpannerIO

Fixes #2304


derrickaw:
Hi @nielm, is this PR still valid?

derrickaw:
Reopen if still valid. Thanks

Issue #2325, opened 2025-04-17
[Bug]: TextImportPipelineIT staging tests permared


Abacn:
### Related Template(s)

TextImportPipeline

### Template Version

N/A

### What happened?

These tests are currently permared. Error:

```
org.apache.beam.it.gcp.spanner.SpannerResourceManagerException: Failed to execute statement.
	at org.apache.beam.it.gcp.spanner.SpannerResourceManager.executeDdlStatements(SpannerResourceManager.java:330)
	at com.google.cloud.teleport.templates.TextImportPipelineIT.testImportCsvToPostgres_UUID(TextImportPipelineIT.java:480)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
[...]

PR #2322, opened 2025-04-15
Regenerate requitements to fix yaml template

Abacn:

PR #2326, opened 2025-04-17
Disable Spanner staging test

Abacn:
Mitigate #2325

Currently Java PR run two set of spanner test for v1 spanner templates. This makes Java PR more noisy. In particular, TextImportPipelineIT staging test has been red for a week.

Disable spanner staging test altogether at once and re-enable them as needed.

CC: @darshan-sj


Abacn:
cancelled Kafka PR and Spanner PR runs (both runs tests in v2) since they are unrelated to this change.

Abacn:
Now there is only one permared test and will be fixed by #2322

Abacn:
R: @damccorm

CC: @darshan-sj

darshan-sj (approved):
LGTM

Abacn:
Thanks for approval! Let me merge this so the next release (will be cut ~12 hr later can be green)

PR #2336, opened 2025-04-23
SpannerIO change to read rows as List of com.google.protobuf.Value objects

lagarwal1:


v1/src/main/java/com/google/cloud/teleport/spanner/spannerio/BatchSpannerReadAsProto.java
```diff
@@ -0,0 +1,386 @@
...
+          partitions =
+              batchTx.partitionReadUsingIndex(
+                  op.getPartitionOptions(),
+                  op.getTable(),
+                  op.getIndex(),
+                  op.getKeySet(),
+                  op.getColumns(),
+                  Options.priority(config.getRpcPriority().get()),
+                  Options.dataBoostEnabled(dataBoostEnabled),
+                  Options.decodeMode(DecodeMode.LAZY_PER_COL));
```
lagarwal1, line 184: DecodeMode lazy is needed to read columns as Value protos

PR #2337, opened 2025-04-23
added beam-sdks-java-extensions-sql-expansion-service

liferoad:


svetakvsundhar (approved):
Thanks, in the description, can you show an image of running `ls` on the cache to prove the jar is present.

Abacn (approved):
Thanks, lgtm

Just wondering if this alone is sufficent to enable sql for beam yaml. If not could be follow up

liferoad:
> Thanks, in the description, can you show an image of running `ls` on the cache to prove the jar is present.

will test this after we release the image next week.

PR #2351, opened 2025-05-05
Added handling for spanner priority in bulk migration

shreyakhajanchi:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -274,4 +275,21 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
...
+        @TemplateParameter.TemplateEnumOption("LOW"),
+        @TemplateParameter.TemplateEnumOption("MEDIUM"),
+        @TemplateParameter.TemplateEnumOption("HIGH")
+      },
+      optional = true,
+      description = "Priority for Spanner RPC invocations",
+      helpText =
+          "The request priority for Cloud Spanner calls. The value must be one of:"
+              + " [`HIGH`,`MEDIUM`,`LOW`]. Defaults to `MEDIUM`.")
+  @Default.Enum("MEDIUM")
```
manitgupta, line 291: Do we need to have consistent defaults for this across templates?

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/3921618efef8968b5500ef80dd8aef1258d9e3cb/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java#L416

If not, let's capture why not in the comments.

shreyakhajanchi: The current behaviour is MEDIUM as default for bulk and had a discussion with @rohitwali , since CDC is related to downtime it makes sense to have it HIGH

manitgupta: I would argue the opposite. During bulk migration, we may need to autoscale the spanner instance in response to the high throughput and quantum of data being loaded. Spanner autoscaler scales on the basis of `high priority` CPU utlization, and not `medium`. Also, since migration is the "most important" activity happening on the spanenr instance at the time of bulk migration, it would again make sense to have `high` as the default.

Deep1998: +1, and if we are changing the default to HIGH, we should put the rationale for defaults in the comments as well.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SourceDbToSpanner.java
```diff
@@ -130,6 +130,7 @@ static SpannerConfig createSpannerConfig(SourceDbToSpannerOptions options) {
         .withProjectId(ValueProvider.StaticValueProvider.of(options.getProjectId()))
         .withHost(ValueProvider.StaticValueProvider.of(options.getSpannerHost()))
         .withInstanceId(ValueProvider.StaticValueProvider.of(options.getInstanceId()))
-        .withDatabaseId(ValueProvider.StaticValueProvider.of(options.getDatabaseId()));
+        .withDatabaseId(ValueProvider.StaticValueProvider.of(options.getDatabaseId()))
+        .withRpcPriority(ValueProvider.StaticValueProvider.of(options.getSpannerPriority()));
```
Deep1998, line 134: Can you please verify that all SpannerConfigs throughout the template are created via createSpannerConfig()?

shreyakhajanchi: yes, I have verified that

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
shreyakhajanchi: Will continue on the discussion to modify the default to `HIGH` if required, merging the PR to unblock the customer

PR #2360, opened 2025-05-07
Disable maven cache

Abacn:
We are using hosted github action runner and the local maven folder is persistent. No need to cache

This caused RC validation beyond RC 2 not effective, as cache keeps putting RC1 into local maven repository

Update: this is currently causing Spanner test failure due to that it is still running on RC1 jars


Abacn:
a drawback is that this will undo #354

Abacn:
R: @damccorm

damccorm (approved):
I think this is fine, but lets monitor workflow runtime and make sure it doesn't go up by a lot.

If it does, one alternative would be to just disable caching when an RC version is in use.

PR #2368, opened 2025-05-13
updated logic to reconsume errors from dlq in reverse replication

shreyakhajanchi:


VardhanThigle (approved):
Looks Good to me.
For the long run, not a request for current PR, let's plan for an IT that produces DLQs,
may be it might get tracked as a part of failure injection.

PR #2372, opened 2025-05-14
Set default container image properties

Abacn:
They are not used in Google provided template release process, but the documented way for external developer staging templates now failing requiring these parameters if not provided


Abacn:
referred here: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/8aa969df01180cbb5f755e2bd365d8e1bdf69527/v2/pom.xml#L200

Abacn:
R: @damccorm

PR #2374, opened 2025-05-14
Add new template Cloud_Datastream_MongoDB_to_Firestore to support migration from MongoDb to Firestore

ninjaAB-5:
Firestore now offers [MongoDB API compatibility](https://cloud.google.com/firestore/mongodb-compatibility/docs/overview?hl=en), and the Firestore team provides a new flex template `Cloud_Datastream_MongoDB_to_Firestore` to support the migration solution via datastream.


v2/datastream-common/src/main/java/com/google/cloud/teleport/v2/datastream/transforms/FormatDatastreamRecordToJson.java
```diff
@@ -147,6 +147,10 @@ public FailsafeElement<String, String> apply(GenericRecord record) {
       outputObject.put("_metadata_schema", getMetadataSchema(record));
       outputObject.put("_metadata_lsn", getPostgresLsn(record));
       outputObject.put("_metadata_tx_id", getPostgresTxId(record));
+    } else if (sourceType.equals("backfill") || sourceType.equals("cdc")) {
```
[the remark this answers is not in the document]
Deep1998: Similar for FormatDatastreamJsonToJson.java

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java
```diff
@@ -0,0 +1,1160 @@
...
+    LOG.info("Processing backfill events");
+    PCollectionTuple backfillResult;
+    if (options.getUseShadowTablesForBackfill()) {
+      // Use shadow tables for backfill (same as CDC processing)
+      backfillResult =
+          splitEvents
+              .get(SplitBackfillAndCdcEventsFn.backfillTag)
+              .apply(
+                  "Process Backfill with Shadow Tables",
+                  ParDo.of(new ProcessChangeEventFn(connectionString, options.getDatabaseName()))
```
Deep1998, line 554: Curious on why are we splitting backfill and cdc separately. If we have shadow table logic for conflict resolution, we should get eventual consistency processing them together?

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/ProcessChangeEventFn.java
```diff
@@ -0,0 +1,224 @@
...
+        // Step 1: Query the shadow collection to see if there is any existing record of this id
+        Object docId = element.getDocumentId();
+        Bson lookupById = eq("_id", docId);
+
+        session = client.startSession();
+        session.startTransaction();
+        LOG.info("Started transaction for document ID: {}", element.getDocumentId());
+
+        Document shadowDoc = shadowCollection.find(session, lookupById).first();
+        LOG.info("Shadow document found: {}", shadowDoc != null ? "yes" : "no");
```
Deep1998, line 88: Its generally not a good practice to put INFO logs inside a processElement() in production as it can significantly impact throughput.

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/ProcessChangeEventFn.java
```diff
@@ -0,0 +1,224 @@
...
+  @Teardown
+  public void teardown() {
+    // Close the MongoClient when the pipeline finishes
+    if (client != null) {
+      client.close();
+    }
+  }
+
+  private static boolean isEventNewerThanShadowDoc(
+      MongoDbChangeEventContext event, Document shadowDoc) {
```
Deep1998, line 213: FYI: we already do something similar in the DatastreamToSpanner template regarding conflict resolution for datastream events across backfill and cdc. While this also largely looks similar, might be worth taking a look at that as well if you have'nt already.

v2/datastream-common/src/main/java/com/google/cloud/teleport/v2/datastream/transforms/FormatDatastreamRecordToJson.java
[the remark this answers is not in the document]
ninjaAB-5: This is related to the way Datastream team formats the data from mongoDB source and is how they are currently represented in avro/json records. I agree it is ideal and descriptive, but it will be more of a Datastream team's decision and out of the scope/control of this template. (Would require some changes to the Datastream's code base to rename these)

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java
ninjaAB-5: This is mainly because leveraging shadow table is expensive from performance's perspective, so we hope to use bulk write on backfills (which can be of large size). The way backfill and cdc events are generated by datastream today guarantees that backfill events always comes first for a certain entry, so the bulk write we do for backfill doesn't really need conflict resolving and is safe to do. Hope that makes sense!

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/ProcessChangeEventFn.java
ninjaAB-5: Thanks! We were referencing the DatastreamToSpanner template as a start and that was helpful. :)

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/ProcessChangeEventFn.java
ninjaAB-5: Good point, removed the info level logs from hot paths. Thanks!

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java
ninjaAB-5: Sorry to be clear, in the template we support both modes - process backfill + cdc together, or backfill first with bulk write followed by cdc with transactional writes, the reply above is more to explain why we have the backfill first mode, and the mode can be controlled over a variable `processBackfillFirst` while launching the job.

Deep1998: Got it that makes sense. 

While datastream would generate the backfill before cdc, because Dataflow reads from GCS, it may not guarantee reading bulk events before CDC events (unless there is some handling in the DatastreamIO code). I'd double check just to be safe.

But you can resolve this comment if that is an acceptable tradeoff.

Deep1998:
Pls fix the unit test errors befor emerging

v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java
ninjaAB-5: Thanks Deep! Yes that's something we identified as well, one option we had to leverage the bulk write performance benefit is to only launch the dataflow template after the datastream backfill is done, then the split would help separating backfill and cdc jobs to process backfills first, otherwise we would call out using all transactional writes for correctness.

v2/datastream-common/src/main/java/com/google/cloud/teleport/v2/datastream/transforms/FormatDatastreamRecordToJson.java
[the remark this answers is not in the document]
ninjaAB-5: The MongoDB source support in Datastream is pretty new (very recently launched for API), so the corresponding infra support for dataflow might not have been covered. Sure, added comments to both files. Thanks!

leongn-google (approved):

liferoad (approved):

PR #2386, opened 2025-05-22
Temporarily disabling Cassandra.allDataTypesTest() until fix for https://github.com/apache/beam/issues/34863 gets merged into dataflow.

VardhanThigle:


manitgupta (dismissed):

Abacn:
https://github.com/apache/beam/issues/34863 is in Beam 2.65.0. The reason test failing is due to maven local cache. I thought this is fixed by #2360. If this isn't the case, we'd need to manually ssh into GitHub Action worker to delete old jars

Abacn:
From https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/spanner-pr.yml?query=event%3Aschedule there are a couple of green run after #2369

PR #2387, opened 2025-05-22
Revert "Temporarily disabling Cassandra.allDataTypesTest() until fix for https://github.com/apache/beam/issues/34863 gets merged into dataflow."

Abacn:
Reverts GoogleCloudPlatform/DataflowTemplates#2386

reason: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2386#issuecomment-2901483233 and https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2386#issuecomment-2901492954


Abacn:
R: @manitgupta @damccorm

Abacn:
Test passed: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/15189614881/job/42726730636?pr=2387 after SSH'd into VM machine to clean up old jar

damccorm:
Have we confirmed that this has been cleaned up on all the workers? It does look like tests are now passing, not sure if we did something or are just hitting a lucky vm now though

damccorm:
Oh race condition - thanks. If we've cleaned this up, then this seems fine to me (pending @VardhanThigle 's review). Thanks for the fix, sorry for the churn @VardhanThigle

PR #2415, opened 2025-06-03
added the pubsub PR workflow

liferoad:


cicd/internal/flags/common-flags.go
```diff
@@ -70,6 +71,19 @@ var (
 			"v2/datastream-to-postgres/",
 			"v2/datastream-to-sql/",
 		},
+		PUBSUB: {"v2/pubsub-binary-to-bigquery/",
+			"v2/pubsub-cdc-to-bigquery/",
+			"v2/pubsub-to-jms/",
+			"v2/pubsub-to-kafka/",
```
Abacn, line 77: actually I think we should not duplicate path. With the change kafka-to-pubsub test will run twice on PubsubPR and KafkaPR

Currently sourcedb-to-spanner test runs on both Spanner PR and DataStream PR

Abacn:
With the past changes the runtime improvement of https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/java-pr.yml?query=event%3Aschedule looks moderate.

liferoad:
> With the past changes the runtime improvement of https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/workflows/java-pr.yml?query=event%3Aschedule looks moderate.

I think it is much better for developers that only touch some specific modules. At least they do not need to run the entire Java PR.

Abacn:
Are there active pubsub template development ongoing? For Spanner / Kafka / Datastream there are teams actively working on it.

On the other hand this increases duplicate tests and friction when changing common modules, which happens on every Beam RC validate and version bump

liferoad:
Close this for now. We can revisit this in the future.

PR #2422, opened 2025-06-05
added a simple jdbc to BQ YAML test

liferoad:


derrickaw (approved):

liferoad:
The test indeed reproduced the issue in https://github.com/apache/beam/issues/35122

image

https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/15476074484/job/43572700007
 
Next, will test this after rebasing this with https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2421

python/src/test/resources/JdbcToBigQueryYamlIT.yaml
```diff
@@ -0,0 +1,16 @@
+pipeline:
+  type: chain
+  transforms:
+    - type: ReadFromPostgres
+      config:
+        driver_class_name: "org.postgresql.Driver"
```
Abacn: ReadFromPostgres already has driver_class_name defined: https://github.com/apache/beam/blob/feb9cda68638b96f2c8fbfbab9d5408cc4801502/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L102

one possibility is that providing driver_class_name manually without providing driver_jars can cause error

liferoad: Perhaps. So where should `driver_jars` be provided?

Abacn: postgres and mysql drivers jar is shipped with expansion service so no need to provide driver jar for these two Jdbc providers

liferoad: ok

PR #2425, opened 2025-06-06
Use more jobs for release

liferoad:


.github/workflows/release.yml
```diff
@@ -100,12 +98,120 @@ jobs:
...
+          --modules-to-build="SPANNER"
+      - name: Upload Site Report
+        uses: ./.github/actions/publish-site-report
+        with:
+          output-zip-file: test-report-spanner
+        if: always()
+
+  finalize-release:
+    name: Finalize Release
+    needs: [release-it-spanner]
```
Abacn, line 217: I think this step needs all 3 jobs as prerequisites?

I see, they run sequentially. Actually this won't work if we only wants to run "release-it-spanner", as it must have the previous two jobs succeeded on same workflow run

liferoad: Yes, but we can retry these two jobs at least.

damccorm: To Yi's point, I think this can be `needs: [release, release-it-datastream, release-it-spanner]`. Then we can get rid of the spanner dependency on datastream above so that they can run in parallel.

An even better end state would be:

1. Release job to create tag (doesn't run any ITs)
2. Job to run most ITs (excluding spanner/datastream) - depends on [1]
3. Job to run spanner ITs - depends on [1]
4. Job to run Datastream ITs - depends on [1]
5. Finalize release job - depends on [1,2,3,4]

That would allow spanner/datastream tests to run in parallel with the main IT suite.

But I don't think we need to block this PR on that. Probably the speedup would be minimal since [2] would consume all of our Dataflow quota anyways.

So I think we can just make the simple change (change this to `needs: [release, release-it-datastream, release-it-spanner]` and `release-it-spanner` to just depend on `release`)

.github/workflows/release.yml
```diff
@@ -100,12 +98,120 @@ jobs:
           --it-private-connectivity="datastream-private-connect-us-central1" \
           --it-spanner-host="https://batch-spanner.googleapis.com" \
           --it-release=true \
-          --it-retry-failures=2
+          --it-retry-failures=2 \
+          --modules-to-build="!SPANNER,!DATASTREAM"
```
damccorm: Have you tried this? I wouldn't expect negation like this to work given https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2903e897f1126a021c37b28d980123bdfddb0260/cicd/internal/flags/common-flags.go#L97

Supporting it doesn't seem too bad

.github/workflows/release.yml
liferoad: I intend to make this sequential since they are competing with each other in terms of resources.

.github/workflows/release.yml
liferoad: I think it already works with https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2903e897f1126a021c37b28d980123bdfddb0260/cicd/internal/flags/common-flags.go#L91 and 

The resulting Maven command will include the following argument:

```javascript
-pl !SPANNER,!DATASTREAM
```

damccorm: I tried this - https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/15536593612/job/43737017093

The command resolves to `mvn -B clean install -f pom.xml -e -am -amd -Dmdep.analyze.skip -Dcheckstyle.skip -Djib.skip -Dmaven.test.skip -Djacoco.skip -DskipShade -T8 --settings=.mvn/settings.xml -pl !SPANNER,!DATASTREAM -am` which fails because `-pl !SPANNER,!DATASTREAM` is not valid.

I think we want this to resolve to a bunch of negated paths like https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2903e897f1126a021c37b28d980123bdfddb0260/cicd/internal/flags/common-flags.go#L47C23-L47C39

```
-pl !v2/datastream-to-spanner/,!v2/spanner-change-streams-to-sharded-file-sink/,...
```

but we should test to make sure it works first

liferoad: you are right. I missed the parsing logic over there. Anyway, get rid of "!" now since we can easily specify the included list. :)

damccorm: Unfortunately, this will not work either - it will lead to:

```
-pl DEFAULT,BIGTABLE,KAFKA
```

which doesn't work either. To make this work, we need to split the string before we do the resolution of known names, not after (we currently do this at the end - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2903e897f1126a021c37b28d980123bdfddb0260/cicd/internal/flags/common-flags.go#L103)

liferoad: I see. Let me fix this.

liferoad: after I played with the go module, I decided to take a much easier way to just split the call into three steps since the module dependencies are really messy.

damccorm (approved):
Thanks!

PR #2426, opened 2025-06-07
[Example Template] Simple LLM Batch Processing Template

Ganeshsivakumar:
This PR introduces a new example Dataflow Flex Template: LlmBatchTextProcessing. It is a simple batch pipeline that reads data from a GCS file, uses an LLM to process the data based on a prompt defining the processing logic, and writes the output back to a GCS file.

## Key Features

- Reads input data from a text file (e.g., GCS) using `TextIO`
- Applies an LLM transformation based on configured model and prompt
- Writes the model outputs to GCS
- Uses OpenAI model configuration via `OpenAiModelOptions`

## Use Cases

This example serves as a starter template for:

- Prompt-based batch processing with LLMs
- Building RAG or summarization pipelines
- Demonstrating LangChain integration with Apache Beam

## Usage

```bash
# cd into DataflowTemplates directory and run the command to build and stage your template 
mvn clean package -PtemplatesStage \
  -DskipTests \
  -DprojectId="<YOUR_PROJECT_ID>" \
  -DbucketName="<YOUR_GCS_BUCKET_NAME>" \
  -DstagePrefix="<YOUR_USERNAME>/<YOUR_TEMPLATE_NAME>" \
  -DtemplateName="LlmBatchTextProcessing" \
  -pl v2/llm-batch-text-processing -am

# run the template
gcloud dataflow jobs run llm-batch-process \
  --gcs-location=gs://<YOUR_GCS_BUCKET_NAME>/<YOUR_USERNAME>/<YOUR_TEMPLATE_NAME>/LlmBatchTextProcessing.json \
  --region=<YOUR_REGION> \
  --parameters inputDataFile=gs://<YOUR_GCS_BUCKET_NAME>/input/input.txt, \
    llmOutputFile=gs://<YOUR_GCS_BUCKET_NAME>/output/output, \
    modelName=<YOUR_MODEL_NAME>, \
    apiKey=<YOUR_OPENAI_API_KEY>, \
    prompt="<YOUR_PROMPT_HERE>"

```

## Notes

This is an example template meant for demonstration and experimentation. It can be extended to support additional LLM providers, structured inputs, and RAG pipelines using embedding transforms.

## Reference

LangchainBeam GitHub: https://github.com/Ganeshsivakumar/langchain-beam

### DIFF
--- a/v2/llm-batch-text-processing/pom.xml
+++ b/v2/llm-batch-text-processing/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+        <groupId>com.google.cloud.teleport.v2</groupId>
+        <artifactId>dynamic-templates</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>llm-batch-text-processing</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.truth</groupId>
+            <artifactId>truth</artifactId>
+            <version>${truth.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.github.ganeshsivakumar</groupId>
+            <artifactId>langchain-beam</artifactId>
+            <version>0.3.0</version>
+        </dependency>
+    </dependencies>
+
+</project>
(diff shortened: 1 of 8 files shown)


liferoad:
Please check the failed workflows.

damccorm (commented):
Hey Ganesh, thanks for putting this together. Unfortunately, I think we are probably not going to be able to take this PR as is for a few reasons:

1) In Dataflow we're standardizing around the RunInference transform. We actually already have an LLM template example using RunInference - https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/51fa0a014f968784eb73e7e538a3781e070688c5/python/src/main/python/streaming-llm - and if we were to add a Java template, it would use a x-lang version of RunInference.

To expand on this, we think this is the right path because:
- It allows for easy interoperability between local and remote inference since most local inference is done in Python
- It is in line with what _most_ (not all) ML engineers we've talked to want
- It has very robust support for both local and remote inference, with proven performance optimizations for both (e.g. for remote inference, it provides autoscaling signals that Dataflow is able to consume)

We have heard a desire for more robust Java remote inference support, but even for that I think long term it will be cleaner to expose that under the RunInference namespace to make things simpler for users (and to allow us to use x-lang when possible). So I'd prefer not to expose a template doing something different.

2) For a Google template, we should probably be using Gemini instead of OpenAI - this is both policy and reflects the general demand we've seen from users. Not a statement about broader demand, but at least Dataflow templates users seem to ask for Gemini more often.

PR #2432, opened 2025-06-09
use scorecard-action@v2.4.2 with GITHUB_TOKEN

liferoad:


.github/workflows/scorecards-analysis.yml
```diff
@@ -27,13 +27,11 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@ce330fde6b1a5c9c75b417e7efc510b822a35564 # v1.1.2
+        uses: ossf/scorecard-action@v2.4.2
```
damccorm, line 30: This may fail - I'm not sure if the previous sha was allow-listed somehow or not. I think I remember that Google enforces this as a security measure for some actions, but I may be getting confused.

Regardless, lets merge and see

liferoad: ok.

PR #2453, opened 2025-06-12
added the update docs workflow

liferoad:


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+        id: changes
+        run: |
+          git status --porcelain
+          if [[ -z $(git status --porcelain) ]]; then
+            echo "no_changes=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.no_changes != 'true'
+        uses: peter-evans/create-pull-request@v7
```
damccorm: This will need to be pinned to a sha - go/github-actions#actions

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Generate Documentation
+        run: |
+          mvn clean prepare-package \
+            -DskipTests \
+            -PtemplatesSpec
+
+      - name: Check for changes
+        id: changes
+        run: |
+          git status --porcelain
```
damccorm: Won't the action automatically handle this? Per https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#-create-pull-request the action will "Check for repository changes in the Actions workspace"

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+          if [[ -z $(git status --porcelain) ]]; then
+            echo "no_changes=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.no_changes != 'true'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
damccorm, line 40: We should grant the token write permissions explicitly - https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token - not sure if this will happen by default or not, but it is something that could be changed in the future at an org level I think

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+          commit-message: 'docs: update generated documentation'
+          title: 'chore(docs): update generated documentation'
+          body: |
+            This PR updates the generated documentation for the templates.
+            This is an automated PR created by the update-docs workflow.
```
damccorm, line 45: Checks won't auto-run when this PR is created, but we want them to run. We should add a line here like:

```
To review this PR, please (1) review the contents to make sure all doc updates are formatted correctly, (2) close and reopen the PR (3) approve/merge if tests pass successfully.
```

This is similar to what we do here - https://github.com/apache/beam/blob/31f9e5838f1eb488e5ab9a1d815e42a2eeb27b53/.github/workflows/update_python_dependencies.yml#L98C218-L98C403

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+          branch: chore/update-generated-docs
+          base: main
+          labels: 'owl-bot-copy'
```
damccorm, line 48: Can we add a team-reviewers parameter so it gets auto-assigned to plat for review?

```suggestion
          labels: 'owl-bot-copy'
          team-reviewers: GoogleCloudPlatform/dataflow-templates-wg
```

Per https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs this may also require updating some actions permissions

.github/workflows/update-docs.yml
liferoad: Good catch. Done.

.github/workflows/update-docs.yml
liferoad: Added the write permission to give it a try.

.github/workflows/update-docs.yml
liferoad: Done.

.github/workflows/update-docs.yml
liferoad: Missed that. Done.

.github/workflows/update-docs.yml
liferoad: I am not sure we need to trigger any checks. This PR only updates .md files. We should always approve and merge it.

damccorm: Yeah, I think that is probably fine. I was thinking about formatting, but it should get auto-picked up.

Eventually we might just want to commit to the `main` branch directly once we've seen this working well.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,48 @@
+name: Update Template Documentation
+
+on:
+  schedule:
+    - cron: '0 16 * * 2' # Every Tuesday at 11 AM EST (16:00 UTC)
+  workflow_dispatch:
+
+permissions:
+  contents: write
```
damccorm, line 9:
```suggestion
  contents: write
  pull-requests: write
```

This is needed for creating the PR, `contents: write` just allows branch creation

damccorm (approved):
One more comment, other than that LGTM - thanks!

.github/workflows/update-docs.yml
liferoad: I see. I thought they were same. Fixed it. Thanks.

PR #2457, opened 2025-06-13
removed adding labels

liferoad:
https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/15640209750/job/44065513216

Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"} -


liferoad:
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2456/files from the previous run. Looks like the PR can be created.

damccorm (approved):
I think if we want to add the label we can with more permissions for the token. But I think getting rid of it is also fine.

PR #2463, opened 2025-06-16
Use private IPs in Spanner to Sourcdb template ITs

darshan-sj:


VardhanThigle:
Please wait for the tests to pass before checkin.

darshan-sj:
> Please wait for the tests to pass before checkin.

All tests passed. Merging.

PR #2464, opened 2025-06-16
Adding support for DateTime column as primary key

VardhanThigle:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -307,14 +307,17 @@ private static TableConfig getTableConfig(
       } else {
         ImmutableSet<IndexType> supportedIndexTypes =
             ImmutableSet.of(
-                IndexType.NUMERIC, IndexType.STRING, IndexType.BIG_INT_UNSIGNED, IndexType.BINARY);
+                IndexType.NUMERIC,
+                IndexType.STRING,
+                IndexType.BIG_INT_UNSIGNED,
+                IndexType.BINARY,
+                IndexType.DATE_TIME);
```
Deep1998, line 314: One thing to note is this datetime is a mysql type but this handling seems generic, which woulud be extended in the future to other jdbc types, where datetime might be represented in some other form or might not exist at all.

Not urgent, but we should see how to account for this as it could get confusing in the future.

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
```diff
@@ -260,6 +260,19 @@ CREATE TABLE tiny_text_pk_table (
    CONSTRAINT PRIMARY KEY (id(20))
 );
 
+CREATE TABLE tiny_text_pk_table (
+                                    id TINYTEXT,
```
Deep1998: Seems like tiny text table sql is duplicated?

Deep1998 (approved):

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
VardhanThigle: Thanks. Fixed in current push.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: That's right.  I reused `Date_TIME` instead of adding a new type. This type exissted as legacy jdbc supports this column type as well.
Will take this up when I add support for additional Pkeys.

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
```diff
@@ -396,6 +402,8 @@ INSERT INTO `tiny_blob_pk_table` (`id`, `tiny_blob_pk_col`) VALUES (FROM_BASE64(
 INSERT INTO `char_pk_table` (`id`, `char_pk_col`) VALUES ('AA==', 'AA=='), ('gAAAAAAAAAA=', 'gAAAAAAAAAA=');
 INSERT INTO `varchar_pk_table` (`id`, `varchar_pk_col`) VALUES ('AA==', 'AA=='), ('gAAAAAAAAAA=', 'gAAAAAAAAAA=');
 INSERT INTO `tiny_text_pk_table` (`id`, `tiny_text_pk_col`) VALUES ('AA==', 'AA=='), ('gAAAAAAAAAA=', 'gAAAAAAAAAA=');
+INSERT INTO `date_time_pk_table` (`id`, `date_time_pk_col`) VALUES ('1000-01-01 00:00:00', '1000-01-01 00:00:00'), ('1000-01-01 00:00:01', '1000-01-01 00:00:01'),
```
rohitwali, line 405: do we need to add a test case for fractional seconds and timezones?

rohitwali (approved):

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
VardhanThigle: We are covering Nanos in the UT.

```
Timestamp start = Timestamp.valueOf("0000-01-01 00:00:00.000000000");
    Timestamp end = Timestamp.valueOf("2041-01-31 23:59:59.999999999");
 ```
 
 I will add fractional seconds to IT in my followup PR.

PR #2476, opened 2025-06-22
Fix the FileBasedDeadLetterQueueReconsumer dup issue

### LINKED ISSUE #2236: [Bug]: FileBasedDeadLetterQueueReconsumer could result in duplicates
### Related Template(s)

Anything using FileBasedDeadLetterQueueReconsumer (as best I can tell, looks like just spanner changestreams + datastreams templates - https://github.com/search?q=repo%3AGoogleCloudPlatform%2FDataflowTemplates%20FileBasedDeadLetterQueueReconsumer&type=code )

### Template Version

latest - seen in 2023-07-18-00_rc00, but also just based on code inspection

### What happened?

The way https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/613b9aa764197a3297dec995c0ce5cba58863a48/v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/dlq/FileBasedDeadLetterQueueReconsumer.java#L91 is written, it could result in duplicates if Dataflow experiences any sort of backlog or slowdown. Specifically, the following scenario could happen:

1) Generate sequence fires, and everything before the reshuffle happens
2) Generate sequence fires again, and everything before the reshuffle happens
3) We remove the file after the reshuffle
4) We try to remove the file after the reshuffle again, which logs but succeeds - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/613b9aa764197a3297dec995c0ce5cba58863a48/v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/dlq/FileBasedDeadLetterQueueReconsumer.java#L143

### Relevant log output

```shell

```

liferoad:
Fixes #2236

This PR changes the process from a "match -> read -> delete" sequence to a "match -> atomic rename -> read -> delete" sequence.

MoveFiles:

 A new MoveFiles step is introduced, which renames each matched dead-letter file by adding a tmp- prefix to its name. his atomic "move" operation effectively claims the file, ensuring that subsequent scans won't find the original file and attempt to process it again.

After the files are moved, a Reshuffle transform is applied to guarantee that the rename operation is completed.


v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/dlq/FileBasedDeadLetterQueueReconsumer.java
```diff
@@ -92,28 +93,39 @@ public PCollection<String> expand(PBegin in) {
...
       public PCollection<String> expand(PCollection<Metadata> input) {
-        // TODO(pabloem, dhercher): Use a Beam standard transform once possible
-        // TODO(pabloem, dhercher): Add a _metadata attribute to track whether a row comes from DLQ.
         TupleTag<String> fileContents = new TupleTag<String>();
-        TupleTag<Metadata> fileMetadata = new TupleTag<Metadata>();
+        TupleTag<ResourceId> fileMetadata = new TupleTag<ResourceId>();
+
+        PCollection<ResourceId> movedFiles =
+            input
+                .apply("MoveFiles", ParDo.of(new MoveFiles()))
```
damccorm, line 114: I think this change introduces the possibility of data loss which is worse. Specifically, we could run into the following scenario:

1) Bundle x contains `file-1`, `file-2`, `file-3`
2) `file-1` is successfully renamed to `tmp-file-1`, but the bundle fails on renaming `file-2`

`file-1` is now orphaned as `tmp-file-1`. I guess the way it is written, we would actually get recurrent errors which is a little better, but still not ideal.

Instead, I'd propose the following tweak:

1) We get a timestamp `FOO` associated with each generate sequence firing. We already get this for free, we'd just need to propogate it through
2) Instead of renaming the file `tmp-<file name>`, we rename it `tmp-FOO-<file name>`
3) Instead of only looking for `<file name>` when doing the rename operation, we look for `<file name>` and rename it OR if `<file name>` is not present, we look for `tmp-FOO-<file name>` and use that instead. If `tmp-FOO-<file name>` and `<file name>` are both not present, we log and move on, assuming that something else has claimed the file.

So the algorithm would be:

```
TriggerConsumeDLQ, AsFilePattern, MatchFiles, <new extract_ts function> -> (ts, file) tuples
in MoveFiles:
   renamed_file_name = "tmp-${ts}-${original_file_name}"
   if exists(renamed_file_name):
      return
   if !exists(original_file_name):
      log.warning('skipping, handled by different pass')
   mv(original_file_name, renamed_file_name)
```

Then the rest of the logic would stay the same.

v2/common/src/main/java/com/google/cloud/teleport/v2/cdc/dlq/FileBasedDeadLetterQueueReconsumer.java
```diff
@@ -92,28 +93,39 @@ public PCollection<String> expand(PBegin in) {
...
-            input.apply(
+            movedFiles.apply(
+                "ConsumeFiles",
                 ParDo.of(new MoveAndConsumeFn(fileContents, fileMetadata))
                     .withOutputTags(fileContents, TupleTagList.of(fileMetadata)));
 
         results
             .get(fileMetadata)
-            .setCoder(MetadataCoder.of())
+            .setCoder(ResourceIdCoder.of())
```
damccorm, line 126: I think this is inevitable, but we are breaking update compatibility. Maybe worth calling out in the PR title so that it gets into the resource notes.

PR #2478, opened 2025-06-22
Reverse replication overrides support

Deep1998:
Reverse replication overrides
- Support Source ddls
- Refactor to not use session file for anytohing other than schema mapping/use sourceddl for type info
- Swap sessino based mapper with schemamapper
- add overrides support


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -186,9 +192,8 @@ public void processElement(ProcessContext c) throws Exception {
         record.setShard(this.shardName);
         qualifiedShard = this.shardName;
       } else {
-        // Skip from processing if table not in session File
-        // TODO: remove dependency on session file when session file is made optional
-        boolean doesTableExist = doesTableExistInSessionFile(tableName);
+        // Skip from processing if table not found at source.
+        boolean doesTableExist = doesTableExistAtSource(tableName);
```
Deep1998, line 210: @manitgupta this was a recent fix for EA? CAn you validate this does nto break any of there requirements?

PR #2481, opened 2025-06-24
Suppress error on cloud build Templates image nonzero return code

Abacn:


liferoad (approved):

Abacn:
Partly revert #2284

> Ideally this should raise an exception, but this returns NZE even for
> successful runs.

actually we do see this happened during staging workflow, and failed the workflow

It is fine because if the image does not exist, it will fail in the future steps

Abacn:
R: @liferoad

liferoad (approved):

kennknowles:
This actually was pretty confusing for a newcomer to the project. I think maybe there could be a way to check the success status via API?

Abacn:
Agree and there was an attempt to solve this (#2284, also noted the current behavior is confusing there), however after that fix the release saw increased flakiness, and partly reverted by this PR.

PR #2484, opened 2025-06-26
[Fix]: Normalize Years and Months while constructing ISO8601 Duration String

sagarwaal:
The change ensures that 
1. Years and months are normalized while constructing IS08601 duration string. For e.g. -
    - 1 year 13 Month  -> P2Y1M (instead of P1Y13M)
     - 2 year 1 Month -> P2Y1M
2. Interval of length 0 is always represented as P0Y
3. Duration string for Interval with non-zero length doesn't have any specifier with 0 quantity, For e.g.
     -    1 hour -> PT1H  (instead of P0DT1H)
     -   1 minute -> PT1M  (instead of P0DT1M)


VardhanThigle (commented):
Thanks  a ton for the change. If Google Sql needs a normalized entry, we should do that in dataflow.

I have triggered an IT flow, if any of the IT data needs to be tweaked for the normalisation, will let you know here.

VardhanThigle (changes requested):
You might need to change 
`.isEqualTo(Value.string("P1000Y1000M3890DT30H31M12.000000009S"))` at line 604 to a normazlied valaue too.

PR #2499, opened 2025-07-03
Add data record count metrics to reverse replication

shreyakhajanchi:


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/FilterRecordsFn.java
```diff
@@ -33,6 +34,8 @@ public FilterRecordsFn(String filtrationMode) {
   @ProcessElement
   public void processElement(ProcessContext c) {
     DataChangeRecord record = c.element();
+    Metrics.counter(FilterRecordsFn.class, "data_record_count_per_record")
```
bharadwaj-aditya: the metric is not clear. Why does it say count per record ?

shreyakhajanchi: actually there is an existing metric `data_record_count` that change stream publishes, but its per transaction and we need a metric per mutation. Should i rename it to mutation_count?

bharadwaj-aditya (approved):
LGTM

PR #2500, opened 2025-07-03
Add change stream timeout parameter to BigtableChangeStreamsToPubSub template

jonnylaw:


v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/bigtablechangestreamstopubsub/BigtableChangeStreamsToPubSub.java
```diff
@@ -274,6 +275,13 @@ public static PipelineResult run(BigtableChangeStreamsToPubSubOptions options) {
           readChangeStream.withMetadataTableTableId(
               options.getBigtableChangeStreamMetadataTableTableId());
     }
+
+    if (options.getBigtableReadChangeStreamTimeoutMs() != null) {
```
ron-gal, line 284: I'd also assert the value is bigger than 0.

jonnylaw: I added it to `validateOptions`, let me know if you'd prefer it be inline.

PR #2503, opened 2025-07-06
Quoting strings in custom transformation in reverse replication

shreyakhajanchi:


shreyakhajanchi:
This is no longer required as we have requested the customer to follow the contract and the long term fix is to use Prepared Statements.

PR #2506, opened 2025-07-07
Construct correct Reverse replication delete DML  when there pkey change between source and spanner.

VardhanThigle:
This PR constructs RR delete event with  source PKey columns that exist in spanner but are not marked as parimary key on Spanner. As the old values are absent in delete CDC event, it relies on the existing stale read methods to get the same. 

Note that the case of mismatch in the way a spanner row read function represents a value and CDC stream represents a value (if any) is out of scope of the current PR.

Testing:
UT coverage

Thanks a ton @shreyakhajanchi  for helping with additional e2e testing to help speed up the work.


bharadwaj-aditya (dismissed):
LGTM

PR #2525, opened 2025-07-08
Add source ddl classes

Deep1998:
Create classes similar to spanner ddl but for source databases which allows storing the schema for migrations. This would be useful to remove dependency on session file for identifying source types.


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SourceInformationSchemaScanner.java
```diff
@@ -0,0 +1,30 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.sourceddl;
+
+/**
+ * Interface for scanning information schema of source databases. Different implementations will
+ * handle different source database types.
+ */
+public interface SourceInformationSchemaScanner {
```
bharadwaj-aditya: nit: This should probably just be sourceSchemaScanner.

Depending on the database, it would figure it if needs to get it from info schema or some other place.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SourceSchema.java
```diff
@@ -0,0 +1,61 @@
...
+    public abstract Builder sourceType(SourceDatabaseType sourceType);
+
+    public abstract Builder tables(ImmutableMap<String, SourceTable> tables);
+
+    public abstract Builder version(String version);
+
+    public abstract SourceSchema build();
+  }
+
+  public SourceTable table(String tableName) {
```
bharadwaj-aditya, line 58: wondering if having this method makes sense.  This can as well be an interface.

bharadwaj-aditya (commented):
Some of these abstract classes seem like shell implementations. They can be interfaces as well to simplify the structure.

Please add some test cases with edge cases , null handling, missing fields etc.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SourceSchema.java
Deep1998: This is emulating the spanner DDL classes we have. It is a helper method to make it easier to access tables in the code which we do a lot. It otherwise tables 2 method calls. And moving this class to an interface would lead to a lot of boilerplate code which we are able to avoid using autovalue.

Deep1998:
> Some of these abstract classes seem like shell implementations. They can be interfaces as well to simplify the structure.
> 
> Please add some test cases with edge cases , null handling, missing fields etc.

Thanks, added more test cases and removed the AbstractSourceInformationSchemaScanner for simplicity. The data model classes (SourceSchema, SourceTable, SourceColumn) are abstract classes because they use AutoValue, which allows us to avoid a lot of boiler plate code, so those I have kept as is.

PR #2529, opened 2025-07-09
Refactor template to avoid using session file for type info

Deep1998:
This PR refactors the reverse template to avoid using session file for anything other than table/column name-mappings. It instead relies of the Ddl and SourceDdl object for schema specific info. This refactoring is necessary to introduce the schemamapper and henceforth, overrides.


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class SchemaUtils {
+
+  public static Ddl buildDdlFromSessionFile(String sessionFile) {
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      Map<String, Object> session = mapper.readValue(new File(sessionFile), Map.class);
+      Map<String, Object> spSchema = (Map<String, Object>) session.get("SpSchema");
```
bharadwaj-aditya: these are used in multiple places here. Suggest to use constants instead of the strings.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
```diff
@@ -0,0 +1,159 @@
...
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      Map<String, Object> session = mapper.readValue(new File(sessionFile), Map.class);
+
+      SourceDatabaseType dbType = SourceDatabaseType.MYSQL;
+      String dbName = "test-db";
+      Map<String, Object> srcSchema = (Map<String, Object>) session.get("SrcSchema");
+      SourceSchema.Builder schemaBuilder = SourceSchema.builder(dbType).databaseName(dbName);
+      ImmutableMap.Builder<String, SourceTable> tablesBuilder = ImmutableMap.builder();
+      if (srcSchema != null) {
```
bharadwaj-aditya: please addd error handling in case an invalid file is passed.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -551,6 +563,27 @@ public static PipelineResult run(Options options) {
       LOG.info("Cassandra config is: {}", shards.get(0));
       shardingMode = Constants.SHARDING_MODE_SINGLE_SHARD;
     }
+    SourceSchemaScanner scanner = null;
+    SourceSchema sourceSchema = null;
+    try {
+      if (options.getSourceType().equals(MYSQL_SOURCE_TYPE)) {
+        Connection connection = createJdbcConnection(shards.get(0));
```
bharadwaj-aditya: Can this be moved behind an interface where session file is an implementation to avoid the if-else blocks ?

bharadwaj-aditya (commented):
This changes are very spread out and hard to quantify or review because of the fan out. It would be cleaner to introduce an interface and swap between session file and alternates. It will keep the implementation and testing clean

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
```diff
@@ -0,0 +1,159 @@
...
+import com.google.cloud.teleport.v2.spanner.type.Type;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+
+public class SchemaUtils {
+
+  public static Ddl buildDdlFromSessionFile(String sessionFile) {
```
rohitwali: will this be reused in Bulk and Live?

rohitwali: also, 1/ may want o qualify the method name with which ddl, spanner or mysql or pass that as argument to the method? 

2/ does this need to be static?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
rohitwali: +1. Same for other strings in this method. Helps with readability as well

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        tableBuilder.endTable();
+      }
+      return ddlBuilder.build();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static SourceSchema buildSourceSchemaFromSessionFile(String sessionFile) {
```
rohitwali: source can be any and not necessarily MySQL. Do you want to build and interface around this API with db specific implementations?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
rohitwali: +1

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
```diff
@@ -80,38 +81,43 @@ public DMLGeneratorResponse getDMLStatement(DMLGeneratorRequest dmlGeneratorRequ
 
     String spannerTableName = dmlGeneratorRequest.getSpannerTableName();
     Schema schema = dmlGeneratorRequest.getSchema();
-
-    if (schema == null
-        || schema.getSpannerToID() == null
-        || schema.getSpSchema() == null
-        || schema.getSrcSchema() == null) {
-      LOG.warn("Schema is invalid or incomplete for table: {}", spannerTableName);
+    Ddl ddl = dmlGeneratorRequest.getDdl();
```
rohitwali: nit: may want to rename this as SpannerDDL. helps read better. if possible, same for `getDdl` API

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
```diff
@@ -80,38 +81,43 @@ public DMLGeneratorResponse getDMLStatement(DMLGeneratorRequest dmlGeneratorRequ
...
     }
 
-    if (sourceTable.getPrimaryKeys() == null || sourceTable.getPrimaryKeys().length == 0) {
+    if (sourceTable.primaryKeyColumns() == null || sourceTable.primaryKeyColumns().size() == 0) {
       LOG.warn(
           "Cannot reverse replicate table {} without primary key. Skipping the record.",
-          sourceTable.getName());
+          sourceTable.name());
       return new DMLGeneratorResponse("");
     }
```
rohitwali, line 116: would it make sense to move these checks in generateDMLResponse API?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
Deep1998: That is the ISchemaMapper object which I will introduce in the follow up PR so that we can swap out session file with overrides, identity mapper etc.

This PR only introduces the Ddl and SourceDdl object and refactors the methods where session file is being used for data type info. Session file is still a required parameter up until this PR.


This bit is more along the reading the source schema in memory which I have refactored into a separate method for now to keep it clean.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
Deep1998: This is a testing helper, this is only used in unit tests to avoid defining a new ddl/sourceddl object for every test. Ill move this to the testing folder to avoid confusion. SO this wont be reused in bulk / live as bulk uses its own source implementation and live does not have any source ddl uses cases as of now.

1) Sure ill update it.
2) Yes static is important as this can be called directly from tests without the need to define an object.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
Deep1998: These methods are only used as helper methods for tests to make reduce the amount of refactoring required across UTs (generating new object structs for tests vs generating from the session file being used).

Ill move these to the test folder to avoid confusion. Ill also make these changes to improve quality nonetheless.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
Deep1998: The sourceddl object is generic for all sources as it only stores a string for the type name. Its a common struct to hold all source schemas, the schema scanners have their own implementations. So this method is source type agnostic.

And also its a test utility so no need to make it too robust.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SchemaUtils.java
Deep1998: This is a test utility to ease building ddl structs from the session file. And some of the tests do use invalid files as they only populate certain parts that they want to test.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
Deep1998: So all that the getDMLStatement() does is do these prechecks and call gnerateDMLResponse(). Moving all of it essentially means getting rid of one of these. There isn't much value add though so I'll just keep it as is for now.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -38,48 +41,59 @@ public class MySQLDMLGenerator implements IDMLGenerator {
...
-          "The spanner table {} was not found in session file, dropping the record",
-          dmlGeneratorRequest.getSpannerTableName());
+          "The spanner table {} was not found in ddl found on spanner. Ddl: {}",
+          spannerTableName,
+          spannerDdl);
       return new DMLGeneratorResponse("");
     }
 
-    SourceTable sourceTable = dmlGeneratorRequest.getSchema().getSrcSchema().get(spannerTableId);
+    if (schema.getToSource().get(spannerTableName).getName() == null) {
```
bharadwaj-aditya, line 69: This logic is repeated in the cassandra generator and not in the mysql generator and will be in any new generator written.
i'm assuming whichever mapping schema is used, you would need a method of this sort, which says getSpannerTable. It would help to do that refactoring at this stage and expose an interface with these simpler methods.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -228,34 +241,31 @@ private static Map<String, String> getColumnValues(
...
       }
       if (customTransformColumns != null && customTransformColumns.contains(colName)) {
         response.put(colName, customTransformationResponse.get(colName).toString());
         continue;
       }
-
-      String colId = entry.getKey();
-      SpannerColumnDefinition spannerColDef = spannerTable.getColDefs().get(colId);
+      String spannerColumnName =
+          schema.getToSpanner().get(sourceTable.name()).getCols().get(colName);
```
bharadwaj-aditya, line 259: similarly - this is just getSpannerColumn(sourceTableName, columnName)

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
Deep1998: Yes i have dnoe the refactoring in the followup PR https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2561. 

This PR does not introduce the interface as i wanted to logically split the XXL PR.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
Deep1998: YEs these changes are also done in the followup PR.

Issue #2489, opened 2025-06-27
[Feature Request]: BigQueryToParquet should allow to create Storage Read session in the current project


mkuthan:
### Related Template(s)

BigQueryToParquet

### What feature(s) are you requesting?

I need to execute the BigQueryToParquet template in Project A while accessing data from Project B. The Dataflow job's service account has permission to create a read session in Project A and to access the specified dataset in Project B.

Currently, the implementation attempts to create the read session in Project B, where the Dataflow service account lacks permission. Additionally, I want the data reading costs from BigQuery to be attributed to Project A instead of Project B.

Could you add functionality to configure the project for creating the read session? Alternatively, please ensure the read session is created in the project where the Dataflow job is executed.

stankiewicz:
Marcin, 
I can confirm that implementation attempts to create the read session to read schema in project b here:
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/bigquery-to-parquet/src/main/java/com/google/cloud/teleport/v2/templates/BigQueryToParquet.java#L118

if you add permission temporarily, can you verify that data reading cost is attributed to project A instead of project B?
Based on BigQuery implementation https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L1718 it should use project A as parent.

mkuthan:
Hi Radek, I'm glad to see you here :) 

I'd love to experiment but getting permission in the project B is off the table this time. 

How do you assess the chances of adding the ability to specify a runtime project and a timeline for implementing this functionality? I would like to know how to plan further work on our side.

stankiewicz:
Hey Marcin:) I think we can change it to follow the behaviour of bigqueryio. I assume that downstream it is creating a stream using project a (or other, based on bigqueryoptions), so it shouldn't be a breaking change for anyone.

mkuthan:
Tested today morning and it seems that the change has not been released yet:

```
latest: Pulling from dataflow-templates/2025-07-08-00_rc00/bigquery-to-parquet
```

Could you release the change, please?

liferoad:
@mkuthan A new weekly release will be done this week. Please keep checking https://github.com/GoogleCloudPlatform/DataflowTemplates/releases.

PR #2530, opened 2025-07-09
fix parent project in BigQueryToParquet

stankiewicz:
fix parent project in BigQueryToParquet - this should be transparent  change as BQIO is following same logic, it was strange that for schema pulling, pipeline is using dataset's project, while for reading it is following options. now it will be aligned.

#2489


stankiewicz:
@liferoad would you be able to review? thank you

liferoad (approved):

PR #2531, opened 2025-07-10
docs: Updates incorrect default value

shuesc1:
Technical Solutions Engineer reported an incorrect value that is surfaced in https://cloud.google.com/dataflow/docs/guides/templates/provided/pubsub-to-datadog#optional-parameters, which pulls from the `README`. Evidence they provided as to correct value (100):
* https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v1/src/main/java/com/google/cloud/teleport/datadog/DatadogEventWriter.java#L61
* https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/854\

More context: b/430467136


damccorm (changes requested):
Oh, looks like the linter is failing. Could you try running 'mvn spotless:apply' to fix these violations?

shuesc1:
Ran `mvn spotless:apply`, but getting some new failing tests and I'm not sure I fully understand the output.

damccorm:
> Ran `mvn spotless:apply`, but getting some new failing tests and I'm not sure I fully understand the output.

This is unrelated flakiness. This change looks good to me, thank you!

PR #2561, opened 2025-07-15
Replace session file usage with ISchemaMapper and add string overrides support

Deep1998:
This PR replaces all usages of session file with the ISchemaMapper and adds stringfile based overrides support

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
@@ -15,7 +15,6 @@
  */
 package com.google.cloud.teleport.v2.templates;
 
-import static com.google.cloud.teleport.v2.spanner.migrations.constants.Constants.CASSANDRA_SOURCE_TYPE;
 import static com.google.cloud.teleport.v2.spanner.migrations.constants.Constants.MYSQL_SOURCE_TYPE;
 
 import com.datastax.oss.driver.api.core.CqlSession;
@@ -32,10 +31,11 @@ import com.google.cloud.teleport.v2.cdc.dlq.StringDeadLetterQueueSanitizer;
 import com.google.cloud.teleport.v2.coders.FailsafeElementCoder;
 import com.google.cloud.teleport.v2.common.UncaughtExceptionLogger;
 import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
-import com.google.cloud.teleport.v2.spanner.migrations.metadata.CassandraSourceMetadata;
-import com.google.cloud.teleport.v2.spanner.migrations.schema.NameAndCols;
-import com.google.cloud.teleport.v2.spanner.migrations.schema.Schema;
-import com.google.cloud.teleport.v2.spanner.migrations.schema.SpannerTable;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.ISchemaMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.IdentityMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaFileOverridesBasedMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaStringOverridesBasedMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SessionBasedMapper;
 import com.google.cloud.teleport.v2.spanner.migrations.shard.CassandraShard;
 import com.google.cloud.teleport.v2.spanner.migrations.shard.Shard;
 import com.google.cloud.teleport.v2.spanner.migrations.spanner.SpannerSchema;
@@ -43,7 +43,6 @@ import com.google.cloud.teleport.v2.spanner.migrations.transformation.CustomTran
 import com.google.cloud.teleport.v2.spanner.migrations.utils.CassandraConfigFileReader;
 import com.google.cloud.teleport.v2.spanner.migrations.utils.CassandraDriverConfigLoader;
 import com.google.cloud.teleport.v2.spanner.migrations.utils.SecretManagerAccessorImpl;
-import com.google.cloud.teleport.v2.spanner.migrations.utils.SessionFileReader;
 import com.google.cloud.teleport.v2.spanner.migrations.utils.ShardFileReader;
 import com.google.cloud.teleport.v2.spanner.sourceddl.CassandraInformationSchemaScanner;
 import com.google.cloud.teleport.v2.spanner.sourceddl.MySqlInformationSchemaScanner;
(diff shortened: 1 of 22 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -430,6 +429,51 @@ public interface Options extends PipelineOptions, StreamingOptions {
...
+            "^\\[([[:space:]]*\\{[[:space:]]*[[:graph:]]+\\.[[:graph:]]+[[:space:]]*,[[:space:]]*[[:graph:]]+\\.[[:graph:]]+[[:space:]]*\\}[[:space:]]*(,[[:space:]]*)*)*\\]$",
+        example =
+            "[{Singers.SingerName, Singers.TalentName}, {Albums.AlbumName, Albums.RecordName}]",
+        helpText =
+            "These are the column name overrides from spanner to source. They are written in the"
+                + "following format: [{SpannerTableName1.SpannerColumnName1, SpannerTableName1.SourceColumnName1}, {SpannerTableName2.SpannerColumnName1, SpannerTableName2.SourceColumnName1}]"
+                + "Note that the SpannerTableName should remain the same in both the spanner and source pair. To override table names, use tableOverrides."
+                + "The example shows mapping SingerName to TalentName and AlbumName to RecordName in Singers and Albums table respectively.")
+    @Default.String("")
+    String getColumnOverrides();
```
bharadwaj-aditya, line 457: Is it possible to get the overrides in one parameters ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -897,4 +911,50 @@ private static SourceSchema fetchSourceSchema(Options options, List<Shard> shard
...
+    }
+
+    ISchemaMapper schemaMapper = new IdentityMapper(ddl);
+    if (hasSessionFile) {
+      schemaMapper = new SessionBasedMapper(options.getSessionFilePath(), ddl);
+    } else if (hasSchemaOverridesFile) {
+      schemaMapper = new SchemaFileOverridesBasedMapper(options.getSchemaOverridesFilePath(), ddl);
+    } else if (hasStringOverrides) {
+      Map<String, String> userOptionsOverrides = new HashMap<>();
+      if (!options.getTableOverrides().isEmpty()) {
```
bharadwaj-aditya: nit: move this checking logic into the creation of the schemaString overrides to eliminate duplicate code

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
```diff
@@ -340,8 +347,12 @@ private static Map<String, PreparedStatementValueObject<?>> getColumnValues(
         response.put(colName, columnValue);
         continue;
       }
-      String spannerColumnName =
-          schema.getToSpanner().get(spannerTable.name()).getCols().get(colName);
+      String spannerColumnName = "";
+      try {
+        spannerColumnName = schemaMapper.getSpannerColumnName("", sourceTable.name(), colName);
+      } catch (NoSuchElementException e) {
+        continue;
```
bharadwaj-aditya, line 354: please add a logger for the columns which have been skipped

bharadwaj-aditya (commented):
Overall looks much cleaner. Thanks for putting this together.

A few comments
1. Lets see if we need different parameters or if we can use a single one
2. We need UT coverage for the schema mappers

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
Deep1998: we could come up with a json like structure to specify the overrides in one parameter insetad of 2. However, we would have to change both al the templates as this convention is followed for live and bulk for now. We could creating a tracking bug to evaluate this decision?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
Deep1998: in case there is an extra column on source (like a default value), this would log once for every row, significantly lowering throughput.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
bharadwaj-aditya: If there is precedent in the other templates, please keep it similar.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
bharadwaj-aditya: Fine to leave this as is if it will come for every row.

But the handling overall seems a bit odd, the distinction between a failure case and an expected case is not clear here.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -66,13 +67,15 @@ public DMLGeneratorResponse getDMLStatement(DMLGeneratorRequest dmlGeneratorRequ
...
     }
 
-    if (schema.getToSource().get(spannerTableName).getName() == null) {
+    String sourceTableName = "";
+    try {
+      sourceTableName = schemaMapper.getSourceTableName("", spannerTableName);
+    } catch (Exception e) {
       LOG.warn(
-          "Equivalent source was not found in session file for spanner table {}, dropping the record",
+          "Equivalent table for {} was not found in source, check schema mapping provided",
```
bharadwaj-aditya: Similar comment here. This would log for every row i guess.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/CassandraDMLGenerator.java
Deep1998: The failure case during mapping will throw a different exception bubbling the the very top where as an expected cases skips silently. The failure should get logged at the top layer inside process element.

bharadwaj-aditya (approved):
LGTM

PR #2562, opened 2025-07-15
fix: read neo4j driver version from maven

Emrehzl94:
This PR adds a way to read Neo4j driver version from Maven by injecting the version into`versions.properties` file during the build. So it can be read from the properties resource at runtime and reported correctly.

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/telemetry/Neo4jTelemetry.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/telemetry/Neo4jTelemetry.java
@@ -18,10 +18,15 @@ package com.google.cloud.teleport.v2.neo4j.telemetry;
 import java.io.InputStream;
 import java.util.Map;
 import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class Neo4jTelemetry {
 
+  private static final Logger LOG = LoggerFactory.getLogger(Neo4jTelemetry.class);
+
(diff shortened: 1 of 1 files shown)


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/telemetry/Neo4jTelemetry.java
```diff
@@ -39,7 +42,30 @@ private static String platform() {
...
+    }
+    return neo4jDriverVersion;
+  }
+
+  private static String loadNeo4jDriverVersion() {
+    var props = new Properties();
+    try (InputStream input = Neo4jTelemetry.class.getResourceAsStream("versions.properties")) {
+      if (input == null) {
+        throw new IllegalStateException("Missing required resource: versions.properties ");
+      }
```
Emrehzl94: I'm not sure if we should hard fail here.

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/telemetry/Neo4jTelemetry.java
```diff
@@ -39,7 +42,30 @@ private static String platform() {
...
+
+      props.load(input);
+
+      var driverVersion = props.getProperty("neo4j-java-driver");
+      if (driverVersion == null) {
+        throw new IllegalStateException("Missing required property: neo4j-java-driver");
+      }
```
Emrehzl94: Same here.

nord-erik (approved):

liferoad (approved):

PR #2568, opened 2025-07-16
Bq to Bt support for sparse data format and user defined cell timestamps

jhambleton:
1) Add functionality for user's to not store columns that have null values and adhere to bigtable's sparse data layout. 
2) Provide users ability to set cell timestamp's based on a column, e.g. SELECT UNIX_MILLIS(event_ts) as cell_timestamp instead of write time. 

Test scenarios and results are captured here: 
https://docs.google.com/document/d/1SLYKavhj8vuw7b0AnLrAvW1RBsC4wckzkRuzK2QE5Uc

### DIFF
--- a/v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
+++ b/v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
@@ -24,9 +24,13 @@ import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Common transforms for Teleport Bigtable templates. */
 public class BigtableConverters {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BigtableConverters.class);
 
   /** Converts from the BigQuery Avro format into Bigtable mutation. */
   @AutoValue
@@ -37,6 +41,10 @@ public class BigtableConverters {
 
     public abstract String rowkey();
 
+    public abstract Boolean skipNullValues();
+
+    public abstract String timestampColumn();
+
     /** Builder for AvroToMutation. */
     @AutoValue.Builder
     public abstract static class Builder {
@@ -44,6 +52,10 @@ public class BigtableConverters {
       public abstract Builder setColumnFamily(String value);
 
       public abstract Builder setRowkey(String rowkey);
+
+      public abstract Builder setSkipNullValues(Boolean setSkipNullValues);
+
+      public abstract Builder setTimestampColumn(String timestampColumn);
 
       public abstract AvroToMutation build();
     }
@@ -58,18 +70,60 @@ public class BigtableConverters {
       Put put = new Put(Bytes.toBytes(rowkey));
 
       List<TableFieldSchema> columns = record.getTableSchema().getFields();
+      Long columnTs = getTimestampColumnMs(row, timestampColumn());
+
       for (TableFieldSchema column : columns) {
         String columnName = column.getName();
-        if (columnName.equals(rowkey())) {
+        if ((columnName.equals(rowkey()))
+            || (columnTs != null && columnName.equals(timestampColumn()))) {
           continue;
         }
 
         Object columnObj = row.get(columnName);
         byte[] columnValue = columnObj == null ? null : Bytes.toBytes(columnObj.toString());
         // TODO(billyjacobson): handle other types and column families
-        put.addColumn(Bytes.toBytes(columnFamily()), Bytes.toBytes(columnName), columnValue);
+
+        // Check if null values should be skipped and if the columnValue is null.
+        // If both are true, the column will not be added.
+        if (!skipNullValues() || null != columnValue) {
+
+          // set cell timestamp if specified and exists
+          if (columnTs != null) {
+            put.addColumn(
+                Bytes.toBytes(columnFamily()), Bytes.toBytes(columnName), columnTs, columnValue);
+          } else {
+            put.addColumn(Bytes.toBytes(columnFamily()), Bytes.toBytes(columnName), columnValue);
+          }
+        }
       }
       return put;
+    }
+
+    /**
+     * Get column that represents the timestamp of the cell in bigtable. The value is expected to be
+     * milliseconds precision with hbase client handling to micros granularity, e.g.
+     * UNIX_MILLIS(timestamp)
+     *
+     * @param row
+     * @param timestampColumnName
+     * @return
+     */
+    private Long getTimestampColumnMs(GenericRecord row, String timestampColumnName) {
+      Long columnTs = null;
+
+      if (timestampColumn() != null && timestampColumn().length() > 0) {
+        columnTs = (Long) row.get(timestampColumnName);
+
+        if (columnTs == null) {
+          LOG.debug(
+              "Ignoring TimestampColumn configuration. Invalid "
+                  + timestampColumnName
+                  + ", value: "
+                  + columnTs);
+        }
+      }
+
+      return columnTs;
     }
   }
 }
--- a/v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
+++ b/v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
@@ -126,7 +126,7 @@ public class BigtableConverters {
       Long columnTs = null;
 
       // timestamp column is optional, if it's specified, retrieve the field value
-      if (timestampColumnName != null && timestampColumnName.length() > 0) {
+      if (!timestampColumnName.isEmpty()) {
         try {
           // Attempt to retrieve the timestamp column value
           Object timestampValue = row.get(timestampColumnName);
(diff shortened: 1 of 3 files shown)


ron-gal (approved):
BigtableConvertersTest tests are failing. Please fix

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
...
+     * UNIX_MILLIS(timestamp)
+     *
+     * @param row
+     * @param timestampColumnName
+     * @return
+     */
+    private Long getTimestampColumnMs(GenericRecord row, String timestampColumnName) {
+      Long columnTs = null;
+
+      if (timestampColumn() != null && timestampColumn().length() > 0) {
```
andre-sampaio: We should either use the `timestampColumnName` parameter or the `timestampColumn()` method. Using them interchangeably risks causing bugs in future iterations

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
...
+        columnTs = (Long) row.get(timestampColumnName);
```
andre-sampaio: I'm not familiar with with the GenericRecord interface and unfortunately the documentation is not explicit about this: what happens if the field with the given name does not exist? Does it return null or throw an exception?

I assume the former but just want to double check

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
...
+
+        if (columnTs == null) {
```
andre-sampaio: The behavior should be documented explicitly in the `timestampColumn` configuration. (e.g. `if an entry does not contain the field then the write time will be used` or something of the sorts)

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
...
+          LOG.debug(
+              "Ignoring TimestampColumn configuration. Invalid "
+                  + timestampColumnName
+                  + ", value: "
+                  + columnTs);
```
andre-sampaio: NIT: if we are here we know this is null

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
       Put put = new Put(Bytes.toBytes(rowkey));
 
       List<TableFieldSchema> columns = record.getTableSchema().getFields();
```
andre-sampaio, line 75: Should we validate that the schema includes the configured `timestampColumn` and throw if it doesn't? Could prevent errors caused by typos.

This of course assumes that the schema for all records would be the same which I'm not sure if holds

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +70,60 @@ public Mutation apply(SchemaAndRecord record) {
...
+      Long columnTs = getTimestampColumnMs(row, timestampColumn());
+
       for (TableFieldSchema column : columns) {
         String columnName = column.getName();
-        if (columnName.equals(rowkey())) {
+        if ((columnName.equals(rowkey()))
+            || (columnTs != null && columnName.equals(timestampColumn()))) {
```
andre-sampaio, line 82: NIT: let's also document that entries that belong to the timestampColumn column will not be included as part of the bt record

andre-sampaio (commented):
Please add a test case or 2 to v2/bigquery-to-bigtable/src/test/java/com/google/cloud/teleport/v2/templates/BigQueryToBigtableIT.java and v2/bigtable-common/src/test/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConvertersTest.java

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: This will throw an exception which would require the user to fix the field config before continuing. From my tests, it is: `Error: AvroRuntimeException: Not a valid schema field: abcd`

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: updated, good catch

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: good catch, added the rowkey to help debug any issues that may arise

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: great point, updated.

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: added, thanks

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: great point, i added a more extensive check with getAndVerifySchemaFields to raise runtimeexceptions for missing column and if the type cannot be casted to a long. If the value is null, we log a warn and use the default timestamp that is added on write in the hbase client. The `timestampColumn` description has been updated.

jhambleton:
finished addressing all comments & added error scenario unit tests

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
jhambleton: example error thrown: 

```Error message from worker: org.apache.beam.sdk.util.UserCodeException: java.lang.IllegalArgumentException: Avro error accessing timestamp column 'abcd' for row with key: 'u5d3slBLeaLm6eCs2h1f2O5jM/sQO3bFsLavqUojDas='. Details: Not a valid schema field: abcd```

File: v2/bigtable-common/src/test/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConvertersTest.java [hunk 7]
```diff
@@ -57,13 +58,16 @@ public class BigtableConvertersTest {
...
+      new Put("r".getBytes()).addColumn("f".getBytes(), "q".getBytes(), "v".getBytes());
+
   /** Generates a short string Avro field. */
   private String generateShortStringField() {
     return String.format(avroFieldTemplate, shortStringField, "string", shortStringFieldDesc);
   }
 
   /** Tests that {@link BigtableConverters.AvroToMutation} creates a Mutation. */
-  @Test
+  // @Test
```
jhambleton: revert this change

v2/bigtable-common/src/main/java/com/google/cloud/teleport/v2/bigtable/transforms/BigtableConverters.java
```diff
@@ -58,18 +73,93 @@ public Mutation apply(SchemaAndRecord record) {
...
+     *     {@code null} if the timestamp column is not found in the row (a warning will be logged).
+     * @throws IllegalArgumentException If the timestamp column is found but cannot be cast to a
+     *     {@link Long}, or if an {@link AvroRuntimeException} occurs during field access.
+     */
+    private Long getAndVerifySchemaFields(
+        String rowkey, GenericRecord row, String timestampColumnName) {
+      Long columnTs = null;
+
+      // timestamp column is optional, if it's specified, retrieve the field value
+      if (timestampColumnName != null && timestampColumnName.length() > 0) {
```
mutianf: nit: 
```suggestion
      if (timestampColumnName != null && !timestampColumnName.isEmpty()) {
```

Also the default value of timestampColumnName is "" so would it ever be null?

jhambleton: great points, removed the unnecessary null check and updated. thanks!

mutianf (approved):

liferoad:
https://console.cloud.google.com/dataflow/jobs/us-west2/2025-07-19_16_56_32-12758550340559396864?project=cloud-teleport-testing

Looks like something is wrong with the xlang container images:
```
docker: Error response from daemon: manifest for gcr.io/cloud-teleport-testing/2025-07-19-23-50-21_it/pubsub-proto-to-bigquery-xlang:latest not found: manifest unknown: Failed to fetch "latest".
```

PR #2571, opened 2025-07-18
[Spanner-to-sourcedb] Not reading generate columns from Spanner during reverse replication

darshan-sj:


darshan-sj:
> Do we need to do this only for vartiual columns?
> 
> image

Yes. We can read the stored columns. I tested with the following schema in spanner:

```
CREATE TABLE gen_col_test (
	id1 INT64 NOT NULL ,        -- From: id1 int(10)
	var1 INT64,                         -- From: var1 int(10)
	var2 INT64,                         -- From: var2 int(10)
	res_always_generated_stored INT64 as (var1 - var2) STORED,  -- From: res_always_generated_stored int(10)
	res_always_generated_virtual INT64 as (var1 + var2), -- From: res_always_generated_virtual int(10)
	res_generated INT64 as (var1 * var2),                -- From: res_generated int(10)
        var1_token TOKENLIST AS (TOKENIZE_NUMBER(`var1`, comparison_type=>'all')) HIDDEN,
        migration_shard_id STRING(50)
) PRIMARY KEY (migration_shard_id, id1);
```

Only virtual columns we cannot read with Read API

VardhanThigle (approved):

PR #2613, opened 2025-08-04
[DatastreamToSpanner] Dataflow failure injection test for live migration template

darshan-sj:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/utils/DataflowFailureInjector.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(DataflowFailureInjector.class);
+
+  /**
+   * Finds and Stops all worker VMs for a Dataflow job to simulate failure.
+   *
+   * @param projectId The Google Cloud project ID.
+   * @param jobId The Dataflow job ID.
+   */
+  public static void abruptlyKillWorkers(String projectId, String jobId)
```
VardhanThigle, line 38: What's the failure message or excaption that the failure injected by `abruptlyKillWorkers` cause?

In general for a few VMs turned down, dataflow is able to provision new VMs.

Is this something that the job can recover from or is it a terminal failure for the job?

v2/datastream-to-spanner/pom.xml
```diff
@@ -97,6 +97,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.google.cloud</groupId>
```
VardhanThigle, line 102: [nit] Should this be test scope?
If it can't be test scoped, could we add a comment that this is needed only for IT?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/utils/DataflowFailureInjector.java
darshan-sj: The job recovers automatically, and there is no error message or exception shown in dataflow job.

v2/datastream-to-spanner/pom.xml
darshan-sj: okay

darshan-sj: Done

VardhanThigle (approved):

PR #2614, opened 2025-08-04
[DO NOT MERGE] Validate Beam 2.67.0rc1

Amar3tto:


Amar3tto:
Python UDF tests are failing which is expected

PR #2626, opened 2025-08-07
Adding python flex template to convert raw text/image objects to ArrayRecord

PAenugula:
Adding python flex template to convert raw text/image GCS objects in bulk to ArrayRecord format. This allows ML researchers to take advantage of Grain pipelines.

- Allows passing in `input_path`, `input_format`, and `output_path`.

### DIFF
--- a/python/src/main/java/com/google/cloud/teleport/templates/python/ArrayRecordConverter.java
+++ b/python/src/main/java/com/google/cloud/teleport/templates/python/ArrayRecordConverter.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.templates.python;
+
+import com.google.cloud.teleport.metadata.Template;
+import com.google.cloud.teleport.metadata.TemplateCategory;
+import com.google.cloud.teleport.metadata.TemplateParameter;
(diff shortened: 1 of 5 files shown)


python/src/main/java/com/google/cloud/teleport/templates/python/ArrayRecordConverter.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.cloud.teleport.metadata.TemplateParameter;
+
+/** template class for ArrayRecordConverter in Python. */
+@Template(
+    name = "ArrayRecord_Converter",
+    category = TemplateCategory.STREAMING,
+    type = Template.TemplateType.PYTHON,
+    displayName = "ArrayRecord Converter Job",
+    description =
+        "The ArrayRecord_Converter Template is used to convert bulk text/image dataset in GCS into ArrayRecord Datasets in GCS"
```
liferoad, line 29: we probably can make this template hidden for now or change the description with something like use as it is and no support.

PAenugula: I think the "use as is" and "no support" is fine. However, would it still require users to locally pull this repository to deploy their own flex template and run?

damccorm: I don't think we can publicly publish a new template and not support it. Per https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/maintainers-guide.md#introducing-new-templates we generally are not accepting new templates, and I don't think we should take this as a result. I'd recommend adding this as a sample (maybe in https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/dataflow/flex-templates) so that customers can take it and build it themselves there if needed.

python/src/main/python/arrayrecord-converter/main.py
```diff
@@ -0,0 +1,141 @@
...
+      write_dir=_WRITE_DIR,
+      file_path_suffix='.arrayrecord',
+      overwrite_extension=False,
+  ):
+
+    ## Upload to GCS
+    def upload_to_gcs(
+        bucket_name, filename, prefix='', source_dir=self._WRITE_DIR
+    ):
+      source_filename = os.path.join(source_dir, filename)
```
liferoad, line 48: this will be expensive since you do this for every element. Better use the existing gcs IO.

PR #2629, opened 2025-08-07
Schema Map Support for Tables

rahulkumar0630:
Updated optional schema map parameter to support table mapping as well


dhercher (dismissed):

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/utils/DatastreamToMySQLDML.java
dhercher, line 1: Alright to leave as is, but consider if this logic could be pushed into the parent class so we don't duplicate (and dont need to implement for each new destination

rahulkumar0630: [Refactored and moved getTargetTableName to DatastreamtoDML.java](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2629/commits/818fe67658d74c7991b12727cd88049327eb3ab5)

rahulkumar0630:
Rebased with main before merging.

PR #2630, opened 2025-08-08
Fix security issues in datastream-to-spanner

manitgupta:
Tested by running a pipeline and migrating a few records..

### DIFF
--- a/v2/datastream-to-spanner/pom.xml
+++ b/v2/datastream-to-spanner/pom.xml
@@ -53,6 +53,21 @@
             <artifactId>json</artifactId>
             <version>${json.version}</version>
         </dependency>
+        <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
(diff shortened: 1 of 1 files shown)


v2/datastream-to-spanner/pom.xml
```diff
@@ -53,6 +53,21 @@
...
             <version>${json.version}</version>
         </dependency>
+        <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
+            <version>1.11.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-avro</artifactId>
```
liferoad, line 63: I suggest you wait for Beam 2.67.0, which fixes these issues.

manitgupta: Does 2.67.0 also fix `commons-beanutils` and `org.eclipse.jetty:jetty-server` issues as well?

Issue #2634, opened 2025-08-11
[Bug]: neo4j integration test failing


Abacn:
### Related Template(s)

googlecloud-to-neo4j

### Template Version

N/A

### What happened?

neo4j integration tests started failing since Aug 8, 2025 due to testcontainer startup timeout: 

https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/16812884401/job/47628672408

I can reproduce it using the following simple code 

```
package org.example;

import org.testcontainers.utility.DockerImageName;
import org.testcontainers.containers.Neo4jContainer;
public class Main {

  public static void main(String[] args) {
    Neo4jContainer container = new Neo4jContainer(DockerImageName.parse("neo4j").withTag("5-enterprise"));
    container.start();
  }
}
```

I get

```
[main] ERROR tc.neo4j:5-enterprise - Could not start container
java.lang.IllegalStateException: Wait strategy failed. Container exited with code 1
```

Likely the newly pushed neo4j containers had breaking changes

https://hub.docker.com/layers/library/neo4j/5-enterprise/images/sha256-50f97951018816afe431f386d62d8bebbcce91eb1aecd03f5af4209deb1f0bf6?context=explore

https://hub.docker.com/layers/library/neo4j/4.4-enterprise/images/sha256-3bac369722a861157b031cff56a8665c5a427857a4d2d1ffe9949f92bc6beafe

### Relevant log output

```shell

```

Abacn:
actually the reproduce above may not be relevant. Adding `container = (Neo4jContainer) container.withEnv("NEO4J_ACCEPT_LICENSE_AGREEMENT", "yes");` both old and new containers started successfully.

Abacn:
It is found this is due to #2624. The reason is unknown.

fbiville:
I see that `org.apache.beam.it.testcontainers.TestContainerResourceManager` has some instrumentation around log consumers, but only for "static" containers (see `org.apache.beam.it.testcontainers.TestContainerResourceManager.Builder#useStaticContainer`).
 
Collecting the actual container logs would actually be helpful to understand exactly what causes the container to not start.

Abacn:
It appears the ConcurrentHashMap in  #2624 caused issue. It causes some race condition on docker. 

[...]

PR #2635, opened 2025-08-11
Pin neo4j test container

Abacn:
mitigate #2634


Abacn:
pin neo4j container does not work. It's not due to neo4j container upgrade.

PR #2645, opened 2025-08-14
Add tag to staging image as well

Abacn:


Abacn:
Tested with us.gcr.io/dataflow-build/streaming-data-generator -> gcr.io/dataflow-build/streaming-data-generator

Abacn:
R: @damccorm

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
```diff
@@ -75,10 +77,11 @@ public void promote() throws IOException, InterruptedException {
     JsonElement parsed = JsonParser.parseString(responseRLO);
     String operation = parsed.getAsJsonObject().get("name").getAsString();
     waitForComplete(operation);
-    addTag(imageTag);
+    addTag(sourcePath, "public-image-latest");
```
damccorm, line 80: Is sourcePath the staging repo? If so, should we tag with something like `no-new-use-public-image-staging-image` so it doesn't get flagged?

Abacn: Thanks, communicating with internal team for clarification, mark as draft at the moment

PR #2648, opened 2025-08-15
Bump hbase-shaded-client to 2.6.3-hadoop3

ahmedabu98:
Fixes a few CVEs like [CVE-2023-44981](https://github.com/advisories/GHSA-7286-pgfv-vxvh)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
     <guava.version>32.0.1-jre</guava.version>
     <hadoop.version>3.4.1</hadoop.version>
     <hamcrest.version>2.1</hamcrest.version>
-    <hbase.client.version>2.5.3-hadoop3</hbase.client.version>
+    <hbase.client.version>2.6.3-hadoop3</hbase.client.version>
     <mockito.version>4.11.0</mockito.version>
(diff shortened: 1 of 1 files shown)


ahmedabu98:
cc @damccorm I think this could fix many vulnerabilities we're seeing

pom.xml
```diff
@@ -68,7 +68,7 @@
     <guava.version>32.0.1-jre</guava.version>
     <hadoop.version>3.4.1</hadoop.version>
     <hamcrest.version>2.1</hamcrest.version>
-    <hbase.client.version>2.5.3-hadoop3</hbase.client.version>
+    <hbase.client.version>2.6.3-hadoop3</hbase.client.version>
```
damccorm, line 71: Should we update https://github.com/apache/beam/blob/b17488ffddbb0923c87e13a709d6c04268bd8915/sdks/java/io/hbase/build.gradle#L37 to match (I see a similar version elsewhere in Beam)?

ahmedabu98: Good idea, opened a PR here: https://github.com/apache/beam/pull/35898

PR #2651, opened 2025-08-18
feat(dataflow): replace cancelJob with forceCancelJob

liferoad:
Add forceCancelJob method to ensure jobs are terminated during cleanup Add tests for forceCancelJob functionality
Update logging messages to reflect force cancel operation

This should speed up the job canceling time.


Abacn:
This likely caused resource leak of GCE Instance templates

image

leak since Aug 18, matches merge time of this PR

derrickaw:
KafkaToBigQuery IT pipeline failed due to maxed Instance Templates of 2,000. Currently deleting the older instance templates, hence the cliff at the end.  Thanks Yi!

PR #2662, opened 2025-08-20
revise metadata link and update docs

derrickaw:
The current metadata link in all the readme documentations is pointing to old location. Updated to new location.

There is a one-off update for useStorageWriteApi, but not related to the goal of this PR.


damccorm (approved):
Thanks!

PR #2671, opened 2025-08-21
fix: remove setuptools wheel

liferoad:
Clean up unused setuptools wheel file from Python installation

### DIFF
--- a/plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
+++ b/plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
@@ -26,7 +26,8 @@ RUN python -m venv /venv \
     && /venv/bin/pip install --no-cache-dir --require-hashes -U -r $REQUIREMENTS_FILE \
     && /venv/bin/pip download --require-hashes --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $REQUIREMENTS_FILE \
     && rm -rf /usr/local/lib/python$PY_VERSION/site-packages \
-    && cp -r /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
+    && cp -r /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/ \
+    && rm -rf /usr/local/lib/python$PY_VERSION/ensurepip/_bundled/setuptools-*.whl
     # change to mv once Beam 2.57.0 is released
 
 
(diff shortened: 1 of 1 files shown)


plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
```diff
@@ -26,7 +26,8 @@ RUN python -m venv /venv \
     && /venv/bin/pip install --no-cache-dir --require-hashes -U -r $REQUIREMENTS_FILE \
     && /venv/bin/pip download --require-hashes --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $REQUIREMENTS_FILE \
     && rm -rf /usr/local/lib/python$PY_VERSION/site-packages \
-    && cp -r /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/
+    && cp -r /venv/lib/python$PY_VERSION/site-packages /usr/local/lib/python$PY_VERSION/ \
+    && rm -rf /usr/local/lib/python$PY_VERSION/ensurepip/_bundled/setuptools-*.whl
```
damccorm, line 30: Will ensurepip still work without this wheel?

liferoad: for xlang templates, I think it does not matter since we do not use `ensurepip` later. But our IT tests should confirm this.

liferoad: All xlang tests are fine.

damccorm: Got it, yeah I was thinking of Yaml but that is a different dockerfile

damccorm: Can we add a todo to remove this once 2.68 is released (should have a beam side fix to this problem)

damccorm (approved):
Thanks - left one comment, but can be merged pending that

plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
liferoad: Done.

liferoad:
The tests kept failing with 
```
"Workflow failed. Causes: One or more operations had an error: 'operation-1755976934789-63d0d3d6e9145-e7286b3d-c603390c': [QUOTA_EXCEEDED] 'Quota 'INSTANCE_TEMPLATES' exceeded.  Limit: 1000.0 in region us-west2.'.
```

Not related to this PR. and the previous run https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/17133187905 also was green.

PR #2684, opened 2025-08-25
SourceDbToSpanner - spanner failure injection test

darshan-sj:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
```diff
@@ -184,6 +185,10 @@ private static void setupLogicalDbMigration(
           configContainer.getSrcTableToShardIdColumnMap(
               tableSelector.getSchemaMapper(), spannerTables);
```
VardhanThigle, line 187: Can we add a top level check to not get into the whole injection code if the pipeline parameter is not set?

The code has many layers, but would this lead to spurious logs of
```
 if (jsonParameter == null || jsonParameter.trim().isEmpty()) {
      LOG.error("JSON parameter for ErrorInjectionPolicyFactory cannot be null or empty.");
      return new NoOpPolicy();
    }

```

Also we probably don't need a NOOP policy if we can entirely avoid the injection frameworks at the root of the pipeline controller.

darshan-sj: Done. Added a top level check to inject the service only if the parameter is set.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
```diff
@@ -184,6 +185,13 @@ private static void setupLogicalDbMigration(
...
 
+      if (options.getFailureInjectionParameter() != null
+          && !options.getFailureInjectionParameter().isBlank()) {
```
VardhanThigle, line 189: [nit] Please consider using [StringUtils.html#isBlank](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence)) to avoid the separate null check.

This can be rolled into a subsequent PR you raise as well.

VardhanThigle (approved):

VardhanThigle (dismissed):
Looks good to me. A nit in recent changes but can be rolled in any subsequent PR as well.
Please ensure tests pass before merge.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/PipelineController.java
darshan-sj: Sure. Will incorporate it in subsequent PRs.

VardhanThigle (approved):

PR #2686, opened 2025-08-25
feat(import): add early index create threshold parameter

liferoad:
Add configurable threshold for determining when to create indexes before data loading to optimize performance. Default value is set to 40.

internal bug: b/440945726

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ImportPipeline.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ImportPipeline.java
@@ -204,6 +204,17 @@ public class ImportPipeline {
     ValueProvider<RpcPriority> getSpannerPriority();
 
     void setSpannerPriority(ValueProvider<RpcPriority> value);
+
+    @TemplateParameter.Integer(
+        order = 12,
+        optional = true,
+        description = "Early Index Create Threshold",
+        helpText =
+            "The threshold for the number of indexes and foreign keys that determines whether to create indexes before data loading. If the total number of indexes and foreign keys is less than this threshold, they will be created before data import for better performance. The default value is 40.")
+    @Default.Integer(40)
+    ValueProvider<Integer> getEarlyIndexCreateThreshold();
+
+    void setEarlyIndexCreateThreshold(ValueProvider<Integer> value);
   }
 
   public static void main(String[] args) {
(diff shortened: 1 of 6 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/ImportPipeline.java
```diff
@@ -204,6 +204,17 @@ public interface Options extends PipelineOptions {
     ValueProvider<RpcPriority> getSpannerPriority();
 
     void setSpannerPriority(ValueProvider<RpcPriority> value);
+
+    @TemplateParameter.Integer(
+        order = 12,
+        optional = true,
+        description = "Early Index Create Threshold",
+        helpText =
+            "The threshold for the number of indexes and foreign keys that determines whether to create indexes before data loading. If the total number of indexes and foreign keys is less than this threshold, they will be created before data import for better performance. The default value is 40.")
```
Abacn: from the code below it sounds the implementation is the opposite,  index created early when larger than the threshold

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
```diff
@@ -660,7 +667,7 @@ public void processElement(ProcessContext c) {
                             if (earlyIndexCreateFlag.get()
                                 && ((createForeignKeyStatements.size()
                                         + createIndexStatements.size())
-                                    >= EARLY_INDEX_CREATE_THRESHOLD)) {
+                                    >= earlyIndexCreateThreshold.get())) {
```
Abacn, line 670: Reading the original internal bug it seems the FR is to disable early creation. Shall we define a special number (0) when it is set to 0 then early creation is disabled?

Abacn (commented):
The change is LGTM. Since this is supposes to improve performance and user should not be worried about the default in general, does the user want to disable it suggest some underlying issue in this feature? As a followup good to confirm with Spanner team.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
liferoad: `earlyIndexCreateFlag` controls this. I do not think we need use 0 here.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportPipeline.java
liferoad: Good catch. Fixed it.

Abacn (approved):

Abacn:
We need to regenerate README.md: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/code-contributions.md#generated-documentation

liferoad:
> We need to regenerate README.md: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/code-contributions.md#generated-documentation

This will be done automatically with https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/update-docs.yml

PR #2687, opened 2025-08-25
Datastream To SQL - Added Default Casing Parameter Functionality

rahulkumar0630:
This PR introduces a new, optional template parameter, defaultCasing, that allows users to control the casing for target table and schema names.

The parameter accepts the following values:

DEFAULT: Converts names to lowercase (e.g., MyTable → mytable).

UPPERCASED: Converts names to uppercase (e.g., MyTable → MYTABLE).

CAPITALIZED: Capitalizes the first letter and lowercases the rest (e.g., MyTable → Mytable).

### DIFF
--- a/v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
+++ b/v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
@@ -226,8 +226,21 @@ public class DataStreamToSQL {
 
     void setDatabaseName(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.Enum(
         order = 13,
+        optional = true,
+        enumOptions = {@TemplateEnumOption("DEFAULT"), @TemplateEnumOption("UPPERCASED"), @TemplateEnumOption("CAPITALIZED")},
+        description = "Toggle for Table Casing",
+        helpText =
+            "A Toggle for table casing behavior. For example,(ie."
+                + "DEFAULT = table1 -> table1, Capitalized = table1 -> Table1")
+    @Default.String("DEFAULT")
+    String getDefaultCasing();
+
+    void setDefaultCasing(String value);
+
+    @TemplateParameter.Text(
+        order = 14,
         optional = true,
         description = "A map of key/values used to dictate schema name changes",
         helpText =
@@ -239,7 +252,7 @@ public class DataStreamToSQL {
     void setSchemaMap(String value);
 
     @TemplateParameter.Text(
-        order = 14,
+        order = 15,
         groupName = "Target",
         optional = true,
         description = "Custom connection string.",
@@ -251,7 +264,7 @@ public class DataStreamToSQL {
     void setCustomConnectionString(String value);
 
     @TemplateParameter.Integer(
-        order = 15,
+        order = 16,
         optional = true,
         description = "Number of threads to use for Format to DML step.",
         helpText =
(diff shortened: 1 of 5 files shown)


v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
```diff
@@ -226,9 +226,26 @@ public interface Options extends PipelineOptions, StreamingOptions {
...
     void setDatabaseName(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.Enum(
         order = 13,
         optional = true,
+        enumOptions = {
+          @TemplateEnumOption("DEFAULT"),
+          @TemplateEnumOption("UPPERCASED"),
+          @TemplateEnumOption("CAPITALIZED")
```
dhercher: what does this do exactly?  convert snake case to camel?

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
```diff
@@ -226,9 +226,26 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setDatabaseName(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.Enum(
         order = 13,
         optional = true,
+        enumOptions = {
+          @TemplateEnumOption("DEFAULT"),
+          @TemplateEnumOption("UPPERCASED"),
```
dhercher: I would just do UPPERCASE, i think its more typical

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
```diff
@@ -226,9 +226,26 @@ public interface Options extends PipelineOptions, StreamingOptions {
 
     void setDatabaseName(String value);
 
-    @TemplateParameter.Text(
+    @TemplateParameter.Enum(
         order = 13,
         optional = true,
+        enumOptions = {
+          @TemplateEnumOption("DEFAULT"),
```
dhercher: we should probably add LOWERCASE (this happens in PG as default, but should have an explicit option?

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
dhercher, line 1: There is also logic in DatastreamIO which runs withLowercaseSourceColumns() in this file.  If we are applying this to columns we need to make sure this is aligned across the template

v2/datastream-to-sql/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSQL.java
rahulkumar0630: Are you talking about TemplateEnumOption? This bounds the user response so that they need to provide one of the three. Any thing else than that will prevent template launch if user decided to use the defaultCasing parameter.

rahulkumar0630: This just capitalizes the table name 

mytable -> Mytable.

dhercher (approved):

PR #2694, opened 2025-08-26
[Yaml]: add dependency requirements etc

derrickaw:
1. Add readme_dependencies.md file to explain the dependency scripts.
2. Add `set -e` to generate_all_dependencies.sh to fail on first failure so that we don't get additional noisy failures that are not the root cause.
3. Add yaml path for requirements creation/update.
4. Add requirements file to yaml folder.


python/README_Dependencies.md
```diff
@@ -0,0 +1,50 @@
+# Python Dependency Management
+
+This document outlines the process for managing Python dependencies for the Python-based and a few other Dataflow templates in this repository.
```
damccorm: We already have https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/contributor-docs/maintainers-guide.md#validating-and-upgrading-beam-versions - could you improve those docs as needed instead of adding a new doc page? If discoverability is a problem, we could try to link this in other places (including in the autogenerated requirements files)

derrickaw: Yea, sorry, I had checked the contribution one, but didn't check the maintainer.  Will improve this one etc.  Thanks

derrickaw:
cc: @chamikaramj

contributor-docs/maintainers-guide.md
```diff
@@ -102,6 +102,28 @@ The remaining files can then be generated by running `sh python/generate_all_dep
 
 **Automated Updates**: The `.github/workflows/update-python-deps.yml` workflow automatically updates these Python dependencies weekly. You can also manually trigger this workflow to update all Python dependencies if needed. For an example of an automated dependency update PR, see https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2676. 
 
+### Prerequisites for Generating Dependencies
```
damccorm, line 105: Thanks! Had one more thought - could we mention that https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/update-python-deps.yml runs weekly now? @liferoad added it last week.

Otherwise, this LGTM - thank you!

derrickaw: done, thanks

damccorm: Oh this already existed right above my comment and I just can't read 🙃 

Thanks for adding the link!

damccorm (approved):
Thanks, this LGTM

PR #2697, opened 2025-08-27
Exclude Deprecated and Hidden Templates from public-image-latest

Abacn:


Abacn:
Tested with

- gcr.io/dataflow-build/kafka-to-gcs (deprecated), has "no-new-use-public-image-latest" tag
- gcr.io/dataflow-build/kafka-to-gcs-flex (not deprecated), has "public-image-latest" tag

Abacn:
R: @liferoad

Issue #2688, opened 2025-08-26
[Bug]: Dataflow job fails due to incorrect source type detection for Datastream CDC data


ksmin23:
### Related Template(s)

cloud_datastream_to_bigquery

### Template Version

2025-08-19-00_rc00

### What happened?

**1. Description:**

I have configured a data pipeline with the following architecture, based on the official documentation at [Implement Datastream and Dataflow for analytics](https://cloud.google.com/datastream/docs/implementing-datastream-dataflow-analytics):

  **Cloud SQL for MySQL -> Datastream -> GCS -> Pub/Sub -> Dataflow -> BigQuery**

**2. Problem:**

On the initial run, the data stored in Google Cloud Storage (GCS) by Datastream correctly contains `"read_method": "mysql-backfill-fulldump"`.

However, for subsequent Change Data Capture (CDC) events, the data is stored with `"read_method": "cdc"`.

This creates an issue when the Dataflow job needs to be restarted. If I cancel the existing Dataflow job and then create a new one with the identical configuration to process the CDC files from GCS, the job fails with an error.

The root cause of the error is that with `"read_method": "cdc"`, Dataflow fails to recognize the source type as MySQL and incorrectly identifies it as MongoDB.

- **Dataflow Worker Error Logs**: see the **Relevant log output**

**3. Suggested Solution:**

To ensure robust source type detection, I suggest modifying Datastream to always include the source database type in the `read_method` for CDC data.

For example, the value should be more specific, such as `"mysql-cdc"` or `"postgresql-cdc"`, to prevent ambiguity for downstream consumers like Dataflow.

### Relevant log output

```shell
org.apache.avro.AvroRuntimeException: Not a valid schema field: schema
	at org.apache.avro.generic.GenericData$Record.get(GenericData.java:282)
	at com.google.cloud.teleport.v2.datastream.transforms.FormatDatastreamRecordToJson.getMetadataSchema(FormatDatastreamRecordToJson.java:242)
	at com.google.cloud.teleport.v2.datastream.transforms.FormatDatastreamRecordToJson.apply(FormatDatastreamRecordToJson.java:155)
	at com.google.cloud.teleport.v2.datastream.transforms.FormatDatastreamRecordToJson.apply(FormatDatastreamRecordToJson.java:55)
	at org.apache.beam.sdk.extensions.avro.io.AvroSource$AvroBlock.readNextRecord(AvroSource.java:610)
[...]

PR #2715, opened 2025-08-29
feat(datastream): add source type override for datastream cdc data

liferoad:
Add support for overriding the source type detection in Datastream CDC data processing. This allows specifying the source type directly when automatic detection from read_method field is not reliable, particularly when the field contains 'cdc'.

Fixes https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/2688

We probably cannot change what `read_method` could contain without any breaking changes. I am wondering whether we can add this optional parameter `datastreamSourceType` to override the source type.


liferoad:
@ksmin23 are you able to review this or help test this?

dhercher (approved):
i might add specific values supported by this flag?  and reason im not totally sure of

liferoad:
> i might add specific values supported by this flag? and reason im not totally sure of

Shall we limit them to 'mysql', 'postgresql', 'oracle'? I am not sure what the source type could be used here. This is why I allow the arbitrary string. @dhercher

liferoad:
I am going to merge this PR for now. We can improve how to make `datastreamSourceType` better in the future if needed.

PR #2739, opened 2025-09-05
ci(release): add package-upgrade label to release notes configuration

liferoad:
Update release.yml and related files to include package-upgrade as a valid label for PRs This ensures dependency updates are properly categorized in release notes

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -12,3 +12,6 @@ changelog:
     - title: Bug fixes
       labels:
         - bug-fix
+    - title: Package Upgrades
+      labels:
+        - package-upgrade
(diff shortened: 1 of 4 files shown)


damccorm (commented):
Thanks! Could we add this as a label here as well - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/d426aaec6078a52223ce112aa8adb6fd3adb0f19/.github/workflows/update-python-deps.yml#L38 (the action takes a labels field)

liferoad:
> Thanks! Could we add this as a label here as well -
> 
> https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/d426aaec6078a52223ce112aa8adb6fd3adb0f19/.github/workflows/update-python-deps.yml#L38
> 
> (the action takes a labels field)

Good idea. Done.

PR #2757, opened 2025-09-10
Test: update dependency versions and jvm config

liferoad:
- Update mongo-java-driver from 3.12.11 to 3.12.14
- Update beam and beam-python versions to 2.69.0-SNAPSHOT
- Add additional JVM opens for reflection access


.mvn/jvm.config
```diff
@@ -4,4 +4,7 @@
 --add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
 --add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
 --add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
+--add-opens=java.base/java.lang=ALL-UNNAMED
```
liferoad, line 7: not needed

PR #2772, opened 2025-09-12
Add artifactRegistry parameter to staging flex template doc

Abacn:
Current documented command only works if the gcp project has container registry path`gcr.io/<project>` . However, container registry is discontinued, and `-DartifactRegisty` is needed to stage the flex template image into  artifact registry


Abacn:
merging for now. batch automated doc update that also done weekly by bot, no need for individual component owner approval

PR #2777, opened 2025-09-15
Update generate_dependencies.sh

liferoad:
Tested: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/17743730781

### DIFF
--- a/python/generate_dependencies.sh
+++ b/python/generate_dependencies.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Copyright 2024 Google Inc. All Rights Reserved.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


python/generate_dependencies.sh
```diff
@@ -15,6 +15,9 @@
 BASE_REQUIREMENTS_PATH=$1
 TARGET_REQUIREMENTS_PATH=$2
 
+# Set default Python version if not provided
+PY_VERSION=${PY_VERSION:-"3.11"}
```
damccorm: Actually, should we just get rid of this entirely? All we're doing with it is using it to create a virtual environment name, we could probably just always call it `python_requirements_gen`

liferoad: Sounds good. Done.

PR #2789, opened 2025-09-18
DataStreamToSpanner - Adding main table read at the beginning of main transaction

darshan-sj:
Fixing commit consistency issue in in Separate shadow table database case, by adding a main table read at the beginning of main read write transaction.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
@@ -27,6 +27,8 @@ import com.google.cloud.teleport.v2.spanner.migrations.exceptions.ChangeEventCon
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.DroppedTableException;
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.InvalidChangeEventException;
 import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -107,4 +110,19 @@ public Mutation getShadowTableMutation() {
   public String getShadowTable() {
     return shadowTable;
   }
+
+  // Fires a read on Data table with lock scanned ranges. Used to acquire exclusive lock on Data row
+  // at the beginning of a readWriteTransaction
+  public void readDataTable(final TransactionContext transactionContext, Ddl dataTableDdl) {
```
bharadwaj-aditya: Please call this method as readExclusive or something to clarify that it will take an elevated lock on the rows.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ShadowTableReadUtils.java
```diff
@@ -101,4 +101,82 @@ public static Statement generateShadowTableReadSQL(
     }
     return stmtBuilder.build();
   }
+
+  // TODO: After beam release, use the latest client lib version which supports setting lock
+  // hints via the read api. SQL string generation should be removed.
+  public static Statement generateDataTableReadSQL(
```
bharadwaj-aditya: Seems like the only difference between the 2 methods is column names generation. Can you extract that out into a method and re-use the existing method ?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
darshan-sj: Done.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ShadowTableReadUtils.java
darshan-sj: Done.

PR #2803, opened 2025-09-21
Making Spanner ITs run on dedicated project

darshan-sj:


.github/workflows/cleanup-spanner-test-infra.yml
```diff
@@ -43,8 +43,27 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r test-infra/requirements.txt
       - name: Run cleanup_pubsub_notifications.py
-        run: python test-infra/cleanup_pubsub_notifications.py
+        run: python test-infra/cleanup_pubsub_notifications.py --project cloud-teleport-testing
```
bharadwaj-aditya, line 48: Shouldn't this be span-cloud-teleport-testing ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/JDBCBaseIT.java
```diff
@@ -150,6 +150,21 @@ private String getMvnBaseRepoPath() {
...
           basePath.replace(new String(new byte[] {27, 91, 48, 109}, StandardCharsets.UTF_8), "");
 
+      // Split by newline and take the last non-empty line to avoid JVM warnings.
+      // The mvn command output can sometimes include unrelated JVM warnings (e.g.
+      // [warning][perf,memops])
+      // mixed with the actual output. The actual path is expected to be the last valid line.
+      // Example output:
+      // [0.005s][warning][perf,memops] Cannot use file /tmp/hsperfdata_runner/109499 ...
+      // /home/runner/.m2/repository
+      String[] lines = basePath.split("\\r?\\n");
```
bharadwaj-aditya: This seems very specific. How do we make sure it covers all cases of output ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
```diff
@@ -670,16 +670,24 @@ public Builder useStaticInstance() {
...
-              || Objects.equals(System.getProperty("spannerInstanceId"), "teleport"))) {
-        this.useStaticInstance = true;
-        List<String> instanceList = TestConstants.SPANNER_TEST_INSTANCES;
-        Random random = new Random();
-        int randomIndex = random.nextInt(instanceList.size());
-        String randomInstanceName = instanceList.get(randomIndex);
-        this.instanceId = randomInstanceName;
-      } else if (System.getProperty("spannerInstanceId") != null) {
+          || Objects.equals(projectId, "span-cloud-teleport-testing")) {
+        if (Strings.isNullOrEmpty(System.getProperty("spannerInstanceId"))
```
bharadwaj-aditya: who sets the spannerInstanceId ?
Can you add the flow to the comments / PR ?

.github/workflows/cleanup-spanner-test-infra.yml
darshan-sj: The span-cloud-teleport-testing project is there in line 67 and 69. We have some tests like DatastreamToSql and the Import/export tests run on old project for which cleanup pubsub notifications and cleanup spanner databases scripts need to be run on the old `cloud-teleport-testing` project.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/JDBCBaseIT.java
darshan-sj: Correct the logic and made it generic. Actually this is a side bug fix, and not directly related to "Making Spanner ITs run on dedicated project" work.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
darshan-sj: The spannerInstanceId is set in the maven command run to execute the integration tests. For the scheduled runs it is set in the go lang code in the cicd/cmd/ folder which generates and executes the tests for the github action workflows.

PR #2839, opened 2025-10-01
Skip reading schema if the value capture mode is NEW_ROW_AND_OLD_VALUES.

jiangzzhu:
When value capture mode is NEW_ROW_AND_OLD_VALUES, the new values will contain all columns of tracked table. Similar to value capture mode NEW_ROW, we don't need to read the information schema to get the new schema.


tianz101:
This PR LGTM, approve

liferoad (approved):

PR #2840, opened 2025-10-01
Only read schema when the col in mod is not in the stored schema.

jiangzzhu:
This is to optimize the performance to reduce the number of information schema read. Before this change, we will read information schema every time (1) the size of new value cols in mod is not equal to the number of cols in the stored schema or (2) the new value col in mod is not in the stored schema. 

Case (1) can happen very often if UPDATE/INSERT does not update all cols in a table, which will introduce a lot of information  schema read. Assume user alter table to add a new col, the first record after the alter table will update the stored information schema immediately.
Case (2) can happen if alter table to add a new col and UPDATE/INSERT update the new added col, which lead to an information schema read to get the new schema at the record commit timestamp.

After the change, we will not update the stored information schema immediately after alter table operation. But we will rely on the first UPDATE/INSERT record which update the alter added col, to update the stored information schema.


tianz101:
This lazy read optimization LGTM, and please ensure the following followups:
1) Make sure the relevant template doc is updated to reflect this.
2) Start with a pipeline with the updated template, test out the add Spanner table and add Spanner column are working correct end to end. Thanks.

PR #2842, opened 2025-10-02
Fix Setuptools Vulnerability

tarun-google:
Fixed Vulnerability bug b/448607999

### DIFF
--- a/python/generate_dependencies.sh
+++ b/python/generate_dependencies.sh
@@ -38,7 +38,7 @@ python3.11 -m venv "$ENV_PATH"
 # allow one-off executions of pip to generate requirements locally without alarming automation
 alias pip_automation="pip"
 
-pip_automation install --upgrade pip==25.0.1 setuptools wheel
+pip_automation install --upgrade pip==25.0.1 setuptools==80.3.1 wheel
 pip_automation install pip-tools
 
 # Install requirements from base file
--- a/python/default_base_python_requirements.txt
+++ b/python/default_base_python_requirements.txt
@@ -1,3 +1,3 @@
 apache-beam[gcp]==2.68.0
 jsonschema==4.25.1 # TODO: remove when correct version could come from beam
-setuptools
+setuptools==80.3.1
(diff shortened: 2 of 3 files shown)


python/src/main/python/job-builder-server/requirements.txt
```diff
@@ -967,11 +951,11 @@ jsonschema==4.25.1 \
     --hash=sha256:3fba0169e345c7175110351d456342c364814cfcf3b964ba4587f22915230a63 \
     --hash=sha256:e4a9655ce0da0c0b67a085847e00a3a51449e1157f4f75e9fb5aa545e122eb85
     # via
-    #   -r ./default_base_yaml_requirements.txt
+    #   -r python/default_base_yaml_requirements.txt
     #   apache-beam
-jsonschema-specifications==2025.4.1 \
-    --hash=sha256:4653bffbd6584f7de83a67e0d620ef16900b390ddc7939d56684d6c81e33f1af \
-    --hash=sha256:630159c9f4dbea161a6a2205c3011cc4f18ff381b189fff48bb39b9bf26ae608
+jsonschema-specifications==2025.9.1 \
```
liferoad: this wont work based on our tests. Caused many flaky cases somehow. The python updates are usually handled by https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/update-python-deps.yml weekly.

python/generate_dependencies.sh
```diff
@@ -38,7 +38,7 @@ python3.11 -m venv "$ENV_PATH"
 # allow one-off executions of pip to generate requirements locally without alarming automation
 alias pip_automation="pip"
 
-pip_automation install --upgrade pip==25.0.1 setuptools wheel
+pip_automation install --upgrade pip==25.0.1 setuptools==80.3.1 wheel
```
tvalentyn, line 41: wouldn't the  latest version of setuptools be installed?

tarun-google: I can do that too. I pushed it to version above vulnerability. Latest version is 80.9.0

tvalentyn: commented on the internal bug, i think the vulnerability alert comes from ensurepip, not setuptools. installing setuptools without version specified should install the latest version and should be preferred.

PR #2861, opened 2025-10-08
feat(kafka): add performance optimization options for Kafka consumers

liferoad:
Add new performance optimization configurations including max poll records, fetch min bytes, and buffer sizes. Introduce optimized Kafka reader methods that support message redistribution and offset deduplication. These changes improve throughput and latency for high-volume Kafka pipelines.


derrickaw:
Reopen if still valid. Thanks

PR #2663, opened 2025-08-20
fix: enforce non-vulnerable version of okio to address

liferoad:
Fix CVE-2023-3635

```
mvn dependency:tree -pl v2/googlecloud-to-clickhouse -am | grep okio                               
```

```
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.6.0:compile
[INFO] |  |     |  +- com.squareup.okio:okio:jar:1.17.6:runtime
[INFO] |  |     \- com.squareup.okio:okio-fakefilesystem:jar:3.4.0:compile
[INFO] |  |        \- com.squareup.okio:okio-fakefilesystem-jvm:jar:3.4.0:compile
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.6.0:compile
[INFO] |  |     |  +- com.squareup.okio:okio:jar:1.17.6:runtime
[INFO] |  |     \- com.squareup.okio:okio-fakefilesystem:jar:3.4.0:compile
[INFO] |  |        \- com.squareup.okio:okio-fakefilesystem-jvm:jar:3.4.0:compile
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.6.0:compile
[INFO] |  |     |  +- com.squareup.okio:okio:jar:1.17.6:runtime
[INFO] |  |     \- com.squareup.okio:okio-fakefilesystem:jar:3.4.0:compile
[INFO] |  |        \- com.squareup.okio:okio-fakefilesystem-jvm:jar:3.4.0:compile
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.6.0:compile
[INFO] |  |     |  +- com.squareup.okio:okio:jar:1.17.6:runtime
[INFO] |  |     \- com.squareup.okio:okio-fakefilesystem:jar:3.4.0:compile
[INFO] |  |        \- com.squareup.okio:okio-fakefilesystem-jvm:jar:3.4.0:compile
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.6.0:compile
[INFO] |  |     |  +- com.squareup.okio:okio:jar:1.17.6:runtime
[INFO] |  |     \- com.squareup.okio:okio-fakefilesystem:jar:3.4.0:compile
[INFO] |  |        \- com.squareup.okio:okio-fakefilesystem-jvm:jar:3.4.0:compile
```

PR #2866, opened 2025-10-09
Remove Okio META-INF files from the templates jars

chamikaramj:
We already upgraded Okio dependency past 1.17.5 in [1] due to CVE-2023-3635 but the META-INF of shaded templates jars seems to contain a reference to this resulting in a false positive for vulnerability trackers.

This PR prevents bundling the relavent META-INF files.

[1] https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2663


chamikaramj:
cc: @liferoad

PR #2867, opened 2025-10-10
changed mysql event ordering in datastream to spanner

bharadwaj-aditya:
Changed ordering check in Datastream-to-spanner

Old: Timestamp, binlog name, binlog position
New: binlog name, binlog position, Timestamp

Timestamp in the binlogs depends on time of DML, not of commit, which was leading to differences in ordering.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/MySqlChangeEventSequence.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/MySqlChangeEventSequence.java
@@ -168,14 +168,14 @@ class MySqlChangeEventSequence extends ChangeEventSequence {
     // For backfill events logfile will be null/empty.
     // These should always be treated as before the CDC events
     if (this.logFile == null && other.getLogFile() == null) {
-      // if two bin log events happen to come, order by time of reading
+      // if two backfill events happen to come, order by time of reading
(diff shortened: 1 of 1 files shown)


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/datastream/MySqlChangeEventSequenceTest.java
```diff
@@ -70,6 +70,26 @@ public void canOrderBasedOnScn() {
...
+  @Test
+  public void canOrderTwoDumpEvents() {
+    MySqlChangeEventSequence dumpEvent1 = new MySqlChangeEventSequence(eventTimestamp, null, -1L);
+    MySqlChangeEventSequence dumpEvent2 =
+        new MySqlChangeEventSequence(eventTimestamp + 1, null, -1L);
+
+    assertTrue(dumpEvent1.compareTo(dumpEvent2) < 0);
+    assertTrue(dumpEvent2.compareTo(dumpEvent1) > 0);
+  }
+
```
darshan-sj, line 92: We didn't have to change any existing tests?

bharadwaj-aditya: No. The existing tests covered the other scenarios did not check the scenario where the order of timestamp was descending but bin log was ascending.

PR #2872, opened 2025-10-10
Yaml: Add yaml syntax UT

derrickaw:
1. Adding yaml syntax UT to verify that the yaml blueprints placed in the yaml folder are syntactically correct, since there will be occasions where we do not have 1:1 mappings with yaml templates.  
2. mvn clean test -pl yaml -am

### DIFF
--- a/yaml/src/test/python/yaml_syntax_test.py
+++ b/yaml/src/test/python/yaml_syntax_test.py
@@ -29,42 +29,80 @@ from jinja2 import Environment, FileSystemLoader, meta
 logging.basicConfig(level=logging.DEBUG,
                     format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
 logging.getLogger('apache_beam').setLevel(logging.DEBUG)
+ 
+ 
+def create_test_method(template_name, yaml_dir):
+    """Creates a test method that validates a single YAML template.
 
+    This factory function generates a test method that will be dynamically added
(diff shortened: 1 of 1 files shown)


derrickaw:
Elastic search IT failure is unrelated

yaml/src/test/python/yaml_syntax_test.py
```diff
@@ -0,0 +1,70 @@
...
+# This will ensure logs are printed to console.
+# Using basicConfig with a format to make logs more informative.
+logging.basicConfig(level=logging.DEBUG,
+                    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+logging.getLogger('apache_beam').setLevel(logging.DEBUG)
+
+
+class YamlSyntaxTest(unittest.TestCase):
+    _logger = logging.getLogger(__name__)
+    def test_all_yaml_files_are_valid_beam_pipelines(self):
```
Abacn: pytest can generate test case dynamically. One can write test case per yaml file. Beam already did that for Yaml integration test: https://github.com/apache/beam/blob/05f6f01a33b3bc9bb5b8b928c6854c98a8068e26/sdks/python/apache_beam/yaml/integration_tests.py#L733

derrickaw: Done, thanks!

Abacn (approved):

PR #2873, opened 2025-10-13
feat(kafka-to-bigquery): Fix dynamic destination connection churn

anpag:
The Kafka-to-BigQuery template, when using DYNAMIC_TABLE_NAMES, was creating a new Storage Write API connection for nearly every record. This was caused by using the non-deterministic `GenericRecord` as the sharding key for `BigQueryIO`'s internal stream cache.

This change fixes the issue by using the deterministic string representation of the Avro schema as the destination key. This ensures that records with the same schema are processed in the same stream, allowing for connection reuse and preventing quota exhaustion.

This change:
- Modifies `BigQueryDynamicDestination` to use `String` as the destination type.
- Uses the Avro schema's canonical string as the destination key.
- Parses the schema string back to a `Schema` object in `getTable` and `getSchema`.
- Uses the deterministic `StringUtf8Coder`.


liferoad:
please run `mvn spotless:apply`

anpag:
I ran `mvn spotless:apply`, so it should be all set. Thanks @liferoad

PR #2881, opened 2025-10-15
add dependencies label for enforcement check

derrickaw:
1. Currently, dependabot PRs seem to always have a failure check for "Enforce labels for release notes / label (pull_request)".
2. Adding "dependencies" label as another valid label for enforcement so that this check passes in the future and a human doesn't need to do it manually.
3. Sorted the labels also.


derrickaw:
#2882 merged, so this PR is not relevant much anymore

PR #2906, opened 2025-10-21
Add delete spanner instance script

Abacn:


Abacn:
https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/18697249155/job/53317769433 works, but due to many backup-ed leaked instances, it timeout out before cleaning up all of them. Merging for now

PR #2909, opened 2025-10-21
update release notes labels logic

derrickaw:
If jobs run in parallel, there is a situation where the release label enforcer gets in front of the label dependabot job and the job will fail.


derrickaw:
Not needed, since Enforcement label workflow reruns on labeling.

PR #2919, opened 2025-10-24
[SpannerToSourceDb] Fixing CDC events processing logic for Reverse replication template

darshan-sj:
Fixing the CDC event processing logic in Reverse replication template

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
@@ -293,7 +293,7 @@ public class SourceWriterFn extends DoFn<KV<Long, TrimmedShardedDataChangeRecord
         Throwable cause = ex.getCause();
         if (cause == null) {
           // If cause is null, then it is a plain Spanner exception
-          outputWithTag(c, Constants.RETRYABLE_ERROR_TAG, cause.getMessage(), spannerRec);
+          outputWithTag(c, Constants.RETRYABLE_ERROR_TAG, ex.getMessage(), spannerRec);
         } else if (cause instanceof InvalidTransformationException) {
           invalidTransformationException.inc();
           outputWithTag(c, Constants.PERMANENT_ERROR_TAG, cause.getMessage(), spannerRec);


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
```diff
@@ -57,4 +61,38 @@ public void write(String sqlStatement) throws SQLException, ConnectionException
...
+      if (connObj == null) {
+        throw new ConnectionException("Connection is null");
+      }
+      connObj.setAutoCommit(false);
+      statement = connObj.createStatement();
+      statement.executeUpdate(sqlStatement);
+
+      // Check if version in shadow table has changed.
+      if (!commitCheck.check()) {
+        // This code path should never execute since Spanner automatically
```
bharadwaj-aditya: I think this comment is incorrect. This code path can execute if the lock is released in the outer transaction, which is why the check is required.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/TransactionalCheck.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+@FunctionalInterface
+public interface TransactionalCheck {
+  /**
+   * Performs the check.
+   *
+   * @return true to proceed with commit, false to trigger a rollback.
+   * @throws Exception if the check itself fails with an Exception.
+   */
+  boolean check() throws Exception;
```
bharadwaj-aditya: it would be good to throw a more specific - typed exception here.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/IDao.java
```diff
@@ -23,4 +23,17 @@ public interface IDao<T> {
...
+   * function. The transaction will be committed only if the callback function returns true. In any
+   * other case including the case where callback returned true but then the transaction could not
+   * be committed, an exception will be thrown.
+   *
+   * @param sqlStatement Query statement.
+   * @param commitCheck Callback function which will be executed and checked before committing the
+   *     transaction.
+   * @throws Exception If the sqlStatement could not be successfully committed.
+   */
+  void writeAndCheck(String sqlStatement, TransactionalCheck commitCheck) throws Exception;
```
bharadwaj-aditya: it should be possible to add the Transactional check as an option in the write method - and pass it as null in cases where it is not needed. That will avoid duplication of code.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShadowTableRecord.java
```diff
@@ -34,4 +34,9 @@ public Timestamp getProcessedCommitTimestamp() {
   public long getRecordSequence() {
     return recordSequence;
   }
+
+  public boolean equals(ShadowTableRecord other) {
```
bharadwaj-aditya, line 41: There needs to be a table level check here as well. If that is not relevant to this class, please add a comment with the reasoning.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/IDao.java
bharadwaj-aditya: I'm not sure why this is created as a generic though.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/TransactionalCheck.java
```diff
@@ -0,0 +1,31 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates.dbutils.dao.source;
+
+/**
+ * A functional interface for performing a check within a transaction. The check must return 'true'
+ * to commit. 'False' or exception would result in rollback. This is used in IDao.
+ */
+@FunctionalInterface
```
bharadwaj-aditya, line 22: This class is not being used as a functional interface. It is being used as a regular interface. 

Is there any value in having this annotation here ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/spanner/SpannerDao.java
```diff
@@ -82,10 +86,33 @@ public ShadowTableRecord getShadowTableRecord(
...
+    Struct row = resultSet.getCurrentRowAsStruct();
+    return new ShadowTableRecord(
+        row.getTimestamp(readColumnList.get(0)), row.getLong(readColumnList.get(1)));
+  }
+
+  public DatabaseClient getDatabaseClient() {
+    return spannerAccessor.getDatabaseClient();
+  }
+
+  public void updateShadowTable(Mutation mutation, TransactionContext transactionContext) {
```
bharadwaj-aditya, line 114: is there any value in keeping this method ? It uses no class level variables and it pretty basic.

bharadwaj-aditya (changes requested):

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
darshan-sj: If the lock is released in the outer transaction, then the read should throw an exception and should not succeed and hence the code inside the if block should not get executed. I'll clarify this in the comment.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/spanner/SpannerDao.java
darshan-sj: IMO all the reads and writes to the database should pass through DAO object. Hence I kept this method.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
darshan-sj: Removed the comment and updated the transactionalCheck interface.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/TransactionalCheck.java
darshan-sj: Updated the TransactionalCheck interface to throw a specific exception.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/IDao.java
darshan-sj: Merged the write and writeAndCheck method.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/ShadowTableRecord.java
darshan-sj: Added a TODO comment to improve the ShadowTableRecord interface to contain tableName.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/TransactionalCheck.java
darshan-sj: The annotation is needed to create and pass the transactionalCheck variable as a lambda function.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -197,81 +204,123 @@ public void processElement(ProcessContext c) {
...
-            outputWithTag(c, Constants.FILTERED_TAG, Constants.FILTERED_TAG_MESSAGE, spannerRec);
-          }
+                      if (!isSourceAhead) {
+                        IDao sourceDao = sourceProcessor.getSourceDao(shardId);
+                        TransactionalCheck check =
+                            () -> {
+                              ShadowTableRecord newShadowTableRecord =
+                                  spannerDao.readShadowTableRecordWithExclusiveLock(
+                                      shadowTableName, primaryKey, ddl, shadowTransaction);
+                              if (!shadowTableRecord.equals(newShadowTableRecord)) {
```
shreyakhajanchi: (!shadowTableRecord.equals(newShadowTableRecord)) will throw a NullPointerException if they are null right?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -197,81 +204,123 @@ public void processElement(ProcessContext c) {
...
+                                keysJson,
+                                spannerRec.getCommitTimestamp(),
+                                spannerRec.getRecordSequence()),
+                            shadowTransaction);
+                      }
+                      return null;
+                    });
+        // Source write metrics should be updated outside the shadowTableTransaction after the
+        // shadow table update is successful.
+        updateSourceWriteMetrics(shardId, spannerRec);
```
shreyakhajanchi: Shouldn't this be inside the `if (!isSourceAhead)` block?

darshan-sj: Are you saying the updateSourceWriteMetrics should be done only if a record is written to source? If we keep it inside the isSourceAhead block and the shadowTable transaction aborts, then we would be counting the writes multiple times. Would that be okay?

shreyakhajanchi: The intention is this - updateSourceWriteMetrics should be done only if a record is written to source. If you see the older logic it was called inside `processRecord` which was called only in case the source was not ahead. I understand the retry bit but that was true earlier also so for now maybe lets keep it to how it was

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
darshan-sj: Yes, it will result in NPE. Corrected it.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
darshan-sj: Makes sense. Moved it back to how it was earlier.

PR #2925, opened 2025-10-24
Apply GCP Application Default authentication when using a Managed Service for Apache Kafka cluster.

fozzie15:
This change allows the streaming data generator to be used in conjunction with Google's Managed Service for Apache Kafka. 

Due to the lack of ability to create UTs/ITs which create and utilize Managed Service resources and the existing flakiness of the kafka IT, this was tested manually. 4 tests were conducted:
1. JSON using local Kafka cluster
2. JSON using Managed Service cluster
3. Avro using local Kafka cluster
4. Avro using Managed Service cluster

These were conducted by modifying the existing Kafka Integration test to utilize these resources as necessary. These cases all pass, and I observed data moving through both kafka clusters.

### DIFF
--- a/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/transforms/StreamingDataGeneratorWriteToKafka.java
+++ b/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/transforms/StreamingDataGeneratorWriteToKafka.java
@@ -50,7 +50,8 @@ import org.joda.time.Instant;
 /** A {@link PTransform} that converts generatedMessages to write to Spanner table. */
 public final class StreamingDataGeneratorWriteToKafka {
 
-  private static final Pattern ManagedKafkaRegex = Pattern.compile(".*managedkafka\\..*\\.goog.*");
+  private static final Pattern ManagedKafkaRegex =
+      Pattern.compile("bootstrap\\..*\\.managedkafka\\..*\\.cloud\\.goog.*");
 
   private StreamingDataGeneratorWriteToKafka() {}
 


v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/transforms/StreamingDataGeneratorWriteToKafka.java
```diff
@@ -49,6 +50,8 @@
 /** A {@link PTransform} that converts generatedMessages to write to Spanner table. */
 public final class StreamingDataGeneratorWriteToKafka {
 
+  private static final Pattern ManagedKafkaRegex = Pattern.compile(".*managedkafka\\..*\\.goog.*");
```
Dippatel98: What is the format of the managed kafka bootstrap address that this matcher is used for?

Is it `bootstrap.<cluster-name>.<region>.managedkafka.<project-name>.cloud.goog:9092`?

fozzie15: Yes, this is the format. It is not documented anywhere on GCP pages (at least from my research on the product), so in the event that parts of it change i've left it mostly generic, while specifically looking for the service name and the goog tag.

Dippatel98: Can we also add `bootstrap` and `cloud` to the regex, I believe those should be relatively permanent as well. A backwards incompatible change should be pretty rare.

fozzie15: added.

PR #2927, opened 2025-10-27
SSL support for Cassandra Bulk Migration.

VardhanThigle:
# Implement Cassandra Client-to-Server TLS Connection support

This commit addresses implements support for Cassandra client-to-server TLS connections .

## Steps to Connect to Cassandra Client-to-Server TLS

To establish a secure client-to-server TLS connection with Cassandra, follow these steps:

1.  Upload the JKS truststore file, which is needed to verify the server-side certificate, to a GCS bucket.
2.  Provide the GCS path to this file using the `extraFilesToStage` parameter. For example:
    ```
    --parameters="extraFilesToStage=gs://your-bucket/path/to/truststore.jks"
    ```
3.  In your driver configuration file, reference the trust store. The path should be `/extra_files/` followed by the filename:
    ```
    advanced.ssl-engine-factory {
      class = DefaultSslEngineFactory
      truststore-path = "/extra_files/truststore.jks"
      truststore-password = "your-truststore-password"
    }
    ```
4.  Ensure that the trust store contains a valid certificate and that the Cassandra hostname matches the Common Name (CN) in the certificate's chain of trust.

## Detailed Overview of the Changes
1. At a high level Client to Server TLS needs the client (Dataflow in this case) to be able to verify the TLS certificate presented by the server (Cassandra in this case)
2. Cassandra Java driver does take care of the verification and uses a JKS trust store presented to it via the driver config file for the same.
3. We use `extra_files_to_stage` as a mechanism to mount the trust store on the dataflow machines - Read [docs/guides/templates/ssl-certificates](https://cloud.google.com/dataflow/docs/guides/templates/ssl-certificates) for details.
4. The extra files are not mounted on the launcher vm by default, so we call Dataflow's JVM initializer to achieve this (this was recommended by dataflow team)
5. For the pipeline run time (after launcher vm), CassandraIO needs a serializable class to provide the SSL options which this PR takes care of.

## Note on Unit Testing
All relevant code paths introduced or modified in this commit are now covered by unit tests. Note that the `main` function of a pipeline can not be unit tested.

Crucially, the unit tests for Cassandra-related components, including those for SSL changes, are executed against an embedded Cassandra instance configured in SSL mode. This ensures that the SSL-related logic is thoroughly tested in a controlled environment.

## Manual Testing
Manual migration from an SSL enabled Cassandra to Spanner has passed.

## Integration test
**TODO(b/455762944)**: Add integration tests to validate the end-to-end Cassandra to Spanner migration with SSL enabled in a deployed Dataflow environment.

### DIFF
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLCustomTransformationsNonShardedIT.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLCustomTransformationsNonShardedIT.java
@@ -75,7 +75,7 @@ public class MySQLCustomTransformationsNonShardedIT extends SourceDbToSpannerITB
   }
 
   @Test
-  public void simpleTest() throws Exception {
+  public void testCustomTransformationsNonSharded() throws Exception {
     loadSQLFileResource(mySQLResourceManager, MYSQL_DDL_RESOURCE);
     createSpannerDDL(spannerResourceManager, SPANNER_DDL_RESOURCE);
     createAndUploadJarToGcs("CustomTransformationAllTypes");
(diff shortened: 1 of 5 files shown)


VardhanThigle:
Its are failing due to [b/456319695](https://b.corp.google.com/456319695)

rohitwali (dismissed):
Thanks for the changes. Pl create a tracking bug for MySQL as well

rohitwali (approved):

PR #2928, opened 2025-10-27
Exposing spannerio maxCommitDelay as a DF job parameter.

sm745052:
- set maxCommitDelay=-1 to use the default setting


manitgupta:
How does this value change with a workload shape? This is something that ***needs*** to be configurable by the user?

VardhanThigle:
> How does this value change with a workload shape? This is something that _**needs**_ to be configurable by the user?

We will have to run more exhaustive tests to get to an out of box default - but it is showing good improvments for shapes with secondary indexes.

PR #2929, opened 2025-10-27
Remove special handling for only PK column tables

manitgupta:
### Description

Fixes b/455473612.

Updated the logic such that tables with only PK columns are not treated as a special case. 

1. The `MySQLDMLGenerator` will **always** generate an `INSERT ON DUPLICATE KEY UPDATE` statement. 
2. For tables which have all columns in PK, the `UPDATE` clause will contain only PK columns.
3. For tables which have subset of columns in PK, the `UPDATE` clause will contain both PK columns and Non-PK columns.

In summary, all columns are of the tables are always added in the `UPDATE` clause, and it is always generated.

### Testing

1. Tested that `ON DUPLICATE KEY UPDATE` works when only PK columns are specified. 

```
CREATE TABLE resource_access (
    user_id INT,
    group_id INT,
    resource_id INT,
    PRIMARY KEY (user_id, group_id, resource_id)
);
```

This works without error on repeated execution.

```
INSERT INTO resource_access (user_id, group_id, resource_id) 
VALUES (101, 5, 99)
ON DUPLICATE KEY UPDATE
    user_id = VALUES(user_id),
    group_id = VALUES(group_id),
    resource_id = VALUES(resource_id);
```

5. Tested the behaviour on both MySQL 5.7 and 8.
6. Added a unit test for the PK only table use-case.
7. Ran a test like so - 
    1. Create a reverse replication stack and inserted an entry `(1,2,3)` into a table with the above schema.
    2. Delete the entire stack.
    3. Delete the row from Spanner. (The row is still present on MySQL).
    4. Create the RR stack again.
    8. Inserted the same row again on Spanner. 
        1. **Without change** - Runs into `The exception while processing shardId: shard1 is java.sql.SQLIntegrityConstraintViolationException: Duplicate entry '1-2-3' for key 'resource_access.PRIMARY` error.
        2. **With change** - No error seen in the Dataflow logs and the `success_record_count` is successfully incremented indicating the record was actually processed (and was a No-Op for MySQL).

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
@@ -121,10 +121,7 @@ public class MySQLDMLGenerator implements IDMLGenerator {
   }
 
   private static DMLGeneratorResponse getUpsertStatement(
-      String tableName,
-      List<String> primaryKeys,
-      Map<String, String> columnNameValues,
-      Map<String, String> pkcolumnNameValues) {
+      String tableName, Map<String, String> allColumnNameValues) {
 
     String allColumns = "";
     String allValues = "";
@@ -132,34 +129,7 @@ public class MySQLDMLGenerator implements IDMLGenerator {
 
     int index = 0;
 
-    for (Map.Entry<String, String> entry : pkcolumnNameValues.entrySet()) {
-      String colName = entry.getKey();
-      String colValue = entry.getValue();
-
-      allColumns += "`" + colName + "`";
-      allValues += colValue;
-      // Add PK columns as well to the UPDATE clause
-      updateValues += " `" + colName + "` = " + colValue;
-
-      // Add comma if not the last item in this loop
-      if (index + 1 < pkcolumnNameValues.size()) {
-        allColumns += ",";
-        allValues += ",";
-        updateValues += ",";
-      }
-      index++;
-    }
-
-    // Add a separator comma if we had PKs AND we have non-PKs to add
-    if (!pkcolumnNameValues.isEmpty() && !columnNameValues.isEmpty()) {
-      allColumns += ",";
-      allValues += ",";
-      updateValues += ",";
-    }
-
-    index = 0;
-
-    for (Map.Entry<String, String> entry : columnNameValues.entrySet()) {
+    for (Map.Entry<String, String> entry : allColumnNameValues.entrySet()) {
       String colName = entry.getKey();
       String colValue = entry.getValue();
       allColumns += "`" + colName + "`";
@@ -167,7 +137,7 @@ public class MySQLDMLGenerator implements IDMLGenerator {
       updateValues += " `" + colName + "` = " + colValue;
 
       // Add comma if not the last item in this loop
-      if (index + 1 < columnNameValues.size()) {
+      if (index + 1 < allColumnNameValues.size()) {
         allColumns += ",";
         allValues += ",";
         updateValues += ",";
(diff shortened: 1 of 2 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -130,34 +130,43 @@ private static DMLGeneratorResponse getUpsertStatement(
...
+      // Add PK columns as well to the UPDATE clause
+      updateValues += " `" + colName + "` = " + colValue;
 
-    if (columnNameValues.size() == 0) { // if there are only PKs
-      // trim the last ','
-      allColumns = allColumns.substring(0, allColumns.length() - 1);
-      allValues = allValues.substring(0, allValues.length() - 1);
+      // Add comma if not the last item in this loop
+      if (index + 1 < pkcolumnNameValues.size()) {
+        allColumns += ",";
```
bharadwaj-aditya: nit: using a string.join method in these scenarios is a more standard way of appending such stuff.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGeneratorTest.java
```diff
@@ -256,6 +256,38 @@ public void primaryKeyNotPresentInSourceSchema() {
...
+        mySQLDMLGenerator.getDMLStatement(
+            new DMLGeneratorRequest.Builder(
+                    modType, tableName, newValuesJson, keyValuesJson, "+00:00")
+                .setSchemaMapper(schemaMapper)
+                .setDdl(ddl)
+                .setSourceSchema(sourceSchema)
+                .build());
+    String sql = dmlGeneratorResponse.getDmlStatement();
+    assertTrue(
+        sql.contains(
```
bharadwaj-aditya: The order of columns in sql will be non-deterministics. This test will probably not work always.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
manitgupta: Surely, but this code follows the appending pattern, so I didn't using `StringJoiner` etc and went for keeping the same syntax/structure.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
```diff
@@ -130,34 +130,43 @@ private static DMLGeneratorResponse getUpsertStatement(
     String allValues = "";
     String updateValues = "";
 
+    int index = 0;
```
bharadwaj-aditya, line 130: Do we need any different handling for PKs and non-pks ?

If we just merge the 2 maps it will probably work fine ?

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGeneratorTest.java
manitgupta: Ahh, makes sense. Let me switch to individual `contains`

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/MySQLDMLGenerator.java
manitgupta: I think we can merge, will take it up.

Issue #2784, opened 2025-09-17
[Bug]: staging artifacts fails silently during integration test


stankiewicz:
### Related Template(s)

all

### Template Version

all

### What happened?

example: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2758
run: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/17644268059/job/50147658864?pr=2758

underlying failure of this job is missing container as staging template had a failed staging task

### Relevant log output

```shell
[pool-1-thread-39] INFO org.apache.beam.it.gcp.TemplateTestBase - Preparing test for GCS_Text_to_BigQuery_Xlang (class com.google.cloud.teleport.v2.templates.TextIOToBigQuery)
[pool-1-thread-39] INFO org.apache.beam.it.gcp.TemplateTestBase - Running command to stage templates: mvn compile package -q -f /home/runner/actions-runner/_work/DataflowTemplates/DataflowTemplates/pom.xml -pl metadata,v2/common,v2/googlecloud-to-googlecloud -am -PtemplatesStage,pluginOutputDir -DpluginRunId=MvheQFXG5S9bBMI7 -DskipShade=false -DskipTests -Dmaven.test.skip -Dcheckstyle.skip -Dmdep.analyze.skip -Dspotless.check.skip -Denforcer.skip -DprojectId=cloud-teleport-testing -Dregion=us-west2 -DbucketName=cloud-teleport-testing-it-gitactions -DgcpTempLocation=cloud-teleport-testing-it-gitactions -DstagePrefix=2025-09-11-15-10-36_IT -DtemplateName=GCS_Text_to_BigQuery_Xlang -DunifiedWorker=null -e
[Thread-31] INFO org.apache.beam.it.gcp.TemplateTestBase - Sep 11, 2025 3:19:45 PM com.google.cloud.teleport.plugin.TemplateSpecsGenerator saveMetadata
[Thread-31] INFO org.apache.beam.it.gcp.TemplateTestBase - INFO: Saving image spec metadata /home/runner/actions-runner/_work/DataflowTemplates/DataflowTemplates/v2/googlecloud-to-googlecloud/target/plugin/MvheQFXG5S9bBMI7/classes/text-to-bigquery-xlang-generated-metadata.json
[...]

PR #2937, opened 2025-10-28
chore(templates-maven-plugin): Add retry logic to template staging

liferoad:
The template staging process for both Classic and Flex templates can occasionally fail due to transient network issues or temporary service unavailability. This can lead to flaky and unreliable builds.

This change introduces a retry mechanism using the Failsafe library. The staging operation will now be retried up to 3 times with a 10-second delay between attempts, making the build process more resilient to these intermittent failures.

Fixes https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/2784

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
@@ -304,10 +304,14 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
   public String stageTemplate(
       TemplateDefinitions definition, ImageSpec imageSpec, BuildPluginManager pluginManager)
       throws MojoExecutionException, IOException, InterruptedException, TemplateException {
+    RetryPolicy<String> retryPolicy =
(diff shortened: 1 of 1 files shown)


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -304,10 +304,14 @@ public void execute() throws MojoExecutionException {
   public String stageTemplate(
       TemplateDefinitions definition, ImageSpec imageSpec, BuildPluginManager pluginManager)
       throws MojoExecutionException, IOException, InterruptedException, TemplateException {
+    RetryPolicy<String> retryPolicy =
+        RetryPolicy.<String>builder().withDelay(Duration.ofSeconds(10)).withMaxRetries(3).build();
     if (definition.isClassic()) {
-      return stageClassicTemplate(definition, imageSpec, pluginManager);
+      return Failsafe.with(retryPolicy)
+          .get(() -> stageClassicTemplate(definition, imageSpec, pluginManager));
```
liferoad, line 311: I remember we might fail silently. If so, this retry might not be useful.

liferoad:
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2943 should be a better solution.

PR #2943, opened 2025-10-30
fix(maven-plugin): Validate container image existence after staging

liferoad:
The `gcloud builds submit` command can occasionally return a non-zero exit code even when the container image is built and pushed successfully. This was causing the template staging process to fail unnecessarily.

This change introduces a new validation step that runs after the build process completes. It uses `gcloud artifacts docker images describe` to explicitly verify that the expected image exists in the Artifact Registry.

This makes the staging process more robust by failing only if the final image artifact is confirmed to be missing, rather than relying solely on the build process's exit code. The error handling for the build process has also been relaxed to a warning for classic templates to accommodate this.

Fixes https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/2784

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
@@ -1139,14 +1139,11 @@ public class TemplatesStageMojo extends TemplatesBaseMojo {
             cloudBuildLogs);
 
     int retval = stageProcess.waitFor();
+    // Ideally this should raise an exception, but this sometimes return NZE even for successful
+    // runs.
     if (retval != 0) {
-      LOG.warn(
-          "Building YAML image nonzero return code {}. This does not necessarily mean an error. "
-              + "Check logs for details. {}",
-          retval,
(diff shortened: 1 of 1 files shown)


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1302,6 +1305,7 @@ private void stageFlexTemplateUsingCloudBuild(
           retval,
           cloudBuildLogs);
     }
+    validateImageExists(imagePathTag, buildProjectId);
```
Abacn: I think we can put these validateImageExists calls inside "retval != 0" block?

liferoad: Good idea.

Abacn (approved):
Thanks!

liferoad:
v2/googlecloud-to-neo4j/target/surefire-reports/TEST-com.google.cloud.teleport.v2.neo4j.templates.DataConversionIT.xml | 4 ✅ | 1 ❌ |  
-- | -- | -- | --


Not related to this PR.

PR #2950, opened 2025-11-03
main and shadow table as a Side Input in spanner-to-sourcedb

sm745052:
Fixes b/456048565

Added main and shadow table as a Side Input in spanner-to-sourcedb so that the job startup time is less.

Context:
when main and shadow table are given as main input, when there are large number of tables, the startup time exceeded 12 min which is the default startup timeout for a dataflow job.

Earlier:
Using 4k tables: Startup time exceeded 12 min

Now:
Tested 4k tables: Startup time was around 4 min

NOTE:
- Also added some TODOs in AssignShardIdFnTests which seem to be a bit off. Will need to correct them in another PR since that is out of scope of this PR.

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
@@ -72,9 +72,9 @@ import org.apache.beam.runners.dataflow.options.DataflowPipelineWorkerPoolOption
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
 import org.apache.beam.sdk.coders.KvCoder;
-import org.apache.beam.sdk.coders.SerializableCoder;
 import org.apache.beam.sdk.coders.StringUtf8Coder;
 import org.apache.beam.sdk.coders.VarLongCoder;
+import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
 import org.apache.beam.sdk.io.FileSystems;
 import org.apache.beam.sdk.io.fs.ResolveOptions.StandardResolveOptions;
 import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig;
@@ -683,11 +683,9 @@ public class SpannerToSourceDb {
             .setCoder(FailsafeElementCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of()));
 
     PCollection<TrimmedShardedDataChangeRecord> dlqRecords =
-        dlqJsonStrRecords
-            .apply(
-                "Convert DLQ records to TrimmedShardedDataChangeRecord",
-                ParDo.of(new ConvertDlqRecordToTrimmedShardedDataChangeRecordFn()))
-            .setCoder(SerializableCoder.of(TrimmedShardedDataChangeRecord.class));
+        dlqJsonStrRecords.apply(
+            "Convert DLQ records to TrimmedShardedDataChangeRecord",
+            ParDo.of(new ConvertDlqRecordToTrimmedShardedDataChangeRecordFn()));
     PCollection<TrimmedShardedDataChangeRecord> mergedRecords = null;
 
     if (options.getFailureInjectionParameter() != null
@@ -708,13 +706,11 @@ public class SpannerToSourceDb {
               // stream data
               .apply("Reshuffle", Reshuffle.viaRandomKey())
               .apply("Filteration", ParDo.of(new FilterRecordsFn(options.getFiltrationMode())))
-              .apply("Preprocess", ParDo.of(new PreprocessRecordsFn()))
-              .setCoder(SerializableCoder.of(TrimmedShardedDataChangeRecord.class));
+              .apply("Preprocess", ParDo.of(new PreprocessRecordsFn()));
       mergedRecords =
           PCollectionList.of(changeRecordsFromDB)
               .and(dlqRecords)
-              .apply("Flatten", Flatten.pCollections())
-              .setCoder(SerializableCoder.of(TrimmedShardedDataChangeRecord.class));
+              .apply("Flatten", Flatten.pCollections());
     } else {
       mergedRecords = dlqRecords;
     }
@@ -761,8 +757,7 @@ public class SpannerToSourceDb {
                     // same source type
                     .withSideInputs(ddlView))
             .setCoder(
-                KvCoder.of(
-                    VarLongCoder.of(), SerializableCoder.of(TrimmedShardedDataChangeRecord.class)))
+                KvCoder.of(VarLongCoder.of(), AvroCoder.of(TrimmedShardedDataChangeRecord.class)))
             .apply("Reshuffle2", Reshuffle.of())
             .apply(
                 "Write to source",
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/changestream/TrimmedShardedDataChangeRecord.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/changestream/TrimmedShardedDataChangeRecord.java
@@ -19,6 +19,11 @@ import com.google.cloud.Timestamp;
 import com.google.gson.annotations.SerializedName;
 import java.io.Serializable;
 import java.util.Objects;
+import org.apache.avro.reflect.AvroEncode;
+import org.apache.avro.reflect.Nullable;
+import org.apache.beam.sdk.coders.DefaultCoder;
+import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
+import org.apache.beam.sdk.io.gcp.spanner.changestreams.encoder.TimestampEncoding;
 import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.Mod;
 import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.ModType;
 
@@ -26,9 +31,13 @@ import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.ModType;
  * Trimmed version of the Apache Beam DataChangeRecord class that only contains the field we need in
  * this pipeline.
  */
-@SuppressWarnings("initialization.fields.uninitialized") // Avro requires the default constructor
+@SuppressWarnings("initialization.fields.uninitialized")
+@DefaultCoder(AvroCoder.class)
 public class TrimmedShardedDataChangeRecord extends java.lang.Object implements Serializable {
+
+  @AvroEncode(using = TimestampEncoding.class)
   private Timestamp commitTimestamp;
+
   private String serverTransactionId;
   private String recordSequence;
   private String tableName;
@@ -36,11 +45,15 @@ public class TrimmedShardedDataChangeRecord extends java.lang.Object implements 
   private ModType modType;
   private long numberOfRecordsInTransaction;
   private String transactionTag;
-  private String shard;
+
+  @Nullable private String shard;
   private boolean isRetryRecord;
 
   @SerializedName("_metadata_retry_count")
   private long metadataRetryCount;
+
+  // AvroCoder requires the default constructor
+  public TrimmedShardedDataChangeRecord() {}
 
   public TrimmedShardedDataChangeRecord(
       com.google.cloud.Timestamp commitTimestamp,
--- a/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/transforms/PreprocessRecordsFnTest.java
+++ b/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/transforms/PreprocessRecordsFnTest.java
@@ -18,7 +18,6 @@ package com.google.cloud.teleport.v2.templates.transforms;
 import com.google.cloud.Timestamp;
 import com.google.cloud.teleport.v2.templates.changestream.TrimmedShardedDataChangeRecord;
 import java.util.Arrays;
-import org.apache.beam.sdk.coders.SerializableCoder;
 import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.DataChangeRecord;
 import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.Mod;
 import org.apache.beam.sdk.io.gcp.spanner.changestreams.model.ModType;
@@ -38,10 +37,7 @@ public class PreprocessRecordsFnTest {
   @Test
   public void basicTest() {
     PCollection<TrimmedShardedDataChangeRecord> output =
-        pipeline
-            .apply(Create.of(getDataChangeRecord()))
-            .apply(ParDo.of(new PreprocessRecordsFn()))
-            .setCoder(SerializableCoder.of(TrimmedShardedDataChangeRecord.class));
+        pipeline.apply(Create.of(getDataChangeRecord())).apply(ParDo.of(new PreprocessRecordsFn()));
 
     PAssert.that(output)
         .containsInAnyOrder(


sm745052:
> It always a great practice to add details of the testing done in the PR description! ;) Please be as elaborate as you want to.

Addressed

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -600,23 +600,25 @@ public static PipelineResult run(Options options) {
...
-                .getDatabase(
-                    spannerMetadataConfig.getInstanceId().get(),
-                    spannerMetadataConfig.getDatabaseId().get())
-                .getDialect(),
-            options.getShadowTablePrefix());
+    // Fetch DDLs and create shadow tables in a DoFn to avoid launcher-side timeout.
+    PCollectionTuple ddlTuple =
+        pipeline.apply(
+            "Process Information Schema",
+            new SpannerToSourceDbProcessInformationSchema(
```
manitgupta: `PTransform` names should end in `*Transform` to reflect that they are PTransforms. Should be named to reflect the action that the class is doing: `SpannerInformationSchemaProcessorTransform`.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -959,7 +959,7 @@ private static SourceSchema fetchSourceSchema(Options options, List<Shard> shard
     return sourceSchema;
   }
 
-  private static ISchemaMapper getSchemaMapper(Options options, Ddl ddl) {
+  public static ISchemaMapper getSchemaMapper(Options options, Ddl ddl) {
```
manitgupta: Why did the access modifier change here?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -182,7 +181,7 @@ public void setup() {
   /** Teardown function disconnects from the Cloud Spanner. */
   @Teardown
   public void teardown() {
-    if (spannerConfig != null) {
+    if (spannerAccessor != null) {
```
manitgupta: What was breaking earlier?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -194,14 +193,31 @@ public void teardown() {
    */
   @ProcessElement
   public void processElement(ProcessContext c) throws Exception {
+    Ddl ddl = c.sideInput(ddlView);
+
+    if (this.schemaMapper == null) {
+      SpannerToSourceDb.Options options =
+          c.getPipelineOptions().as(SpannerToSourceDb.Options.class);
+      this.schemaMapper = SpannerToSourceDb.getSchemaMapper(options, ddl);
+    }
```
manitgupta: This seems like an anti-pattern and is very confusing. 

1. `@ProcessElement` is called for every element. This code would result in a schemaMapper being initialized at the time of processing an element. Logically, schema mapping should be initialized prior to processing elements.
2. You should not be importing `SpannerToSourceDb` here.  This looks like an inverted import.


Overall the code flow seems very odd to me. Please take a look.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -180,6 +181,15 @@ public void teardown() throws Exception {
...
   @ProcessElement
   public void processElement(ProcessContext c) {
+    Ddl ddl = c.sideInput(ddlView);
+    Ddl shadowTableDdl = c.sideInput(shadowTableDdlView);
+
+    if (this.schemaMapper == null) {
+      SpannerToSourceDb.Options options =
+          c.getPipelineOptions().as(SpannerToSourceDb.Options.class);
+      this.schemaMapper = SpannerToSourceDb.getSchemaMapper(options, ddl);
+    }
```
manitgupta: Same comment as earlier.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterTransform.java
```diff
@@ -95,17 +92,17 @@ public SourceWriterTransform.Result expand(
...
                         this.sourceSchema,
                         this.shadowTablePrefix,
                         this.skipDirName,
                         this.maxThreadPerDataflowWorker,
                         this.source,
-                        this.customTransformation))
+                        this.customTransformation,
+                        this.ddlView,
+                        this.shadowTableDdlView))
+                .withSideInputs(ddlView, shadowTableDdlView)
```
manitgupta, line 121: Why is `ddlView` and `shadowTableDdlView` passed both as a parameter and as a SideInput? Won't the `sideInput` already be available inside the `DoFn`?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
```diff
@@ -0,0 +1,155 @@
...
+      SpannerConfig shadowTableSpannerConfig,
+      String shadowTablePrefix) {
+    this.spannerConfig = spannerConfig;
+    this.shadowTableSpannerConfig = shadowTableSpannerConfig;
+    this.shadowTablePrefix = shadowTablePrefix;
+  }
+
+  @Override
+  public PCollectionTuple expand(PBegin p) {
+    return p.apply("Create empty", Create.of((Void) null))
```
manitgupta: This is a confusing name. Root transforms are generally called `Pulse` or something similar.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
```diff
@@ -0,0 +1,155 @@
...
+        .apply(
+            "Create Shadow tables and return Information Schemas",
+            ParDo.of(
+                    new ProcessInformationSchemaFn(
+                        spannerConfig, shadowTableSpannerConfig, shadowTablePrefix))
+                .withOutputTags(MAIN_DDL_TAG, TupleTagList.of(SHADOW_TABLE_DDL_TAG)));
+  }
+
+  static class ProcessInformationSchemaFn extends DoFn<Void, Ddl> {
```
manitgupta: Nit: My personal preference is to have a DoFns which non-trivial amount of logic into their own class. It makes the `PTransform` using these `DoFns` easy to extend in the future without increasing the class length too much.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
```diff
@@ -0,0 +1,155 @@
...
+        spannerAccessor.close();
+      }
+      if (shadowTableSpannerAccessor != null) {
+        shadowTableSpannerAccessor.close();
+      }
+    }
+
+    @ProcessElement
+    public void processElement(ProcessContext c) {
+      LOG.info("Starting processing of Information Schema...");
```
manitgupta: Is this needed? Let's add logs so that they add value when something goes wrong..

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
sm745052: it was not breaking but this seems more logical

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterTransform.java
sm745052: we need a reference for the sideinput which we get via the constructor. LMK if i am missing something.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
sm745052: ack

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
sm745052: ack

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SpannerToSourceDbProcessInformationSchema.java
sm745052: ack

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
sm745052: 1. Now the schema mapper is constructed for each invocation of process element
2. Circular import is avoided

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
sm745052: ack

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -250,7 +283,7 @@ public void processElement(ProcessContext c) throws Exception {
 
   @NotNull
   private Map<String, Object> getSpannerRecordMapAndUpdateDeletedValue(
-      TrimmedShardedDataChangeRecord record, String tableName, String keysJsonStr)
+      TrimmedShardedDataChangeRecord record, String tableName, String keysJsonStr, Ddl ddl)
```
bharadwaj-aditya, line 269: does the DDL need to be passed everywhere ? It can be a class level variable here.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -569,7 +604,7 @@ private Object getColumnValueFromRow(Column column, Value value) throws Exceptio
     }
   }
 
-  private boolean doesTableExistAtSource(String tableName) {
+  private boolean doesTableExistAtSource(String tableName, ISchemaMapper schemaMapper) {
```
bharadwaj-aditya, line 590: same comment here. The schema mapper can be a class level variable here like it was previously

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFnTest.java
```diff
@@ -215,17 +240,20 @@ public void testProcessElementInsertModForMultiShard() throws Exception {
...
     assignShardIdFn.setSpannerAccessor(spannerAccessor);
     ObjectMapper mapper = new ObjectMapper();
     mapper.enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
     assignShardIdFn.setMapper(mapper);
-    assignShardIdFn.setShardIdFetcher(
-        ShardingLogicImplFetcher.getShardingLogicImpl("", "", "", schemaMapper, "skip"));
+    assignShardIdFn.setSchemaMapper(schemaMapper);
 
-    assignShardIdFn.processElement(processContext);
+    assignShardIdFn.processElement(processContext); // TODO: This should be called only once
```
bharadwaj-aditya: what is the TODO here ?

sm745052: To remove one of the two calls to processElement.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
sm745052: earlier it was initialized during startup. Now we are getting the ddl via sideinput. IIUC, there are 2 ways to go
1. Cache the ddl from side input in a class variable like,
```
if(ddl==None){
   ddl = c.sideInput(ddlView);
}
```
2. Use it directly without caching.


There might be a value in not having the sideInput cached. This [documentation](https://beam.apache.org/documentation/patterns/side-inputs/#slowly-updating-side-input-using-windowing) is a commonly used pattern for sideinputs where they can change. Although this is not the case in our code today, caching it might prohibit us to change the sideinput in future.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFnTest.java
bharadwaj-aditya: Why is this being called twice ? Maybe update the reason in the comment

sm745052: It seems like a logical error which doesn't have impact on test since later it is checking if it is called atleast once. I tested by removing one call and that passed the test too.

```
verify(processContext, atLeast(1)).output(eq(KV.of(key, record)));
```

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
sm745052: same reason as with DDL

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
bharadwaj-aditya: That is not as relevant here because this will be a static/global side input.

This should ideally just come from the schema mapper, but that might be a larger change. I think it is fine to go this way for now.

bharadwaj-aditya (dismissed):
Minor comment. LGTM

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/SchemaMapperUtils.java
```diff
@@ -0,0 +1,68 @@
...
+package com.google.cloud.teleport.v2.templates.utils;
+
+import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.ISchemaMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.IdentityMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaFileOverridesBasedMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaStringOverridesBasedMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SessionBasedMapper;
+
+public class SchemaMapperUtils {
```
manitgupta, line 25: Possible to add a unit test for this now that it is refactored out? (We should have already written it this way and this is not a blocker for your PR but it will be good citizenship if you add that in :) - please do so only if you have the time for it.)

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/ProcessInformationSchemaFn.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.ddl.InformationSchemaScanner;
+import com.google.cloud.teleport.v2.templates.utils.ShadowTableCreator;
+import org.apache.beam.sdk.io.gcp.spanner.SpannerAccessor;
+import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProcessInformationSchemaFn extends DoFn<Void, Ddl> {
```
manitgupta, line 35: Looks like other `Fn` classes have unit tests added. Can you add unit test for this? 

Please also have a look at any other new classes you have created in this PR and add unit tests for them.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -90,48 +92,59 @@ public class SourceWriterFn extends DoFn<KV<Long, TrimmedShardedDataChangeRecord
...
   private transient SpannerDao spannerDao;
-  private final Ddl ddl;
-  private final Ddl shadowTableDdl;
   private final SourceSchema sourceSchema;
   private final String shadowTablePrefix;
   private final String skipDirName;
   private final int maxThreadPerDataflowWorker;
   private final String source;
-  private SourceProcessor sourceProcessor;
+  private transient SourceProcessor sourceProcessor;
```
manitgupta: In general its dangerous to make these unrelated changes in a PR targeting a specific feature...was there a reason you made this change?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -194,21 +206,42 @@ public void teardown() {
...
+    IShardIdFetcher shardIdFetcher = this.shardIdFetcher;
+    if (shardIdFetcher == null) {
+      shardIdFetcher =
+          ShardingLogicImplFetcher.getShardingLogicImpl(
+              customJarPath,
+              shardingCustomClassName,
+              shardingCustomParameters,
+              schemaMapper,
+              skipDirName);
+    }
```
manitgupta: Lets follow a similar code flow to `SourceWriterFn` here as well.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/SchemaMapperUtils.java
sm745052: Done

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
sm745052: Done. Added raw session files for that ... there is was a problem in ShardingLogicImplFetcher class where the shardidfetched is a static variable which was causing issues in the tests .. so added a function to reset the static variable for the tests

PR #2957, opened 2025-11-03
[TEST] test(mysql): add waitForMySqlReady for sourcedb tests

liferoad:


VardhanThigle (approved):

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/SourceDbToSpannerITBase.java
```diff
@@ -388,4 +402,44 @@ private String driverClassNameFrom(JDBCResourceManager jdbcResourceManager) {
       throw new IllegalArgumentException(e);
     }
   }
+
+  /**
+   * Simple health-check for MySQL that retries {@code SELECT 1} with exponential backoff. Ensures
+   * the container or proxy is ready before applying schema or performing writes.
+   */
+  private void waitForMySqlReady(JDBCResourceManager jdbcResourceManager) {
+    if (!(jdbcResourceManager instanceof MySQLResourceManager)) {
```
VardhanThigle, line 411: This change is good to have and harmless but it might not be directly helping the current issue as much as the timeout paramter tuning since:
1. The test case is able to create schema and tables in the container before dataflow job runs even in failing runs
2. In failing runs Dataflow attempts to connect to mysql fails a few times, retries and it succeeds but by that tine dataflow polling timeout kicks in .

The timeout seems to help - though I am still wondering what has degraded the network in IT project in recent times - would it be just the volume of tests?

darshan-sj: I reduced the number of tests and the same failure happened there as well - https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2954/checks 

But this change has consistently passed for 4 times. We should just merge this to unblock all the PRs.

VardhanThigle: Thanks @liferoad  and @darshan-sj 
Taking this change in to help unblock.

We Will continue to monitor and finetune as need be.

PR #2967, opened 2025-11-05
Dedup image scan result for identical images built on same dependency

Abacn:


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
```diff
@@ -256,7 +257,13 @@ String getDigestFromTag(String tag) throws IOException, InterruptedException {
     }
     // The response is expected to be just the digest, e.g., "sha256:..."
     // Trim any leading/trailing whitespace.
-    return response.trim();
+    response = response.trim();
+    if (response.startsWith("sha256:")) {
+      return response;
+    }
+    // gcloud container doesn't fail on image not found
```
Abacn, line 264: currently if there isn't already an image with public-image-latest tag, staging template command would fail for `gcr.io` registries. Fixed here.

Abacn:
Tested with ` mvn verify -PtemplatesRelease   -DprojectId=...   -DbucketName=...   -DlibrariesBucketName=...   -DstagePrefix=moss-test-1   -Dmaven.test.skip -T8 -e   -DstagingArtifactRegistry=...   --settings=.mvn/settings.xml   -pl v2/jdbc-and-googlecloud -am` 4 containers are published on same dependencies, scanvola only run once. This saves ~5 min. (Should save more for xlang templates, as Python image scan takes longer).

liferoad (approved):

PR #2969, opened 2025-11-06
Unique tag

Abacn:


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/PromoteHelper.java
```diff
@@ -227,7 +227,7 @@ String getDigestFromTag(String tag) throws IOException, InterruptedException {
             "images",
             "list-tags",
             targetPath, // This is the image name, e.g., us.gcr.io/my-project/my-image
-            "--filter=tags:" + tag,
+            "--filter=tags=" + tag,
```
liferoad, line 230: WARNING: --filter : operator evaluation is changing for consistency across Google APIs.  tags=public-image-latest currently does not match but will match in the near future.  Run `gcloud topic filters` for details.

Abacn: read this warning it suggests this would work for newer gcloud cli. Also this outputs to stderr so it does not affect digest resolving (in stdout)

PR #2962, opened 2025-11-04
fix(build): update surefire report processing to handle time format

liferoad:
* disable cache when using cloud build
* sed changes:
Before sed:

```
<testsuite>
  <testcase name="test1" time="1,147.352"/>
  <testcase name="test2" time="1,234,567.890"/>
  <testcase name="test3" time="123.456"/>
  <testcase name="test4" time="1,234"/>
  <testcase name="test5" time="1,234,567"/>
  <testcase name="test6" time="0.123"/>
</testsuite>
```

After sed:
```
<testsuite>
  <testcase name="test1" time="1147.352"/>
  <testcase name="test2" time="1234567.890"/>
  <testcase name="test3" time="123.456"/>
  <testcase name="test4" time="1234"/>
  <testcase name="test5" time="1234567"/>
  <testcase name="test6" time="0.123"/>
</testsuite>
```

TODO:

* remove the version pins for jsonschema
* explore using cache with the template name as the part of URLs

PR #2984, opened 2025-11-13
Add auto yaml to java generation

derrickaw:
1. Update KafkaToBigQuery.yaml to jive with existing KafkaToBigQuery.java template.
2. Add generation script to convert yaml blueprint to java interface.
~~3. Add workflow to inspect each pull request opened to rerun generation script and create or modify java templates.~~
4. Add java template file as a baseline file.
5. Add generic readme on how to add new blueprints, templates, etc.
6. Add unit tests for this generation script.
7. mvn clean install -PtemplatesValidate -pl yaml -am

### DIFF
--- a/yaml/README.md
+++ b/yaml/README.md
@@ -13,7 +13,7 @@ Place this file in [here](https://github.com/GoogleCloudPlatform/DataflowTemplat
 1.  **Generate YAML Template (Optional):** If desired, create the YAML template using the generation script.
 
 ```shell
-python yaml/scripts/generate_yaml_java_templates.py yaml/src/main/yaml
+python yaml/src/main/python/generate_yaml_java_templates.py yaml/src/main/yaml
 ```
 
 1.  **Create Integration Test:** Develop an integration test to validate the functionality of the new template.


.github/workflows/yaml_blueprint_to_java_template.yml
```diff
@@ -0,0 +1,71 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow generates or modifies Java template files based on changes
+# to YAML files in a pull request.
+
+name: YAML to Java Template Generation or Modification
+
+on:
+  pull_request:
```
damccorm: I think we probably want to run this on different conditions (probably commits to this branch on the main branch or on a schedule). We should also add a workflow_dispatch trigger

.github/workflows/yaml_blueprint_to_java_template.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  job-generate-templates:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5.0.0
+        with:
+          fetch-depth: 0
```
damccorm: Any reason for this? This will slow clones

.github/workflows/yaml_blueprint_to_java_template.yml
damccorm: Oh, I see - you're committing it back to the pull request branch. In general, I think it is probably better to have this as a standalone thing rather than tying it to PRs - that way it can be run if we update the underlying generation infrastructure (or make similar other changes)

damccorm: I also think this won't work on forks since I don't think the workflow has permissions on forks. I'd recommend following the pattern in https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/update-python-deps.yml

damccorm: Alternatively, you could require that a user generate these as part of their pull request and fail the check if they don't (similar to how linting checks work). To do this, you could do all the same things you're already doing, but finish with `git diff --exit-code` to fail the job if there are differences detected.

In general, I'd stay away from modifying a user's branch though, since it is surprising and will run into those permission problems.

tarun-google: i like the final suggestion, at least for the initial version. Let us give instructions for users to generate java interface from yaml using your script(We did similar setup for ManagedIO docs). Committing to working branch might not be good. I am happy not to have a validation check (PR vs script output) too. We might not want to create Template for every Blueprint.

.github/workflows/yaml_blueprint_to_java_template.yml
derrickaw: When I was iterating on this and with my understanding, it was required to have the full history to know about the branch and its context for pushing back to the same PR.  Its a mute point now based on other comments, so will remove. Thanks.

.github/workflows/yaml_blueprint_to_java_template.yml
derrickaw: For the initial version:
1. Removed the workflow.
2. Added a readme file explaining how to add the yaml blueprint, interface, IT, etc.
3. Added mvn spotless process to the generate script from the workflow file to minimize that probable check failure.

Thanks

tarun-google (approved):
LGTM

yaml/src/main/yaml/KafkaToBigQuery.yaml
```diff
@@ -1,60 +1,113 @@
...
+      description: "Kafka Bootstrap Server list"
       help: "Kafka Bootstrap Server list, separated by commas."
+      example: "localhost:9092,127.0.0.1:9093"
       required: true
       type: text
+      order: 1
 
     - name: "kafkaReadTopics"
-      help: "Kafka topic to read the input from."
+      description: "Kafka topic(s) to read the input from."
```
tarun-google, line 39: QQ: Are we displaying these new fields in UI too?

derrickaw: This yaml file isn't used currently in the UI for Job Builder.  The template fields in the java template are used currently in the Create Job from template.

yaml/README.md
```diff
@@ -0,0 +1,20 @@
+# Creating a YAML Template
+
+This document outlines the steps required to create a new YAML template for Dataflow.
+More detailed instructions on how to run the template can be found in the associated README files for each template.
+
+## Steps
+
+1.  **Add YAML Blueprint:** Create the YAML blueprint file that defines the template's structure and parameters.
```
chamikaramj: Add a link on contribution instructions to this Github repo ?

yaml/README.md
```diff
@@ -0,0 +1,20 @@
...
+Place this file in [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/yaml/src/main/yaml).
+
+2.  **Generate YAML Template (Optional):** If desired, create the YAML template using the generation script.
+
+```shell
+python yaml/scripts/generate_yaml_java_templates.py yaml/src/main/yaml
+```
+
+3.  **Create Integration Test:** Develop an integration test to validate the functionality of the new template.
+Place this file in [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/yaml/src/test/java).
```
chamikaramj, line 20: Any existing instructions on creating the integration tests for Flex templates that we can link to ?

yaml/scripts/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,242 @@
...
+import sys
+
+from pathlib import Path
+
+import yaml
+import textwrap
+
+
+def get_java_type(param_type):
+    """Maps a YAML parameter type to a Java type."""
```
chamikaramj: Why are we limited to these three types here ?

yaml/scripts/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,242 @@
...
+        return 'String'
+    elif param_type == 'integer':
+        return 'Integer'
+    elif param_type == 'boolean':
+        return 'Boolean'
+    else:
+        return 'String'
+
+def get_template_parameter_type(param_type):
+    """Maps a YAML parameter type to a TemplateParameter annotation type."""
```
chamikaramj: Ditto.

yaml/src/main/python/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,242 @@
...
+        return e
+
+def generate_java_interface(yaml_path, java_path):
+    """Generates a Java interface file from a YAML blueprint."""
+
+    # Read the YAML file and do some replacements
+    with open(yaml_path, 'r') as f:
+        content = f.read()
+        # Remove Jinja variables before parsing
+        content = re.sub(r'{{.*?}}', '', content)
```
chamikaramj, line 86: Probably we should have some validation to make sure we do not remove required params ?

yaml/src/main/python/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,242 @@
...
+      helpText = "{wrapped_help_text}",
+      example = "{example}"
+    )
+"""
+        # required param
+        if param.get('required', False):
+            param_code += "  @Validation.Required\n"
+
+        # default param
+        if 'default' in param:
```
chamikaramj, line 128: Ditto regarding types.

yaml/src/main/python/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,242 @@
...
+    # Sort the list of YAML files
+    yaml_files = sorted(list(yaml_files))
+
+    if not yaml_files:
+        print(f"No YAML files found in {input_path}", file=sys.stderr)
+        sys.exit(1)
+
+    # Convert each YAML file to a Java interface
+    try:
+        for yaml_path in yaml_files:
```
chamikaramj, line 210: Is it possible that we might have other YAML files that are not intended to be pipelines ?

yaml/README.md
derrickaw: done, thanks

yaml/README.md
derrickaw: done, thanks

yaml/src/main/python/generate_yaml_java_templates.py
derrickaw: Under normal situation jinja variables are rendered before the they are loaded.  In this case we do not care about the jinja variables since this exercise really only cares about the metadata of the file, but it prevents us from loading the file correctly, so just doing a dummy replace for now.

yaml/src/main/python/generate_yaml_java_templates.py
derrickaw: The idea is that this location would only contain the golden yaml blueprints.

yaml/scripts/generate_yaml_java_templates.py
derrickaw: The template parameter types are limited based on this - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/metadata/src/main/java/com/google/cloud/teleport/metadata/TemplateParameter.java#L24

Updated a little.

yaml/scripts/generate_yaml_java_templates.py
derrickaw: Goes back to the limitation of https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/metadata/src/main/java/com/google/cloud/teleport/metadata/TemplateParameter.java#L24

Updated a little.

yaml/src/main/python/generate_yaml_java_templates.py
derrickaw: see other comments, thanks

yaml/src/main/python/generate_yaml_java_templates.py
```diff
@@ -0,0 +1,231 @@
...
+            text=True)
+        print("Successfully ran mvn spotless:apply.")
+    except FileNotFoundError as e:
+        print("Error: 'mvn' command not found. Please ensure Maven is installed and in your PATH.")
+        return e
+    except subprocess.CalledProcessError as e:
+        print(f"Error running mvn spotless:apply: {e}", file=sys.stderr)
+        return e
+
+def generate_java_interface(yaml_path, java_path):
```
chamikaramj, line 68: Is it possible to add some unit testing to cover these functions ?

chamikaramj:
Lemme know when this is ready for merging.

yaml/src/main/python/generate_yaml_java_templates.py
derrickaw: Yep, shifted the file to under yaml/src/main/python, instead of yaml/scripts and then created a test script.  Thanks

derrickaw:
Hi @chamikaramj, ready to merge. Thanks.

chamikaramj:
I don't have access to merge. Probably @damccorm can do it.

chamikaramj:
(nvm, could merge after resolving comments)

Abacn:
PubSubToBigQueryAutoIT, PubSubToBigQueryIT, KafkaToBigQueryYamlIT appear to become increasingly flaky after this change. The last 3 scheduled runs of Java PR workflow all failed.

e.g.

https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/20027395602

```
testPubsubToBigQueryWithPythonUdf

 java.lang.IllegalArgumentException: Error staging template

Image gcr.io/cloud-teleport-testing/pubsub-to-bigquery-xlang:2025-12-08-13-06-15_IT was not created properly. Check the build logs for more details.
```

update:

the error on cloud build is

```
error building image: error building stage: failed to get filesystem from image: GET https://gcr.io/v2/dataflow-templates-base/java17-template-launcher-base-distroless/blobs/sha256:fd4aa3667332c2a4837e1b5f395b0555b3c4ec299e38166ae93ee84bad01befa: TOOMANYREQUESTS: Quota exceeded for quota metric 'Requests per project in the US multi-region' and limit 'Requests per project in the US multi-region per minute' of service 'artifactregistry.googleapis.com' for consumer 'project_number:167633336322'.
```

likely we need to re-enable cache for Cloudbuild's `gcr.io/kaniko-project/executor` step

https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/2962/files#r2495990868

derrickaw:
Seems like reenabling the cache has fixed the flakiness - 

image

tarun-google:
I have also encountered staging issue for KafkaToBigQueryYamlIT template for a couple of runs during my PR, It is successful now.

PR #2993, opened 2025-11-16
Fix TrimmedShardedDataChangeRecord to use AvroCoder everywhere in the template.

aasthabharill:
b/459651641

#### Problem
We currently use SerializableCoder in Reverse Replication template which is quite slow and is contributing to the OOMs we've been seeing while running reverse jobs. 

#### Fix
Fix TrimmedShardedDataChangeRecord to use AvroCoder everywhere in the template.


v2/spanner-to-sourcedb/pom.xml
```diff
@@ -140,6 +140,18 @@
...
+            <artifactId>classic-templates</artifactId>
+            <version>1.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud.teleport</groupId>
+            <artifactId>classic-templates</artifactId>
+            <version>1.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
manitgupta: Why is this required?

aasthabharill: without this block i get this:

Error:  /home/runner/actions-runner/_work/DataflowTemplates/DataflowTemplates/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/changestream/TrimmedShardedDataChangeRecord.java:[19,73] package com.google.cloud.teleport.spanner.spannerio.changestreams.encoder does not exist

aasthabharill: I'm replacing this by adding the TimestampEncoding file to v2 itself so it can be self sufficient, please tal and lmk if that is a better approach.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/changestream/encoder/TimestampEncoding.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.avro.io.Encoder;
+import org.apache.avro.reflect.CustomEncoding;
+
+/**
+ * This encoder/decoder writes a com.google.cloud.Timestamp object as a pair of long and int to avro
+ * and reads a Timestamp object from the same pair. The long stores the number of seconds since
+ * January 1, 1970, 00:00:00 UTC. A negative value is the number of seconds before January 1, 1970,
+ * 00:00:00 UTC. The int stores the fractional seconds components, in the range 0..999999999.
+ */
+public class TimestampEncoding extends CustomEncoding<Timestamp> {
```
manitgupta: Why can't you use the TimestampEncoder that is used by the `DatachangeRecord` inside Apache beam? That will have a dependency either in beam or on a lower dependency library, right?

manitgupta:
I think you can close this PR for now since we need to do a round of functional and load testing before we can merge this. 
The right owner can take up those items (along with the fix).

PR #2998, opened 2025-11-17
[tests] Add integration tests for migrating to a Spanner database with default timezone different from normal to bulk, live, and reverse Spanner migration templates

nmemond:
This adds integration tests for the new schema migration option to set a default timezone for the target Spanner database implemented in https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/1229 for all three Spanner data migration pillars.

Because the default timezone can only be set when the DB does not contain any tables, I elected to create a new suite for these tests to avoid impacting any existing tests. The one exception to this was for the reverse migration, which already included a timezone test which I simply re-used.


v2/datastream-to-spanner/src/test/resources/DataStreamToSpannerTimezoneIT/DateData.avro
VardhanThigle, line 1: A quick Question
Was this generated from an actual Datastream? If no, we might want to quickly run datastream and collect the Avros for this schema.

Here are the mappings of Datastream for Avro Types: https://docs.cloud.google.com/datastream/docs/unified-types.

[Edit] Removed the PG part of this review comment as current focus is on mySql as a source.

nmemond: Ah, no, it wasn't. I copied and modified one of the existing avro files, `DataStreamToSpannerDDLIT/mysql-cdc1-AllDatatypeColumns.avro`. I assumed it had the correct mappings, I'll look into fixing this. Thanks!

nmemond: I've updated the avro file to one that I got from running a datastream against the schema. Note that the actual table data is the same, I believe because MySQL `DATETIME` is mapped to Unified type `TIMESTAMP` when the target is CloudStorage (and presumably Spanner).

VardhanThigle: That's right. Thanks for updating the file.

v2/sourcedb-to-spanner/src/test/resources/TimezoneIT/mysql-schema.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE DateData (
+    id INT NOT NULL,
+    timestamp_column TIMESTAMP,
+    datetime_column DATETIME,
+ PRIMARY KEY(id));
+
+SET time_zone = 'Australia/Brisbane';
```
VardhanThigle, line 7: @shreyakhajanchi 
How does default time zone interplay with start and end timestamps for reverse replication?

The [documentation](https://docs.cloud.google.com/spanner/docs/set-default-time-zone#:~:text=Some%20functions%2C%20such%20as%20TIMESTAMP,database%20to%20customize%20this%20behavior.) does not talk a lot about CDC.

Also, what happens in case shadow table is in same and or different databases?

VardhanThigle: Will take this over chat.

PR #3003, opened 2025-11-19
[spanner-to-sourcedb] Custom Transformation fails when SQL PK col is dropped in Spanner

aasthabharill:
[b/458636564](https://b.corp.google.com/issues/458636564)

## Scenario: 
SQL schema has column opponent_id which is part of a composite PK. This column is dropped in Spanner (and thus from spanner pk as well).

## Error: 
When we try to use custom transformation to populate this column during reverse replication, it errors out with message: SQLException: Field 'opponentId' doesn't have a default value despite the value for opponent_id being generated by the custom transformer

## Fix: 
Check custom transformation response when checking for PK col before looking for it in Spanner table.

## Tests:
Template Container built in gs://ea-functional-tests/templates/flex/Spanner_to_SourceDb

#### SQL source
Ran [dataflow job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2025-11-24_01_20_13-15593556438418894092;step=;logsSeverity=ERROR;graphView=0?project=span-cloud-ck-testing-external&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))) with above container.

#### 1. When column lgmMode exists in CloudSQL and is part of PK but was dropped on Spanner 
Expected: bug fix for this flow - row should migrate to sql without error
SQL Query:

```
INSERT INTO `unified_lgm_event_data` (`ddrkey`, `userId`, `slotId`, `eventId`, `status`, `attempts`, `round`, `intAttrib0`, `intAttrib1`, `intAttrib2`, `intAttrib3`, `intAttrib4`, `intAttrib5`, `intAttrib6`, `intAttrib7`, `intAttrib8`, `intAttrib9`, `stringAttrib0`, `stringAttrib1`, `stringAttrib2`, `stringAttrib3`, `stringAttrib4`, `lastAttemptedTime`, `timesAttemptedInInterval`, `lastCompletedTime`, `createTime`, `lastUpdateTime`) VALUES (-2, -2, 0, 0, 0, 0, 0, -2, -1, -1, -1, -1, -1, -1, -1, -1, -1, '', '', '', '', '', 0, 0, 0, 0, 0);
```

#### 2. Regression Testing: When column round exists in CloudSQL and is NOT part of PK but was dropped on Spanner 
Expected: no change in code flow and row should migrate to sql without error

```
INSERT INTO `fcas_user_matches` (`ddrkey`, `userId`, `matchId`, `type`, `subType`, `subSubType`, `startTime`, `state`, `eventId`) VALUES (2, 23, 32, 32, 1, 1, 1, 1, 1);
```

#### Note on Integration test: 
- Integration test for SQL source was also updated to include this test flow, and Ignore flag was temporarily disabled to check if it passes. 
- Test run revealed pre-existing assertion errors that are fixed in this PR
- Expected result: The dataflow job created for the test passes without SQLException error.
- The test is ultimately flaky and times out repeatedly so Ignore flag was added again

### DIFF
--- a/v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceIT/cassandra-transformation-schema.sql
+++ b/v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceIT/cassandra-transformation-schema.sql
@@ -1,10 +1,9 @@
 DROP TABLE IF EXISTS customers;
 CREATE TABLE customers (
-    id int,
+    id int PRIMARY KEY,
     full_name text,
     first_name text,
     last_name text,
     empty_string text,
-    null_key text,
-    PRIMARY KEY (id, full_name)
+    null_key text
 );
--- a/v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbCustomTransformationIT/mysql-schema.sql
+++ b/v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbCustomTransformationIT/mysql-schema.sql
@@ -9,7 +9,7 @@ CREATE TABLE Users1 (
 
 CREATE TABLE AllDatatypeTransformation (
     varchar_column VARCHAR(20) NOT NULL,
-    pk_column INT NOT NULL,
+    source_only_pk INT NOT NULL,
     tinyint_column TINYINT,
     text_column TEXT,
     date_column DATE,
@@ -27,5 +27,5 @@ CREATE TABLE AllDatatypeTransformation (
     bool_column TINYINT(1),
     binary_column VARBINARY(150),
     bit_column BIT(8),
-    PRIMARY KEY (pk_column, varchar_column)
+    PRIMARY KEY (source_only_pk, varchar_column)
 );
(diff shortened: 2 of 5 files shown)


v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbCustomTransformationIT/mysql-schema.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE Users1 (
 
 CREATE TABLE AllDatatypeTransformation (
     varchar_column VARCHAR(20) NOT NULL,
+    pk_column INT NOT NULL,
```
shreyakhajanchi: nit: the column name is misleading as there are 2 pk columns

v2/spanner-to-sourcedb/src/test/resources/SpannerToCassandraSourceIT/cassandra-transformation-schema.sql
```diff
@@ -1,9 +1,10 @@
 DROP TABLE IF EXISTS customers;
 CREATE TABLE customers (
-    id int PRIMARY KEY,
+    id int,
```
shreyakhajanchi: in case of cassandra we only support like to like migration , so schema change can be removed in case of cassandra

PR #3009, opened 2025-11-20
chore: Added latency metrics to Write to source DoFn

pratickchokhani:


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -192,6 +192,7 @@ public void processElement(ProcessContext c) {
       skippedRecordCountMetric.inc();
       outputWithTag(c, Constants.SKIPPED_TAG, Constants.SKIPPED_TAG_MESSAGE, spannerRec);
     } else {
+      Stopwatch timer = Stopwatch.createStarted();
```
manitgupta, line 195: Will this create a new `Stopwatch` object per element inside the `@ProcessElement`? 
Given a new object will be created per element, will it create extra GC/memory pressure? Looks like the `StopWatch` will be discarded after the element is processed. 

It would be great if there are more instance of this being used elsewhere or we have confidence that this will not cause issues. Let me know what you think.

Alternatively, wouldn't it just be safer to use a static method like `System.currentTimeMillis();` to get the time and subtract to find the elapsed time?

pratickchokhani: This is being used in SpannerIO.writeMutations (http://shortn/_UL54yQwHLq). 
I figured that this function is more heavily used so it should be okay.

Also, the class itself is very lean containing only the following variables:
    private boolean isRunning;
    private long elapsedNanos;
    private long startTick;

manitgupta: Thanks for the reference. 
Also, please refer to the apache beam's SpannerIO in general instead of the duplicate version in DataflowTemplates, which might be stale - 

https://github.com/search?q=repo%3Aapache%2Fbeam%20Stopwatch.createStarted()%3B&type=code

PR #3016, opened 2025-11-21
Support Float as primary key for Mysql source

NicolasRicheYopp:
In the SourceDB -> Spanner flow, add support for Float type primary key for Mysql source.

- [x] _UniformSplitter_: Add a boundary extractor and splitter (split the range of value in 2) for Float value 
  - [x] splitFloat Implementation

  **Because Float is an approximate values, we need additional changes:**
  - [x] Add support for approximate value (Float, Double) in isSplittable function. Equals when diff < configurable delta Uphttps://github.com/GoogleCloudPlatform/DataflowTemplates/blob/a73e0f07a719c668396e09de5195e7a9a0b756ce/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java#L147
  - [x] The delta mentioned above will be used by default. But if the Mysql column has a specific precision - FLOAT(size, d)  - d being the number of decimals. This granularity should carry over as delta to use. We need to add a 'delta' value in **PartitionColumn** https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/a73e0f07a719c668396e09de5195e7a9a0b756ce/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java#L26
  - [x] Tests
- [x] _Source IndexType_: Add a new IndexType.Float which map internally to Java Float.class
  - [x] Implementation
  - [x] Tests
- [x] _MysqlDialectAdapter_: Map Mysql source FLOAT column type to the newly created IndexType above
  - [x] Implementation
  - [x] Tests
- [x] Add Integration tests in _v2/sourcedb-to-spanner/src/test/resources/DataTypesIT_

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
@@ -380,6 +380,7 @@ public final class MysqlDialectAdapter implements DialectAdapter {
           .put("BOOL", IndexType.NUMERIC)
           .put("YEAR", IndexType.NUMERIC)
           .put("DATE", IndexType.DATE)
+          .put("DECIMAL", IndexType.DECIMAL)
           // Float is listed as numeric types in Mysql Ref
           // https://dev.mysql.com/doc/refman/8.4/en/numeric-types.html
           // But here the end goal is to map to a Java Float.class,
@@ -433,8 +434,9 @@ public final class MysqlDialectAdapter implements DialectAdapter {
         @Nullable String collation = rs.getString(InformationSchemaStatsCols.COLLATION_COL);
         @Nullable String padSpace = getPadSpaceString(rs);
         int numericScale = rs.getInt(InformationSchemaStatsCols.NUMERIC_SCALE_COL);
+        boolean hasNumericScale = !rs.wasNull();
         logger.debug(
-            "Discovered column {} from index {}, isUnique {}, isPrimary {}, cardinality {}, ordinalPosition {}, character-set {}, collation {}, pad-space {}, numeric-scale {}",
+            "Discovered column {} from index {}, isUnique {}, isPrimary {}, cardinality {}, ordinalPosition {}, character-set {}, collation {}, pad-space {}, numericScale {}",
             colName,
             indexName,
             isUnique,
@@ -486,6 +488,7 @@ public final class MysqlDialectAdapter implements DialectAdapter {
                 .setIndexType(indexType)
                 .setCollationReference(collationReference)
                 .setStringMaxLength(stringMaxLength)
+                .setNumericScale(hasNumericScale ? numericScale : null)
                 .setDecimalStepSize(decimalStepSize)
                 .build());
       }
@@ -720,10 +723,11 @@ public final class MysqlDialectAdapter implements DialectAdapter {
     public static final String CHAR_MAX_LENGTH_COL = "cols.CHARACTER_MAXIMUM_LENGTH";
     public static final String CHARACTER_SET_COL = "cols.CHARACTER_SET_NAME";
     public static final String COLLATION_COL = "cols.COLLATION_NAME";
-    public static final String NUMERIC_SCALE_COL = "cols.NUMERIC_SCALE";
 
     // TODO(vardhanvthigle): MySql 5.7 is always PAD space and does not have PAD_ATTRIBUTE Column.
     public static final String PAD_SPACE_COL = "collations.PAD_ATTRIBUTE";
+
+    public static final String NUMERIC_SCALE_COL = "cols.NUMERIC_SCALE";
 
     public static ImmutableList<String> colList() {
       return ImmutableList.of(
@@ -736,8 +740,8 @@ public final class MysqlDialectAdapter implements DialectAdapter {
           CHAR_MAX_LENGTH_COL,
           CHARACTER_SET_COL,
           COLLATION_COL,
-          NUMERIC_SCALE_COL,
-          PAD_SPACE_COL);
+          PAD_SPACE_COL,
+          NUMERIC_SCALE_COL);
     }
 
     private InformationSchemaStatsCols() {}
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
@@ -100,6 +100,11 @@ public abstract class Boundary<T extends Serializable>
   }
 
   @Nullable
+  Integer numericScale() {
+    return partitionColumn().numericScale();
+  }
+
+  @Nullable
   BigDecimal decimalStepSize() {
     return partitionColumn().decimalStepSize();
   }
@@ -271,6 +276,11 @@ public abstract class Boundary<T extends Serializable>
 
     public Builder<T> setCollation(CollationReference value) {
       this.partitionColumnBuilder().setStringCollation(value);
+      return this;
+    }
+
+    public Builder<T> setNumericScale(Integer value) {
+      this.partitionColumnBuilder().setNumericScale(value);
       return this;
     }
 
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
@@ -23,6 +23,7 @@ import com.google.common.collect.ImmutableMap;
 import java.io.Serializable;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import java.sql.Date;
 import java.sql.Timestamp;
 import java.time.Duration;
@@ -57,7 +58,7 @@ public class BoundarySplitterFactory {
               BigDecimal.class,
               (BoundarySplitter<BigDecimal>)
                   (start, end, partitionColumn, boundaryTypeMapper, processContext) ->
-                      splitBigDecimal(start, end))
+                      splitBigDecimals(start, end, partitionColumn))
           .put(String.class, (BoundarySplitter<String>) BoundarySplitterFactory::splitStrings)
           .put(
               BYTE_ARRAY_CLASS,
@@ -174,14 +175,26 @@ public class BoundarySplitterFactory {
     return (start & end) + ((start ^ end) >> 1);
   }
 
-  private static BigDecimal splitBigDecimal(BigDecimal start, BigDecimal end) {
-    BigInteger startBigInt = (start == null) ? null : start.toBigInteger();
-    BigInteger endBigInt = (end == null) ? null : end.toBigInteger();
+  @VisibleForTesting
+  protected static BigDecimal splitBigDecimals(
+      BigDecimal start, BigDecimal end, PartitionColumn partitionColumn) {
+    Preconditions.checkNotNull(
+        partitionColumn, "Trying to split BigDecimals without partition column information.");
+    Preconditions.checkNotNull(
+        partitionColumn.numericScale(), "Trying to split BigDecimals without numeric scale.");
+    int scale = partitionColumn.numericScale();
+    BigInteger startBigInt = bigDecimalToBigInt(start, scale);
+    BigInteger endBigInt = bigDecimalToBigInt(end, scale);
+
     BigInteger split = splitBigIntegers(startBigInt, endBigInt);
     if (split == null) {
       return null;
     }
-    return new BigDecimal(split);
+    return new BigDecimal(split, scale);
+  }
+
+  private static BigInteger bigDecimalToBigInt(BigDecimal value, int scale) {
+    return value == null ? null : value.setScale(scale, RoundingMode.UNNECESSARY).unscaledValue();
   }
 
   private static Date splitDates(Date start, Date end) {
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapterTest.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapterTest.java
@@ -291,7 +291,8 @@ public class MysqlDialectAdapterTest {
             "binary",
             "year",
             "bool",
-            "date");
+            "date",
+            "decimal");
     ImmutableList<SourceColumnIndexInfo> expectedSourceColumnIndexInfos =
         getExpectedSourceColumnIndexInfosForBasicIndexes(false);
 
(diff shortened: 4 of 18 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -263,4 +268,29 @@ protected static Timestamp instantToTimestamp(Instant instant) {
   private static Timestamp splitTimestamps(Timestamp start, Timestamp end) {
     return instantToTimestamp(splitInstants(timeStampToInstant(start), timeStampToInstant(end)));
   }
+
+  private static Float splitFloats(Float start, Float end) {
+      if (start == null && end == null) {
+          return null;
```
NicolasRicheYopp: BoundarySplitter interface comments does not specify what should be the behavior when both start and end parameters are null :

>  *  Implementations must not assume that that start is less than end. It depends on the
>        specific ordering used by the database schema.
>  * Implementations must be overflow safe.
>   * Implementations must guarantee that the splitter is idempotent. The same boundary must get
>       same split point through the lifetime of the migration.
>   * Implementations must guarantee that the database would teat the splitpoint as being
>       in-between the start and end as per the ordering and collation used for the partition
>       column.

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ef38a8092bd3fcd804656744a7894cb5af87909b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitter.java#L40

For consistency, I then followed the behavior used by other split functions: 
When both start and end params are null the split functions return null.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -375,6 +375,10 @@ private ImmutableMap<String, SourceColumnType> getTableCols(
           .put("TINYTEXT", IndexType.STRING)
           .put("DATETIME", IndexType.TIME_STAMP)
           .put("TIMESTAMP", IndexType.TIME_STAMP)
+          // Float is listed as numeric types in Mysql Ref https://dev.mysql.com/doc/refman/8.4/en/numeric-types.html
+          // But here the end goal is to map to a Java Float.class,
+          // we need a distinct Source IndexType to map to Float.class
+          .put("FLOAT", IndexType.FLOAT)
```
NicolasRicheYopp, line 388: Note: Only works for FLOAT(p <= 24).
For FLOAT(p > 24) it should be mapped to Double to not lose precision during the conversion.

WIP: To confirm, Mysql or/and JDBC may return the column type automatically as DOUBLE if p> 24

**EDIT: (p) only matters a table creation and does not exist in column once it created. More details in the comment below.**

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
VardhanThigle: That's right.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
NicolasRicheYopp: From Mysql Docs 

> MySQL also supports this optional precision specification, but the precision value in [FLOAT(p)](https://dev.mysql.com/doc/refman/8.4/en/floating-point-types.html) is used only to determine storage size. A precision from 0 to 23 results in a 4-byte single-precision FLOAT column. A precision from 24 to 53 results in an 8-byte double-precision DOUBLE column.
> https://dev.mysql.com/doc/refman/8.4/en/floating-point-types.html

Float(p), p only matters at table creation time. Mysql then assign a column of type FLOAT or DOUBLE depending of size of p (=<24 or >24).

So once the table is created, p does not exists: The column is either a FLOAT or a DOUBLE. The code above is correct.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -263,4 +268,29 @@ protected static Timestamp instantToTimestamp(Instant instant) {
...
+      }
+      if (end == null) {
+          end = Float.MAX_VALUE;
+      }
+
+      // Calculate overflow safe mid-point
+
+      // If signs are different, simple addition is safe from overflow
+      // because the values cancel each other out towards zero.
+      if ((start < 0 && end > 0) || (start > 0 && end < 0)) {
```
VardhanThigle: For float and double we also need to enhance[ isSplittable](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/a73e0f07a719c668396e09de5195e7a9a0b756ce/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java#L147) to stop the splitting at appropriate precision.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -104,14 +105,45 @@ public static <T extends Serializable> Builder<T> builder() {
...
+
   /**
    * Checks if two boundaries can be merged with each other.
    *
    * @param other other boundary.
    * @return true if ranges are mergable.
    */
   public boolean isMergable(Boundary<?> other) {
-    return Objects.equal(this.end(), other.start()) || Objects.equal(this.start(), other.end());
+    return areValuesEqual(this.end(), other.start()) || areValuesEqual(this.start(), other.end());
```
NicolasRicheYopp, line 145: @VardhanThigle I assumed isMergeable need to updated as well (same logic than for isSplittable, Objects.equal does not work for float point types)

I will update unit tests for that if confirmed that isMergeable need to updated as isSplittable.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -463,6 +477,7 @@ private ImmutableList<SourceColumnIndexInfo> getTableIndexes(
                 .setIndexType(indexType)
                 .setCollationReference(collationReference)
                 .setStringMaxLength(stringMaxLength)
+                .setDecimalStepSize(decimalStepSize)
```
NicolasRicheYopp, line 492: Data flow for carrying a Mysql decimal precision of a column. 
Example: FLOAT(p, d)

The decimalStep is carried as a BigDecimal. The same property will be used for DOUBLE(p, d) AND DECIMAL(size, d).

- The value is injected in SourceColumnIndexInfo.
- Then SourceColumnIndexInfo carry over the value into PartitionColumn
- The Boundary receives the ParitionColumn

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapterTest.java
```diff
@@ -343,16 +358,16 @@ public void testDiscoverIndexesBasic() throws SQLException, RetriableSchemaDisco
...
-    SourceColumnIndexInfo.builder()
-        .setColumnName("testColYear")
-        .setIndexName("primary")
-        .setIsUnique(true)
-        .setIsPrimary(true)
-        .setCardinality(100L)
-        .setIndexType(IndexType.NUMERIC)
-        .setOrdinalPosition(6)
-        .build();
+                .build(),
```
NicolasRicheYopp: Year was not include in the expected result list ( ; vs expected , ).
Not related to this PR but fixing it here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
NicolasRicheYopp: Done

NicolasRicheYopp:
@VardhanThigle I made the changes we discussed during the meeting:
- Move default decimalStep from Boundary to MysqlDialectAdapter
- Add tests of Boundary.isMergable with float values (check that it works with the provided granularity of DecimalStepSize parameter)

I think the PR is now ready but of course let me know if you see anything missing.

Thank you

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -97,21 +98,44 @@ Integer stringMaxLength() {
...
   }
 
+  /**
+   * Custom equality check that relies on Object.equals() for most types. But switches to
+   * delta/min-step based comparison for: - Floating-point types (Float, Double) - Source DB type
+   * where there is a defined precision Examples: Float(p, d) TODO: Double(p, d).
+   */
+  public boolean areValuesEqual(Object valueA, Object valueB) {
+    if (valueA instanceof Float f1 && valueB instanceof Float f2) {
+      BigDecimal b1 = new BigDecimal(f1.toString());
```
VardhanThigle, line 127: Would it be possible to use `Float` or `Double` here?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactoryTest.java
```diff
@@ -321,6 +321,56 @@ public void testTimeStampSplitting() {
...
+    // -10 to -20 -> -15 (Same as -20 to -10)
+    assertThat(splitter.getSplitPoint(-10.0f, -20.0f, null, null, null)).isEqualTo(-15.0f);
+    // 100 to -100 -> 0 (Same as -100 to 100)
+    assertThat(splitter.getSplitPoint(100.0f, -100.0f, null, null, null)).isEqualTo(0.0f);
+    // 100 to -50 -> 25 (Same as -50 to 100)
+    assertThat(splitter.getSplitPoint(100.0f, -50.0f, null, null, null)).isEqualTo(25.0f);
+
+    // 3. Null Handling
+    // Both Null the splitter does not process and return null
+    assertThat(splitter.getSplitPoint(null, null, null, null, null)).isEqualTo(null);
```
VardhanThigle, line 418: Thanks for the detailed UT
Can we also add a few cases around FLOAT.MIN (either as step size or as range boundary or both)?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -97,21 +98,44 @@ Integer stringMaxLength() {
...
     return (new AutoValue_Boundary.Builder<T>()).setSplitIndex("1").setBoundaryTypeMapper(null);
   }
 
+  /**
+   * Custom equality check that relies on Object.equals() for most types. But switches to
+   * delta/min-step based comparison for: - Floating-point types (Float, Double) - Source DB type
+   * where there is a defined precision Examples: Float(p, d) TODO: Double(p, d).
+   */
+  public boolean areValuesEqual(Object valueA, Object valueB) {
+    if (valueA instanceof Float f1 && valueB instanceof Float f2) {
```
VardhanThigle, line 126: If this is not used outside this package, can we make it protected (and possibly `@visibleForTest`)?

Please cover cases where both are not type float or one is of type flot and other is not for a complete coverage of the lines of code here. Currently the coverage for this if is partial.

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
```diff
@@ -459,8 +464,8 @@ SET time_zone = 'Asia/Kolkata';
 INSERT INTO `timestamp_pk_table` (`id`, `timestamp_pk_col`) VALUES ('2005-01-01 05:31:54.123456', '2005-01-01 05:31:54.123456');
 SET time_zone = SYSTEM;
 INSERT INTO `year_pk_table` (`id`, `year_pk_col`) VALUES (1901, 1901), (2000, 2000);
-
 INSERT INTO `bool_pk_table` (`id`, `bool_pk_col`) VALUES (TRUE, TRUE), (FALSE, FALSE);
+INSERT INTO `float_pk_table` (`id`, `float_pk_col`) VALUES (45.56, 45.56), (3.4E38, 3.4E38), (-3.4E38, -3.4E38);
```
VardhanThigle: Just wanted to be double sure given the complexity of Float type:

Would it be possible to add a few values close to the precision diff and a few close to boundary (min, max) here?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
nmemond: Yep, will update.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactoryTest.java
nmemond: Sure, will do.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
nmemond: Makes sense, will do.

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-data-types.sql
nmemond: Yeah, that's a good idea I think, I'll update this.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
nmemond: Actually, after updating this I think it's probably best to stick with BigDecimal to avoid cases where the difference between `f1` and `f2` can't be exactly represented by a float or double. For example, `1.03f - 1.02f` returns `0.00999999f` instead of exactly `0.01f`, and so `1.03f` and `1.02f` would return as equal if we used `Float` here (assuming step size of 0.01).

VardhanThigle: Makes sense. Thanks.

PR #3017, opened 2025-11-24
Implementing DataflowMetricCounterCheck Condition check

darshan-sj:


darshan-sj:
This change is not required now. We have implemented the test using DLQ event counter which directly reads the DLQ entires from GCS bucket and returns the count which is more reliable than the metric counters.

PR #3018, opened 2025-11-24
[DatastreamToSpanner] Integration test for CDC processing algorithm for Live migration template

darshan-sj:


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,227 @@
...
+    int numRows = 100;
+    int burstIterations = 10000;
+
+    // generate Load
+    CDCCorrectnessTestUtil testUtil = new CDCCorrectnessTestUtil();
+    testUtil.generateLoad(numRows, burstIterations, sourceDBResourceManager);
+
+    FlexTemplateDataflowJobResourceManager.Builder flexTemplateBuilder =
+        FlexTemplateDataflowJobResourceManager.builder(testName)
+            .addEnvironmentVariable("numWorkers", 10)
```
bharadwaj-aditya: It would be better to use constants in these scenarios.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,227 @@
...
+
+    // Wait for at least 1 row to appear in Spanner
+    ConditionCheck conditionCheck =
+        SpannerRowsCheck.builder(spannerResourceManager, USERS_TABLE).setMinRows(1).build();
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(20)), conditionCheck);
+    assertThatResult(result).meetsConditions();
+
+    DataflowFailureInjector.abruptlyKillWorkers(jobInfo.projectId(), jobInfo.jobId());
```
bharadwaj-aditya: this should happen a few times during the duration of the test. Probably after the job starts. Considering the job will start after about 3 minutes, this should be called after a short while. It would be better if this checks running status before killing the workers

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,227 @@
...
+
+    // Wait for Forward migration job to be in running state
+    assertThatPipeline(jobInfo).isRunning();
+
+    // Wait for at least 1 row to appear in Spanner
+    ConditionCheck conditionCheck =
+        SpannerRowsCheck.builder(spannerResourceManager, USERS_TABLE).setMinRows(1).build();
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(20)), conditionCheck);
```
bharadwaj-aditya, line 196: This result is overwritten below. What is the purpose of the wait here ?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,227 @@
...
+    Thread.sleep(600000);
+
+    long expectedRows = sourceDBResourceManager.getRowCount(USERS_TABLE);
+    // Wait for exact number of rows as source to appear in Spanner
+    conditionCheck =
+        SpannerRowsCheck.builder(spannerResourceManager, USERS_TABLE)
+            .setMinRows((int) expectedRows)
+            .setMaxRows((int) expectedRows)
+            .build();
+    result =
```
bharadwaj-aditya, line 234: The wait here will be invoked after the sleep of 600 seconds. Essentially this would mean the test would take over 30 minutes. Is that the expectation here ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
```diff
@@ -0,0 +1,400 @@
...
+                  } catch (SQLException e) {
+                    LOG.error("Thread processing failed", e);
+                    throw new RuntimeException(e);
+                  }
+                }));
+      }
+
+      // Wait for all threads to finish
+      for (Future<?> future : futures) {
+        future.get(2, TimeUnit.HOURS); // Wait for each task to complete
```
bharadwaj-aditya, line 98: This is a very long duration. This would be in addition to the time taken to complete processing and evaluate the results. Over 2.5 hours per test.

Is there a sample run where we know how long this will take ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+          // Insert the row
+          user.mutateRandomly();
+          insertUser(conn, user);
+          isRowPresent = true;
+          counter.incrementAndGet();
+        }
+        // Else: Do nothing (row remains absent)
+      }
+    }
+    System.out.print(".");
```
bharadwaj-aditya: remove the system out. Use loggers

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+        DriverManager.getConnection(
+            resourceManager.getUri(),
+            resourceManager.getUsername(),
+            resourceManager.getPassword())) {
+      for (int k = 0; k < numRows; k++) {
+        // 1. Create Unique Random ID
+        int id;
+        User user;
+        do {
+          id = getRandom().nextInt(1000000);
```
bharadwaj-aditya: The method above is generating the ids. Those should be used. Why are different ids being generated at 2 places ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+    int id;
+    String firstName;
+    String lastName;
+    int age;
+    boolean status;
+    long col1;
+    long col2;
+
+    static final List<String> UPDATABLE_COLUMNS =
+        ImmutableList.of("first_name", "last_name", "age", "status", "col1", "col2");
```
bharadwaj-aditya: using constants for instead of these names is a good practice

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,227 @@
...
+        DatastreamResourceManager.builder(testName, PROJECT, REGION)
+            .setCredentialsProvider(credentialsProvider);
+    if (System.getProperty("privateConnectivity") != null) {
+      datastreamResourceManagerBuilder.setPrivateConnectivity(
+          System.getProperty("privateConnectivity"));
+    }
+    datastreamResourceManager = datastreamResourceManagerBuilder.build();
+    Stream stream =
+        createDataStreamResources(
+            artifactBucket, gcsPrefix, sourceConnectionProfile, datastreamResourceManager);
```
rohitwali, line 142: should this also go in the set up?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
rohitwali: is this relevant for this test?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import org.apache.beam.it.gcp.cloudsql.CloudSqlResourceManager;
+import org.apache.beam.it.gcp.spanner.SpannerResourceManager;
+import org.apache.beam.it.jdbc.AbstractJDBCResourceManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CDCCorrectnessTestUtil {
```
rohitwali: s/CDCCorrectnessTestUtil/FuzzyCDCLoadGenerator/? or something similar?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
```diff
@@ -0,0 +1,400 @@
...
+    ExecutorService executor = Executors.newFixedThreadPool(totalRows);
+    try {
+      List<Future<?>> futures = new ArrayList<>();
+
+      for (int i = 0; i < totalRows; i++) {
+        futures.add(
+            executor.submit(
+                () -> {
+                  // Each thread gets its own connection to ensure thread safety
+                  try (Connection conn = DriverManager.getConnection(jdbcUrl, username, password)) {
```
rohitwali, line 87: would it make sense to use a connection pooling library instead?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
```diff
@@ -0,0 +1,400 @@
...
+        resourceManager.getPassword());
+  }
+
+  /** Process a single row: Insert -> Burst Operation. */
+  private void processSingleRow(Connection conn, int burstIterations) throws SQLException {
+    // 1. Create Unique Random ID
+    int id;
+    User user;
+    do {
+      id = getRandom().nextInt(1000000);
```
rohitwali: what is 1 million here?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Line 198 to 203 is the dataflow IT framework way for waiting for a condition. The condition we are waiting for here is 1 row to appear in spanner (defined at line 198). 

Once this condition is met we move on in the test case. The next time we wait at line 219 is on a different condition the variables could be reused.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+    // 2. Insert initial row
+    insertUser(conn, user);
+    counter.incrementAndGet();
+    boolean isRowPresent = true;
+
+    for (int i = 0; i < burstIterations; ++i) {
+      // If row is present, then update the row with 75% probability and delete the row with 25%
+      // probability
+      if (isRowPresent) {
+        if (getRandom().nextDouble() < 0.75) {
```
rohitwali: 1/ didnt we want to do insert/update/deletes with equal probability?
2/ given its a util, shouldnt this be a configuration that is passed to it?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
rohitwali: should we use a logger instead?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Done. Modified the code to kill workers 4 times at the interval of 20 seconds each. This actually happens after the job starts. The conditional wait at Line 198 to 203 is for waiting till 1 row appears in spanner indicating the pipeline has fully started.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: This is wait till numberOfRows in spanner is = expectedRows. This could return immediately if the Spanner already contains expectedRows (If all the processing completed within 10 minutes of wait above). 20 minutes here is the timeout for the condition wait. If spanner doesn't contain expected number of rows within 20 minutes, then this would fail.

There are two 10 minutes sleep in this test case at Line 215 and Line 231. The dataflow is taking around 10 minutes to process all the 1 million records. the second 10 minutes sleep below is buffer. 

Overall the test is taking around 1 hour 5 minutes to run.
35 minutes to generate load on MySQL
20 minutes for the dataflow job
remaining is for setting up all the resources. Like creating MySQL and spanner databases, datastream, pubsub and GCS resources etc.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: This is taking 35 minutes when I execute locally. I have put 2 hours timeout here conservatively.

Overall the test is taking around 1 hour 5 minutes to run.
35 minutes to generate load on MySQL
20 minutes for the dataflow job
remaining is for setting up all the resources. Like creating MySQL and spanner databases, datastream, pubsub and GCS resources etc.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: This method is a residual method. I have removed this.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
```diff
@@ -0,0 +1,400 @@
...
+      ps.executeUpdate();
+    }
+  }
+
+  void updateUser(Connection conn, User u) throws SQLException {
+    // Randomly select a column to update.
+    String columnToUpdate =
+        User.UPDATABLE_COLUMNS.get(getRandom().nextInt(User.UPDATABLE_COLUMNS.size()));
+
+    String sql = "UPDATE `Users` SET " + columnToUpdate + " = ? WHERE id = ?";
```
rohitwali: would it make sense to decouple the Users entity from this class?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Done. Moved to setup.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Yes it is relevant. This is equivalent to downscaling and upscaling dataflow workers. And it induces work item assignment rebalancing.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: generating a random Id between [0, 1000000) .

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: 1. If we just do Insert or update or delete randomly, Many of the inserts would be no op if the row already exists and many of the deletes would be no op if the row is already deleted, but it would still consume CPU cycles, network bandwidth etc. It doesn't make sense to have such load. insert only when the row doesn't exist, delete or update only when the row exists.
2. This whole class is written very specifically for the CDC algorithm correctness IT. I don't see it to be used for other test cases and hence chose to not make it generic and config driven. For example, this class doesn't support multiple tables, ordering of events between the tables etc. Making it generic is over engineering and doesn't make sense to me. If there is a use case in future where we want to re-use this class, then we can enhance it at that time.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: A connection pool library is useful when the connection utilization is very dynamic. In our case, we get 100 (num of rows) connections at the beginning and they remain in use till the end of load generation. I used the HikariCP and checked, it is actually taking much more time and degrading the performance.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
rohitwali: so our template code will perform worse than the test? is that worth poking?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
rohitwali: i guess i followed that part. what I am trying to understand is
1/ why not 1 billion? how did we pick a million?
2/ could this be range be different for different tests?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
rohitwali: 1. it makes sense to handle insert in insert and delete on delete, which IIUC we are already handling. but still unclear to me why that would mean equal probably can be used. is there a doc that walks through how the current distribution is picked? 
2. I was assuming this class to provide load generation capabilities for CDC tests. I am ok with just having support for one table, but even for that it makes sense to support different workload distributions (Open/Closed Principle). are there any other aspects of this class that are CDC correctness specific that I may be missing?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -171,7 +153,7 @@ public void liveMigrationCrossDbTxnCdcTest()
     int burstIterations = 10000;
 
     // generate Load
-    CDCCorrectnessTestUtil testUtil = new CDCCorrectnessTestUtil();
+    testUtil = new FuzzyCDCLoadGenerator();
```
rohitwali: nit: may want to use a better name

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCCorrectnessTestUtilTest.java
```diff
@@ -0,0 +1,325 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** Unit tests for {@link FuzzyCDCLoadGenerator}. */
+@RunWith(JUnit4.class)
+public class FuzzyCDCCorrectnessTestUtilTest {
```
rohitwali: Should this reflect the class under Test?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import org.apache.beam.it.gcp.cloudsql.CloudSqlResourceManager;
+import org.apache.beam.it.gcp.spanner.SpannerResourceManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FuzzyCDCLoadGenerator {
```
rohitwali, line 45: should this be under the main package or test package?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+      } else {
+        // If row is not present, then insert the row with 75% probability and do nothing with 25%
+        // probability
+        if (getRandom().nextDouble() < 0.75) {
+          // Insert the row
+          user.mutateRandomly();
+          user.insert(conn);
+          isRowPresent = true;
+          counter.incrementAndGet();
```
rohitwali: where is this used?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Done.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: This is in spanner-common module and is imported in datastream-to-spanner and spanner-to-sourcedb modules. It is correctly placed in main package since we cannot import from the test package.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: For the template, we need to control the maximum number of connections to the sourcedb. We cannot do that with directly creating a connection. Not using a connection pool is not an option in our template.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: 1. The id range doesn't matter for mySQL but for Spanner it helps keep the data distributed into different splits. I thought 100 ids in the range of 0- 1 million is a sufficient distribution for optimum performance from Spanner. I updated it to max value of INT which is ~ 2 billion.
2. Currently it can be same for both the Live and reverse CDC tests. It is not locked in. We can parameterise it when the need arrises.

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCCorrectnessTestUtilTest.java
darshan-sj: Fixed! A miss during renaming. Thanks for catching!

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/FuzzyCDCLoadGenerator.java
darshan-sj: I was using this to count the number of events which happened and I was asserting the metric counter of the dataflow to match this number. I removed the assertion after realising that the counter values can be inaccurate. This is still useful for debugging purposes. I have logged the counter value at the end of data generation. While debugging any issues, we can compare this value with the metric counters in the dataflow job.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/testutils/failureinjectiontesting/CDCCorrectnessTestUtil.java
darshan-sj: Updated the algorithm and parameterised the updateProbability. PTAL.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,239 @@
...
+    ConditionCheck conditionCheck =
+        SpannerRowsCheck.builder(spannerResourceManager, USERS_TABLE).setMinRows(1).build();
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(createConfig(jobInfo, Duration.ofMinutes(20)), conditionCheck);
+    assertThatResult(result).meetsConditions();
+
+    // Kill the dataflow workers multiple times to induce work item assignment rebalancing and
+    // inturn increase the chance of same key being processed by multiple workers parallelly.
+    DataflowFailureInjector.abruptlyKillWorkers(jobInfo.projectId(), jobInfo.jobId());
```
bharadwaj-aditya: it would make sense to keep killing workers every 20-30 seconds until the end of the job. Any reason to stop after 3 iterations ?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
```diff
@@ -0,0 +1,239 @@
...
+@TemplateIntegrationTest(DataStreamToSpanner.class)
+@RunWith(JUnit4.class)
+public class DataStreamToSpannerCDCFT extends DataStreamToSpannerFTBase {
+
+  private static final String SPANNER_DDL_RESOURCE = "DataStreamToSpannerCDCFT/spanner-schema.sql";
+  public static final String USERS_TABLE = "Users";
+  public static final HashMap<String, String> USERS_TABLE_COLUMNS =
+      new HashMap<>() {
+        {
+          put("id", "INT NOT NULL");
```
bharadwaj-aditya: better to use the constants in the user class here and in all other references

darshan-sj: Done.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/failureinjectiontesting/DataStreamToSpannerCDCFT.java
darshan-sj: Done.

PR #3019, opened 2025-11-24
Add support for Postgres to Iceberg Blueprint

tarun-google:
1. Supports Postgres to Iceberg Blueprint
2. Added ability to import options, so this can be re-used for multiple pipelines
3. Added ability to remove unused fields. 
4. Added Templates and Tests for the Blueprint
5. Added test coverage for main.py 

Next:
1. Add Template and Tests[Done]
2. Add KMS encryption as pipeline level feature for both blueprints and templates.
3. main.py should have test coverage[Done]

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/GcsResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/GcsResourceManager.java
@@ -295,9 +295,16 @@ public final class GcsResourceManager implements ArtifactClient, ResourceManager
 
   @Override
   public synchronized void cleanupAll() {
-    // If bucket was created by manager, simply delete the bucket without worrying about other
+    // If bucket was created by manager, simply delete the bucket without worrying
+    // about other
     // cleanups.
     if (hasNonStaticBucket) {
+      // List and delete all objects in the bucket
+      Page<Blob> blobs = client.list(bucket);
+      for (Blob blob : blobs.iterateAll()) {
+        blob.delete();
+      }
+
       boolean deleted = client.delete(bucket);
       if (deleted) {
         LOG.debug("Bucket '{}' was deleted", bucket);
--- a/it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/storage/GcsResourceManagerTest.java
+++ b/it/google-cloud-platform/src/test/java/org/apache/beam/it/gcp/storage/GcsResourceManagerTest.java
@@ -434,13 +434,15 @@ public final class GcsResourceManagerTest {
     GcsResourceManager client2 = new GcsResourceManager(client, BUCKET, TEST_CLASS, true);
 
     when(client.delete(BUCKET)).thenReturn(true);
+    TestBlobPage allPages = createPages(ImmutableList.of());
+    when(client.list(BUCKET)).thenReturn(allPages);
 
     // Act
     client2.cleanupAll();
 
     // Assert
     verify(client).delete(BUCKET);
-    verify(client, never()).list(anyString(), any(BlobListOption.class));
+    verify(client).list(BUCKET);
     verify(client, never()).delete(anyIterable());
   }
 
@@ -450,13 +452,15 @@ public final class GcsResourceManagerTest {
     GcsResourceManager client2 = new GcsResourceManager(client, BUCKET, TEST_CLASS, true);
 
     when(client.delete(BUCKET)).thenReturn(false);
+    TestBlobPage allPages = createPages(ImmutableList.of());
+    when(client.list(BUCKET)).thenReturn(allPages);
 
     // Act
     client2.cleanupAll();
 
     // Assert
     verify(client).delete(BUCKET);
-    verify(client, never()).list(anyString(), any(BlobListOption.class));
+    verify(client).list(BUCKET);
     verify(client, never()).delete(anyIterable());
   }
 


tarun-google:
Waiting for approval from one of the TL

yaml/src/main/yaml/PostgresToIceberg.yaml
```diff
@@ -0,0 +1,55 @@
+template:
+  name: "Postgres_To_Iceberg_Yaml"
+  display_name: "Postgres to Iceberg (Yaml)"
+  description: "A template for Postgres to Iceberg."
+  flex_container_name: "postgres-to-iceberg-yaml"
+  category: "BATCH"
+
+  options_file:
+    - "jdbc_options"
```
chamikaramj: Given that we have separate schema-transforms for Databases (MySQL, Postgre etc.) should we also have separate options files ? This will also better translate to managed replacement when we do it.

yaml/src/main/python/options/iceberg_options.yaml
```diff
@@ -0,0 +1,79 @@
...
+        type: text
+      - order: 2
+        name: "catalog_name"
+        description: "Name of the catalog containing the table."
+        help: "The name of the Iceberg catalog that contains the table."
+        example: "my_hadoop_catalog"
+        required: true
+        type: text
+      - order: 3
+        name: "catalog_properties"
```
chamikaramj: Is it possible to give users some pre-defined values when selecting the options for the blueprint ? For example, catalog type is one of several supported values so it's better if users can just select than typing that in. Same for other similar options we can just list.

yaml/src/main/python/main.py
```diff
@@ -43,6 +64,54 @@ def _get_pipeline_yaml():
         pipeline_yaml = fin.read().decode()
     return pipeline_yaml
 
+def _get_pipeline_with_options(yaml_pipeline):
```
chamikaramj, line 67: Let's make sure that this logic is unit tested.

yaml/src/main/python/options/iceberg_options.yaml
tarun-google: I am not sure. 

Firstly, I believe we are using same yaml to support Blueprints and Templates. We do not want to be opinionated for Templates, that restrict users from some paths. Also, catalog properties is a map that changes significantly based on type, i am not sure if giving a few default ones changes the things, if we do, we have to accept arbitrary individual key-value pairs(again not sure for Templates, might getaway with Blueprints). Best thing we can do is to point users guides on expected inputs by Iceberg documentation and Dataflow example 

If we really want to reduce effort of users on this end the way is create something like PostgresToIcebergRestCatalog, PostgresToIcebergHadoopCatalog, PostgresToIcebergHiveCatalog etc..,

chamikaramj: Sg, we can track this as a feature for the future.

chamikaramj (dismissed):
Thanks. LGTM.

tarun-google:
@chamikaramj please re-review the PR. had to bring multiple changes to get this working

1. There is a bug on Iceberg Hadoop catalog with GCS with few java versions https://github.com/apache/iceberg/issues/12232
2. So, moved to Iceberg Rest Catalog. Upgraded the Iceberg version from 1.6 -> 1.10 for new AuthManager. 
3. Added some changes to GcsResourceManager, as its not cleaning the bucket properly.

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/GcsResourceManager.java
```diff
@@ -295,9 +295,16 @@ public void registerTempDir(String dirName) {
 
   @Override
   public synchronized void cleanupAll() {
-    // If bucket was created by manager, simply delete the bucket without worrying about other
+    // If bucket was created by manager, simply delete the bucket without worrying
+    // about other
     // cleanups.
     if (hasNonStaticBucket) {
+      // List and delete all objects in the bucket
+      Page<Blob> blobs = client.list(bucket);
```
chamikaramj, line 303: Is this required before deleting the bucket ? If so how did it work before ? Note that listing and deleting all objects can be very time consuming.

yaml/src/main/yaml/PostgresToIceberg.yaml
```diff
@@ -0,0 +1,55 @@
...
+        partition_column: {{ partitionColumn }}
+        table: {{ location }}
+        num_partitions: {{ numPartitions }}
+        fetch_size: {{ fetchSize }}
+        connection_init_sql: {{ connectionInitSql }}
+        disable_auto_commit: {{ disableAutoCommit }}
+        output_parallelization: {{ outputParallelization }}
+        type: {{ jdbcType | default("postgres") }}
+
+    - type: WriteToIceberg
```
chamikaramj, line 40: Just to confirm, can this blueprint be executed end to end for direct extract-load type use-cases without a transform in the middle (to a new Iceberg table) ?

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/storage/GcsResourceManager.java
tarun-google: I believe usecases are going though else part with have the list and delete already. In my use case when we directly drop the non empty bucket i get exception.

yaml/src/main/yaml/PostgresToIceberg.yaml
tarun-google: yes

PR #3024, opened 2025-11-25
[spanner-to-sourcedb] DELETEs: Error fetching shard id column

aasthabharill:
[b/458068271](https://b.corp.google.com/issues/458068271)
[b/458070941](https://b.corp.google.com/issues/458070941)

## Problem: 
Whenever a row has column with NULL value, the reverse replication fails with the following error messages:
1. AssignShardId step in dataflow: ```Error fetching shard Id column: Illegal call to getter of null value```
2. DLQ entry: ```"error_message":"No shard identified for the record"```

## Fix: 
marshalSpannerValues calls getter functions to get value of the column, but these function throw a NullPointerException when the value is NULL. This is handled incorrectly and causes the above error. So the fix is to catch a NULL value at the beginning of the function itself.

## Tests:
1. Unit test updated - failed without the fix with the expected error and passed with it.
2. Integration test updated to include NULL values in the row.

### Dataflow job with container built on fixed code.
Template Container built in gs://ea-functional-tests/templates/flex/Spanner_to_SourceDb

Ran [dataflow job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2025-11-26_02_33_03-2924712538454216688;graphView=0?project=span-cloud-ck-testing-external&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))) with above container.

#### 1. double datatype 
SQL Query: column skill of type double is kept NULL, column variance of type double is non-NULL

```
INSERT INTO ut_scl_squad (ddrKey, gameSpaceId, ownerPersId, teammateIndex, teammatePersId, `variance`, lastUpdateTime) VALUES (3002, 17613, 172377253, 202, 1269841562, 1234.56, 2720290009);
DELETE FROM ut_scl_squad WHERE ddrKey = 3002 AND gameSpaceId = 17613 AND ownerPersId = 172377253;
```

#### 2. timestamp datatype
SQL Query: 
- column createdTime of type timestamp is kept non-NULL, column squadName of type varchar is NULL
```
INSERT INTO ut_showoff (ddrkey, showoffId, userId, createdTime, `count`) VALUES (2002, 2002, -285444577, TIMESTAMP('2025-01-01T12:00:00Z'), 2444035295);
DELETE FROM ut_showoff WHERE ddrkey = 2002 AND showoffId = 2002;
```

- column createdTime of type timestamp is kept NULL
```
INSERT INTO ut_showoff (ddrkey, showoffId, userId) VALUES (202, 202, -28577);
DELETE FROM ut_showoff WHERE ddrkey = 202 AND showoffId = 202;
```

#### 3. Reserved Keywords
SQL Query: column content of type varbinary is kept NULL
```
INSERT INTO sedges (ddrkey, created, `from`, `to`, `type`, `value`, flags) VALUES (2001, 2001, 'nodeA', 'nodeB', 'edgeType1', 'edgeValue1', -1326272220);
DELETE FROM sedges WHERE ddrkey = 2001 AND created = 2001;
```

#### 4. INTERLEAVING
- INTERLEAVE IN PARENT ON DELETE CASCADE: on deleting parent on spanner, both parent and child records from sql were also deleted.
- INTERLEAVE IN: on deleting parent on spanner, the parent row was also deleted from sql.

#### 5. Regression testing: Custom sharding logic
- Flow when theres no migration_shard_id column and pipeline uses ddrkey value to compute shard is not affected by this change.

### DIFF
--- a/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbDatatypeIT.java
+++ b/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbDatatypeIT.java
@@ -260,6 +260,24 @@ public class SpannerToSourceDbDatatypeIT extends SpannerToSourceDbITBase {
             .to(Value.bytes(ByteArray.copyFrom("varbinary")))
             .set("bit_column")
             .to(Value.bytes(ByteArray.copyFrom("a")))
+            .set("null_string_column")
+            .to((String) null)
+            .set("null_int_column")
+            .to((Long) null)
+            .set("null_date_column")
+            .to((Date) null)
+            .set("null_float_64_column")
+            .to((Double) null)
+            .set("null_float_32_column")
+            .to((Float) null)
+            .set("null_numeric_column")
+            .to((BigDecimal) null)
+            .set("null_timestamp_column")
+            .to((Timestamp) null)
+            .set("null_blob_column")
+            .to((ByteArray) null)
+            .set("null_bool_column")
+            .to((Boolean) null)
             .build();
     spannerResourceManager.write(m);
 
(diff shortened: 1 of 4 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/AssignShardIdFn.java
```diff
@@ -337,6 +337,11 @@ private Map<String, Object> fetchSpannerRecord(
    */
   private void marshalSpannerValues(
       ObjectNode newValuesJsonNode, String tableName, String colName, Struct row) {
+    if (row.isNull(colName)) {
```
bharadwaj-aditya, line 340: Is this handled in scenarios where the custom transform returns the shard id field as well ?

VardhanThigle (approved):
Good catch! Thanks for the fix.

PR #3035, opened 2025-12-01
[datastream-to-spanner] Unable to convert field timestamp to long

aasthabharill:
[b/458064173](https://b.corp.google.com/issues/458064173)

#### Problem
We create some new columns to hold metadata in the shadow tables that are created. These shadow tables also have the Primary Key columns from the data table. When the data table has columns of the same name as the metadata columns we create, it gets overwritten because of which we get data type conversion errors. For eg. a primary key column "timestamp TIMESTAMP" gets overwritten to "timestamp INT64".

#### Fix
To ensure that we don't have these kind of clashes, we now dynamically compute the name of the shadow column by checking it against the list of primary key column names. In case a column with the same name is found, we add "shadow_" as a prefix iteratively till we get a unique column name for the shadow table. 

### Testing:
- Built container with changes: gs://ea-functional-tests/templates-aastha-2025-12-09/flex/Cloud_Datastream_to_Spanner
-  Confirmed in DDL of Shadow table that original "timestamp" column of type TIMESTAMP is present along with new columns: shadow_timestamp, log_file and log_position
- Ran a live migration job with above template and ran INSERT, UPDATE and DELETE statements containing "timestamp TIMESTAMP" column as Primary Key in data table - and these passed without errors

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
@@ -78,6 +78,10 @@ public abstract class ChangeEventContext {
     this.changeEvent = changeEvent;
     this.dataTable = changeEvent.get(DatastreamConstants.EVENT_TABLE_NAME_KEY).asText();
     Table table = ddl.table(this.dataTable);
+    if (table == null) {
+      throw new DroppedTableException(
+          "Table from change event does not exist in Spanner. table=" + this.dataTable);
+    }
     Set<String> existingPrimaryKeyColumnNames =
         table.primaryKeys().stream().map(k -> k.name()).collect(Collectors.toSet());
 


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/DatastreamConstants.java
```diff
@@ -74,10 +74,10 @@ public class DatastreamConstants {
   public static final String ORACLE_TIMESTAMP_KEY = "_metadata_timestamp";
 
   public static final Pair<String, String> ORACLE_TIMESTAMP_SHADOW_INFO =
-      Pair.of("timestamp", "INT64");
+      Pair.of("shadow_timestamp", "INT64");
```
VardhanThigle: can this be dynamically generated?

If that's tough, can we add a nonce to this name?

one possibility would be` smt_shadow_timestamp_4b01b`
The `nonce` `4b01b` came from [UUID generator](https://www.uuidgenerator.net/version4) indended to keep things random. Based on max length of what a column name can support in spanner you can increase this to more than 4 characters to further reduce the possibility that a customer runs into this.

You should use the same nonce and prefix for all the shadow table columns and we shoudl document this in SMT's public docs.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -62,22 +62,32 @@ public abstract class ChangeEventContext {
   // Data table for the change event.
   protected String dataTable;
 
+  // The following fields store the "safe" names (to avoid collision with data column names) of the
+  // shadow table columns.
+  protected String safeTimestampColumn;
```
VardhanThigle: [nit Java Style] please make this and similar variables as `final` as we don't expect them to get modified outside the constructer.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/DatastreamConstants.java
VardhanThigle: Thanks for dynamically generating this by the escaping/prefixing logic.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/spanner/ShadowTableCreator.java
```diff
@@ -83,16 +83,38 @@ Table constructShadowTable(Ddl informationSchema, String dataTableName, Dialect
...
+   *
+   * @param baseShadowColumnName the initial desired name for the shadow column
+   * @param existingPrimaryKeyColumnNames a set of existing column names in the data table (should
+   *     be lowercase)
+   * @return a unique and safe column name
+   */
+  public static String getSafeShadowColumnName(
+      String baseShadowColumnName, Set<String> existingPrimaryKeyColumnNames) {
+    String safeName = baseShadowColumnName;
+    while (existingPrimaryKeyColumnNames.stream().anyMatch(safeName::equalsIgnoreCase)) {
```
VardhanThigle: [No Code change needed] Really appreciate the attention to detail shown in the prefixing logic here.

VardhanThigle (dismissed):
LGTM

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventConvertor.java
```diff
@@ -119,6 +119,9 @@ static Mutation.WriteBuilder changeEventToShadowTableMutationBuilder(
     String shadowTableName = shadowTablePrefix + tableName;
     try {
       Table table = ddl.table(shadowTableName);
+      if (table == null) {
```
bharadwaj-aditya: Will this cause the migration to stop ? Is this different from the previous behaviour ?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
aasthabharill: since they are getting modified out of the constructor (in ShadowTableCreator.java), i cannot assign them as final

VardhanThigle: Ok Just realized that we are setting super class variables in the inherited class.
Why not pass them as constructor params in that case?

VardhanThigle (approved):

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -78,10 +78,6 @@ protected ChangeEventContext(
     this.changeEvent = changeEvent;
     this.dataTable = changeEvent.get(DatastreamConstants.EVENT_TABLE_NAME_KEY).asText();
     Table table = ddl.table(this.dataTable);
-    if (table == null) {
-      throw new DroppedTableException("Table not found in DDL: " + this.dataTable);
```
VardhanThigle: If we are sure that this won't be null, please add a comment and a [@NotNull](https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html) on the class member that this is handled by the caller. If it's just a comment and annotation, you can roll that in in a subsequent PR as well. 


If we just remove the null check, line following this `table.primaryKeys()` will crash anyways.
There are many such dropped table checkers in RR code - a few of which emit a log to warn the user of a dropped table case and let the pipeline continue for other tables.

VardhanThigle (approved):

shreyakhajanchi (approved):
LGTM. Please merge once the tests pass

PR #3037, opened 2025-12-01
Add integration tests to live spanner migration for table and index limits

nmemond:
This adds integration tests to verify table and index limits during live spanner data migration.

This includes tests for the following limits:
- max size of a STRING cell
- max size of data per cell
- max size of a primary key
- max size of an index
- total size of non-key columns (currently ignored because of OOM issues with low-provisioned VMs/DBs)


VardhanThigle:
Can we please rebase this?
Will try re-running the workflow.

nmemond:
> Can we please rebase this? Will try re-running the workflow.

Done.

VardhanThigle (approved):

VardhanThigle (approved):

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToSpannerTableAndIndexLimitsIT.java
```diff
@@ -0,0 +1,273 @@
...
+                    mySQLResourceManager.getUsername(),
+                    mySQLResourceManager.getPassword(),
+                    mySQLResourceManager.getPort())
+                .setAllowedTables(Map.of(mySQLResourceManager.getDatabaseName(), TABLES))
+                .build();
+
+        ADDITIONAL_JOB_PARAMS.putAll(
+            new HashMap<>() {
+              {
+                put("dlqMaxRetryCount", "1");
```
darshan-sj, line 137: @nmemond  Can you please explain why "dlqMaxRetryCount" is set to "1" here?

cc: @VardhanThigle

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
```diff
@@ -0,0 +1,277 @@
...
+                    mySQLResourceManager.getUsername(),
+                    mySQLResourceManager.getPassword(),
+                    mySQLResourceManager.getPort())
+                .setAllowedTables(Map.of(mySQLResourceManager.getDatabaseName(), TABLES))
+                .build();
+
+        ADDITIONAL_JOB_PARAMS.putAll(
+            new HashMap<>() {
+              {
+                put("dlqMaxRetryCount", "1");
```
darshan-sj, line 138: @nmemond  Can you please explain why "dlqMaxRetryCount" is set to "1" here?

cc: @VardhanThigle

darshan-sj:
We should never use the static variable ADDITIONAL_JOB_PARAMS to pass parameters to the test. It was a mistake to introduce it in the first place. I'm trying to understand why specifically we have set "dlqMaxRetryCount" to "1" here. And since the ADDITIONAL_JOB_PARAMS is a static map in DatastreamToSpannerBaseIT it is getting set to all the tests and the tests have become flaky.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
nmemond: Ah, I believe this was likely a misunderstanding on my part. There are some rows that I expect will fail to migrate (intentionally, to test that certain limits can't be exceeded), and so I didn't want them continually being retried. I was not aware that this would impact every other test as well, apologies for that.

Are we better off just removing this completely then? I have a follow-up PR built on top of this one that I can make the change in, just let me know, thanks!

darshan-sj: I didn't understand this part > "There are some rows that I expect will fail to migrate (intentionally, to test that certain limits can't be exceeded), and so I didn't want them continually being retried."

Can you please elaborate?

nmemond: Yes, there are rows that we expect to fail to migrate and my expectation was that these would end up in the deadletter queue, so I didn't want them to be retried a bunch of times (since they'd just fail again).

For example in the `testCellSize` test, we write three rows to the source DB, but two of them have data that exceeds Spanner's allowed limits, and so we only expect one of the rows to end up in the target Spanner DB.

That being said, based on what you're saying, I likely misunderstood the purpose of that parameter as well as the preferred way to set it.

darshan-sj: Understood. You can pass a new parameter which will be local to this test in the hashmap here - 

https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3037/files#diff-ff4c29c82258d5352f252f0f9f040a3c6d9d5d414e0d75d7b5094d7a7e8c74caR150

But using ADDITIONAL_JOB_PARAMS (https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3037/files#diff-ff4c29c82258d5352f252f0f9f040a3c6d9d5d414e0d75d7b5094d7a7e8c74caR135) to pass the parameter is wrong. The ADDITIONAL_JOB_PARAMS is a static variable in the base class and it changes the parameter for every sub class. The ADDITIONAL_JOB_PARAMS should not have been created in the first place. 

I'll correct this in my PR. But I have one more comment below.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
```diff
@@ -0,0 +1,277 @@
...
+    assertThatPipeline(jobInfo).isRunning();
+
+    LOG.info("Waiting for pipeline to process data for cell size test...");
+    PipelineOperator.Result result =
+        pipelineOperator()
+            .waitForCondition(
+                createConfig(jobInfo, Duration.ofMinutes(10)),
+                SpannerRowsCheck.builder(
+                        spannerResourceManager, String.format("\"%s\"", LARGE_CELL_TABLE))
+                    .setMinRows(1)
```
darshan-sj, line 227: You are writing 3 rows to source (as per the sql resource file). You expect that 2 of them to be written in Spanner. 1 of them would fail. Then why are you waiting for only 1 row to appear in spanner here? And then you are asserting a row with id=3 at line 238. This is a streaming pipeline and there are no guarantees on the order in which the rows would be processed and written to spanner. What if the row with id not equal to 3 gets processed and the processing of row with id=3 is delayed by few minutes for some reason. You would wait for 1 row to appear and immediately proceed to asserts. And the assert at line 238 would fail. This would lead to flakiness. I think this problem is there in the PG dialect test also. Please correct this in subsequent PRs.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+    SpannerAsserts.assertThatStructs(rows)
+        .hasRecordsUnorderedCaseInsensitiveColumns(
+            List.of(
+                Map.ofEntries(
+                    // `assertThatStructs` converts the `rows` structs to records which eventually
+                    // calls
+                    // `com.google.cloud.spanner.Value.StringImpl::valueToString`, which truncates
+                    // strings to 33 characters
+                    entry("pk_col1", "A".repeat(33) + "..."),
```
darshan-sj, line 207: This is wrong way to assert. What if the implementation in `com.google.cloud.spanner.Value.StringImpl::valueToString` or treatment of String columns in `assertThatStructs` changes? This assert would fail. 

There are only 7 columns. You can read the value of each column from the struct retrieved from the Spanner and assert the exact contents of it fully. You don't need to use `assertThatStructs` compulsorily.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
```diff
@@ -0,0 +1,277 @@
...
+    assertThatResult(result).meetsConditions();
+    List<Struct> rows = null;
+    try {
+      rows = spannerResourceManager.readTableRecords(LARGE_CELL_TABLE, List.of("id"));
+    } catch (Exception e) {
+      LOG.error("Exception while reading spanner rows from {}", LARGE_CELL_TABLE, e);
+      throw e;
+    }
+    SpannerAsserts.assertThatStructs(rows)
+        .hasRecordsUnorderedCaseInsensitiveColumns(List.of(Map.ofEntries(entry("id", 3))));
```
darshan-sj, line 238: Similar to assertion comment in the first test method -- we should assert the largeCell column value here.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java
nmemond: Hmm, this is not quite correct. We do in fact expect 2 rows to fail to migrate and only 1 to succeed, which is why we check for 1 row here and then only assert on the 1 expected row below (which is the row with id 3). Ordering should play no part here: if row 3 is the first to be processed, it'll be migrated and we can happily proceed. If any other row is first, they'll fail, and the check won't pass (we won't get to the asserts at all).

nmemond:
Thanks for the feedback @darshan-sj. I likely won't have time to address them as I've started on another project and am no longer allocated to the Spanner migration work. I'll follow-up with Ross to see who's going to be taking this over to close out the remaining PRs, they may able to address these concerns.

PR #3045, opened 2025-12-03
 Support for Bool as a primary key

gpalanisamyYopp:


v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-spanner-schema.sql
```diff
@@ -300,3 +300,8 @@ CREATE TABLE IF NOT EXISTS timestamp_pk_table (
                                                   id TIMESTAMP NOT NULL,
                                                   timestamp_pk_col TIMESTAMP NOT NULL,
 ) PRIMARY KEY(id);
+
+CREATE TABLE IF NOT EXISTS bool_pk_table (
+  id INT64 NOT NULL,
+  bool_pk_col INT64 NOT NULL,
+) PRIMARY KEY(id);
```
NicolasRicheYopp, line 334: My understanding: 

The final Spanner table schema is defined by the spanner-migration-tool, not the local mapping by the Dataflow template.
Mysql BOOL column maps to a Spanner BOOL ref: https://googlecloudplatform.github.io/spanner-migration-tool/data-types/mysql.html


```sql
CREATE TABLE IF NOT EXISTS bool_pk_table (
  id BOOL NOT NULL,
  bool_pk_col BOOL NOT NULL,
) PRIMARY KEY(id);
```

I think the full flow is :
- A. source-spanner-migration-tool do the schema migration
- B. Dataflow source-to-spanner migrate the data: The mapping  BOOL -> IndexType.Numeric, Java.Long we do in dataflow is used to handle the data and migrate the data with Java, but does not create the schema. So in integration tests we create the spanner tables with the mapping set by A.

nmemond: Correct, both of these columns should have type `BOOL`.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDataTypesIT.java
```diff
@@ -273,6 +273,7 @@ private Map<String, List<Map<String, Object>>> getExpectedData() {
             "2005-01-01T00:01:54.123456000Z",
             "2037-12-30T23:59:59Z",
             "2038-01-18T23:59:59Z"));
+    expectedData.put("bool_pk", createRows("bool_pk", "0", "2", "1"));
```
nmemond: The test data that's being written only includes two values (true and false), those are the values we should expect (see also Nic's comment about the spanner table schema, which should map to `BOOL`).

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDataTypesIT.java
```diff
@@ -273,6 +273,7 @@ private Map<String, List<Map<String, Object>>> getExpectedData() {
             "2005-01-01T00:01:54.123456000Z",
             "2037-12-30T23:59:59Z",
             "2038-01-18T23:59:59Z"));
+    expectedData.put("bool_pk", createRows("bool_pk", "FALSE", "TRUE"));
```
NicolasRicheYopp: Depending if the row order is enforced or not, it might fails as the value are inserted in this order True, False

> INSERT INTO `bool_pk_table` (`id`, `bool_pk_col`) VALUES (TRUE, TRUE), (FALSE, FALSE);

https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3045/files#diff-3cb5689d27a0d68ae9e4ecb4c10e7a3a661b3f2fad67f566049b411538bab7faR432

nmemond: The row order shouldn't matter, I don't think. In the end, this gets turned into a Map anyways, so we can't really control the order either way. (Note, this is only true for these PK tables, since the PK is the same data type and value as the "value" column. The other tables use an auto-increment integer as the id, so the order does matter for those).

VardhanThigle:
Please resolve the merge conflicts and fix the spotless and checksyle issues.

VardhanThigle:
Spotless still needs to be fixed.

VardhanThigle:
Spotless still needs to be fixed.

v2/sourcedb-to-spanner/src/test/resources/DataTypesIT/mysql-spanner-schema.sql
```diff
@@ -323,6 +323,9 @@ CREATE TABLE IF NOT EXISTS timestamp_pk_table (
                                                   timestamp_pk_col TIMESTAMP NOT NULL,
 ) PRIMARY KEY(id);
 
+CREATE TABLE IF NOT EXISTS bool_pk_table (
+  id BOOL NOT NULL,
+  bool_pk_col BOOL NOT NULL,
```
NicolasRicheYopp, line 328: Uncomplete? (Missing closing ')' )

VardhanThigle:
PLease fix the integration test.

VardhanThigle:
Please rebase the PR on top of the tree to get the integration tests to passing state.

nmemond:
I unfortunately don't have permissions to push to the fork and the developer that owns this is out of office for the rest of the year. I've re-created the PR but up-to-date with main in #3112.

VardhanThigle:
Closing this as #3112  is raised to cover this.

PR #3072, opened 2025-12-08
chore: IAM Validation added for fail-safe.

pratickchokhani:
**New IAM Validation Utilities**: Introduced a set of new utility classes (IAMCheckResult, IAMPermissionsChecker, IAMRequirementsCreator, IAMResourceRequirements) to programmatically check IAM permissions for Google Cloud Platform resources.
**Cloud Resource Manager API Integration**: Added the google-api-services-cloudresourcemanager dependency to the pom.xml to enable interaction with the Cloud Resource Manager API for permission checks.
**Spanner Permission Validation**: Integrated a validateRequiredPermissions method into the Import and Export pipeline template to perform pre-execution IAM checks for Spanner resources, ensuring necessary permissions are present for the pipeline to run successfully.

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
@@ -28,6 +28,8 @@ import com.google.cloud.teleport.spanner.iam.IAMPermissionsChecker;
 import com.google.cloud.teleport.spanner.iam.IAMRequirementsCreator;
 import com.google.cloud.teleport.spanner.iam.IAMResourceRequirements;
 import com.google.cloud.teleport.spanner.spannerio.SpannerConfig;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
 import java.util.Collections;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
 import org.apache.beam.sdk.Pipeline;
@@ -40,6 +42,8 @@ import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.options.ValueProvider.NestedValueProvider;
 import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Dataflow template that exports a Cloud Spanner database to Avro files in GCS.
@@ -78,6 +82,7 @@ import org.apache.beam.sdk.transforms.SerializableFunction;
       "In addition to the Identity and Access Management (IAM) roles necessary to run Dataflow jobs, you must also have the <a href=\"https://cloud.google.com/spanner/docs/export#iam\">appropriate IAM roles</a> for reading your Cloud Spanner data and writing to your Cloud Storage bucket."
     })
 public class ExportPipeline {
+  private static final Logger LOG = LoggerFactory.getLogger(ExportPipeline.class);
 
   /** Options for Export pipeline. */
   public interface ExportPipelineOptions extends PipelineOptions {
@@ -295,24 +300,29 @@ public class ExportPipeline {
   }
 
   private static void validateRequiredPermissions(ExportPipelineOptions options) {
-    IAMResourceRequirements spannerRequirements =
-        IAMRequirementsCreator.createSpannerResourceRequirement();
+    try {
+      IAMResourceRequirements spannerRequirements =
+          IAMRequirementsCreator.createSpannerResourceRequirement();
 
-    GcpOptions gcpOptions = options.as(GcpOptions.class);
+      GcpOptions gcpOptions = options.as(GcpOptions.class);
 
-    IAMPermissionsChecker iamPermissionsChecker =
-        new IAMPermissionsChecker(gcpOptions.getProject(), gcpOptions);
-    IAMCheckResult missingPermission =
-        iamPermissionsChecker.check(Collections.singletonList(spannerRequirements));
-    if (missingPermission.isSuccess()) {
-      return;
+      IAMPermissionsChecker iamPermissionsChecker =
+          new IAMPermissionsChecker(gcpOptions.getProject(), gcpOptions);
+      IAMCheckResult missingPermission =
+          iamPermissionsChecker.check(Collections.singletonList(spannerRequirements));
+      if (missingPermission.isPermissionsAvailable()) {
+        return;
+      }
+      String errorString =
+          "For resource: "
+              + missingPermission.getResourceName()
+              + ", missing permissions: "
+              + missingPermission.getMissingPermissions()
+              + ";";
+      throw new RuntimeException(errorString);
+    } catch (GeneralSecurityException | IOException e) {
+      LOG.error("Error while validating permissions for spanner", e);
+      throw new RuntimeException(e);
     }
-    String errorString =
-        "For resource: "
-            + missingPermission.getResourceName()
-            + ", missing permissions: "
-            + missingPermission.getMissingPermissions()
-            + ";";
-    throw new RuntimeException(errorString);
   }
 }
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
@@ -23,10 +23,16 @@ import com.google.cloud.teleport.metadata.TemplateCreationParameter;
 import com.google.cloud.teleport.metadata.TemplateParameter;
 import com.google.cloud.teleport.metadata.TemplateParameter.TemplateEnumOption;
 import com.google.cloud.teleport.spanner.ExportPipeline.ExportPipelineOptions;
+import com.google.cloud.teleport.spanner.iam.IAMCheckResult;
+import com.google.cloud.teleport.spanner.iam.IAMPermissionsChecker;
+import com.google.cloud.teleport.spanner.iam.IAMRequirementsCreator;
+import com.google.cloud.teleport.spanner.iam.IAMResourceRequirements;
 import com.google.cloud.teleport.spanner.spannerio.SpannerConfig;
+import java.util.Collections;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
+import org.apache.beam.sdk.extensions.gcp.options.GcpOptions;
 import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
@@ -263,6 +269,7 @@ public class ExportPipeline {
             .withDatabaseId(options.getDatabaseId())
             .withRpcPriority(options.getSpannerPriority())
             .withDataBoostEnabled(options.getDataBoostEnabled());
+
     p.begin()
         .apply(
             "Run Export",
@@ -275,6 +282,7 @@ public class ExportPipeline {
                 options.getShouldExportRelatedTables(),
                 options.getShouldExportTimestampAsLogicalType(),
                 options.getAvroTempDirectory()));
+    validateRequiredPermissions(options);
     PipelineResult result = p.run();
     if (options.getWaitUntilFinish()
         &&
@@ -284,5 +292,27 @@ public class ExportPipeline {
         options.as(DataflowPipelineOptions.class).getTemplateLocation() == null) {
       result.waitUntilFinish();
     }
+  }
+
+  private static void validateRequiredPermissions(ExportPipelineOptions options) {
+    IAMResourceRequirements spannerRequirements =
+        IAMRequirementsCreator.createSpannerResourceRequirement();
+
+    GcpOptions gcpOptions = options.as(GcpOptions.class);
+
+    IAMPermissionsChecker iamPermissionsChecker =
+        new IAMPermissionsChecker(gcpOptions.getProject(), gcpOptions);
+    IAMCheckResult missingPermission =
+        iamPermissionsChecker.check(Collections.singletonList(spannerRequirements));
+    if (missingPermission.isSuccess()) {
+      return;
+    }
+    String errorString =
+        "For resource: "
+            + missingPermission.getResourceName()
+            + ", missing permissions: "
+            + missingPermission.getMissingPermissions()
+            + ";";
+    throw new RuntimeException(errorString);
   }
 }
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMCheckResult.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMCheckResult.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,146 @@
...
+      List<String> granted = testIamPermissionsResponse.getPermissions();
+      return granted == null ? Collections.emptyList() : granted;
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to check project permissions", e);
+    }
+  }
+
+  private CloudResourceManager createCloudResourceManagerService()
+      throws IOException, GeneralSecurityException {
+    if (resourceManagerForTesting != null) {
```
rohitwali: do you want to inject it into the class?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Checks IAM permissions for a single requirement list.
+   *
+   * @param requiredPermissions required permissions.
+   * @return Result of the check. It contains list of missing permissions.
+   */
+  private IAMCheckResult check(
+      CloudResourceManager resourceManager, List<String> requiredPermissions) {
+    HashSet<String> grantedPermissions =
+        new HashSet<>(checkPermission(resourceManager, projectIdResource, requiredPermissions));
```
rohitwali: are we only checking at project level? should we fallback to more granular level as well?
e.g., if user/role has required permissions at say, spanner instance or database level?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,146 @@
...
+
+    List<String> missingPermissions =
+        requiredPermissions.stream()
+            .filter(p -> !grantedPermissions.contains(p))
+            .collect(Collectors.toList());
+
+    return new IAMCheckResult(projectIdResource, missingPermissions);
+  }
+
+  private List<String> checkPermission(
```
rohitwali, line 91: what is the difference between this and previous "check" method? their signature looks similar

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,146 @@
...
+  public IAMCheckResult check(List<IAMResourceRequirements> requirements) {
+    try {
+      CloudResourceManager resourceManager = createCloudResourceManagerService();
+
+      List<String> permissionList =
+          requirements.stream()
+              .map(IAMResourceRequirements::getPermissions)
+              .flatMap(Collection::stream)
+              .toList();
+      return check(resourceManager, permissionList);
```
rohitwali: this method looks light and naming is a bit ambiguous. do you want to merge the methods or refactor this a bit?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Checks IAM permissions for a list of requirements.
+   *
+   * @param requirements List of resources and required permissions.
+   * @return List of results, only missing permissions are included. Empty list indicate all the
+   *     requirements are met.
+   */
+  public IAMCheckResult check(List<IAMResourceRequirements> requirements) {
+    try {
+      CloudResourceManager resourceManager = createCloudResourceManagerService();
```
rohitwali: should this be injected at class level?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMRequirementsCreator.java
```diff
@@ -0,0 +1,45 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.spanner.iam;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+public class IAMRequirementsCreator {
+  private static final List<String> SPANNER_PERMISSIONS =
```
rohitwali: curious: why do we have a single list? shouldnt the list be use case specific?
for example, I see same list for both import and export. shouldnt their permission list be different?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMResourceRequirements.java
```diff
@@ -0,0 +1,40 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.spanner.iam;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Represents the IAM permissions required on a specific GCP resource. */
+public class IAMResourceRequirements {
```
rohitwali, line 25: this class seems like a holder for an arraylist.
how are we thinking of growing/extending this class, if in future?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMRequirementsCreator.java
rohitwali: should this be a list that caller (user of the module) passes into this module?

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
```diff
@@ -285,4 +293,26 @@ public static void main(String[] args) {
       result.waitUntilFinish();
     }
   }
+
+  private static void validateRequiredPermissions(ExportPipelineOptions options) {
```
rohitwali, line 302: i see this code is duplicated across pipelines. would it make sense to push it down as an API in the IAM checker module?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: This class is designed for the check method to be called only once. This reduces IAM validation call. 

Caller is suppose to aggregate all the permissions and call the Check method.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: The testIam API is not to specific on how to validate permissions at instance or database level.

Few of my experiments failed on that part. We can do a followup on figuring out if granular permission is required.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: This require resource name. This was made from the perspective when we are checking permission at resource level. 

Currently, there isn't much difference.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: We can do that. But that will only help in testing purpose. 

My thinking was that if IAMpermissionsChecker is initialized by the user at class level and is never destroyed, then the resource manager will also never be destroyed. 

Hence, initialized later.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: Fixed.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMRequirementsCreator.java
pratickchokhani: The goal of this class is to have separate lists. For the export pipeline, we can have a separate list within this class. 

The thought process is that, we do not need too many distinct lists and there will be duplicates.

For export, I can create a instances list and databases create or drop. But rest should be there.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMResourceRequirements.java
pratickchokhani: This will grow when we will be checking permissions per resource.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
pratickchokhani: It needs to be per-Template. For Import and export the code is same. But this will change when we shift to other templates.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
```diff
@@ -285,4 +293,26 @@ public static void main(String[] args) {
...
+    IAMResourceRequirements spannerRequirements =
+        IAMRequirementsCreator.createSpannerResourceRequirement();
+
+    GcpOptions gcpOptions = options.as(GcpOptions.class);
+
+    IAMPermissionsChecker iamPermissionsChecker =
+        new IAMPermissionsChecker(gcpOptions.getProject(), gcpOptions);
+    IAMCheckResult missingPermission =
+        iamPermissionsChecker.check(Collections.singletonList(spannerRequirements));
+    if (missingPermission.isSuccess()) {
```
bharadwaj-aditya: the naming is a little confusing. Does missingPermission.isSuccess mean the permission is there or the check failed or does it mean the check succeeded ?

Try to avoid double negatives where possible.

bharadwaj-aditya: Based on the code reading - i would suggest renaming this to permissionCheck or something rather than missing permissions.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMCheckResult.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public String getResourceName() {
+    return resourceName;
+  }
+
+  public List<String> getMissingPermissions() {
+    return new ArrayList<>(missingPermissions);
+  }
+
+  public boolean isSuccess() {
```
bharadwaj-aditya: nit: rename method to permissionsAvailable or something clearer.
isSuccess could be misunderstood as the check API is successful or so.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
bharadwaj-aditya: Maybe add this in the docs to make it clear that it is an expensive call it should not be called frequently.

Though, just double checking if there is any harm in making it a class level variable as suggested ?

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
```diff
@@ -0,0 +1,135 @@
...
+      HashSet<String> grantedPermissions =
+          new HashSet<>(checkPermission(resourceManager, projectIdResource, permissionList));
+
+      List<String> missingPermissions =
+          permissionList.stream()
+              .filter(p -> !grantedPermissions.contains(p))
+              .collect(Collectors.toList());
+
+      return new IAMCheckResult(projectIdResource, missingPermissions);
+    } catch (IOException | GeneralSecurityException e) {
```
bharadwaj-aditya: Add an info logger here with the error message. It is possible that the parent hides this in another error.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
bharadwaj-aditya: It seems like a helper method. In which case, it might make sense to call the other method from here (or vice versa) to not duplicate the flows.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMRequirementsCreator.java
bharadwaj-aditya: In that case call this the DEFAULT_SPANNER_PERMISSIONS or so. Again with comments to cover the thought process.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMResourceRequirements.java
bharadwaj-aditya: +1 to Rohit's comment here. Please remove if not needed or add docs on where it might be needed.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
rohitwali: we are mixing testing code in main code. injecting would ensure decoupling and avoids doing stuff like `if test code`. just a bit cleaner to read IMO

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
rohitwali: >The testIam API is not to specific on how to validate permissions at instance or database level.

Could you say more about this? The TestIamPermissionsRequest takes resource as an argument, which can be instance or database. are you saying there is no way to specify them?

my concern here is that this may cause permission failures even though user may have required permissions and hence may be confusing to them.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
rohitwali: could you pl help be understand better with an example, on how it might deviate? and may be leave a comment for future code readers

manitgupta (commented):
1. Only the `IAMPermissiomsChecker` seems to have a UT, do we need UTs for the other classes that have been added?
2. Did you get a resolution on the placement of this lib? There was a discussion on `v1` v/s `v2` and how that would work out.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: Have moved this to constructur.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMPermissionsChecker.java
pratickchokhani: Have moved this to constructor.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMRequirementsCreator.java
pratickchokhani: Done.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMResourceRequirements.java
pratickchokhani: Have added the comment.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
pratickchokhani: Done.

v1/src/main/java/com/google/cloud/teleport/spanner/iam/IAMCheckResult.java
pratickchokhani: Done.

derrickaw:
Reopen if still valid. Thanks

PR #3087, opened 2025-12-09
add DECIMAL type support as primary key

gpalanisamyYopp:


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -163,13 +163,25 @@ private static Long splitLongs(Long start, Long end) {
...
+    if (start == null && end == null) {
+      return null;
+    }
+    if (start == null) {
+      start = BigDecimal.ZERO;
+    }
+    if (end == null) {
+      end = BigDecimal.ZERO;
+    }
+    int scale = partitionColumn.getScale();
```
VardhanThigle: Are we adding `getScale()` in [PartitionColum](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/96d239d6123580ff52f05d78307a7cbee43b7e40/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java#L45) in some other PR?

If it's not yet added, please add a `scale()` method there and also add a checkState similar to the [one](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/96d239d6123580ff52f05d78307a7cbee43b7e40/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java#L74) that's there

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -163,13 +163,26 @@ private static Long splitLongs(Long start, Long end) {
...
+      return null;
+    }
+    if (start == null) {
+      start = BigDecimal.ZERO;
+    }
+    if (end == null) {
+      end = BigDecimal.ZERO;
+    }
+
+    int scale = Math.max(start.scale(), end.scale());
```
NicolasRicheYopp, line 195: I think the scale should come from Mysql, not built-in scale of the Java BigDecimal.

Ref the float PR: Where decimalStepSize is added to SourceColumnIndexInfo https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3016/files#diff-d524f968e49d31273f6f392c3823cb46bada4c10b698b3591193449a0b235c4bR487

Note: Float should be merged first, to reuse that decimalStepSize property for BigDecimal

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -376,6 +376,7 @@ private ImmutableMap<String, SourceColumnType> getTableCols(
           .put("TINYTEXT", IndexType.STRING)
           .put("DATETIME", IndexType.TIME_STAMP)
           .put("TIMESTAMP", IndexType.TIME_STAMP)
+          .put("DECIMAL", IndexType.NUMERIC)
```
NicolasRicheYopp: Should be IndexType.DECIMAL (to call the BigDecimal splitter)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -452,6 +453,16 @@ private ImmutableList<SourceColumnIndexInfo> getTableIndexes(
           stringMaxLength = null;
         }
 
+        BigDecimal decimalStepSize = null;
+        if (indexType.equals(IndexType.FLOAT) || indexType.equals(IndexType.DECIMAL)) {
```
VardhanThigle, line 457: I am not sure if we have `FLOAT` type checked in yet.
Or should I hold on this reivew till `FLOAT` is merged in

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
```diff
@@ -452,6 +453,16 @@ private ImmutableList<SourceColumnIndexInfo> getTableIndexes(
...
+          if (numericScale > 0) {
+            decimalStepSize = BigDecimal.ONE.scaleByPowerOfTen(-numericScale);
```
VardhanThigle, line 459: Do we really need to couple this with stepSize or could we handle the scale as is?

My worry here is any rounding error in this exponentiation.

VardhanThigle:
Could we please fix the build?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
rossd-bitquill: This PR has some overlap with the `float` PR.  It would likely be simpler to start with the `float` PR, and then we will rebase this PR on top of that one.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java
rossd-bitquill: The scale is explicitly required to calculate the StepSize, as the StepSize is used to calculate the smallest possible difference between two values (later on in the Splitter) and then check for inequality of two values by comparing the difference to the smallest possible difference between two values.

I don't see a way to decouple these two concepts, but I also don't expect to have rounding issues given we are using a BigDecimal and a Java built-in method to perform the exponentiation.  As I understand the BigDecimal type, it is explicitly designed to handle such calculations without loss of precision.

nmemond: @VardhanThigle I think I see what you mean, and I expect working with the scale directly would actually simplify things a bit (I'm thinking it should allow us to use `setScale` and then `unscaledValue` to get a corresponding BigInteger, and then split with that). I'm going to work on swapping out the decimal step size in favour of just the scale itself.

nmemond: I would also say, I personally think it makes more sense for the `SourceColumnIndexInfo` class to have just the numeric scale, and then this gets used in various ways depending on the column type (since at the end of the day, the idea of "scale" is the same for decimal, double, and float, we just make use of it in different ways). But that can be a later improvement/refactoring.

nmemond:
@VardhanThigle I unfortunately don't have permissions to push to the fork and the developer that owns this is out of office for the rest of the year. I've re-created the PR but up-to-date with main and with build and issues fixed in #3119. Please close out this PR.

VardhanThigle:
Should this be rolled back as it has already been taken care of in other PRs.

VardhanThigle:
Closing as discussed over chat. This has been taken care of in other PRs.

PR #3088, opened 2025-12-09
Fix Auto Generate script to consider option files

tarun-google:
1. Extended Template Auto Generation script to fetch options from files
2. Updated PostgresToIceberg YAML Template with the Auto generate script
Screenshot 2025-12-09 at 2 06 05 PM

### DIFF
--- a/yaml/src/main/yaml/PostgresToIceberg.yaml
+++ b/yaml/src/main/yaml/PostgresToIceberg.yaml
@@ -17,7 +17,7 @@ template:
       "The Output Iceberg table need not exist, but the storage must exist and passed through catalog_properties."
    }
   streaming: false
-  hidden: true
+  hidden: false
 
   options_file:
     - "postgres_options"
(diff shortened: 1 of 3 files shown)


yaml/src/main/yaml/PostgresToIceberg.yaml
```diff
@@ -1,9 +1,23 @@
...
+  yamlTemplateFile: "PostgresToIceberg.yaml"
+  filesToCopy: >
+    {"template.yaml", "main.py", "requirements.txt", "options/postgres_options.yaml", "options/iceberg_options.yaml"}
+  contactInformation: "https://cloud.google.com/support"
+  requirements: {
+     "The Input Postgres instance and table must exist.",
+      "The Output Iceberg table need not exist, but the storage must exist and passed through catalog_properties."
+   }
+  streaming: false
+  hidden: true
```
tarun-google: @derrickaw marking this as hidden, let me know if we want to plan/test somethings before exposing to public

derrickaw: long as we have IT, its ok to keep visible.

derrickaw:
Can you run that script on the KafkaToBigQuery.yaml and verify everything looks good and include the changes in this PR.  Thanks.

PR #3097, opened 2025-12-10
Add Template and Blueprint support for Kafka to Iceberg

tarun-google:
1. Add kafka-Iceberg YAML Blueprint and Template
2. Add Service Options tuned for kafka-Iceberg pipelines

### DIFF
--- a/yaml/src/main/java/com/google/cloud/teleport/templates/yaml/KafkaToIcebergYaml.java
+++ b/yaml/src/main/java/com/google/cloud/teleport/templates/yaml/KafkaToIcebergYaml.java
@@ -280,4 +280,27 @@ public interface KafkaToIcebergYaml {
       example = "60")
   @Validation.Required
   Integer getTriggeringFrequencySeconds();
+
+  @TemplateParameter.Text(
+      order = 25,
+      name = "sdfCheckpointAfterDuration",
+      optional = true,
+      description = "Dataflow Pipeline Option: Duration after which to checkpoint stateful DoFns.",
+      helpText =
+          "Duration after which to checkpoint stateful DoFns. For example: 30s. Documentation: https://docs.cloud.google.com/dataflow/docs/reference/service-options",
+      example = "30s")
+  @Default.String("30s")
+  String getSdfCheckpointAfterDuration();
+
+  @TemplateParameter.Integer(
+      order = 26,
+      name = "sdfCheckpointAfterOutputBytes",
+      optional = true,
+      description =
+          "Dataflow Pipeline Option: Output bytes after which to checkpoint stateful DoFns.",
+      helpText =
+          "Output bytes after which to checkpoint stateful DoFns. For example: 536870912. Documentation: https://docs.cloud.google.com/dataflow/docs/reference/service-options",
+      example = "536870912")
+  @Default.Integer(536870912)
+  Integer getSdfCheckpointAfterOutputBytes();
 }
--- a/yaml/src/main/yaml/KafkaToIceberg.yaml
+++ b/yaml/src/main/yaml/KafkaToIceberg.yaml
@@ -27,6 +27,22 @@ template:
     - iceberg_common_options
     - iceberg_write_options
     - iceberg_streaming_write_options
+    - name: "sdfCheckpointAfterDuration"
+      label: "SDF Checkpoint After Duration"
+      description: "Dataflow Pipeline Option: Duration after which to checkpoint stateful DoFns."
+      help: "Duration after which to checkpoint stateful DoFns. For example: 30s. Documentation: https://docs.cloud.google.com/dataflow/docs/reference/service-options"
+      example: "30s"
+      default: "30s"
+      required: false
+      type: text
+    - name: "sdfCheckpointAfterOutputBytes"
+      label: "SDF Checkpoint After Output Bytes"
+      description: "Dataflow Pipeline Option: Output bytes after which to checkpoint stateful DoFns."
+      help: "Output bytes after which to checkpoint stateful DoFns. For example: 536870912. Documentation: https://docs.cloud.google.com/dataflow/docs/reference/service-options"
+      example: "536870912"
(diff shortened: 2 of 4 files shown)


yaml/src/main/yaml/KafkaToIceberg.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+  options_file:
+    - "kafka_options"
+    - "iceberg_options"
+
+  parameters:
+    - kafka_read_options
+    - iceberg_common_options
+    - iceberg_write_options
+    - iceberg_streaming_write_options
```
tomstepp, line 29: Sorry if I'm missing it, but where are the iceberg options defined?

yaml/src/main/yaml/KafkaToIceberg.yaml
```diff
@@ -0,0 +1,67 @@
...
+        message_name: {{ message_name }}
+        offset_deduplication: {{ offset_deduplication }}
+        redistribute_by_record_key: {{ redistribute_by_record_key }}
+        redistribute_num_keys: {{ redistribute_num_keys }}
+        redistributed: {{ redistributed }}
+        schema: {{ schema }}
+
+    - type: WriteToIceberg
+      name: WriteToIceberg
+      config:
```
tomstepp, line 74: Can we include `direct_write_byte_limit` here? Maybe a default limit could be 64 MB or something.

yaml/src/main/yaml/KafkaToIceberg.yaml
tarun-google: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/yaml/src/main/python/options/iceberg_options.yaml

yaml/src/main/yaml/KafkaToIceberg.yaml
tarun-google: okay, you just added this to beam 2 weeks ago, Beam release is in progress, i can wait and add this

yaml/src/main/yaml/KafkaToIceberg.yaml
tomstepp: Ah thanks, I was accidentally checking Apache Beam instead of GCP templates.

yaml/src/main/yaml/KafkaToIceberg.yaml
tomstepp: Sounds good, thanks!

tarun-google: DataflowTemplates Beam upgrade is pushed to after code freeze. So, i will take this as an improvement.

yaml/src/main/yaml/KafkaToIceberg.yaml
```diff
@@ -0,0 +1,74 @@
...
+        drop: {{ drop }}
+        keep: {{ keep }}
+        only: {{ only }}
+        partition_fields: {{ partitionFields }}
+        table_properties: {{ tableProperties }}
+        triggering_frequency_seconds: {{ triggeringFrequencySeconds }}
+
+options:
+  streaming: true
+  dataflow_service_options:
```
tarun-google, line 88: all the cross lang transforms are routed to runner_v2 so directly passing v2 args

yaml/src/main/java/com/google/cloud/teleport/templates/yaml/KafkaToIcebergYaml.java
```diff
@@ -0,0 +1,283 @@
...
+
+  @TemplateParameter.Text(
+      order = 8,
+      name = "format",
+      optional = true,
+      description = "The encoding format for the data stored in Kafka.",
+      helpText =
+          "The encoding format for the data stored in Kafka. Valid options are: RAW,STRING,AVRO,JSON,PROTO. For example: JSON",
+      example = "JSON")
+  @Default.String("JSON")
```
tomstepp, line 129: Curious why we default to JSON. I'm not sure what's most common. Looks like this is done in a few places.

yaml/src/main/yaml/KafkaToIceberg.yaml
```diff
@@ -0,0 +1,74 @@
...
+    - sdf_checkpoint_after_duration=30s
```
tomstepp: Is there some way for a customer to override these defaults?

Is there some way to provide the matching unbounded source flags?

derrickaw (approved):

yaml/src/main/java/com/google/cloud/teleport/templates/yaml/KafkaToIcebergYaml.java
tarun-google: not sure, following the default, but customers can surely overwrite.

yaml/src/main/yaml/KafkaToIceberg.yaml
tarun-google: all the cross lang transforms are routed to runner_v2(by Beam dataflow runner sdk) so directly passing v2 args, there is no use of passing v1 flags.

Also, customers can still pass additional_experimental args. sorry, let me test override, i am not sure, what gets picked.

tarun-google: Added sdf_checkpoint_after_duration as a jinja param, So we show documentation to user that we are setting it and allow them to override these args

tomstepp (approved):

yaml/src/main/yaml/KafkaToIceberg.yaml
tomstepp: It is possible to run on legacy (v1) runner with both Kafka source and Iceberg sink, so it would be nice to have similar defaults for legacy and portable (v2).

tomstepp: This could be a follow up though.

PR #3101, opened 2025-12-11
Clarify that MongoDB CDC to BigQuery template does not read from MongoDB directly

Abacn:


derrickaw:
Seems stale. Reopen if still valid. Thanks

PR #3107, opened 2025-12-12
Add integration tests to verify table and index limits for live spanner migration from PostgreSQL

nmemond:
This PR is built on top of #3037, that PR should be reviewed and merged first.

This adds integration tests to verify table and index limits during live spanner data migration from a PostgreSQL DB.

This includes tests for the following limits:
- max size of a STRING cell
- max size of data per cell
- max size of a primary key
- max size of an index
- total size of non-key columns (currently ignored because of OOM issues with low-provisioned VMs/DBs)


nmemond:
Closing in favour of #3265.

PR #3111, opened 2025-12-15
[Fix IT] Fix CassandraAllDataTypesIT failing test

aasthabharill:
https://b.corp.google.com/issues/468909414

#### RCA
- reference.conf file using to construct cassandra configuration comes from org/apache/cassandra which is at version 4.18.1 but all other related classes (for eg. TypedDriverOption) come from com/datastax/oss which is at version 4.17.0. This mismatch is causing newer added options/parameters to fail. 

#### Fix:
com.datastax.oss version upgrade to 4.18.1 - this is actually invalid ever since datastax moved to apache so we now have to use the new library org.apache.cassandra with updated version 4.18.1

#### Debugging:
- When i checked the reference.conf file in the java driver core we build (/.m2/repository/org/apache/cassandra/java-driver-core/4.18.1/java-driver-core-4.18.1.jar!/reference.conf) it had the line ```preferred-remote-dcs = [""]```. 
- This reference.conf is automatically merged with our specified .conf file ([Documentation](https://docs.datastax.com/en/developer/java-driver/4.2/manual/core/configuration/index.html): "Since application.conf inherits from reference.conf, you only need to redeclare what you override")
- We strictly check if all the options specified in the final configuration are "valid" configurations by using the class TypedDriverOption (path in jar: /Users/aasthabharill/.m2/repository/com/datastax/oss/java-driver-core/4.17.0/java-driver-core-4.17.0.jar!/com/datastax/oss/driver/api/core/config/TypedDriverOption.class) 

- Problem: 
  - All the updated versions (specifically 4.18) of the java driver are in apache/cassandra (mentioned on datastax website)
  - apache/cassandra is actually used to get the [reference.conf](https://github.com/apache/cassandra-java-driver/blob/ded2985f14e9774cb46dd0104b1f045613bc0279/core/src/main/resources/reference.conf#L583) file we're using which has the ```preferred-remote-dcs = [""]``` field.
  - but we're using TypedDriverOption of datastax version 4.17 which did not have ```preferred-remote-dcs``` in its [reference.conf](https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/reference/index.html)
  - Because we're using the above outdated TypedDriverOption to get a list of all the "valid" options - we will keep getting errors as and when new options are added.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
@@ -54,33 +54,7 @@ public final class CassandraDriverConfigLoader {
    */
   public static DriverConfigLoader loadFile(String path) throws FileNotFoundException {
     URL url = loadSingleFile(path);
-    LOG.info("aastha: Loaded Cassandra Driver config from path {}", path);
-    try {
-      LOG.info(
-          "aastha: Config content:\n{}",
-          com.google.common.io.Resources.toString(url, java.nio.charset.StandardCharsets.UTF_8));
-    } catch (java.io.IOException e) {
-      LOG.warn("aastha: Failed to read config content for logging", e);
-    }
-
-    URL referenceUrl =
-        CassandraDriverConfigLoader.class.getClassLoader().getResource("reference.conf");
-    LOG.info("aastha: reference.conf location: {}", referenceUrl);
-
-    // Log reference.conf content if found
-    if (referenceUrl != null) {
-      try {
-        LOG.info(
-            "aastha: reference.conf content:\n{}",
-            com.google.common.io.Resources.toString(
-                referenceUrl, java.nio.charset.StandardCharsets.UTF_8));
-      } catch (java.io.IOException e) {
-        LOG.warn("aastha: Failed to read reference.conf content for logging", e);
-      }
-    } else {
-      LOG.warn("aastha: reference.conf not found in classpath!");
-    }
-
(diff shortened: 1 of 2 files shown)


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/CassandraDriverConfigLoader.java
```diff
@@ -136,14 +136,8 @@ protected static void putInOptionsMap(
 
     TypedDriverOption option = OPTIONS_SUPPORTED_BY_DRIVER.get(optionName);
     if (Objects.equal(option, null)) {
-      LOG.error(
-          "Unknown Cassandra Option {}, Options supported by driver = {}",
-          optionName,
```
VardhanThigle, line 141: We added this check when we had discovered issues where the code was not setting options as expected.

Removing this can stop tests from failing when they actually should.

Let's discuss this offline and find the best route.

manitgupta: +1

aasthabharill: Thanks for calling out, i'll keep this in mind! I had kept this PR in draft as I'm still analyzing the code—this was just a preliminary check. I'll ping you once the PR is ready

VardhanThigle: Thanks. Sorry I missed that this PR was in draft.

manitgupta (approved):
Let’s wait for all the Cassandra tests to pass before merging.

PR #3138, opened 2025-12-21
Iam test

pratickchokhani:


v1/src/main/java/com/google/cloud/teleport/spanner/ExportTransform.java
```diff
@@ -184,6 +177,19 @@ public ExportTransform(
...
 
+      PCollection<Void> validationSignal =
+              begin
+                      .apply("Trigger Validation", Create.of((Void) null))
+                      .apply(
+                              "Validate Config",
+                              ParDo.of(
+                                      new DoFn<Void, Void>() {
+                                          @ProcessElement
+                                          public void processElement(ProcessContext c) {
```
bharadwaj-aditya, line 188: This validation will happen after the pipeline starts, which is going to be pretty similar to running the import itself.

PR #3140, opened 2025-12-22
[Draft] Break direct dep between live and bulk modules

manitgupta:


manitgupta:
This PR is breaking for other templates `datastream-mongodb-to-firestore`. Shelving this idea for now. We can consider creating a duplicate constant file to limit the change radius to Spanner templates.

PR #3150, opened 2025-12-23
Add PubSubToBigTable blueprint/template

derrickaw:
1. Adds a new PubSubToBigTable blueprint and template
2. Adds a helper support in pubsub
3. ~~Needs a windowing improvement before merging - pending PR - https://github.com/apache/beam/pull/37174~~
4. We previously expected the pending PR above to be needed for this PR, but the PR is only for multi-line support as single line string json is already supported.  Hence why the tests passed on the PR.

### DIFF
--- a/yaml/src/test/java/com/google/cloud/teleport/templates/yaml/PubSubToBigTableYamlIT.java
+++ b/yaml/src/test/java/com/google/cloud/teleport/templates/yaml/PubSubToBigTableYamlIT.java
@@ -19,17 +19,20 @@ import static com.google.common.truth.Truth.assertThat;
 import static org.apache.beam.it.truthmatchers.PipelineAsserts.assertThatPipeline;
 import static org.apache.beam.it.truthmatchers.PipelineAsserts.assertThatResult;
 
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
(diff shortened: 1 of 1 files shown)


yaml/src/main/yaml/PubSubToBigTable.yaml
```diff
@@ -0,0 +1,87 @@
...
+      example: "projects/your-project-id/topics/your-error-topic-name"
+      required: true
+      type: text
+      order: 1
+
+
+pipeline:
+  type: composite
+  transforms:
+    - type: ReadFromPubSub
```
tarun-google, line 51: why not have an un-opinionated pipeline? can you dump all the fields allowed by IO here? the same with bigtable, write to iceberg. Opinionated Templates can make the usage impossible for some customers. 

https://github.com/apache/beam/blob/ca27526d759ea617166958940671019dea6547e2/sdks/python/apache_beam/yaml/yaml_io.py#L307

derrickaw: I was trying to mimic the existing Dataflow templates and provide the main functionality.  If we wanted to add more flexibility, we can add that in a future PR.  Your preference still to do in this PR?  Thanks.

PR #3152, opened 2025-12-23
Add integration tests to live spanner migration for checking PostgreSQL data type support

nmemond:
This adds integration tests to check data type mappings when doing a live migration from PostgreSQL to Spanner for both dialects.

Note that some of the type mappings fail to migrate as expected. The checks for those are still included for completeness, but they're ignored to avoid failing the tests.

### DIFF
--- a/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerDataTypesIT.java
+++ b/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerDataTypesIT.java
@@ -321,6 +321,8 @@ public class PostgreSQLDatastreamToSpannerDataTypesIT extends DataStreamToSpanne
             "bit_varying",
             "bit_varying_to_string",
             "bytea",
+            "json",
+            "json_to_string",
             "macaddr",
             "macaddr8",
             "uuid_to_bytes",


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerDataTypesIT.java
```diff
@@ -0,0 +1,670 @@
...
+  private static HashSet<PostgreSQLDatastreamToSpannerDataTypesIT> testInstances = new HashSet<>();
+
+  @Before
+  public void setUp() throws IOException {
+    skipBaseCleanup = true;
+    synchronized (PostgreSQLDatastreamToSpannerDataTypesIT.class) {
+      testInstances.add(this);
+      if (!initialized) {
+        LOG.info("Setting up PostgreSQL resource manager...");
+        postgresResourceManager = PostgresResourceManager.builder(testName).build();
```
VardhanThigle: Would this need a static Resource maanger?

A static resource manager is a statically provisioned cloud PG instance in the test project. In this context it will also need flags to set wal level correctly and [ip allowlisting](https://docs.cloud.google.com/datastream/docs/ip-allowlists) for datastream access.

Reasons:
Currently the test is failing for `wal_level` for the containarized source.
Even if we get past that, I doubt if datastream would have access to migrate CDC from the containerised source.




Note:
If you need to test this with a static resource manager locally, you will need to provision one and pass the correct mvn parameters for it's address.

Please refer [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/96d239d6123580ff52f05d78307a7cbee43b7e40/v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/DatastreamToSpannerReservedKeywordsPostgresIT.java#L83) for an example of using static source.

nmemond: Ah yes, that makes complete sense. I'll update.

VardhanThigle:
Failure in asserts

```
pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Asserting type: boolean_to_string
[pool-1-thread-60] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Loading columns [id, col] from teleport3.testpos_20251225_064337_vk3rlh.t_boolean_to_string
[pool-1-thread-60] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Loaded 3 records from teleport3.testpos_20251225_064337_vk3rlh.t_boolean_to_string
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [1, false]
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [2, true]
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [3, null]
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Asserting type: json_to_string
[pool-1-thread-60] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Loading columns [id, col] from teleport3.testpos_20251225_064337_vk3rlh.t_json_to_string
[pool-1-thread-60] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Loaded 3 records from teleport3.testpos_20251225_064337_vk3rlh.t_json_to_string
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [1, {"duplicate_key": 2}]
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [2, {"null_key": null}]
[pool-1-thread-60] INFO com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT - Found row: [3, null]
Error:  Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1,665.32 s <<< FAILURE! - in com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT
[INFO] Running com.google.cloud.teleport.v2.templates.SeparateShadowTableDatabaseStringOverridesIT
[pool-1-thread-38] INFO org.apache.beam.it.gcp.TemplateTestBase - Starting integration test com.google.cloud.teleport.v2.templates.SeparateShadowTableDatabaseStringOverridesIT.migrationTestWithRenameTableAndColumns
Dec 25, 2025 6:40:23 AM com.google.api.client.googleapis.services.AbstractGoogleClient <init>
WARNING: Application name is not set. Call Builder#setApplicationName.
[pool-1-thread-38] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Not creating Spanner instance - reusing static teleport2
[pool-1-thread-38] INFO org.apache.beam.it.gcp.spanner.SpannerResourceManager - Creating database shadow__20251225_064337_z9eumg in instance teleport2.
```

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerDataTypesIT.java
```diff
@@ -0,0 +1,670 @@
...
+    result.put("int2_to_string", createRows("-32768", "32767", "3", "NULL"));
+    result.put("int4", createRows("-2147483648", "2147483647", "4", "NULL"));
+    result.put("int4_to_string", createRows("-2147483648", "2147483647", "4", "NULL"));
+    result.put("int8", createRows("-9223372036854775808", "9223372036854775807", "5", "NULL"));
+    result.put(
+        "int8_to_string", createRows("-9223372036854775808", "9223372036854775807", "5", "NULL"));
+    result.put("json", createRows("{\"duplicate_key\":1}", "{\"null_key\":null}", "NULL"));
+    result.put(
+        "json_to_string",
+        createRows("{\"duplicate_key\": 1, \"duplicate_k...", "{\"null_key\": null}", "NULL"));
```
VardhanThigle: Should this be `duplicate_key`?

VardhanThigle: This might be fixed with the correction pushed current diff. Thanks.

Please fix spotless.

nmemond: Spotless should be fixed now. Sorry about that, I thought I had run it but obviously missed it!

VardhanThigle: No Worries :)

VardhanThigle:
There seems to be a Assert in matching 

`PostgreSQLDatastreamToSpannerDataTypesIT.testPostgreSqlDataTypes`

```
expected that contains unordered record (and case insensitive) {COL={"duplicate_key":1}, ID=1}, but only had [{col={"duplicate_key":2}, id=1}, {col={"null_key":null}, id=2}, {col=NULL, id=3}]
	at com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT.validateResult(PostgreSQLDatastreamToSpannerDataTypesIT.java:372)
	at com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT.testPostgreSqlDataTypes(PostgreSQLDatastreamToSpannerDataTypesIT.java:263)
```

`PostgreSQLDatastreamToSpannerDataTypesIT.testPostgreSqlDataTypesPGDialect`

```
expected that contains unordered record (and case insensitive) {COL=08:00:2b:01:02:03:04:05, ID=1}, but only had [{col=NULL, id=1}, {col=NULL, id=2}]
	at com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT.validateResult(PostgreSQLDatastreamToSpannerDataTypesIT.java:372)
	at com.google.cloud.teleport.v2.templates.PostgreSQLDatastreamToSpannerDataTypesIT.testPostgreSqlDataTypesPGDialect(PostgreSQLDatastreamToSpannerDataTypesIT.java:311)
```

anowardear062-svg:
Support and proses issues

anowardear062-svg:
This the update on the place

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerDataTypesIT.java
```diff
@@ -0,0 +1,674 @@
...
+      if (combinedCondition == null) {
+        combinedCondition = c;
+      } else {
+        combinedCondition.and(c);
+      }
+      numCombinedConditions += 1;
+      if (numCombinedConditions >= 3) {
+        conditions.add(combinedCondition);
+        combinedCondition = null;
+        numCombinedConditions = 0;
```
darshan-sj, line 464: Why are you grouping 3 conditions with `and` and then using the Chained condition on the grouped ones? Any particular reason for having this pattern?

We should not use chained condition check like this. Chained condition check introduces 15s delay in between each check and with so many of them chained, the test takes a lot of time to execute and could sometimes timeout if the migration is a little slow. 

I'm working on fixing the already checked in tests with this wrong pattern -
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3380

nmemond: The nice thing about the Chained condition is that it avoids re-running conditions that have already passed, but of course the downside is as you mentioned: regardless of whether a condition passes or fails, there's a 15 second delay before the next check runs.

The combination of Chained and `and` was an effort to strike a balance between the two, given there are quite a few conditions. Maybe 3 was too small of a grouping, but I do think the idea itself is sound. Otherwise, there's a chance we end up having to re-run dozens of conditions over and over because a later condition is failing. In the worst case, we re-run every condition except the last one (if that's the one that's failing). Using a fixed group size, we can decide how many conditions will re-run in that worst case: for a group size of X, in the worst case we will re-run X-1 passing conditions every check.

darshan-sj: I don't think that re-running conditions which have already passed is a problem here. The tables don't contain a lot of rows and the count query should return within milliseconds. Even if we have 100s of tables, querying all of them is not that much inefficient operation. On the other hand waiting for 15s seconds between each query is bad and it is adding up a lot of delay here. Please don't use this pattern in your further PRs and use "and" across all the conditions.

PR #3159, opened 2025-12-26
[SourceDbToSpanner] Correcting Binary datatype handling in SourceDbToSpanner template on Spanner write errors

darshan-sj:
Merged all the Bulk+Live retryDLQ test cases into a single IT.

Also, found and fixed the following bugs:

1. DLQ file overwrite issue was still there since the same `dlqTransform` with same UUID was used to write the Spanner write errors, transformation errors and filtered events and they were overwriting each others' files. - Fixed this bug.
2. Fixed BYTES and BYTES ARRAY handling for spanner write errored DLQ events.
3. Fixed an issue in GenericRecordTypeConverter for BIT datatype handling when there is custom transformation present in the pipeline.

### DIFF
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLCustomTransformationsNonShardedIT.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLCustomTransformationsNonShardedIT.java
@@ -113,7 +113,7 @@ public class MySQLCustomTransformationsNonShardedIT extends SourceDbToSpannerITB
     row.put("enum_column", "1");
     row.put("bool_column", true);
     row.put("varbinary_column", "AQIDBAUGBwgJCgsMDQ4PEBESExQ=");
-    row.put("bit_column", "Ew==");
+    row.put("bit_column", "DQ==");
     row.put("binary_column", "AQIDBAUGBwgJCgsMDQ4PEBESExQ=");
     row.put("char_column", "newchar");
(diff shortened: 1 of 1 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -273,7 +274,29 @@ protected FailsafeElement<String, String> mutationToDlqElement(Mutation m) {
     Map<String, Value> mutationMap = m.asMap();
     for (Map.Entry<String, Value> entry : mutationMap.entrySet()) {
       Value value = entry.getValue();
-      json.put(entry.getKey(), value == null ? null : String.valueOf(value));
+      Object val = null;
+      if (value != null && !value.isNull()) {
+        switch (value.getType().getCode()) {
+          case BYTES:
+            val = Hex.encodeHexString(value.getBytes().toByteArray());
```
VardhanThigle, line 276: Can we add a case for `bytesArray` as well?
We support arrays for Cassandra.

VardhanThigle (dismissed):
Overall looks good to me. Please consider adding a case for bytesArray.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
darshan-sj: Done.

VardhanThigle (approved):

PR #3182, opened 2026-01-05
Live and Bulk Retry: Data conversion errors when NULL is passed in non-textual datatype 

aasthabharill:
https://b.corp.google.com/issues/458072376
https://b.corp.google.com/issues/458072905

**Alternative implementation:**
instead of manually checking in each toDatatype function, we could add this check to containsValue function that is used by all datatypes

Reason for rejection: a user can pass "NULL" string purposely to a string column - this should not be converted to null by my logic. containsValue function is called in method toString and thus this conversion cannot be added to the common function and needs to be added for each datatype seperately wherever relevant.
This would not change the existing behaviour for string data type. It would only change NULL handling for other datatypes


darshan-sj (approved):
LGTM

PR #3197, opened 2026-01-06
Bump Beam to 2.70.0

Abacn:


Abacn:
R: @shunping

python/default_base_python_requirements.txt
```diff
@@ -1,2 +1,2 @@
-apache-beam[gcp]==2.69.0
```
shunping, line 1: Do we need "==2.70.0" in the new file?

python/default_base_yaml_requirements.txt
```diff
@@ -1,2 +1,2 @@
-apache-beam[dataframe,gcp,test,yaml]==2.69.0
```
shunping, line 1: Same as above.

python/default_base_python_requirements.txt
shunping: Per our offline discussion, this change is done so we won't have to change the beam version in file in the future.

Abacn: these version numbers were introduced in https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/03e50babd03c6eca32cd958c5084f6e2b7a10892#diff-e6550ce2bfc6ec90e1b09f1bb69b7753ae931a58080373851388e100e8bff956

and prior to that, new versions were propagated automatically. It shouldn't be necessary otherwise we need to bump these values each time

PR #3204, opened 2026-01-08
[DO NOT MERGE] Validate Beam 2.71.0rc1

damccorm:


damccorm:
Currently what I am seeing is that all tests are passing except some spanner related tests. These are failing with:

```
Timeout in polling result file: gs://dataflow-staging-us-west2-269744978479/staging/template_launches/2026-01-08_12_48_39-9625519560852535969/operation_result.
Service account: 269744978479-compute@developer.gserviceaccount.com
Image URL: gcr.io/cloud-teleport-testing/spanner-changestreams-to-bigquery:2026-01-08-20-42-15_IT
Troubleshooting guide at https://cloud.google.com/dataflow/docs/guides/troubleshoot-templates#timeout-polling
```

Example: https://pantheon.corp.google.com/dataflow/jobs/us-west2/2026-01-08_12_48_39-9625519560852535969;logsSeverity=ERROR;graphView=1?project=cloud-teleport-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=13802955&mods=dm_deploy_from_gcs

I'm not currently seeing anything in the logs that gives much of a hint about what is happening. I have confirmed that if I remove `SpannerIO.ReadChangeStream` from the pipeline, that allows it to succeed, so there is something odd going on there. There are no obvious changes in https://github.com/apache/beam/tree/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner

@Abacn have you seen anything like this while doing this upgrade before? If not, I will keep looking

Abacn:
In the past there were situations that spanner tests due to breaking changes in Spanner client that get upgraded as GCP BOM upgrade happened in Beam.

"Timeout in polling result file: gs://dataflow-staging-us-west2-269744978479/staging/template_launches/2026-01-08_12_48_39-9625519560852535969/operation_result."

This sounds template launcher failed.

>  if I remove SpannerIO.ReadChangeStream from the pipeline

Sounds like pipeline expansion gets stuck. if we can reproduce it locally then could get a hint. Also reach out to Spanner CDC team to investigate.

damccorm:
> In the past there were situations that spanner tests due to breaking changes in Spanner client that get upgraded as GCP BOM upgrade happened in Beam.

Thanks, this is good context - was already starting on the other steps here, thanks!

damccorm:
Interestingly, I'm not able to repro the problem. For example, here is the same pipeline that fails as a template running with the 2.71.0 rc version - https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-01-12_09_09_16-11282152439616596575;step=;graphView=0?project=apache-beam-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=13802955&mods=dm_deploy_from_gcs

vs the same pipeline running in a template: https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-01-12_08_29_11-15499653623879337782;logsSeverity=INFO;graphView=1?project=apache-beam-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=13802955&mods=dm_deploy_from_gcs

I have pretty definitively proven that something with the spanner change streams connector is the problem though. The minimal repro I have right now is changing the run method in SpannerChangeStreamsToBigQuery.java to:

```
  public static PipelineResult run(SpannerChangeStreamsToBigQueryOptions options) {
    setOptions(options);
    validateOptions(options);
    Pipeline pipeline = Pipeline.create(options);
    DeadLetterQueueManager dlqManager = buildDlqManager(options);
    String spannerProjectId = OptionsUtils.getSpannerProjectId(options);

    String dlqDirectory = dlqManager.getRetryDlqDirectoryWithDateTime();
    String tempDlqDirectory = dlqManager.getRetryDlqDirectory() + "tmp/";

    Timestamp startTimestamp =
        options.getStartTimestamp().isEmpty()
            ? Timestamp.now()
            : Timestamp.parseTimestamp(options.getStartTimestamp());
    Timestamp endTimestamp =
        options.getEndTimestamp().isEmpty()
            ? Timestamp.MAX_VALUE
            : Timestamp.parseTimestamp(options.getEndTimestamp());

    SpannerConfig spannerConfig =
        SpannerConfig.create()
            .withProjectId(spannerProjectId)
            .withInstanceId(options.getSpannerInstanceId())
            .withDatabaseId(options.getSpannerDatabase())
            .withRpcPriority(options.getRpcPriority());
    if (!Strings.isNullOrEmpty(options.getSpannerHost())) {
      spannerConfig =
          spannerConfig.withHost(ValueProvider.StaticValueProvider.of(options.getSpannerHost()));
    }
    // Propagate database role for fine-grained access control on change stream.
    if (options.getSpannerDatabaseRole() != null) {
      spannerConfig =
          spannerConfig.withDatabaseRole(
              ValueProvider.StaticValueProvider.of(options.getSpannerDatabaseRole()));
    }

    SpannerIO.ReadChangeStream readChangeStream =
        SpannerIO.readChangeStream()
            .withSpannerConfig(spannerConfig)
            .withMetadataInstance(options.getSpannerMetadataInstanceId())
            .withMetadataDatabase(options.getSpannerMetadataDatabase())
            .withChangeStreamName(options.getSpannerChangeStreamName())
            .withInclusiveStartAt(startTimestamp)
            .withInclusiveEndAt(endTimestamp)
            .withRpcPriority(options.getRpcPriority());

    PCollection<DataChangeRecord> dataChangeRecord =
        pipeline
            .apply("Read from Spanner Change Streams", readChangeStream)
            .apply("Reshuffle DataChangeRecord", Reshuffle.viaRandomKey());

    return pipeline.run();
  }
```

I can then launch a template with:

```
mvn clean verify   -PtemplatesIntegrationTests   -Dproject="apache-beam-testing"   -DartifactBucket="apache-beam-damccorm"   -Dregion=us-central1   -pl v2/googlecloud-to-googlecloud -am   -Dtest=SpannerChangeStreamsToBigQueryIT#testSpannerChangeStreamsToBigQueryAddTable
```

This successfully launches the template when I don't change `pom.xml` at all (Template version is set to 2.70.0) though the test fails because of my changes. When I apply this PRs changes to `pom.xml`, it fails to launch the template at all. There is something wrong with the connector in 2.71.0. It is possibly just a permissions issue, but we need to investigate further.

damccorm:
I've narrowed the problem to this line: https://github.com/apache/beam/blob/d0223389f47f8085477e113391d7ae5961bc0bbc/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java#L2105

Specifically, having:

```
SpannerAccessor.getOrCreate(spannerConfigWithCredential)
```

in the template run method causes hangs. Interestingly, we still log after this, so it seems like some resource is not getting cleaned up correctly

Abacn:
> There is something wrong with the connector in 2.71.0. 

Try to revert com.google.cloud:google-cloud-spanner to 6.95.0 for templates?

https://github.com/apache/beam/commit/8ff2c940321cc8adb49e2fc590a1d35fd7113add

if it worked we then need to revert #36995 (and the line in #37142 s well) to unblock the release, and ask Spanner team to take a further look

> we still log after this

Log agent runs in a parallel container inside launcher VM and it is not affected by stuck pipeline expansion

EDIT:

~I cannot reproduce it locally either. What I did is to launch a pipeline locally with this modified code:~

Actually it can be reproduced: https://github.com/Abacn/DataflowTemplates/commit/26c4218bb7a4af1dc963216fa7ae6929a0ea5ce3

Running with IDEA, there are threads preventing main method exiting:

image

Abacn:
Update: confirmed that downgrading `com.google.cloud:google-cloud-spanner` to 6.104 resolves the stuck main function:


My minimum reproduce:

build.gradle
```
plugins {
    id 'application'
}

configurations.all {
    resolutionStrategy {
        force "org.slf4j:slf4j-api:1.7.36"
    }
}

apply plugin : "java"

def beam_version = "2.70.0"
group = 'com.github.abacn'

ext {
    javaMainClass = "com.google.spanner.v1.SpannerTest"
}

application {
    mainClassName = javaMainClass
}

repositories {
    mavenCentral()
}

dependencies {
    implementation(group: 'org.apache.beam', name: 'beam-sdks-java-io-google-cloud-platform', version: "$beam_version")
    implementation("com.google.cloud:google-cloud-spanner:6.104.0")
}
```

main function:

```
  public static void main(String[] argv) {
    SpannerConfig config = SpannerConfig.create()
        .withProjectId("cloud-teleport-testing")
        .withInstanceId("beam-test")
        .withDatabaseId("test-271");
    SpannerAccessor accessor = SpannerAccessor.getOrCreate(config);
    // accessor.close();
  }
```

- Spanner 6.104.0 x without close --- run successful
- Spanner 6.105.0 x without close ---- stuck
- Spanner 6.105.0 x with close --- run successful

I think we need to revert this line

https://github.com/apache/beam/blob/133034612813f20d76e0c081d5058f04daf10544/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L617

and build a new RC

before that we can override implementation("com.google.cloud:google-cloud-spanner:6.104.0") dependency in DataflowTemplates to use 6.104.0 (declare it in pom, and exclude it as a transient dep for io-google-cloud-platform) to confirm

damccorm:
Yep, this showed up for me as well. Thanks!

> Spanner 6.105.0 x with close --- run successful

This is a good find, thanks for digging in. I was guessing we weren't closing connections but hadn't had time to dig in yet. We could probably consider rolling forward instead. I did a quick audit and the only places this is misused are:

- https://github.com/apache/beam/blob/133034612813f20d76e0c081d5058f04daf10544/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java#L2105
- https://github.com/apache/beam/blob/133034612813f20d76e0c081d5058f04daf10544/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/changestreams/dao/DaoFactory.java#L119 (a few places in this file)

damccorm:
I'll put together a PR for tomorrow; if we can avoid diverging from the BOM version, I do prefer that.

sakthivelmanii:
@rahul2393 can you please check this issue? my primary suspect is related to enabling gRPC gcp by default.

damccorm:
> Yep, this showed up for me as well. Thanks!
> 
> > Spanner 6.105.0 x with close --- run successful
> 
> This is a good find, thanks for digging in. I was guessing we weren't closing connections but hadn't had time to dig in yet. We could probably consider rolling forward instead. I did a quick audit and the only places this is misused are:
> 
> * https://github.com/apache/beam/blob/133034612813f20d76e0c081d5058f04daf10544/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java#L2105
> * https://github.com/apache/beam/blob/133034612813f20d76e0c081d5058f04daf10544/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/changestreams/dao/DaoFactory.java#L119 (a few places in this file)

Actually, a roll forward isn't as easy as I thought. It is straightforward in SpannerIO, but in DaoFactory there are several places where we initialize a SpannerAccessor, then use it to create a DB client which we continue to pass around beyond the scope of the method. So we can't just call close there, we need to close the connection only when we're tearing down the calling context.

For now, I'll just revert the spanner version, but we'll need more investigation

damccorm:
https://github.com/apache/beam/pull/37305 for the revert in Beam

damccorm:
Validating RC2 here - https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3222 - I'll leave this one open for now though for investigation

PR #3207, opened 2026-01-09
Move neo4j test resource manager to test class level

Abacn:
Attempt to mitigate a recurring issue #2634 

currently neo4j test container is setup before per test run, and it is flaky. I suspect the concurrently running containers has some race condition and could cause

```
[pool-1-thread-12] ERROR tc.neo4j:5-enterprise - Could not start container
org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException
```

seen in tests

### DIFF
--- a/.github/workflows/java-pr.yml
+++ b/.github/workflows/java-pr.yml
@@ -57,8 +57,112 @@ env:
 permissions: write-all
 
 jobs:
+  spotless_check:
+    name: Spotless
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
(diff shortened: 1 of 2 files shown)


Abacn:
Test passsed: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/20867377749?pr=3207

```
[INFO] googlecloud-to-neo4j ............................... SUCCESS [21:50 min]
```

where `Attempting to clean up Neo4j manager.` log appears 6 times (there are 6 test classes in total).

Previously this log appeared 28 times, which means test container spined up 28 times.

revert changes (that comment out other tests) and ready for review

fbiville (changes requested):
a bunch of integration test classes are gone, is there a specific reason why? they covered specific features of the template.

PR #3219, opened 2026-01-14
Fixing Nulls becoming 0 for Cassandra and Astra

VardhanThigle:
## Issue
fixes b/475697267
Cassandra/AstraDB migration were making null integers 0, due to the behavior of the `Row` interface.
## Fix
Added an additional null check before calling the extractor.
## Test
1. Covered all lines of code change in UT as well as IT
2. Tested Null values on a real Astra DB instance.


n0rm4l-me:
Awesome, thanks for the quick fix!

PR #3222, opened 2026-01-15
Update to Beam 2.71.0

damccorm:


damccorm:
This is running into the same problems as the last rc :( I'm not sure why yet. I did confirm that https://repository.apache.org/content/repositories/orgapachebeam-1421/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.71.0/ should contain the correct Spanner version...

damccorm:
`mvn help:effective-pom -Dartifact=org.apache.beam:beam-sdks-java-google-cloud-platform-bom:2.71.0` still shows 

```
      <dependency>
        <artifactId>google-cloud-spanner</artifactId>
        <version>6.105.0</version>
      </dependency>
```

which is probably the issue. I'll dig in more here.

.github/actions/setup-env/action.yml
```diff
@@ -70,6 +70,10 @@ runs:
       with:
         cache-key: ${{ inputs.java-cache-key }}
         java-version: ${{ inputs.java-version }}
+    - name: Temporarily clear java cache
+      shell: bash
+      run: |
+        rm -rf /home/runner/.m2/repository/org/apache/beam
```
damccorm: I had to do this to remove cached versions of the sdk, but it should have no real impact

pom.xml
```diff
@@ -49,9 +49,8 @@
     <jacoco.version>0.8.13</jacoco.version>
 
     <!-- Beam and linked versions -->
-    <beam.version>2.70.0</beam.version>
-    <beam-python.version>2.70.0</beam-python.version>
-    <beam-maven-repo></beam-maven-repo>
```
Abacn, line 54: Just noting here, when doing 2.72 validation`<beam-maven-repo></beam-maven-repo>` line needs to be added back

damccorm: Thanks, didn't mean to cut this. I added it back

PR #3223, opened 2026-01-15
Fix surefire version inconsistency and smoke tests

Abacn:
Another action item for https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/20950867982

After https://github.com/apache/maven-surefire/issues/2707, floating points written no longer uses "," separator. However due to version mismatch the written raw report (with older surfire) still has commas, but the report plugin (on newer version) doesn't expect that.

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-report-plugin:3.5.3:report-only (default-cli) on project kafka-to-bigquery: Execution default-cli of goal org.apache.maven.plugins:maven-surefire-report-plugin:3.5.3:report-only failed: Error parsing JUnit XML report /home/runner/actions-runner/_work/DataflowTemplates/DataflowTemplates/v2/kafka-to-bigquery/target/surefire-reports/TEST-com.google.cloud.teleport.v2.templates.KafkaToBigQueryFlexAvroIT.xml: Failed to parse time value: For input string: "9,318.045" -> [Help 1]
```

suggesting one test takes 9,318.045 s, but unfortunately the report wasn't uploaded and prevent further investigation

After bumping surefire version, #3224 was found and also fixed here

### DIFF
--- a/cicd/cmd/run-it-smoke-tests/main.go
+++ b/cicd/cmd/run-it-smoke-tests/main.go
@@ -49,7 +49,7 @@ func main() {
 	// Run integration tests
 	mvnFlags = workflows.NewMavenFlags()
 	err = workflows.MvnVerify().Run(
-		mvnFlags.IncludeDependencies(),
+		mvnFlags.DoNotIncludeDependencies(),
 		mvnFlags.SkipDependencyAnalysis(),
 		mvnFlags.SkipCheckstyle(),
 		mvnFlags.SkipJib(),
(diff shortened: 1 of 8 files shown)


Abacn:
After bumping surefire version, it is found part of smoke test has never been run: #3224, and also fixed in this PR

cicd/cmd/run-it-smoke-tests/main.go
```diff
@@ -49,7 +49,7 @@ func main() {
 	// Run integration tests
 	mvnFlags = workflows.NewMavenFlags()
 	err = workflows.MvnVerify().Run(
-		mvnFlags.IncludeDependencies(),
+		mvnFlags.DoNotIncludeDependencies(),
```
Abacn, line 52: after fixing smoke test tags, smoke tests for spanner-pr starting to fail due to some parameter not set by Bigtable resource manager. This is because we run irrelevant tests in all dependency modules.

After the change it will still get covered by Java PR's smoke test step

Abacn:
R: @derrickaw

Issue #3224, opened 2026-01-15
[Bug]: Smoke tests misconfigured


Abacn:
### Related Template(s)

N/A

### Template Version

N/A

### What happened?

smoke tests never run certain test suites. It was configured here:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ff9870f0f89981531ab84c3dddc71087aabbdfbc/pom.xml#L795

This is an invalid expression. In surefire-plugin 2.x, it results in only `${direct-runner.tests}` run. In surefire-plugin 3.x, it fails with invalid expression.

After fixing expression, it is found IcebergResourceManagerIT fails:

https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/21044330452/job/60516308539#logs

### Relevant log output

```shell

```

Abacn:
Likely due to the static member variables setup in `Junit.@Before` causing conflicts when tests run in parallel:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/ff9870f0f89981531ab84c3dddc71087aabbdfbc/it/iceberg/src/test/java/com/google/cloud/teleport/it/iceberg/IcebergResourceManagerIT.java#L54

Abacn:
Move to P3 for tracking the TODO in IcebergResourceManagerIT and re-enable the test after it's fixed.

tarun-google:
i think its due to not randomizing the warehouse directory. All tests might be using the same directory. I will prepare a fix. I will also remove static for mutable variables.

PR #3226, opened 2026-01-16
Fix IcebergResourceManagerIT by randomizing warehouse

tarun-google:
Fix https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/3224


tarun-google:
Failed test is not relevant. We can take this live

PR #3227, opened 2026-01-16
add yaml manifest file

derrickaw:
1. add yaml manifest file
2. group gson dependency and update to non-vulnerability of 2.13.2

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
@@ -302,25 +302,22 @@ public class TemplatesReleaseMojo extends TemplatesBaseMojo {
                           String.join("/", stagePrefix, yamlBlueprintsGCSPath, fileName);
                       BlobId blobId = BlobId.of(bucketNameOnly(bucketName), objectName);
 
-                      // Process only new files that don't exist in GCS already
-                      if (storage.get(blobId) == null) {
-                        BlobInfo blobInfo = BlobInfo.newBuilder(blobId).build();
-                        try (InputStream inputStream = Files.newInputStream(path)) {
-                          storage.create(blobInfo, inputStream);
-                          LOG.info(
-                              "Uploaded blueprint {} to gs://{}/{}",
-                              fileName,
-                              bucketNameOnly(bucketName),
-                              objectName);
-                        } catch (IOException e) {
-                          throw new RuntimeException("Error reading file " + fileName, e);
-                        }
-                      } else {
-                        LOG.info(
-                            "Skipping existing blueprint: gs://{}/{}",
-                            bucketNameOnly(bucketName),
-                            objectName);
-                      }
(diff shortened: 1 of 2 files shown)


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
```diff
@@ -288,10 +306,23 @@ public void execute() throws MojoExecutionException {
...
+                        blueprints.add(new Blueprint(fileName, objectName));
                       } catch (IOException e) {
                         throw new RuntimeException("Error reading file " + fileName, e);
                       }
                     });
+            String manifestObjectName =
+                String.join("/", stagePrefix, yamlBlueprintsGCSPath, yamlManifestName);
+            BlobId manifestBlobId = BlobId.of(bucketNameOnly(bucketName), manifestObjectName);
+            BlobInfo manifestBlobInfo = BlobInfo.newBuilder(manifestBlobId).build();
+            storage.create(
```
chamikaramj: Should there be any retrying here ?

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
```diff
@@ -288,10 +306,23 @@ public void execute() throws MojoExecutionException {
...
+                manifestBlobInfo,
```
chamikaramj: What if some of the blueprints were exported previously ? Should we skip those and try to rest ?

derrickaw: added a skip, thanks

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
derrickaw: added a retry, thanks

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
chamikaramj: Thanks. Another question though. What if we actually want to update an existing blueprint ?

I think we should always update blueprints in GCS with the source of truth in Github when this script is executed.

May be instead of skipping, we should just overwrite every blueprint in retries so that retries do not fail. If this is the default behavior for the GCS copy operation you are doing we are good and can just drop the skipping logic.

derrickaw: Removed the skip logic and added blueprint retry logic regardless if there or not already. Thanks.

derrickaw:
R: @chamikaramj

chamikaramj:
LGTM. Thanks!

chamikaramj (approved):

PR #3233, opened 2026-01-17
clean up some repeat properties in submodule poms

derrickaw:
1. Some submodule poms have duplicate properties, so removing those and moving to parent pom.
2. Some properties in submodule poms are already covered by the parent pom, so removing those.


Abacn (approved):
Thanks!

PR #3237, opened 2026-01-20
add batch write flow control support

mutianf:
Add an option to enable bulk write flow control. Default to false. Also fix incorrect settings in the template for write options.


Abacn:
PubSubCdcToBigQueryIT failed not related to this change. Let me merge it for now.

PR #3242, opened 2026-01-22
Remove notthreadsafe causing test running sequentially

Abacn:
The annotation was added in #1275. The race was later fixed in https://github.com/apache/beam/pull/33979. We can remove the annotation for now.

Fixes Release workflow timeout


Abacn:
Test passed: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/21255311640

Abacn:
R: @derrickaw

Issue #3243, opened 2026-01-22
[Feature Request]: Support UDFs in the PubsubToRedis template.


tvalentyn:
### Related Template(s)

pubsub-to-redis

### What feature(s) are you requesting?

The template code extends the JavascriptTextTransformer interface, but doesn't yet implement the UDF support in the templates code:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/63091f159e1889e7234988c96a2515498bf9d46e/v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java#L133

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/63091f159e1889e7234988c96a2515498bf9d46e/v1/src/main/java/com/google/cloud/teleport/templates/common/JavascriptTextTransformer.java#L74

We can see how other templates implement UDF support when using this interface and add such support and tests.

PR #3245, opened 2026-01-22
Removes mentioning UDF support in PubsubToRedis, since this has not yet been implemented.

tvalentyn:
Pubsub-To-Redis template does not implement custom UDF support yet, even though it previously extended the `JavascriptTextTransformerOptions` interface.

This change removes the unsupported interface and regenerates the docs.

FR to add support: https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/3243

### DIFF
--- a/v2/pubsub-to-redis/README_Cloud_PubSub_to_Redis.md
+++ b/v2/pubsub-to-redis/README_Cloud_PubSub_to_Redis.md
@@ -6,10 +6,6 @@ Pub/Sub subscription and writes the message payload to Redis. The most common us
 case of this template is to export logs to Redis Enterprise for advanced
 search-based log analysis in real time.
 
-Before writing to Redis, you can apply a JavaScript user-defined function to the
-message payload. Any messages that experience processing failures are forwarded
-to a Pub/Sub unprocessed topic for further troubleshooting and reprocessing.
-
 For added security, enable an SSL connection when setting up your database
 endpoint connection.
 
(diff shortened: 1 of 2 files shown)


v2/pubsub-to-redis/README_Cloud_PubSub_to_Redis.md
```diff
@@ -37,20 +33,6 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 * **redisSinkType**: The Redis sink. Supported values are `STRING_SINK, HASH_SINK, STREAMS_SINK, and LOGGING_SINK`. For example, `STRING_SINK`. Defaults to: STRING_SINK.
 * **connectionTimeout**: The Redis connection timeout in milliseconds.  For example, `2000`. Defaults to: 2000.
 * **ttl**: The key expiration time in seconds. The `ttl` default for `HASH_SINK` is -1, which means it never expires.
-* **javascriptTextTransformGcsPath**: The Cloud Storage URI of the .js file that defines the JavaScript user-defined function (UDF) to use. For example, `gs://my-bucket/my-udfs/my_file.js`.
```
tvalentyn, line 40: @damccorm should we go ahead with changes to this file in light of line 23 or you think there is a different way to change this? Looks like this content is also in the source code for https://docs.cloud.google.com/dataflow/docs/guides/templates/provided/pubsub-to-redis .

damccorm: This file is autogenerated, so we can't change it correctly. Can we just remove the Javascript options here - https://github.com/tvalentyn/DataflowTemplates/blob/7573c22421ac390fccd269d8c1c1ec0fe268fa33/v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java#L129 - and then regenerate the docs with https://github.com/tvalentyn/DataflowTemplates/blob/main/contributor-docs/code-contributions.md#generated-documentation ?

Technically its breaking, but only if you were relying on functionality that doesn't work anyways

tvalentyn: got it - done, thanks!

PR #3248, opened 2026-01-22
update yaml path for uploading blueprints etc

derrickaw:
helps b/477930132

### DIFF
--- a/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
+++ b/plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
@@ -158,7 +158,7 @@ public class TemplatesReleaseMojo extends TemplatesBaseMojo {
   protected boolean publishYamlBlueprints;
 
   @Parameter(
-      defaultValue = "yaml/src/main/yaml",
+      defaultValue = "src/main/yaml",
       property = "yamlBlueprintsPath",
       readonly = true,
       required = false)
--- a/plugins/templates-maven-plugin/src/test/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojoTest.java
+++ b/plugins/templates-maven-plugin/src/test/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojoTest.java
@@ -63,7 +63,7 @@ public class TemplatesReleaseMojoTest {
   public void setUp() throws IOException {
     mojo = new TemplatesReleaseMojo();
     mavenProject = mock(MavenProject.class);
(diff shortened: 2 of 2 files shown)


plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesReleaseMojo.java
```diff
@@ -263,7 +263,7 @@ public void execute() throws MojoExecutionException {
         LOG.info(
             "Trying to upload Job Builder blueprints to bucket '{}'...",
             bucketNameOnly(bucketName));
-        Path yamlPath = Paths.get(project.getBasedir().getAbsolutePath(), yamlBlueprintsPath);
+        Path yamlPath = Paths.get(project.getBasedir().getParent(), yamlBlueprintsPath);
```
Abacn: the original problem was yaml blueprint was attempted for unrelated modules (v1). After the change it would still be the case.

I think the problem is that TemplatesReleaseMojo will get executed for every module, and this line will result in invalid path for all but  `yaml` module

> Path yamlPath = Paths.get(project.getBasedir().getAbsolutePath(), "yaml/src/main/yaml");

I think we can keep this line, and just skip if yamlPath does not exist

derrickaw: Done, thanks

Abacn: Still, it seems defaultValue needs to be "src/main/yaml". Would be nice if we can confirm this via local test

derrickaw: I think the tests cover that part.  It was being within a different submodule that was throwing it off.   I updated the exception test to expect a warning now if the path doesn't exist. Thoughts? Thanks

Abacn: Tests are good, thanks.

I mean, we should probably change default value of  `yamlBlueprintsPath` to "src/main/yaml", and keep `project.getBasedir().getAbsolutePath()`. Because "src/main/yaml" is the relative path to the module root

derrickaw: gotcha

derrickaw: updated, thanks

PR #3249, opened 2026-01-23
[DataStreamToSpanner] Improvement: Adding group by key step before Spanner write step

darshan-sj:
Improvement: Adding group by key step before Spanner write step

This would serialize events happening on same key in Dataflow job and hence avoid lock contention in Spanner when there are too many events corresponding to a small set of keys.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
@@ -101,16 +101,16 @@ public class SpannerTransactionWriter
   @Override
   public SpannerTransactionWriter.Result expand(
       PCollection<FailsafeElement<String, String>> input) {
-    PCollectionTuple groupByResults =
+    PCollectionTuple keyedEvents =
         input.apply(
-            "Produce Key value pairs with key as Primary Key Hash",
+            "Key By PK Hash",
             ParDo.of(new CreateKeyValuePairsWithPrimaryKeyHashDoFn(ddlView))
                 .withSideInputs(ddlView)
                 .withOutputTags(
                     DatastreamToSpannerConstants.SUCCESSFUL_KEYED_EVENT_TAG,
                     TupleTagList.of(List.of(DatastreamToSpannerConstants.PERMANENT_ERROR_TAG))));
     PCollectionTuple spannerWriteResults =
-        groupByResults
+        keyedEvents
             .get(DatastreamToSpannerConstants.SUCCESSFUL_KEYED_EVENT_TAG)
             .setCoder(
                 KvCoder.of(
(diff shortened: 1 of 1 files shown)


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/CreateKeyValuePairsWithPrimaryKeyHashDoFn.java
```diff
@@ -0,0 +1,84 @@
...
+      JsonNode changeEvent = mapper.readTree(msg.getOriginalPayload());
+      Ddl ddl = c.sideInput(ddlView);
+
+      String tableName = changeEvent.get(DatastreamConstants.EVENT_TABLE_NAME_KEY).asText();
+      com.google.cloud.spanner.Key primaryKey =
+          ChangeEventSpannerConvertor.changeEventToPrimaryKey(
+              tableName, ddl, changeEvent, /* convertNameToLowerCase= */ true);
+      String finalKeyString = tableName + "_" + primaryKey.toString();
+      Long finalKey = (long) finalKeyString.hashCode();
+      c.output(KV.of(finalKey, msg));
```
VardhanThigle, line 79: Should we hash the finalyKeyString and make the hash as the index of the KV<>?

As in should we make it `KV<Long, Message>`?

This would also help us control number of buckets. 
`KV<Long, Message>` is a more common pattern I have seen in dataflow templates.

darshan-sj: Yes. I have made it KV<Long, Message> only. I'm using finalKeyString.hashCode() and taking the hash value as long for key.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
```diff
@@ -93,29 +101,53 @@ public SpannerTransactionWriter(
...
-                    new SpannerTransactionWriterDoFn(
-                        spannerConfig,
-                        shadowTableSpannerConfig,
-                        ddlView,
-                        shadowTableDdlView,
-                        shadowTablePrefix,
-                        sourceType,
-                        isRegularRunMode))
-                .withSideInputs(ddlView, shadowTableDdlView)
+            "Produce Key value pairs with key as Primary Key Hash",
```
bharadwaj-aditya: please use a smaller name. This won't show up on the DF graph properly.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
```diff
@@ -93,29 +101,53 @@ public SpannerTransactionWriter(
   @Override
   public SpannerTransactionWriter.Result expand(
       PCollection<FailsafeElement<String, String>> input) {
-    PCollectionTuple spannerWriteResults =
+    PCollectionTuple groupByResults =
```
bharadwaj-aditya: naming nit: It is a reshuffle to ensure we process by key, there is no group by  in this operation. Something like Kv write results or similar is a more appropriate name.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
darshan-sj: Done.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriter.java
darshan-sj: Done.

bharadwaj-aditya (approved):
LGTM

PR #3255, opened 2026-01-26
feat(Spanner): Live Migration: Remove generated columns from change events during transformation and add tests for this behavior

pratickchokhani:
If a primary is Generated Column within Spanner and the change stream event is DELETE, then DML will be used
Else Mutation is used.

For DML, we include all the non-generated columns in the where clause. This is so because we do not have the list of dependent columns.

If column is Generated Column in Spanner, then the column is omitted from both DML or Mutations.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.Options;
 import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TransactionRunner;
 import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.ChangeEventConvertorException;
@@ -279,7 +280,7 @@ class SpannerTransactionWriterDoFn
             c, changeEventContext, currentChangeEventSequence, shadowTableDdl, ddl);
       } else {
         processSingleDatabaseTransaction(
-            c, changeEventContext, currentChangeEventSequence, shadowTableDdl);
+            c, changeEventContext, currentChangeEventSequence, shadowTableDdl, ddl);
       }
       com.google.cloud.Timestamp timestamp = com.google.cloud.Timestamp.now();
       c.output(timestamp);
@@ -378,7 +379,8 @@ class SpannerTransactionWriterDoFn
       ProcessContext c,
       ChangeEventContext changeEventContext,
       ChangeEventSequence currentChangeEventSequence,
-      Ddl shadowDdl) {
+      Ddl shadowDdl,
+      Ddl ddl) {
 
     spannerAccessor
         .getDatabaseClient()
@@ -403,7 +405,14 @@ class SpannerTransactionWriterDoFn
                     skippedEvents.inc();
                     return null;
                   }
-                  // Apply shadow and data table mutations.
+                  // Execute DML if applicable
+                  Statement dataDml = changeEventContext.getDataDmlStatement(ddl);
+
+                  if (dataDml != null) {
+                    transaction.executeUpdate(dataDml);
+                  }
+
+                  // Apply shadow and data table mutations (only if they exist)
                   transaction.buffer(changeEventContext.getMutations());
                   isInTransaction.set(false);
                   return null;
@@ -504,8 +513,17 @@ class SpannerTransactionWriterDoFn
                                       "Shadow table sequence changed during transaction");
                                 }
 
+                                // Execute Data DML if applicable
+                                Statement dataDml =
+                                    changeEventContext.getDataDmlStatement(dataTableDdl);
+                                if (dataDml != null) {
+                                  mainTxn.executeUpdate(dataDml);
+                                }
+
                                 // Write to main table
-                                mainTxn.buffer(changeEventContext.getDataMutation());
+                                if (changeEventContext.getDataMutation() != null) {
+                                  mainTxn.buffer(changeEventContext.getDataMutation());
+                                }
                                 return null;
                               });
 
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
@@ -21,16 +21,19 @@ import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TransactionContext;
+import com.google.cloud.spanner.Value;
 import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.ddl.IndexColumn;
 import com.google.cloud.teleport.v2.spanner.ddl.Table;
 import com.google.cloud.teleport.v2.spanner.migrations.convertors.ChangeEventSpannerConvertor;
+import com.google.cloud.teleport.v2.spanner.migrations.convertors.ChangeEventTypeConvertor;
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.ChangeEventConvertorException;
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.DroppedTableException;
 import com.google.cloud.teleport.v2.spanner.migrations.exceptions.InvalidChangeEventException;
 import com.google.cloud.teleport.v2.spanner.migrations.spanner.SpannerReadUtils;
 import com.google.cloud.teleport.v2.templates.spanner.ShadowTableCreator;
 import com.google.common.collect.ImmutableMap;
-import java.util.Arrays;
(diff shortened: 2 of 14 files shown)


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/spanner/ShadowTableCreator.java
```diff
@@ -70,6 +70,9 @@ Table constructShadowTable(Ddl informationSchema, String dataTableName, Dialect
             .filter(col -> primaryKeyColNames.contains(col.name()))
             .collect(Collectors.toList());
     for (Column col : primaryKeyCols) {
+      // Generated expression should be removed from the shadow table columns.
```
shreyakhajanchi: // This is requried as shadow table only contains primary key columns. - this comment is unclear. Can you explain a bit more?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -107,23 +110,93 @@ protected void convertChangeEventToMutation(Ddl ddl, Ddl shadowTableDdl)
...
+  private boolean hasGeneratedPK(Table table) {
+    for (IndexColumn keyColumn : table.primaryKeys()) {
+      if (table.column(keyColumn.name()).isGenerated()) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private Statement generateDeleteDml(Table table, String tableName, JsonNode event)
```
shreyakhajanchi, line 162: Will this work for PostgreSQL dialect as well? I think it might break for PostgreSQL dialect

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -107,23 +110,93 @@ protected void convertChangeEventToMutation(Ddl ddl, Ddl shadowTableDdl)
...
     this.primaryKey =
         ChangeEventSpannerConvertor.changeEventToPrimaryKey(
             changeEvent.get(DatastreamConstants.EVENT_TABLE_NAME_KEY).asText(),
             ddl,
             changeEvent,
             /* convertNameToLowerCase= */ true);
-    this.dataMutation = ChangeEventConvertor.changeEventToMutation(ddl, changeEvent);
+
+    String changeType = getChangeType(changeEvent);
+    boolean isDelete = DatastreamConstants.DELETE_EVENT.equalsIgnoreCase(changeType);
```
shreyakhajanchi, line 127: Datastream also has `UPDATE-DELETE` do we need to handle that as well?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
```diff
@@ -473,6 +482,13 @@ void processCrossDatabaseTransaction(
                       .run(
                           (TransactionRunner.TransactionCallable<Void>)
                               mainTxn -> {
+                                // Execute Data DML if applicable
+                                Statement dataDml =
+                                    changeEventContext.getDataDmlStatement(dataTableDdl);
+                                if (dataDml != null) {
+                                  mainTxn.executeUpdate(dataDml);
+                                }
+
```
shreyakhajanchi: should we not do this after acquiring lock and validating sequence, at the same place we write mutations as an `else` clause?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/transform/ChangeEventTransformerDoFnTest.java
```diff
@@ -758,4 +758,62 @@ public void testProcessElementWithSpannerMutationAndCustomTransformation() throw
...
+        ChangeEventTransformerDoFn.create(
+            schema, null, null, null, "mysql", customTransformation, false, ddlView, spannerConfig);
+    changeEventTransformerDoFn.setMapper(mapper);
+    changeEventTransformerDoFn.setSpannerAccessor(spannerAccessor);
+    changeEventTransformerDoFn.setChangeEventSessionConvertor(changeEventSessionConvertor);
+    changeEventTransformerDoFn.processElement(processContextMock);
+
+    verify(changeEventSessionConvertor, times(1))
+        .transformChangeEventData(changeEvent, databaseClientMock, ddl);
+  }
```
shreyakhajanchi, line 825: is the assert missing here?

v2/datastream-to-spanner/src/test/resources/MySQLDataTypesIT/mysql-generated-col.sql
```diff
@@ -0,0 +1,16 @@
+DELETE FROM `generated_pk_column_table` WHERE `first_name_col` = "AA";
```
shreyakhajanchi, line 1: why do we have this separate file, can these statements not be a part of mysql-data-types.sql like other DMLs are?

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToSpannerDataTypesAndExpressionIT.java
```diff
@@ -189,20 +192,36 @@ public void testMySqlDataTypes() throws Exception {
     assertThatResult(result).meetsConditions();
 
     validateResult(spannerResourceManager, expectedData);
+
+    LOG.info("Executing MySQL DML script...");
```
shreyakhajanchi, line 197: It might be better to create a new IT for generated columns or atleast rename this IT to be clearer.
@darshan-sj thoughts?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventTypeConvertor.java
```diff
@@ -49,6 +51,55 @@ public class ChangeEventTypeConvertor {
...
   private static final Pattern NUMERIC_PATTERN = Pattern.compile("-?\\d+(\\.\\d+)?");
 
+  public static Value toValue(
+      JsonNode changeEvent, Type columnType, String key, boolean requiredField)
+      throws ChangeEventConvertorException {
+
+    switch (columnType.getCode()) {
+      case BOOL:
+      case PG_BOOL:
+        return Value.bool(ChangeEventTypeConvertor.toBoolean(changeEvent, key, requiredField));
```
shreyakhajanchi: since the methods are local now, maybe call `toBoolean` directly rather than `ChangeEventTypeConvertor.toBoolean`

shreyakhajanchi (commented):
Added some comments. Please add details of manual test run in the description for completeness.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
pratickchokhani: Have validated this and have also added an Integration test where the primary key is updated.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventTypeConvertor.java
pratickchokhani: Done.

v2/datastream-to-spanner/src/test/resources/MySQLDataTypesIT/mysql-generated-col.sql
pratickchokhani: The first sql script is executed and then the data is asserted. 
After that this script is executed to validate updates.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java
pratickchokhani: Done.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
pratickchokhani: Have added a TODO for PostgreSQL dialect.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/spanner/ShadowTableCreator.java
pratickchokhani: Have updated the comment.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/transform/ChangeEventTransformerDoFnTest.java
pratickchokhani: Have added the assert.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToSpannerDataTypesAndExpressionIT.java
pratickchokhani: Have renamed the test.

PR #3262, opened 2026-01-27
Add support for UDF function transformations in PubSubToRedis template

wladabutenko:
Adding optional UDF support to the PubSubToRedis pipeline, enabling user-defined transformations before writing to Redis

fixes: #3243

### DIFF
--- a/v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
+++ b/v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
@@ -24,25 +24,37 @@ import com.google.cloud.teleport.metadata.Template;
 import com.google.cloud.teleport.metadata.TemplateCategory;
 import com.google.cloud.teleport.metadata.TemplateParameter;
 import com.google.cloud.teleport.metadata.TemplateParameter.TemplateEnumOption;
+import com.google.cloud.teleport.v2.coders.FailsafeElementCoder;
 import com.google.cloud.teleport.v2.common.UncaughtExceptionLogger;
 import com.google.cloud.teleport.v2.templates.io.RedisHashIO;
 import com.google.cloud.teleport.v2.templates.transforms.MessageTransformation;
+import com.google.cloud.teleport.v2.transforms.FailsafeElementTransforms.ConvertFailsafeElementToPubsubMessage;
+import com.google.cloud.teleport.v2.transforms.JavascriptTextTransformer;
+import com.google.cloud.teleport.v2.values.FailsafeElement;
+import com.google.common.base.Strings;
 import java.util.Map;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.PipelineResult;
+import org.apache.beam.sdk.coders.CoderRegistry;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubIO;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder;
 import org.apache.beam.sdk.io.redis.RedisConnectionConfiguration;
 import org.apache.beam.sdk.io.redis.RedisIO;
 import org.apache.beam.sdk.options.Default;
-import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.options.Validation;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.TupleTag;
 import org.apache.beam.sdk.values.TypeDescriptors;
+import org.apache.commons.lang3.ArrayUtils;
 import org.checkerframework.checker.initialization.qual.Initialized;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.UnknownKeyFor;
@@ -93,6 +105,9 @@ import org.slf4j.LoggerFactory;
       "The Pub/Sub to Redis template is a streaming pipeline that reads messages from a Pub/Sub subscription and "
           + "writes the message payload to Redis. The most common use case of this template is to export logs to Redis "
           + "Enterprise for advanced search-based log analysis in real time.",
+      "Before writing to Redis, you can apply a JavaScript user-defined function to the message payload. Any "
+          + "messages that experience processing failures are forwarded to a Pub/Sub unprocessed topic for further "
+          + "troubleshooting and reprocessing.",
       "For added security, enable an SSL connection when setting up your database endpoint connection."
     },
     optionsClass = PubSubToRedis.PubSubToRedisOptions.class,
@@ -102,6 +117,7 @@ import org.slf4j.LoggerFactory;
         "https://cloud.google.com/dataflow/docs/guides/templates/provided/pubsub-to-redis",
     requirements = {
       "The source Pub/Sub subscription must exist prior to running the pipeline.",
+      "The Pub/Sub unprocessed topic must exist prior to running the pipeline.",
       "The Redis database endpoint must be accessible from the Dataflow workers' subnetwork.",
     },
     preview = true,
@@ -117,13 +133,26 @@ public class PubSubToRedis {
   /** The log to output status messages to. */
   private static final Logger LOG = LoggerFactory.getLogger(PubSubToRedis.class);
 
+  /** The tag for the main output for the UDF. */
+  public static final TupleTag<FailsafeElement<PubsubMessage, String>> UDF_OUT =
+      new TupleTag<FailsafeElement<PubsubMessage, String>>() {};
+
+  /** The tag for the dead-letter output of the udf. */
+  public static final TupleTag<FailsafeElement<PubsubMessage, String>> UDF_DEADLETTER_OUT =
+      new TupleTag<FailsafeElement<PubsubMessage, String>>() {};
+
+  /** Pubsub message/string coder for FailsafeElement. */
+  public static final FailsafeElementCoder<PubsubMessage, String> FAILSAFE_ELEMENT_CODER =
+      FailsafeElementCoder.of(PubsubMessageWithAttributesCoder.of(), StringUtf8Coder.of());
+
   /**
    * The {@link PubSubToRedisOptions} class provides the custom execution options passed by the
    * executor at the command-line.
    *
    * <p>Inherits standard configuration options.
    */
-  public interface PubSubToRedisOptions extends PipelineOptions {
+  public interface PubSubToRedisOptions
+      extends JavascriptTextTransformer.JavascriptTextTransformerOptions {
     @TemplateParameter.PubsubSubscription(
         order = 1,
         groupName = "Source",
@@ -224,6 +253,17 @@ public class PubSubToRedis {
     Long getTtl();
 
     void setTtl(Long ttl);
+
+    @TemplateParameter.PubsubTopic(
+        order = 9,
+        description = "Output deadletter Pub/Sub topic",
+        helpText =
+            "The Pub/Sub topic to forward unprocessable messages to. Messages that fail UDF transformation or cannot be written to Redis are forwarded here.",
+        example = "projects/<PROJECT_ID>/topics/<TOPIC_NAME>")
+    @Validation.Required
+    String getDeadletterTopic();
+
+    void setDeadletterTopic(String deadletterTopic);
   }
 
   /** Allowed list of sink types. */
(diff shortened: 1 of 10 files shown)


tarun-google:
LGTM, Thanks for adding this missing feature!

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -269,73 +325,427 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+    }
+  }
+
+  /**
+   * The {@link PubSubMessageTransform} transform processes PubSubMessages using an optional UDF.
+   *
+   * <p>If a UDF is provided, it will be applied to the message payload. Messages that fail UDF
+   * processing will be sent to the dead-letter output.
+   */
+  public static class PubSubMessageTransform
```
tvalentyn: I'd prefer to keep the UDF implementation logic similar to the logic we have in other templates. I am seeing that we are using an existing shared construct FailsafeJavascriptUdf - is this something we can reuse here as well? See, for example:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/c9b2238d852f15e0039ce1c761498ff4699204bd/v2/pubsub-binary-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/PubsubProtoToBigQuery.java#L431 ?

tvalentyn: Similarly for other portions of the change, I'd expect we follow some pattern already established in other templates, for consistency and to simplify the review. If we are to follow a different pattern, let's call out the differences and reasons.

wladabutenko: Done. 
Integration Tests are following same pattern as [https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/pubsub-to-jms/src/test/java/com/google/cloud/teleport/v2/templates/PubsubToJmsIT.java](url) template, as I couldn't find Redis Resource Manager

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1469,13 +1469,13 @@ private static void copyJavaArtifacts(
...
   private void performVulnerabilityScanAndGenerateUserSBOM(
-      String imagePathTag, String buildProjectId, File buildDir, TemplateType imageType)
+      String imagePathTag, String buildProjectId, File buildDir, Template.TemplateType imageType)
       throws IOException, InterruptedException {
     LOG.info("Generating user SBOM and Performing security scan for {}...", imagePathTag);
 
     // Continuous scanning is expensive. Images are built on identical dependencies and only differ
     // by entry point. We only need to check once.
-    ImmutablePair<String, TemplateType> uniqueImage =
+    ImmutablePair<String, Template.TemplateType> uniqueImage =
```
tvalentyn: Why are we changing this file?

wladabutenko: I was unable to stage template due to java.lang.NoClassDefFoundError Caused by: java.lang.ClassNotFoundException: TemplateType after pulling the latest changes. I can rollback this file to previous version and rebuild template one more time

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -271,24 +316,26 @@ public static PipelineResult run(PubSubToRedisOptions options) {
 
     /*
      * Steps: 1) Read PubSubMessage with attributes and messageId from input PubSub subscription.
-     *        2) Extract PubSubMessage message to PCollection<String>.
-     *        3) Transform PCollection<String> to PCollection<KV<String, String>> so it can be consumed by RedisIO
-     *        4) Write to Redis using SET
+     *        2) Apply JavaScript UDF transformation to message payload (if configured).
```
tvalentyn, line 317: are the steps 2-4 on the left hand side no longer accurate for some reason?

wladabutenko: True, I will update those

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
tvalentyn: I am still not following the plan here - why are we adding a new PubSubMessage transform?

v2/pubsub-to-redis/src/test/java/com/google/cloud/teleport/v2/templates/PubSubToRedisIT.java
```diff
@@ -0,0 +1,288 @@
+/*
+ * Copyright (C) 2023 Google LLC
```
tvalentyn, line 2: Have you run this test?

v2/pubsub-to-redis/terraform/Cloud_PubSub_to_Redis/dataflow_job.tf
```diff
@@ -34,231 +34,238 @@ variable "region" {
 }
 
 variable "inputSubscription" {
-  type        = string
-  description = "Pub/Sub subscription to read the input from, in the format of 'projects/your-project-id/subscriptions/your-subscription-name' (Example: projects/your-project-id/subscriptions/your-subscription-name)"
-
+  type = string
```
tvalentyn, line 37: for my information, were changes to this file autogenerated?

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
wladabutenko: Similarly to ‎PubsubProtoToBigQuery::RunUdf class [https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/pubsub-binary-to-bigquery/src/main/java/com/google/cloud/teleport/v2/templates/PubsubProtoToBigQuery.java](url), return a PCollectionTuple which allows return two streams from single transform, success and failure, failed data moving to the dead-letter topic. 
if no UDF is configured, handles the raw messages logic allowing them to go through without modification, as Redis can accept raw strings.

v2/pubsub-to-redis/terraform/Cloud_PubSub_to_Redis/dataflow_job.tf
wladabutenko: Yes

v2/pubsub-to-redis/src/test/java/com/google/cloud/teleport/v2/templates/PubSubToRedisIT.java
wladabutenko: Yes

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
wladabutenko: Actually, steps 2-4 describes only STRING_SINK scenario, excluding HASH_SINK, LOGGING_SINK and STREAMS_SINK. I would left existing one as it's more common

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +385,145 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+
+    @Override
+    public PCollectionTuple expand(PCollection<PubsubMessage> input) {
+      // Map incoming messages to FailsafeElement so we can recover from failures
+      // across multiple transforms.
+      PCollection<FailsafeElement<PubsubMessage, String>> failsafeElements =
+          input.apply("MapToRecord", ParDo.of(new PubsubMessageToFailsafeElementFn()));
+
+      // Apply UDF if provided
+      if (javascriptTextTransformGcsPath != null) {
```
tvalentyn: I would suggest to  fold the PubSubMessageTransform into applyUdf; the core logic in PubSubMessageTransform still deals with the applying the UDF, and I feel like extracting this into a separate transform just compicates the control flow; for example, we should be able to only check once whether we are going with the udf flow or no udf flow around line 407, instead of having to repeat
```

       // Apply UDF if provided
      if (javascriptTextTransformGcsPath != null) {

```

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
tvalentyn: yes, your new comments look more accurate.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +385,145 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+    if (!useJavascriptUdf) {
+      return messages;
+    }
+
+    PCollectionTuple udfResult =
+        messages.apply(
+            "Apply UDF",
+            new PubSubMessageTransform(
+                options.getJavascriptTextTransformGcsPath(),
+                options.getJavascriptTextTransformFunctionName(),
```
tvalentyn: i am seeing some other templates have validation of these params, like:

```
if (useJavascriptUdf && Strings.isNullOrEmpty(options.getJavascriptTextTransformFunctionName())) {
          "JavaScript function name cannot be null or empty if file is set");
```

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
tvalentyn: yeah, it wasn't immediately obvious why this is needed and it is causing merge conflicts now

wladabutenko: Resolved.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
wladabutenko: Added.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
wladabutenko: Done.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -392,8 +390,7 @@ public static PipelineResult run(PubSubToRedisOptions options) {
    *
    * <p>If no UDF is configured, returns the input messages unchanged.
    *
-   * <p>This follows the same pattern as {@code PubsubProtoToBigQuery.runUdf} and {@code
-   * PubSubToMongoDB.PubSubMessageToJsonDocument}.
+   * <p>This follows the same pattern as {@code PubsubProtoToBigQuery.runUdf}.
```
tvalentyn, line 393: Let's make this  comment in the code, not part of the docstring.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/io/RedisHashIO.java
```diff
@@ -28,41 +27,43 @@
...
 import redis.clients.jedis.Transaction;
 
 public abstract class RedisHashIO {
 
   public static WriteHash write() {
-
-    return (new AutoValue_RedisHashIO_WriteHash.Builder())
-        .setConnectionConfiguration(RedisConnectionConfiguration.create())
-        .build();
+    return new WriteHash(RedisConnectionConfiguration.create(), null);
```
tvalentyn, line 37: Why are we changing this file?

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +383,108 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+    if (!useJavascriptUdf) {
+      return messages;
+    }
+
+    if (Strings.isNullOrEmpty(options.getJavascriptTextTransformFunctionName())) {
+      throw new IllegalArgumentException(
+          "JavaScript function name cannot be null or empty if file is set");
+    }
+
+    PCollectionTuple udfResult = performUdf(messages, options);
```
tvalentyn: do we really need this helper? as a reader of this code, i'd be puzzled why we have both 'applyUdf' and 'performUdf'.  `performUdf` is two lines of code that belongs to the same level of abstraction, so i'd argue this helper increases the complexity here rather than reduces it.

v2/pubsub-to-redis/terraform/Cloud_PubSub_to_Redis/dataflow_job.tf
tvalentyn: thanks, for my information/for future reference, what command did you use to regenerate this?

tvalentyn:
please also check spotless errors.

tvalentyn:
continued in  https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3338

PR #3263, opened 2026-01-27
add addition label to release label workflow

derrickaw:
1. Existing release label workflow considers 5 different labels with one being `improvement`.  To me `improvement` would be about improving on existing code to make it faster, better logic, etc.
2. There is another label `addition` that seems to be geared more to new features and such, so adding that label to the release label check.
3. Also, making the labels in alphanumeric order.
4. If reviewer thinks we should not do this, should we just delete the addition label to minimize confusion with the release label workflow? Thanks.

### DIFF
--- a/.github/workflows/release-notes-labels.yml
+++ b/.github/workflows/release-notes-labels.yml
@@ -13,6 +13,8 @@ jobs:
         with:
           mode: minimum
           count: 1
+          # Note: If updating these labels, also update .github/release.yml 
+          # and contributor-docs/code-contributions.md.
           labels: |
             addition
             bug-fix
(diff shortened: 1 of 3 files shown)


.github/workflows/release-notes-labels.yml
```diff
@@ -14,9 +14,10 @@ jobs:
           mode: minimum
           count: 1
           labels: |
+            addition
+            bug-fix
             ignore-for-release
-            new-template
             improvement
-            bug-fix
+            new-template
```
damccorm, line 23: Please update https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/cb18ecca6551d4ca9029947f3b14382c35e0450e/.github/release.yml#L8 and https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/cb18ecca6551d4ca9029947f3b14382c35e0450e/contributor-docs/code-contributions.md?plain=1#L492 as well (and maybe add a comment to this file pointing to those locations)

PR #3268, opened 2026-01-28
Yaml templates sharing container image

Abacn:
Part of #2279

PR #3285, opened 2026-01-31
Fix yaml templates setting template sdk language == yaml

Abacn:
#3268 is supposed to making yaml templates build container once for all templates under yaml/. It was ineffective due to an early "return" leftover, making the PR effectively a no-op for yaml templates so far.

After removing the line, all tests but PubsubToBigtable passed. It is found this is due to a Beam bug gets triggered:

https://github.com/apache/beam/issues/37475

`--jinja_variable_flags=...,projectId,...` options passed to beam yaml.main to filter out `--project` pipeline option, causing

```
ValueError: Pipeline has validations errors: Missing required option: project.
```

However this currently does not occur because the pipeline json has always setting

```"sdkInfo": {
    "language": "Python"
  }
```

for Yaml templates. This json was generated by `gcloud dataflow flex-templates build` command. After this change, it is generated by templates-maven-plugin, same as Java/Xlang templates.

But this would cause Dataflow backend to add `--jinja_variable_flags` flag to template launcher, thus triggering the Beam bug.

This PR then
- Removed the line that prevented #3268 being effective, and

- Workaround for https://github.com/apache/beam/issues/37475 such that dropping `jinja_variable_flags` at the moment, recover the behavior prior to the first change

### DIFF
--- a/yaml/src/main/python/main.py
+++ b/yaml/src/main/python/main.py
@@ -53,7 +53,7 @@ def _clean_undefined_recursively(d):
 
 
 def _get_template_name(argv):
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(allow_abbrev=False)
     parser.add_argument('--label', '--labels', dest='labels', action='append', default=None)
     arg_label, _ = parser.parse_known_args(argv)
     template_name = None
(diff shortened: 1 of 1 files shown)


Abacn:
Yaml test failure due to https://github.com/apache/beam/issues/37475

update:

what happens is that after #3268, the overriden template json has set

```
"sdkInfo": {
    "language": "YAML"
  }
```

while the template json generated by `gcloud dataflow flex-template build` set to `language: Python`. Setting to yaml makes Dataflow backend adding a `jinja_variable_flags` flag passing to main.py, triggered this bug

Abacn:
R: @derrickaw

Abacn:
```
Dataflow is unable to process the managed transform(s) ref_AppliedPTransform_WriteToBigQuery-beam-schematransform-org-apache-beam-bigquery_write-v1_11(beam:schematransform:org.apache.beam:bigquery_write:v1) due to an internal error. Because the upgrade failed, the new job or the update request has been rejected. To troubleshoot, see https://cloud.google.com/dataflow/docs/guides/common-errors#managed-transforms-upgrade.

Unknown error in operation 'operation-1770059035967-649dbfc11ff55-1a1e4de0-7716ad57': [ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS] 'The zone 'projects/cloud-teleport-testing/zones/us-west2-b' does not have enough resources available to fulfill the request.  '(resource type:compute)'.'.
```

stockout issue again, not related to the change

PR #3296, opened 2026-02-05
feat: upgrade to latest release of import-spec

fbiville:
BigQueryAction and HttpAction have been removed from import-spec and re-added here.

SourceTransformations has also been removed, and reimplemented as an entity extension.

Node references of relationship targets are now strongly typed, and may contain key mapping overrides.

This upgrade needs a more recent Jackson version than what templates pull, so this commit also makes sure to pull a version compatible with Jackson.

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapper.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapper.java
@@ -186,11 +186,14 @@ class TargetMapper {
             ? null
             : findNodeTarget(mappings.getJSONObject("target"), nodes).get();
     var sourceTransformations = parseSourceTransformations(edge);
+    var dependencies = jobIndex.getDependencies(targetName);
+    dependencies.remove(sourceNodeReference);
+    dependencies.remove(targetNodeReference);
     return new RelationshipTarget(
         getBooleanOrDefault(edge, "active", true),
         targetName,
         getStringOrDefault(edge, "source", DEFAULT_SOURCE_NAME),
-        jobIndex.getDependencies(targetName),
+        dependencies,
         relationshipType,
         writeMode,
         nodeMatchMode,
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/ModelUtils.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/ModelUtils.java
@@ -250,8 +250,11 @@ public class ModelUtils {
           "Expected node or relationship target when gathering key properties, found: %s"
               .formatted(entity.getClass()));
     }
-    return Stream.concat(keyFields, uniqueFields)
-        .collect(Collectors.toCollection(LinkedHashSet::new));
+    var keys = keyFields.collect(Collectors.toCollection(LinkedHashSet::new));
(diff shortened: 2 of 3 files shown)


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/actions/preload/PreloadBigQueryAction.java
```diff
@@ -39,10 +38,7 @@ public void configure(Action action, ActionContext context) {
   @Override
   public List<String> execute() {
     List<String> msgs = new ArrayList<>();
-    String sql = action.getSql();
-    if (StringUtils.isEmpty(sql)) {
-      throw new RuntimeException("SQL Query not provided for preload query action.");
-    }
+    String sql = action.sql();
```
dhrudevalia, line 41: Is the empty case condition no longer required?

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapper.java
```diff
@@ -183,17 +185,21 @@ private static RelationshipTarget parseEdge(
         !mappings.has("target")
             ? null
             : findNodeTarget(mappings.getJSONObject("target"), nodes).get();
+    var sourceTransformations = parseSourceTransformations(edge);
```
dhrudevalia, line 188: nit: for these parseX functions since we're doing a few `is null then empty list` checks, do we want to push that down to the parse function so it just returns an empty list directly?

though that would maybe increase the complexity of this pr

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jRowWriterTransform.java
```diff
@@ -133,7 +133,11 @@ public PCollection<Row> expand(@NonNull PCollection<Row> input) {
   }
 
   private ReportedSourceType determineReportedSourceType() {
-    Source source = importSpecification.findSourceByName(target.getSource());
+    Source source =
```
dhrudevalia, line 136: Should we add a `isPresent()` validation check before `get()` after `findFirst()`, or is that an existing validation in importSpec.

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/actions/preload/PreloadBigQueryAction.java
fbiville: there's an explicit validator for this now

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/model/helpers/TargetMapper.java
fbiville: that could work, as long as there is only a single entity extension to retrieve (which is the case today, source transformation is currently the only one)

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jRowWriterTransform.java
fbiville: it's a validation in import-spec :)

PR #3320, opened 2026-02-11
fix: fix file descriptor leak due to unclosed driver

fbiville:


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/transforms/Neo4jRowWriterTransform.java
```diff
@@ -179,7 +180,7 @@ private String getCypherQuery() {
       return query;
     }
 
-    var capabilities = connectionSupplier.get().capabilities();
+    Neo4jCapabilities capabilities = getNeo4jCapabilities();
```
fbiville:
```suggestion
    var capabilities = getNeo4jCapabilities();
```

Abacn:
```
v2/cdc-parent/cdc-change-applier/target/surefire-reports/TEST-com.google.cloud.dataflow.cdc.applier.CdcToBigQueryChangeApplierPipelineIT.xml
```

not related to the change. Merging for now

PR #3323, opened 2026-02-12
AstraDBToBigQueryIT fix - different region and dataset issues

derrickaw:
1. Tries to fix - https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/21922736267/job/63323185228?pr=3321
2. Can't find dataset, so creates one in the same region as IT.
3. Also, makes sure the region is same throughout for AstraDB, BigQuery, and execution location.

### DIFF
--- a/v2/astradb-to-bigquery/src/test/java/com/google/cloud/teleport/v2/astradb/templates/AstraDbToBigQueryIT.java
+++ b/v2/astradb-to-bigquery/src/test/java/com/google/cloud/teleport/v2/astradb/templates/AstraDbToBigQueryIT.java
@@ -37,6 +37,7 @@ import java.net.http.HttpResponse;
 import java.time.Duration;
 import org.apache.beam.it.common.PipelineLauncher;
 import org.apache.beam.it.common.PipelineOperator;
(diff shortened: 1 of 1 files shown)


v2/astradb-to-bigquery/src/test/java/com/google/cloud/teleport/v2/astradb/templates/AstraDbToBigQueryIT.java
```diff
@@ -69,7 +69,7 @@ public class AstraDbToBigQueryIT extends TemplateTestBase implements Serializabl
 
   private static final String ASTRA_DB = "dataflow_integration_tests";
 
-  private static final String ASTRA_DB_REGION = "us-east1";
+  private static final String ASTRA_DB_REGION = System.getProperty("region", "us-east1");
```
Abacn: This is a shortcut. We should fetch region for it framework lib: TestProperties.region().

derrickaw: done, thanks

PR #3329, opened 2026-02-13
fix: use ISO-8601 date format in MongoDB to BigQuery template to maintain precision of timestamps (currently trimmed to seconds)

### LINKED ISSUE #3308: [Bug]: ISODate serialization truncates millisecond precision
### Related Template(s)

mongodb_to_bigquery

### Template Version

2026-02-03-00_rc01

### What happened?

I have MongoDB to BigQuery dataflow job.

When there are ISODate (example: `ISODate('2026-02-03T15:31:41.924Z')`) in Mongo documents, then in BigQuery those appear in descriptive (long) format like "Feb 3, 2026, 3:31:41 PM".

This error is related to how "Read Documents" transforms MongoDB Documents to JSON here: https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/mongodb-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/transforms/JavascriptDocumentTransformer.java#L150

By default it uses Relaxed JSON mode: https://github.com/mongodb/mongo-java-driver/blob/main/bson/src/main/org/bson/Document.java#L418

Gemini suggested that:
> If JsonWriterSettings aren't explicitly forced to ISO-8601, the MongoDB Java driver defaults to "Relaxed" mode. In this mode, it uses the JVM's default Locale to stringify dates.
Locale.US produces: Feb 3, 2026, 3:31:41 PM
ISO Standard would be: 2026-02-03T15:31:41.924Z

Could someone take a look at this?

### Relevant log output

```shell

```

pilis:
This MR fixes #3308 

⚠️ This changes are not backward compatible. Breaking change is change of timestamp format.

MongoDB ISODate fields were being serialized using the JVM's default locale, resulting in descriptive date formats like "Feb 3, 2026, 3:31:41 PM" instead of ISO-8601 format, thanks to which it maintains the precision of the source data - currently, ISODate has truncated precision to seconds.

This change configures MongoDB's JsonWriterSettings to use JsonMode.RELAXED, which outputs dates in ISO-8601 format: {"$date": "2026-02-03T15:31:41.924Z"}.

Changes:
- Added EXTENDED_JSON_WRITER_SETTINGS constant to MongoDbUtils
- Updated getTableSchema() to use MongoDB JsonWriterSettings for all user options (FLATTEN, JSON, NONE)
- Updated JavascriptDocumentTransformer to use consistent date serialization
- Added unit tests to verify ISO-8601 date format
- Updated integration test for consistency

Fixes locale-dependent date serialization issues in BigQuery output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/v2/googlecloud-and-mongodb/README_MongoDB_to_BigQuery.md
+++ b/v2/googlecloud-and-mongodb/README_MongoDB_to_BigQuery.md
@@ -28,6 +28,7 @@ on [Metadata Annotations](https://github.com/GoogleCloudPlatform/DataflowTemplat
 
 * **KMSEncryptionKey**: Cloud KMS Encryption Key to decrypt the mongodb uri connection string. If Cloud KMS key is passed in, the mongodb uri connection string must all be passed in encrypted. For example, `projects/your-project/locations/global/keyRings/your-keyring/cryptoKeys/your-key`.
 * **filter**: Bson filter in json format. For example, `{ "val": { $gt: 0, $lt: 9 }}`.
+* **useIso8601DateFormat**: If true, MongoDB date and timestamp values are serialized as ISO-8601 strings (for example, `2026-02-03T15:31:41.924Z`), preserving millisecond precision. If false (the default), the previous serialization format is used, which trims timestamps to second precision and depends on the JVM locale.
 * **useStorageWriteApi**: If `true`, the pipeline uses the BigQuery Storage Write API (https://cloud.google.com/bigquery/docs/write-api). The default value is `false`. For more information, see Using the Storage Write API (https://beam.apache.org/documentation/io/built-in/google-bigquery/#storage-write-api).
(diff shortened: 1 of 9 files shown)


pilis:
Hey @Abacn— mind taking a look or kicking off CI here? 🙏

Fixes #3329 — ISODate was losing millisecond precision in the MongoDB → BigQuery template because dates were serialized with the JVM default locale. Switched to MongoDB's `JsonWriterSettings` (`JsonMode.RELAXED`) so they come out as proper ISO-8601 (`{"$date":"2026-02-03T15:31:41.924Z"}`).

Just rebased onto `main` (picks up the move to `v2/googlecloud-and-mongodb/`), so it's clean and mergeable. Spotless + `MongoDbUtilsTest` pass locally.

⚠️ Note: breaking change to the timestamp format. I need this for a client, so any chance of a review soon would be hugely appreciated 🙏

pilis:
@GoogleCloudPlatform/dataflow-templates-wg could I get a reviewer assigned here? No code owner is configured for v2/googlecloud-and-mongodb/, so none was auto-assigned. cc @Abacn @jrmccluskey

Update based on the earlier hesitation around the format change: this is now non-breaking. ISO-8601 serialization is opt-in behind a new useIso8601DateFormat parameter that defaults to false, so existing pipelines are unchanged. Added unit tests (incl. the fallback path) and regenerated the template docs; Spotless + MongoDbUtilsTest pass locally.

Since this is a fork PR, the workflow approval resets on each push — this is the final push, so whenever a maintainer can hit "Approve and run" + review, it should be good to go. 🙏

PR #3337, opened 2026-02-16
feat(Spanner): Add workflow dispatch options to run specific test suites or individual tests within Spanner PR

pratickchokhani:


anowardear062-svg (approved):

anowardear062-svg (approved):

anowardear062-svg (approved):

manitgupta:
What is the use-case for this? Curious to know why we would need to run a specific test via the CI/CD flow?

pratickchokhani:
> What is the use-case for this? Curious to know why we would need to run a specific test via the CI/CD flow?

This is only applicable when the workflow is executed manually from the Actions page. While testing integration test, this will reduce the execution time to ~15m from the current time of ~1.5 hr.

Only to be used for debugging purpose.

manitgupta:
Can't we run the ITs directly using Maven commands? 
Something like this - 

```
mvn clean verify -PtemplatesIntegrationTests -pl v2/spanner-to-sourcedb -am -Dtest=SpannerToSourceDbCustomShardIT -DartifactBucket="manit-testing" -Dproject="span-cloud-testing" -DspannerInstanceId="manit-testing"
```

.github/workflows/spanner-pr.yml
```diff
@@ -167,6 +174,11 @@ jobs:
   java_integration_tests_templates:
     name: Dataflow Templates Integration Tests
     needs: [java_integration_smoke_tests_templates]
+    if: |
```
pratickchokhani, line 180: Reasoning: if previous tests are skipped, due to the respective if condition, then integration test should be executed.

pratickchokhani:
> Can't we run the ITs directly using Maven commands? Something like this -
> 
> ```
> mvn clean verify -PtemplatesIntegrationTests -pl v2/spanner-to-sourcedb -am -Dtest=SpannerToSourceDbCustomShardIT -DartifactBucket="manit-testing" -Dproject="span-cloud-testing" -DspannerInstanceId="manit-testing"
> ```

We can but, for some tests other resources like Cassandra, Mysql or other instances are required. Thought process was that using this those setup will not be required.

manitgupta:
Thanks, got it. We should document this class of tests and how to run them somewhere.
CC: @darshan-sj

pratickchokhani:
> Thanks, got it. We should document this class of tests and how to run them somewhere. CC: @darshan-sj

Will be updating the Integration test playbook.

.github/workflows/spanner-load-tests.yml
```diff
@@ -21,6 +21,11 @@ on:
   # at 02:00 weekly on every Tuesday.
   - cron: '0 2 * * 2'
   workflow_dispatch:
+    inputs:
+      specific_test:
+        description: 'Specific Integration Test (passed to -Dtest=)'
```
darshan-sj: This should be Specific Load Test? Or just Specific Test.

.github/workflows/spanner-pr.yml
```diff
@@ -101,6 +106,7 @@ jobs:
   java_unit_tests:
     name: Unit Tests
     needs: [java_build]
+    if: github.event_name != 'workflow_dispatch' || github.event.inputs.specific_test == ''
```
darshan-sj, line 110: I didn't understand this check. If this condition is true, the unit test executes? 

We are saying if event == workflow_dispatch && specific_test != "" then skip the UT. -> That makes sense. Please add a comment explaining it.

cicd/cmd/run-it-tests/main.go
```diff
@@ -75,7 +75,9 @@ func main() {
 		flags.CloudProxyPostgresPort(),
 		flags.CloudProxyPassword(),
 		flags.UnifiedWorkerHarnessContainerImage(),
-		flags.CloudProxyPassword())
+		flags.CloudProxyPassword(),
+		mvnFlags.SpecificTest(flags.TestToRun()),
```
darshan-sj, line 79: If TestToRun is empty, this would add -Dtest="" ? Won't that skip all the tests?

pratickchokhani: If TestToRun is empty, then SpecificTest will return empty string. This is the convention followed across.

.github/workflows/spanner-pr.yml
pratickchokhani: Have updated with comments.

darshan-sj (approved):

anowardear062-svg (approved):

PR #3338, opened 2026-02-16
Add support for UDF function transformations in PubSubToRedis template

wladabutenko:
Added optional UDF support to the PubSubToRedis pipeline, enabling user-defined transformations before writing to Redis.

### DIFF
--- a/v2/pubsub-to-redis/src/test/java/com/google/cloud/teleport/v2/templates/PubSubToRedisTest.java
+++ b/v2/pubsub-to-redis/src/test/java/com/google/cloud/teleport/v2/templates/PubSubToRedisTest.java
@@ -239,6 +239,7 @@ public class PubSubToRedisTest {
     // TestPipeline rule requires pipeline.run() to be called
     pipeline.run();
   }
+
   @Test
   public void testRedisStringSink() {
     String key = "testStringKey";


v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +383,103 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+        .apply(
+            "Extract Transformed Payload",
+            ParDo.of(
+                new DoFn<FailsafeElement<PubsubMessage, String>, PubsubMessage>() {
+                  @ProcessElement
+                  public void processElement(ProcessContext c) {
+                    FailsafeElement<PubsubMessage, String> element = c.element();
+                    c.output(
+                        new PubsubMessage(
+                            element.getPayload().getBytes(),
```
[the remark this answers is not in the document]
tvalentyn: Seems like a reasonable concern.

I think we should have helpers for these PubsubMessage<->Failsafe manipulations in a shareable location:

I am seeing these already:
https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/common/src/main/java/com/google/cloud/teleport/v2/transforms/PubSubToFailSafeElement.java 
 https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/googlecloud-to-elasticsearch/src/main/java/com/google/cloud/teleport/v2/elasticsearch/transforms/PubsubMessageToFailsafeElementFn.java

Then, helpers should use consistent encodings.

v2/pubsub-to-redis/terraform/Cloud_PubSub_to_Redis/dataflow_job.tf
```diff
@@ -34,231 +34,238 @@ variable "region" {
...
-  description = "Key expiration time in sec (ttl, default for HASH_SINK is -1 i.e. no expiration)"
-  default     = null
+  type = number
+  description = "The key expiration time in seconds. The `ttl` default for `HASH_SINK` is -1, which means it never expires."
+  default = null
+}
+
+variable "deadletterTopic" {
+  type = string
+  description = "The Pub/Sub topic to forward unprocessable messages to. Messages that fail UDF transformation or cannot be written to Redis are forwarded here. For example, `projects/<PROJECT_ID>/topics/<TOPIC_NAME>`"
```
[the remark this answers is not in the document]
tvalentyn: thoughts on this one? what do other templates do?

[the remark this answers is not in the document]
wladabutenko: I think this is due to the nature of Redis as it treats the data as an binary-safe, opaque block of text. So for STRING_SINK type it does not validate whether the string malformed/correct, it's takes the raw characters and stores them as a Redis string value.

Based on [v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/transforms/MessageTransformation.java](url) the entire message payload mapped into fieldKey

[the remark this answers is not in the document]
tvalentyn: is there a failure mode where a message is not written to Reddis for whatever reason? if so, will such message be forwarded to DLQ?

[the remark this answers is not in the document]
tvalentyn: The comment (which I agree with) is that if only messages for which UDF failed to run are forwarded to DLQ, then the docstring should not say ...Messages that fail UDF transformation **or cannot be written to Redis** are forwarded here...

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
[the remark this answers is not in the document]
wladabutenko: Added

v2/pubsub-to-redis/terraform/Cloud_PubSub_to_Redis/dataflow_job.tf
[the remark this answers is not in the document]
wladabutenko: Done.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/transforms/MessageTransformation.java
```diff
@@ -115,37 +115,19 @@ public void processElement(
...
-      objectMapper.enable(DeserializationFeature.FAIL_ON_TRAILING_TOKENS);
-      if (!isValidJSON(element)) {
-        LOG.warn("Invalid JSON received, forwarding to DLQ. MessageId: {}", messageId);
-        receiver.get(LOGS_DEADLETTER_OUT).output(pubsubMessage);
-        return;
-      }
-
-      // readValue requires a JSON object ({}). Valid JSON that is not an object
-      // (e.g. [], null, "string") would throw here, so we catch and DLQ it.
+      // Catches invalid JSON, non-object JSON ([], null, "string")
```
tvalentyn: can this be unit-tested?

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/transforms/MessageTransformation.java
```diff
@@ -49,6 +51,17 @@ public class MessageTransformation {
...
 
+  /** The tag for the main (valid) output of the {@link MessageToRedisLogs} transform. */
+  public static final TupleTag<KV<String, KV<String, String>>> LOGS_MAIN_OUT =
+      new TupleTag<KV<String, KV<String, String>>>() {};
+
+  /**
+   * The tag for the dead-letter output of the {@link MessageToRedisLogs} transform. Messages that
+   * contain invalid JSON are forwarded here instead of being silently dropped.
+   */
+  public static final TupleTag<PubsubMessage> LOGS_DEADLETTER_OUT =
```
tvalentyn: > Messages that
   * contain invalid JSON are forwarded here instead of being silently dropped.

It would be safer to make this improvement in a separate change and let's focus this PR on UDF support. Then, it is easier to track and rollback changes independently if necessary.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +394,103 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+                .build())
+        .apply("Write Failed UDF", writeUdfFailures(options));
+
+    // Extract the successfully transformed messages
+    return udfResult
+        .get(UDF_OUT)
+        .apply(
+            "Extract Transformed Payload",
+            ParDo.of(
+                new DoFn<FailsafeElement<PubsubMessage, String>, PubsubMessage>() {
```
tvalentyn, line 454: Can ConvertFailsafeElementToPubsubMessage be used here as well?

tvalentyn: ah, maybe we can't do that if that helper only extracts the unmodified message.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/transforms/MessageTransformation.java
wladabutenko: I've reset the brunch back to 4 commit and move last 3 commits to a new brunch. We will open a separate PR for DLQ LOGGING_SINK feature, as we discussed.

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
wladabutenko: we also need to preserve original attributes (messageId, publish time, etc.)

wladabutenko:
R: @stankiewicz

v2/pubsub-to-redis/src/test/java/com/google/cloud/teleport/v2/templates/PubSubToRedisIT.java
```diff
@@ -0,0 +1,288 @@
...
+      Function<LaunchConfig.Builder, LaunchConfig.Builder> paramsAdder) throws IOException {
+    // Arrange
+    String testId = testName + "-hash";
+    TopicName tc = pubsubResourceManager.createTopic(testId);
+    SubscriptionName subscription = pubsubResourceManager.createSubscription(tc, "sub-" + testId);
+
+    redisClient = createRedisClient();
+
+    LaunchConfig.Builder options =
+        paramsAdder.apply(
```
stankiewicz, line 193: missing dlq

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -224,6 +253,17 @@ public interface PubSubToRedisOptions extends PipelineOptions {
...
     void setTtl(Long ttl);
+
+    @TemplateParameter.PubsubTopic(
+        order = 9,
+        description = "Output deadletter Pub/Sub topic",
+        helpText =
+            "The Pub/Sub topic to forward unprocessable messages to. Messages that fail UDF transformation are forwarded here.",
+        example = "projects/<PROJECT_ID>/topics/<TOPIC_NAME>")
+    @Validation.Required
+    String getDeadletterTopic();
```
stankiewicz: dlq is not required, but it's only needed for UDF. I would relax this requirement and fail pipeline if someone is using UDF but not providing deadletter. 
Also I would recommend renaming it to OutputDeadletterTopic

v2/pubsub-to-redis/src/main/java/com/google/cloud/teleport/v2/templates/PubSubToRedis.java
```diff
@@ -338,4 +383,103 @@ public static PipelineResult run(PubSubToRedisOptions options) {
...
+        .apply(
+            "Extract Transformed Payload",
+            ParDo.of(
+                new DoFn<FailsafeElement<PubsubMessage, String>, PubsubMessage>() {
+                  @ProcessElement
+                  public void processElement(ProcessContext c) {
+                    FailsafeElement<PubsubMessage, String> element = c.element();
+                    c.output(
+                        new PubsubMessage(
+                            element.getPayload().getBytes(java.nio.charset.StandardCharsets.UTF_8),
```
stankiewicz: import StandardCharsets

wladabutenko:
All comments addressed, PTAL

stankiewicz:
it tests are passing for PubSubToRedis

tvalentyn (approved):

tvalentyn:
Thaks, @wladabutenko . we can mere once tests pass.

tvalentyn:
thanks, @stankiewicz for help reviewing.

PR #3374, opened 2026-02-18
yaml/python workflow update - add missing pom file change detection

derrickaw:
1. During the new yaml/python workflow creation - #3241, missed the pom files.
2. Adding pom file detection except in v1 and v2 folders.


.github/workflows/yaml-pr.yml
Abacn: we can deduplicate this line?

derrickaw: Done

PR #3380, opened 2026-02-19
Fixing flaky Spanner tests

darshan-sj:
Fixing flaky tests - 

1. MySQLDatastreamToSpannerDataTypesAndExpressionIT
2. PostgreSQLDatastreamToSpannerDataTypesIT
3. SpannerToMySqlDataTypesPGDialectIT
4. SpannerToMySqlDataTypesIT
5. DatastreamToSpannerReservedKeywordsPostgresIT

Issues found and fixed:

1. The tests were using ChainedConditionCheck and chaining 23 conditions. Chained condition check executes checks one after the other with 15 seconds wait in between. Hence, even if all the underlying 23 conditions were satisfied, the whole condition check was taking 23*15 = 345 secs ~ 6 mins. and hence we were hitting timeout. Changed the condition check to use "AND" condition instead of using chained condition. "AND" condition doesn't put waits in between.
2. Added a wait for Cutover minutes in `MySQLDatastreamToSpannerDataTypesAndExpressionIT`. The test contains update statements and the Number of rows in destination matching the number of rows in source does not suffice to conclude that the migration is complete. 
3. Removed incorrect number of rows assert in the destination for unsupported tables in `SpannerToMySqlDataTypesPGDialectIT` and `SpannerToMySqlDataTypesIT` (by adding the unsupported tables to [ignore list](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/a196aaffa58134965b4331d21ba3178cb2fd09b7/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToMySqlDataTypesPGDialectIT.java#L186)
4. Implemented Postgres Replication Slot Management in CloudPostgresResourceManager - implemented a managed PostgreSQL replication slot and publication creation within CloudPostgresResourceManager. This simplifies integration tests by handling the lifecycle of the pg replication slot resources automatically and robustly. This prevents "tuple concurrently updated" errors during resource creation/deletion. Also added corresponding changes to tests.

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudPostgresResourceManager.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudPostgresResourceManager.java
@@ -142,66 +142,94 @@ public class CloudPostgresResourceManager extends CloudSqlResourceManager {
    * @return ReplicationInfo containing the names of the created slot and publication.
    */
   public ReplicationInfo createLogicalReplication(List<String> tableNames) {
-    // Generate unique names
-    // Postgres max identifier length is 63
-    // Pattern: prefix + "_" + date + "_" + randomUUID (no hyphens)
-    String dateId =
-        DateTimeFormatter.ofPattern("yyyyMMdd_HHmmss").format(java.time.LocalDateTime.now());
-    String uuid = java.util.UUID.randomUUID().toString().replace("-", "");
+    synchronized (CloudPostgresResourceManager.class) {
+      // Generate unique names
+      // Postgres max identifier length is 63
+      // Pattern: prefix + "_" + date + "_" + randomUUID (no hyphens)
+      String dateId =
+          DateTimeFormatter.ofPattern("yyyyMMdd_HHmmss").format(java.time.LocalDateTime.now());
+      String uuid = java.util.UUID.randomUUID().toString().replace("-", "");
 
-    String pubPrefix = "pub";
-    String slotPrefix = "slot";
+      String pubPrefix = "pub";
+      String slotPrefix = "slot";
 
-    String publicationName = String.format("%s_%s_%s", pubPrefix, dateId, uuid);
-    String replicationSlotName = String.format("%s_%s_%s", slotPrefix, dateId, uuid);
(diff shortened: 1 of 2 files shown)


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MySQLDatastreamToSpannerDataTypesIT.java
pratickchokhani: Similar to other PR, shouldn't this be ConditionCheck?

darshan-sj: Corrected the code.

VardhanThigle (dismissed):

pratickchokhani (dismissed):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudPostgresResourceManager.java
```diff
@@ -113,27 +126,205 @@ public synchronized long getRowCount(String tableName) {
...
+          LOG.debug("Granting replication privilege to current user...");
+          statement.execute("ALTER USER CURRENT_USER WITH REPLICATION");
+          connection.commit();
+
+          LOG.debug(
+              "Acquiring lock {} for replication setup (Create Publication)...",
+              REPLICATION_SETUP_LOCK_ID);
+          statement.execute("SELECT pg_advisory_xact_lock(" + REPLICATION_SETUP_LOCK_ID + ")");
+
+          if (tableNames == null || tableNames.isEmpty()) {
```
manitgupta, line 196: Wow, now we need locks to execute tests :P

manitgupta (approved):

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/cloudsql/CloudPostgresResourceManager.java
darshan-sj: Yeah. The replication slot resources are PG instance level objects. They can cause concurrency issues if two tests try to execute in parallel. :)

PR #3381, opened 2026-02-19
Correcting Dataflow jobs resource leaks

darshan-sj:
shutdownHook gets executed just before the whole maven process exits and not tied to individual test threads.

Removing the setupShutdownHook boolean variable and always setting up shutdownHook to turndown Dataflow job before exit. There is no usecase where we don't want to setup a ShutdownHook to turn down dataflow job.


darshan-sj:
@damccorm @Abacn Please review

PR #3385, opened 2026-02-19
Fix staging Python templates and staging log not shown

Abacn:
Fix b/485853563

### DIFF
--- a/plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
+++ b/plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
@@ -59,7 +59,6 @@ COPY --from=python-base /venv /root/.apache_beam/cache/venvs/py-$PY_VERSION-beam
 # Copy required shared libraries from python-base
 COPY --from=python-base /lib/$CHIPSET_ARCH/ld-*so* /lib64/
 COPY --from=python-base /lib/$CHIPSET_ARCH/lib*so* /lib/$CHIPSET_ARCH/
-COPY --from=python-base /usr/lib/$CHIPSET_ARCH/libffi* /usr/lib/$CHIPSET_ARCH/
 COPY --from=python-base /usr/local/lib/libpython$PY_VERSION* /usr/local/lib/
 
 # Copy minimal commands from python-base needed to execute template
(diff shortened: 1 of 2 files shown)


plugins/core-plugin/src/main/resources/Dockerfile-template-xlang
```diff
@@ -59,7 +59,6 @@ COPY --from=python-base /venv /root/.apache_beam/cache/venvs/py-$PY_VERSION-beam
 # Copy required shared libraries from python-base
 COPY --from=python-base /lib/$CHIPSET_ARCH/ld-*so* /lib64/
 COPY --from=python-base /lib/$CHIPSET_ARCH/lib*so* /lib/$CHIPSET_ARCH/
-COPY --from=python-base /usr/lib/$CHIPSET_ARCH/libffi* /usr/lib/$CHIPSET_ARCH/
```
Abacn, line 62: it appears the COPY statement triggered the issue can be removed. The previous statement, `/lib/$CHIPSET_ARCH/lib*so*` also contains libffi.so.8

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
```diff
@@ -365,10 +365,10 @@ private String getSpecPath(
 
             try {
               Process exec = Runtime.getRuntime().exec(mavenCmd);
-              IORedirectUtil.redirectLinesLog(exec.getInputStream(), LOG);
```
Abacn, line 368: `Runtime.getRuntime().exec(mavenCmd)` is unblocking. Therefore these two redirectlinesLog call won't capture logs. This has caused log not shown for staging command

darshan-sj: Should this change be done here as well? - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/917cf5de27d7d97219dca21fce372cd65634989f/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/dataflow/FlexTemplateDataflowJobResourceManager.java#L217

Abacn:
R: @derrickaw

Abacn:
R: @darshan-sj

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/TemplateTestBase.java
Abacn: Yes. let me first this first to make tests green. Will fix in follow ups.

Abacn:
Spanner PR attempt 1 failed single test not related to the change. Merging for now.

PR #3392, opened 2026-02-20
Dataflow Template for Firestore to Firestore Documents Copy (Clone of PR 3098)

jingqizz:
Implements a basic Dataflow template for copying documents from one Firestore database to another. (Patched from #3098)

Co-authored-by: pacoavila808 <pacoavila@google.com>

### DIFF
--- a/v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
+++ b/v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
@@ -203,7 +203,7 @@ public class FirestoreToFirestore {
       // 2. Apply FirestoreIO to get partitions (as RunQueryRequests)
       PCollection<RunQueryRequest> partitionedQueries =
           partitionQueryRequests.apply(
-              "GetPartitions",
+              "Get Partitions",
               FirestoreIO.v1()
                   .read()
                   .partitionQuery()
@@ -216,7 +216,7 @@ public class FirestoreToFirestore {
       // 3. Execute each partitioned query
       PCollection<RunQueryResponse> responses =
           partitionedQueries.apply(
-              "QueryDocumentsInPartitions",
+              "Query Documents in Partitions",
               FirestoreIO.v1()
                   .read()
                   .runQuery()
@@ -228,7 +228,7 @@ public class FirestoreToFirestore {
 
       // 4. Process the documents from the responses
       PCollection<Document> documents =
-          responses.apply("ExtractDocuments", ParDo.of(new RunQueryResponseToDocumentFn()));
+          responses.apply("Extract Documents", ParDo.of(new RunQueryResponseToDocumentFn()));
 
       // 5. Prepare documents for writing to the destination database
       PCollection<Write> writes =
@@ -237,7 +237,7 @@ public class FirestoreToFirestore {
 
       // 6. Write documents to the destination Firestore database
       writes.apply(
-          "WriteDocumentsToDestination",
+          "Write Documents to Destination",
           FirestoreIO.v1()
               .write()
               .withProjectId(destinationProjectId)
@@ -276,14 +276,14 @@ public class FirestoreToFirestore {
       Query query =
           Query.newBuilder().addKind(KindExpression.newBuilder().setName("__kind__")).build();
       return p.apply(
-              "FindAllCollectionGroups",
+              "Find All Collection Groups",
               DatastoreIO.v1()
                   .read()
                   .withProjectId(options.getSourceProjectId())
                   .withDatabaseId(options.getSourceDatabaseId())
                   .withQuery(query))
           .apply(
-              "ExtractCollectionGroupNames",
+              "Extract Collection Group Names",
               ParDo.of(
                   new DoFn<Entity, String>() {
                     @ProcessElement
@@ -302,6 +302,6 @@ public class FirestoreToFirestore {
         Arrays.stream(options.getCollectionGroupIds().split(","))
             .map(String::trim)
             .collect(Collectors.toList());
-    return p.apply("CreateCollectionGroups", Create.of(collectionGroupIdsList));
+    return p.apply("Create Collection Groups", Create.of(collectionGroupIdsList));
   }
 }


v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,289 @@
...
+          collectionIdsList,
+          maxNumWorkers,
+          readTime);
+
+      // 1. Construct the PartitionQuery requests for the collections.
+      PCollection<PartitionQueryRequest> partitionQueryRequests =
+          p.apply(Create.of(collectionIdsList))
+              .apply(
+                  new CreatePartitionQueryRequestFn(
+                      sourceProjectId, sourceDatabaseId, maxNumWorkers > 1 ? maxNumWorkers : 20L));
```
pacoavila808: nit: We should probably store the default max workers in a constant, and include it in the documentation of the flag as well.

(update: looks like this is a shared flag, what's the default as is?)

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,289 @@
...
+      String destinationDatabaseId = options.getDestinationDatabaseId();
+
+      List<String> collectionIdsList;
+      String collectionIds = options.getCollectionIds();
+      if (collectionIds.isEmpty()) {
+        try {
+          collectionIdsList = getAllCollectionIds(sourceProjectId, sourceDatabaseId);
+        } catch (Exception e) {
+          LOG.error("Failed to list collections: {}", e.getMessage(), e);
+          return;
```
pacoavila808: I think we actually want to throw the exception here to make sure the pipeline clearly fails in case this happens.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,289 @@
...
+import org.slf4j.LoggerFactory;
+
+@Template(
+    name = "Cloud_Firestore_to_Firestore",
+    category = TemplateCategory.BATCH,
+    displayName = "Firestore to Firestore",
+    description = {
+      "The Firestore to Firestore template is a batch pipeline that reads documents from one"
+          + " <a href=\"https://cloud.google.com/firestore/docs\">Firestore</a> database and writes"
+          + " them to another Firestore database.",
```
pacoavila808: Can you add a callout that this does not support Enterprise edition databases as the source?

v2/firestore-to-firestore/src/test/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestoreIT.java
```diff
@@ -0,0 +1,141 @@
...
+
+    sourceFirestoreResourceManager.write(collectionId, inputData);
+
+    LaunchConfig options =
+        LaunchConfig.builder(testName, SPEC_PATH)
+            .addParameter("sourceProjectId", PROJECT)
+            .addParameter("sourceDatabaseId", SOURCE_DATABASE_ID)
+            .addParameter("destinationProjectId", PROJECT)
+            .addParameter("destinationDatabaseId", DESTINATION_DATABASE_ID)
+            .addParameter("collectionIds", collectionId)
```
pacoavila808: Can you add a test case that doesn't specify collectionIds, and verify that all collections are copied?

it/google-cloud-platform/pom.xml
```diff
@@ -107,16 +107,28 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-spanner</artifactId>
-            <version>6.104.0</version>
+            <version>6.102.0</version>
```
pacoavila808: Is this needed? Ideally we wouldn't downgrade the spanner version here as this looks like a shared dependency.

(sorry can't remember if this was part of my original changes and what the context was).

it/google-cloud-platform/pom.xml
```diff
@@ -107,16 +107,28 @@
...
         </dependency>
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-pubsub</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-pubsublite</artifactId>
+        </dependency>
```
pacoavila808: Similar here I think this is unrelated to our change, remove and very it's not needed?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,289 @@
...
+    try {
+      UncaughtExceptionLogger.register();
+
+      LOG.info("Starting Firestore to Firestore pipeline...");
+
+      Options options = PipelineOptionsFactory.fromArgs(args).withValidation().as(Options.class);
+      validateOptions(options);
+      LOG.info("Pipeline options parsed and validated.");
+
+      Pipeline p = Pipeline.create(options);
```
pacoavila808, line 163: Can you add a pipeline option to log errors to a given GCS path? See https://docs.cloud.google.com/dataflow/docs/guides/templates/provided/cloud-storage-to-firestore for an example.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
jingqizz: I found `maxNumWorkers` in metadata/src/main/java/com/google/cloud/teleport/metadata/options/DefaultTemplateOptions.java.

Seems like there is no explicit value, so I presume the default will be `null`?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
jingqizz: Throw the RuntimeException here instead. Thanks!

it/google-cloud-platform/pom.xml
jingqizz: oops sorry I think this is because I cloned from #3098, but there are some recent version changes for sure :)

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
pacoavila808: Looking at a few other examples, they actually use DataflowPipelineWorkerPoolOptions instead. We should probably use that for consistency (sorry I don't remember where DataflowPipelineOptions came from.

From our example, we also need to check that the value is > 0 in case the user did not set it, in which case we need to set a default value. https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/743da1d2ea9c4114ff700da8f7668d653ccee6d1/v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java#L855

I think we need to:
1. Check the value above is >0, and set the default if so, then use that same value here.
2. Pull the default into a constant, and update the value to something a little more reasonable (sorry, also don't remember where 20 came from, possibly AI). Looking at other Firestore examples, 500 is a reasonable default. https://docs.cloud.google.com/dataflow/docs/guides/templates/provided/cloud-storage-to-firestore#console

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+                  .withProjectId(destinationProjectId)
+                  .withDatabaseId(destinationDatabaseId)
+                  .batchWrite()
+                  .withDeadLetterQueue()
+                  .withRpcQosOptions(rpcQosOptions)
+                  .build());
+
+      // 7. Log errors to GCS if errorWritePath is provided
+      String errorWritePath = options.getErrorWritePath();
+      if (!Strings.isNullOrEmpty(errorWritePath)) {
```
pacoavila808: I think other pipelines require this? In general it's probably best if we require it to avoid the foot gun of potentially not having error output.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+              "QueryDocumentsInPartitions",
+              FirestoreIO.v1()
+                  .read()
+                  .runQuery()
+                  .withProjectId(sourceProjectId)
+                  .withDatabaseId(sourceDatabaseId)
+                  .withReadTime(readTime)
+                  .withRpcQosOptions(rpcQosOptions)
+                  .build());
+
```
pacoavila808, line 228: Are there similar error tracking methods that we can use to track read errors like we're doing below?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+      // 7. Log errors to GCS if errorWritePath is provided
+      String errorWritePath = options.getErrorWritePath();
+      if (!Strings.isNullOrEmpty(errorWritePath)) {
+        LOG.info("Error logging to GCS is enabled: {}", errorWritePath);
+        writeFailures
+            .apply(
+                "FilterFailedWrites",
```
pacoavila808: nit: This isn't really a filter, right, but should be named "MapFailedWritesToErrorMessage" or similar?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+        optional = true,
+        description = "Error Write Path",
+        helpText =
+            "The Cloud Storage path to write error logs to. Path should be in the format gs://<bucket>/<folder>/.",
+        example = "gs://your-bucket/errors/")
+    String getErrorWritePath();
+
+    void setErrorWritePath(String value);
+
+    void setReadTime(String readTime);
```
pacoavila808, line 150: nit: can you move this back up to be below the getReadTime() function?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+    category = TemplateCategory.BATCH,
+    displayName = "Firestore to Firestore",
+    description = {
+      "The Firestore to Firestore template is a batch pipeline that reads documents from one"
+          + " <a href=\"https://cloud.google.com/firestore/docs\">Firestore</a> database and writes"
+          + " them to another Firestore database. It is not supported for Enterprise edition databases"
+          + " to be the source.",
+      "Data consistency is guaranteed only at the end of the pipeline when all data has been"
+          + " written to the destination database.\n",
+    },
```
pacoavila808, line 63: Can you add a callout to the fact that errors are written to GCS? See https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/743da1d2ea9c4114ff700da8f7668d653ccee6d1/v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java#L119 for example

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,345 @@
...
+import org.slf4j.LoggerFactory;
+
+@Template(
+    name = "Cloud_Firestore_to_Firestore",
+    category = TemplateCategory.BATCH,
+    displayName = "Firestore to Firestore",
+    description = {
+      "The Firestore to Firestore template is a batch pipeline that reads documents from one"
+          + " <a href=\"https://cloud.google.com/firestore/docs\">Firestore</a> database and writes"
+          + " them to another Firestore database. It is not supported for Enterprise edition databases"
```
pacoavila808: nit; Maybe put the callout that Enterprise databases on a new line to draw a little more attention to it.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,361 @@
...
+
+      p.run();
+      LOG.info("Pipeline.run() called.");
+    } catch (Exception e) {
+      LOG.error("Failed to run pipeline: {}", e.getMessage(), e);
+      throw e;
+    }
+  }
+
+  private static String getDocumentName(Write write) {
```
pacoavila808: nit; Maybe getDocumentNameFromWriteFailure to make it clear what this is for?

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,361 @@
...
+
+      // 7. Log errors to GCS if errorWritePath is provided
+      String errorWritePath = options.getErrorWritePath();
+      PCollection<String> allErrors =
+          PCollectionList.of(runQueryErrors)
+              .and(writeErrors)
+              .apply("MergeErrorCollections", Flatten.<String>pCollections());
+      if (!Strings.isNullOrEmpty(errorWritePath)) {
+        LOG.info("Error logging to GCS is enabled: {}", errorWritePath);
+        allErrors.apply(
```
pacoavila808: I think we should also make sure the pipeline exits in an error state if there are errors. This will prevent users from thinking all the documents were copied when in reality they weren't. e.g. something like:

```
      // 9. Count the total number of errors
      PObject<Long> numErrors = allErrors.apply("CountErrors", Count.globally());

      // Run the pipeline
      p.run();
      LOG.info("Pipeline.run() called.");

      // 10. Check if there were any errors
      long errorCount = numErrors.getValue();
      if (errorCount > 0) {
        LOG.error("Pipeline completed with {} errors.", errorCount);
        throw new RuntimeException("Pipeline failed due to " + errorCount + " errors during read/write operations.");
      } else {
        LOG.info("Pipeline completed successfully with no errors.");
      }
```

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/RunQueryResponseToDocumentFn.java
```diff
@@ -21,11 +21,20 @@
...
 public class RunQueryResponseToDocumentFn extends DoFn<RunQueryResponse, Document> {
+  public static final TupleTag<Document> DOCUMENT_TAG = new TupleTag<Document>() {};
+  public static final TupleTag<String> ERROR_TAG = new TupleTag<String>() {};
+
   @ProcessElement
   public void processElement(ProcessContext c) {
     RunQueryResponse response = c.element();
-    if (response != null && response.hasDocument()) {
-      c.output(response.getDocument());
+    try {
```
pacoavila808: Can you update v2/firestore-to-firestore/src/test/java/com/google/cloud/teleport/v2/transforms/RunQueryResponseToDocumentFnTest.java so that the new error handling code is covered by tests?

jingqizz: Sorry I was just updating the test cases! :) 

`testRunQueryResponseToDocument_skipsMissingDocument` should cover the case where (1) there is no document in the response (2) the output error includes "response is null or has no document"

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,365 @@
...
+                            getDocumentName(failure.getWrite()), failure.getStatus());
+                      }));
+
+      // 7. Log errors to GCS if errorWritePath is provided
+      String errorWritePath = options.getErrorWritePath();
+      PCollection<String> allErrors =
+          PCollectionList.of(runQueryErrors)
+              .and(writeErrors)
+              .apply("MergeErrorCollections", Flatten.<String>pCollections());
+      if (!Strings.isNullOrEmpty(errorWritePath)) {
```
pacoavila808: This is still marked as optional on the pipeline?

pacoavila808: See https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/743da1d2ea9c4114ff700da8f7668d653ccee6d1/v2/datastream-mongodb-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamMongoDBToFirestore.java#L739, if not set, the pipeline uses a default location. It doesn't just skip writing error output.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,300 @@
...
+    name = "Cloud_Firestore_to_Firestore",
+    category = TemplateCategory.BATCH,
+    displayName = "Firestore to Firestore",
+    description = {
+      "The Firestore to Firestore template is a batch pipeline that reads documents from one"
+          + " <a href=\"https://cloud.google.com/firestore/docs\">Firestore</a> database and writes"
+          + " them to another Firestore database. ",
+      "It is not supported for Enterprise edition databases to be the source.",
+      "Data consistency is guaranteed only at the end of the pipeline when all data has been"
+          + " written to the destination database. Any errors that occur during operation are recorded in error queues. \n",
```
pacoavila808: Nit: Remove "Any errors that occur during operation are recorded in error queues." since we reverted that.

pacoavila808 (approved):

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/transforms/RunQueryResponseToDocumentFn.java
jingqizz: Mark this as resolved.

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
```diff
@@ -0,0 +1,307 @@
...
+                  .withProjectId(sourceProjectId)
+                  .withDatabaseId(sourceDatabaseId)
+                  .withReadTime(readTime)
+                  .withRpcQosOptions(rpcQosOptions)
+                  .build());
+
+      // 3. Execute each partitioned query
+      PCollection<RunQueryResponse> responses =
+          partitionedQueries.apply(
+              "QueryDocumentsInPartitions",
```
leongn-google: I can't remember, but do these step names have to exclude spaces and needs to be camel case? Or is it convention? 

It might be better if they are human readable so they show up in the console better

pacoavila808: I don't think it's required, but UpperCamelCase is the casing that's general convention in other templates and documentation e.g. https://beam.apache.org/documentation/programming-guide/#applying-transforms.

With that said, I see other templates using spaces, and I agree that is more readable in the UI, so I've added spaces to make them more human readable.

leongn-google (approved):

v2/firestore-to-firestore/src/main/java/com/google/cloud/teleport/v2/templates/FirestoreToFirestore.java
jingqizz: Resolve the comment for merging

PR #3402, opened 2026-02-24
Add metadata for sharding in Bulk Migration and relax change event column verification in Live Migration

aasthabharill:
https://b.corp.google.com/issues/484901616

### 1. Don’t throw error on finding dropped source columns in change event
- This strict check is unnecessary and breaks code. If a user doesn’t pass session file, our existing code has no mechanism to drop these columns from change event before this strict verification is carried out. 
- Change: Relax this check and drop irrelevant columns from change event

### 2. Bulk DLQ writes `_metadata_shard_id` in all DLQ entries
- If Spanner doesn't have a `shardIdColumn`, this information is completely lost in the DLQ.
- This is important for the user to have **better debuggability** when the records are consumed and re-written during retries in retryDLQ mode. 
- What is the need for this _metadata field if we're adding `shardIdColumn` in DLQ? shardIdColumn is only populated if Spanner has a dedicated column for shardId. For cases where there's no such column, this information is lost. 
- Callout: While bulk writes all the errors in the subfolder of the shard id (i.e. shard1 errors are written in shard1 folder), this information gets lost once Live Migration retryDLQ consumes it. 

### 3. Bulk DLQ adds shardIdColumn and _metadata_shard_id_column_name if Spanner DDL has a shardIdColumn
- `shardIdColumn` should conditionally be added based on presence of a column on Spanner as this is consumed as a source column by live migration code. If Spanner has this column, we provide this column value upfront via the DLQ. If it doesn't, theres no need to provide it as an additional column. (Even if we were to pass this, if it doesnt exist on Spanner it will be removed from the change event as per code in Live Migration)
- Clarification: We need to pass `shardIdColumn` most importantly to handle the edge case where Bulk DLQ only contains source row information but the corresponding Spanner schema has a `shardIdColumn`. Therefore we can't completely remove this field from the DLQ. 
- Note that an alternative was to use `_metadata_shard_id` to get the `shardIdColumn` value by editing Live Migration code to consume it, but this seems overly complicated. I wanted to decouple bulk retry from Live migration as much as possible without exposing live migration code to be aware of fields only relevant to Bulk DLQ. (Live migration code doesn't need `_metadata_shard_id` as it already provides datastream details that can be used to derive source)
- `_metadata_shard_id_column_name` is relevant in Live migration code as it makes the pipeline aware where the shard id is stored, and is used in metrics and DLQ downstream.
- Callout: I have edited `populateShardId` such that it will overwrite `_metadata_shard_id_column_name`, but note that this function is only called if session file is passed. In case of overrides, or no file being passed, this `_metadata_shard_id_column_name` needs to be populated from the DLQ itself (This might be fixed later but added in the DLQ for completeness and robustness)

## Testing
Major flows:
1. Non-sharded case (no shard file)
2. Sharded case with shard column (shard_id) in spanner schema
3. Sharded case without shard column

All the above tests will be tested with all different permutations of these:
1. Error types:
a. Transformation error (source row written to DLQ)
b. Spanner Writer error (Spanner mutation written to DLQ)

2. Configuration: (relevant as code paths are different)
a. Session file
b. Overrides

## Results from testing 
Everything works as expected.
- **Non sharded:** No shard field was added. On fixing issue, row is migrated successfully. 
- **Sharded without shard column:** `_metadata_shard_id` populated, no other metadata/column was added. `_metadata_shard_id` is retained even after live retry. On fixing issue, row is migrated successfully. 
- **Sharded with shard column:** `migration_shard_id `, `_metadata_shard_id_column_name ` and `_metadata_shard_id` are all populated. All flags are retained even after live retry. On fixing issue, row is migrated successfully. 

One Limitation was found: 
- Since overrides currently don't support passing `shardIdColumn`, the pipeline is not aware which column contains the shard id. (created b/487903664 to track this issue)
- This PR's code does work if the shard id column name is `migration_shard_id` as that was passed as the default value in case `schemaMapper.getShardIdColumnName` returned null. Thats why the above test passed. 
- Any other name will not work - i.e. the column and shard_id_column_name metadata will not be added to the DLQ and **it will break in retryDLQ mode.**
- This is NOT an issue within the scope of this bug and should be dealt with seperately

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
@@ -331,32 +331,7 @@ public class DeadLetterQueue implements Serializable {
       putValueToJson(json, entry.getKey(), val);
     }
 
-    // Attempt to extract and populate shard ID metadata
-    try {
-      if (this.shardId != null && !this.shardId.isEmpty()) {
-        json.put(Constants.EVENT_SHARD_ID, this.shardId);
-        populateShardIdColumnAndValue(json, m.getTable(), this.shardId);
-      } else {
-        // Just try to find if the mutation has the shard id column and populate it in metadata
-        // We know that if the mutation has the shard id column, it must have the shard id value
-        String shardIdColName = getShardIdColumnName(m.getTable());
-        if (mutationMap.containsKey(shardIdColName)) {
-          Value shardIdValue = mutationMap.get(shardIdColName);
-          if (shardIdValue != null && !shardIdValue.isNull()) {
-            String shardIdStr = shardIdValue.toString();
-            if (shardIdValue.getType().getCode() == Type.Code.STRING) {
-              shardIdStr = shardIdValue.getString();
-            } else if (shardIdValue.getType().getCode() == Type.Code.INT64) {
-              shardIdStr = Long.toString(shardIdValue.getInt64());
-            }
-            json.put(Constants.EVENT_SHARD_ID, shardIdStr);
-            json.put(Constants.SHARD_ID_COLUMN_NAME, shardIdColName);
-          }
-        }
-      }
-    } catch (Exception e) {
-      LOG.warn("Failed to populate shard ID metadata for DLQ mutation", e);
-    }
+    populateShardIdMetadataForMutation(json, m, mutationMap);
 
     return FailsafeElement.of(json.toString(), json.toString())
         .setErrorMessage("SpannerWriteFailed");
@@ -394,6 +369,31 @@ public class DeadLetterQueue implements Serializable {
     }
   }
 
+  private void populateShardIdMetadataForMutation(
+      JSONObject json, Mutation m, Map<String, Value> mutationMap) {
+    // Attempt to extract and populate shard ID metadata
+    try {
+      if (this.shardId != null && !this.shardId.isEmpty()) {
+        json.put(Constants.EVENT_SHARD_ID, this.shardId);
+        populateShardIdColumnAndValue(json, m.getTable(), this.shardId);
+      } else {
+        // Just try to find if the mutation has the shard id column and populate it in metadata
+        // We know that if the mutation has the shard id column, it must have the shard id value
+        String shardIdColName = getShardIdColumnName(m.getTable());
+        if (mutationMap.containsKey(shardIdColName)) {
+          Value shardIdValue = mutationMap.get(shardIdColName);
+          if (shardIdValue != null && !shardIdValue.isNull()) {
+            String shardIdStr = shardIdValue.toString();
+            json.put(Constants.EVENT_SHARD_ID, shardIdStr);
+            json.put(Constants.SHARD_ID_COLUMN_NAME, shardIdColName);
+          }
+        }
+      }
+    } catch (Exception e) {
+      LOG.warn("Failed to populate shard ID metadata for DLQ mutation", e);
+    }
+  }
+
   private void populateShardIdColumnAndValue(
       JSONObject json, String spannerTableName, String shardIdValue) {
     String shardIdColName = getShardIdColumnName(spannerTableName);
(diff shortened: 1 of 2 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -71,17 +72,30 @@ public class DeadLetterQueue implements Serializable {
 
   private final ISchemaMapper schemaMapper;
 
+  private final String shardId;
```
bharadwaj-aditya, line 75: Is an instance of the DLQ created per shard ?

It does not seem like it as all the other parameters are generic.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -309,6 +331,33 @@ protected FailsafeElement<String, String> mutationToDlqElement(Mutation m) {
       putValueToJson(json, entry.getKey(), val);
     }
 
+    // Attempt to extract and populate shard ID metadata
+    try {
```
bharadwaj-aditya: nit: Move this into a different method, this method is becoming too large.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -309,6 +331,33 @@ protected FailsafeElement<String, String> mutationToDlqElement(Mutation m) {
...
+        // We know that if the mutation has the shard id column, it must have the shard id value
+        String shardIdColName = getShardIdColumnName(m.getTable());
+        if (mutationMap.containsKey(shardIdColName)) {
+          Value shardIdValue = mutationMap.get(shardIdColName);
+          if (shardIdValue != null && !shardIdValue.isNull()) {
+            String shardIdStr = shardIdValue.toString();
+            if (shardIdValue.getType().getCode() == Type.Code.STRING) {
+              shardIdStr = shardIdValue.getString();
+            } else if (shardIdValue.getType().getCode() == Type.Code.INT64) {
+              shardIdStr = Long.toString(shardIdValue.getInt64());
```
bharadwaj-aditya: It might make sense to populate the value as a string here. The DLQ flow will need to parse it anyway.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -344,4 +393,38 @@ private void putValueToJson(JSONObject json, String key, Object value) {
...
+    String shardIdColName = null;
+    if (schemaMapper != null) {
+      try {
+        if (spannerTableName != null) {
+          shardIdColName = schemaMapper.getShardIdColumnName("", spannerTableName);
+          if (shardIdColName != null && !shardIdColName.isEmpty()) {
+            return shardIdColName;
+          }
+          shardIdColName =
+              "migration_shard_id"; // we can fallback to this as we explicitly check the Spanner
```
bharadwaj-aditya: use constants for this.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -215,12 +229,21 @@ protected FailsafeElement<String, String> rowContextToDlqElement(RowContext r) {
...
+    }
     if (r.row().shardId() != null) {
-      // Added default to not fail in the DLQ flow if the src table is not found in
-      // map
-      json.put(
-          srcTableToShardIdColumnMap.getOrDefault(r.row().tableName(), "migration_shard_id"),
-          r.row().shardId());
+      json.put(Constants.EVENT_SHARD_ID, r.row().shardId());
+      populateShardIdColumnAndValue(json, spannerTable, r.row().shardId());
+    } else if (this.shardId != null && !this.shardId.isEmpty()) {
```
bharadwaj-aditya, line 244: how does this work in the non-sharded flow ?

bharadwaj-aditya (approved):
Minor comments - but LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
aasthabharill: Bulk DLQ is written for each shard in separate folders. This DeadLetterQueue class is called by the MigrateTableTransform which takes shardId in its constructor - so it should be safe to pass shardId in the DLQ file as it is not shard agnostic even in the current system

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
aasthabharill: in non-sharded flow, `r.row().shardId() = null` and `this.shardId.isEmpty()` will return true as MigrateTableTransform has `shardId=""` for those cases

Issue #3421, opened 2026-02-27
[Bug]: StreamingDataGenerator fails with IllegalStateException: zip file closed on Dataflow workers


darshan-sj:
### Related Template(s)

streaming_data_generator

### Template Version

2026-02-23-00_rc00

### What happened?

The StreamingDataGenerator template intermittently fails during worker startup with a java.lang.IllegalStateException: zip file closed. This occurs during the initialization of the JsonDataGeneratorImpl within the MessageGeneratorFn setup phase.

**Root Cause**
The issue is caused by the org.reflections library (a dependency of json-data-generator) performing a classpath scan of the worker's "fat JAR" to register functions.

When multiple DoFn instances are initialized in parallel on a multi-core Dataflow worker, they concurrently attempt to scan the same JAR file. This leads to a race condition where one thread may close the ZipFile handle or a stream while another thread is still iterating through the entries, resulting in the IllegalStateException.

**Proposed Fix**
Thread-Safe Singleton Initialization: Implement a static provider for JsonDataGenerator using double-checked locking. This ensures the expensive and sensitive classpath scan is performed exactly once per Worker JVM, regardless of the number of threads.

### Relevant log output

```shell
exception: "java.lang.RuntimeException: org.apache.beam.sdk.util.UserCodeException: java.lang.IllegalStateException: zip file closed
	at org.apache.beam.runners.dataflow.worker.IntrinsicMapTaskExecutorFactory$1.typedApply(IntrinsicMapTaskExecutorFactory.java:202)
	at org.apache.beam.runners.dataflow.worker.IntrinsicMapTaskExecutorFactory$1.typedApply(IntrinsicMapTaskExecutorFactory.java:171)
	at org.apache.beam.runners.dataflow.worker.graph.Networks$TypeSafeNodeFunction.apply(Networks.java:67)
	at org.apache.beam.runners.dataflow.worker.graph.Networks$TypeSafeNodeFunction.apply(Networks.java:54)
	at org.apache.beam.runners.dataflow.worker.graph.Networks.replaceDirectedNetworkNodes(Networks.java:91)
	at org.apache.beam.runners.dataflow.worker.IntrinsicMapTaskExecutorFactory.create(IntrinsicMapTaskExecutorFactory.java:114)
	at org.apache.beam.runners.dataflow.worker.BatchDataflowWorker.doWork(BatchDataflowWorker.java:252)
[...]

PR #3422, opened 2026-02-27
[StreamingDataGenerator] Fix IllegalStateException: zip file closed in StreamingDataGenerator

darshan-sj:
This PR fixes a critical worker-side crash in the StreamingDataGenerator template. The issue occurs when multiple DoFn instances initialize simultaneously on a multi-core Dataflow worker, leading to a race condition during classpath scanning.

**The Problem**
The MessageGeneratorFn uses JsonDataGeneratorImpl, which relies on the Reflections library to register custom functions. This library scans the worker's "fat JAR" to find these functions.

- In the current implementation, every DoFn instance calls new JsonDataGeneratorImpl() in its @Setup method.
- On workers with multiple vCPUs, these threads concurrently attempt to iterate through the same JAR/Zip entries.
- This results in a java.lang.IllegalStateException: zip file closed when one thread interferes with the underlying file handle of another.

**The Fix**

- Thread-Safe Static Singleton: Introduced a getSharedGenerator() method using the double-checked locking pattern. This ensures that only the first thread to reach the setup phase performs the classpath scan. All other threads on the same worker JVM reuse the cached generator.
- Worker Stability: By serializing the initialization, we eliminate the race condition on the ZIP file handle.
- Performance Optimization: Reduces worker bootstrap time and CPU overhead by eliminating redundant, expensive classpath scans (which can take several seconds for large shaded JARs).

**Changes**

- Updated @Setup logic in StreamingDataGenerator.java to initialize the transient data generator field via singleton initialisation pattern.

Fixes: https://github.com/GoogleCloudPlatform/DataflowTemplates/issues/3421

### DIFF
--- a/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datagenerator/DataGenerator.java
+++ b/it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datagenerator/DataGenerator.java
@@ -39,6 +39,7 @@ import org.slf4j.LoggerFactory;
 /** Helper class for starting a Streaming Data Generator Dataflow template job. */
 public class DataGenerator {
   private static final Logger LOG = LoggerFactory.getLogger(DataGenerator.class);
+  // Revert to latest release path after next release. (b/488254537)
   private static final String SPEC_PATH =
       "gs://cloud-teleport-testing/templates/flex/Streaming_Data_Generator";
   private static final String PROJECT = TestProperties.project();
--- a/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
+++ b/v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/StreamingDataGenerator.java
@@ -672,6 +672,10 @@ public class StreamingDataGenerator {
 
     @Setup
     public void setup() {
+      // Use a static singleton to prevent concurrent classpath scanning.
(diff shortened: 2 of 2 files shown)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/datagenerator/DataGenerator.java
```diff
@@ -39,8 +39,9 @@
 /** Helper class for starting a Streaming Data Generator Dataflow template job. */
 public class DataGenerator {
   private static final Logger LOG = LoggerFactory.getLogger(DataGenerator.class);
+  // Revert to latest release path after next release. (b/488254537)
```
damccorm, line 42: Do we need to revert? It may make more sense to just always use the version we publish to cloud-teleport-testing, that will help us catch any issues we might push

darshan-sj: Do we have an automation which builds and pushes templates to cloud-teleport-testing? I'm not aware. I manually staged the template to cloud-teleport-testing and put the link here.

damccorm: Oh - I see, I thought this was already being done automatically. Please ignore my comment then, we don't need to build new infra for this :)

PR #3434, opened 2026-03-03
more pom properties refactoring

derrickaw:
1. Noticed more duplicate pom properties in #3431/changes, so just refactoring a little more.
2. All these removals are already covered in the root pom file.


Abacn (approved):
Thanks!

PR #3438, opened 2026-03-04
Added foreign key and indexes in spanner-common to be consumed by data generator pipeline

shreyakhajanchi:
This PR refactors the schema scanning logic to capture all indexes (both unique and non-unique) and detailed foreign key information from MySQL sources. 

It also modifies Spanner DDL to return index objects instead of strings


shreyakhajanchi:
> This might be outside the context of current PR as it's adding things to existing code but curious why have we stringified the index?

Not sure, why that was done earlier, but there is no usage of it, therefore modified it to be Index object instead of String

PR #3399, opened 2026-02-24
Graph Size Optimization Milestone-1 via Multi-Table (within a shard) Read.

VardhanThigle:
# PR Description: Graph Size Optimization via Multi-Table Read Orchestration

## Executive Summary
This project ensures that `sourcedb-to-spanner` Dataflow template to support large-scale database migrations (5,000+ tables) within a single Dataflow job. The current milestone is aimed at supporting migration of multiple tables within a given shard. The core optimization decouples the Dataflow graph size from the number of tables being processed, moving from an **O(N) graph complexity** (where N is the number of tables) to an **O(1) constant-size graph**. The code is structured in a way to easily accomodate multi-shard use case as a followup. 

## Scope of this PR
This is PR-0 and it's aimed at giving a complete picture of the proposed changes to the reviewers. 

## The Problem: Graph Size Bottleneck
Previously, the `ReadWithUniformPartitions` (RWUP) transform instantiated a unique set of Dataflow transforms for every table. For migrations involving thousands of tables, this resulted in:
1.  **Deployment Failures:** Dataflow graphs exceeding the maximum size limit (e.g., >10MB).
2.  **Performance Degradation:** Severe overhead in graph construction, job submission, and worker-side transform management.
3.  **Scalability Limits:** Inability to migrate physical database shards containing many tables in a single unified pipeline.

## The Solution: Multi-Table Orchestration
We have refactored the partitioning and reading engine to be **data-driven** rather than **configuration-driven**. Instead of many table-specific graphs, we now use a single, generic graph that processes a heterogeneous stream of `Range` objects. Each object carries its own context (table name, partition columns, etc.), and the internal `DoFn`s dynamically generate SQL queries based on this context.

#### 1. Constant-Size Dataflow Graph
The new architecture uses a single iterative loop that processes ranges for all tables simultaneously. This ensures the graph remains the same size whether processing 1 table or 5,000 tables. There will be multiple instances of this graph as per the dependency levels, but there will be a natural limit on how long the chain of interleaving (or FKs) could practically grow.

#### 2. The "Graduation Lane"
To prevent "straggler" tables (large or complex tables) from blocking smaller tables, we introduced the **Graduation Lane**. 
- Tables that are easily partitioned "graduate" early from the iterative splitting loop.
- Graduated ranges move immediately to the read phase, while complex ranges continue to be refined in the loop.
- This dramatically improves resource utilization and overall pipeline throughput.

#### 3. Parallelized Schema Discovery
[...]

PR #3441, opened 2026-03-05
Graphsize Multi Table Reader.

VardhanThigle:
# PR Description: Adding MultiTable Read Transform to optimize Graph Size

## Summary
This PR introduces the **Multi-Table Read Engine** and integrates it into the `sourcedb-to-spanner` template. It is a critical component of the O(1) Graph Size optimization, enabling the processing of 5,000+ tables without hitting Dataflow graph limits.

This is (second) child of [pull/3399](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3399). Please refer to the parent PR to get a complete context on the changes involved.

## Key Changes
- **`MultiTableReadAll` Transform:** A new high-performance reader that handles heterogeneous ranges from multiple tables using connection pooling.
- **`IoWrapper` Refactor:** Updated the interface to support grouped table readers, allowing for unified partitioning across multiple tables.
- **Memory-Safe Batching:** Introduced `AssignSyntheticKeyDoFn` to enable keyed-batching of range combinations, preventing OOM errors when processing millions of range objects.
- **Lineage Integration:** Full support for Beam Lineage API, reporting all unique source tables read during the job.

## Design Considerations
- **Thread Safety:** All `DoFn`s (including `MultiTableReadFn`) follow the Beam lifecycle where instances are thread-confined. Connections are managed per-bundle to optimize overhead.
- **Cassandra Compatibility:** While the primary focus is JDBC, the `IoWrapper` changes have been applied to Cassandra to maintain full project integrity.
- **Error Resilience:** Added fail-safe logic in range splitting to ensure that metadata errors in a single table do not crash the entire migration job.

## Validation
- **Unit Tests:** >98% coverage on new transforms.
- **Scale Test:** Successfully validated with a 5,000-table migration (MySQL & PostgreSQL) - on Parent PR #3399  (We still need parallel schema discovery to reach this number).
- **Regression:** Verified single-table migration remains unaffected.
- ** Existing Integration tests** - Existing Integration tests like MySQLAllDatatypes will run a 109 tables migration in. a single stage Dataflow job on this PR.

#### 1. Constant-Size Dataflow Graph
The new architecture uses a single iterative loop that processes ranges for all tables simultaneously. This ensures the graph remains the same size whether processing 1 table or 5,000 tables. There will be multiple instances of this graph as per the dependency levels, but there will be a natural limit on how long the chain of interleaving (or FKs) could practically grow.

#### 2. The "Graduation Lane"
To prevent "straggler" tables (large or complex tables) from blocking smaller tables, we introduced the **Graduation Lane**. 
- Tables that are easily partitioned "graduate" early from the iterative splitting loop.
- Graduated ranges move immediately to the read phase, while complex ranges continue to be refined in the loop.
- This dramatically improves resource utilization and overall pipeline throughput.

#### 3. MultiTableReadAll
A custom `PTransform` designed to consume the final collection of ranges from all tables and execute the reads efficiently. It manages connection pools per worker and integrates with Beam's Lineage API for enhanced observability.

---

### Before: O(N) Graph Complexity
*(Each box represents a set of transforms created for EVERY table)*

existing_architecture

### After: O(1) Constant Graph Size

new_architecture

### Core Logic
- **`ReadWithUniformPartitions` Refactor:** Shifted from single-table fields to a `List<TableSplitSpecification>`.
- **`RangeClassifierDoFn`:** Logic for directing ranges to either further refinement or the graduation lane.
- **`MultiTableReadAll` / `MultiTableReadFn`:** New efficient reader for heterogeneous range collections.

## Todos
These items will be covered in followup activities
1. Monitoring/metrics for per-table ranges  produced and read.
2. Multi Shard support.

## Out of scope
These items are out of scope of the current work:

1. Checkpointed migrations.
2. Fixed graph size for multi level migrations.
3. Any architectural changes.


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/AssignSyntheticKeyDoFn.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2024 Google LLC
```
manitgupta: Nit: New files should have 2026 I guess?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
```diff
@@ -0,0 +1,387 @@
...
+   * @param queryProvider the query provider value provider.
+   * @return a new transform instance.
+   */
+  public MultiTableReadAll<ParameterT, OutputT> withQueryProvider(
+      ValueProvider<QueryProvider> queryProvider) {
+    checkArgument(
+        queryProvider != null, "withQueryProvider(queryProvider) called with null queryProvider");
+    return toBuilder().setQueryProvider(queryProvider).build();
+  }
+
```
manitgupta, line 209: We should evaluate if `@AutoValue` is really driving any value in delcaring this `PTransform `? Here, `withQueryProvider` encapsulates some business logic and then ends up calling `setQueryProvider()` just to set a member field indirectly. Maybe we should just set the member  in `withQueryProvider` directly. 

One way to think about it is that do we ever expect a customer of `MultiTableReadAll` transform to use `setQueryProvider` directly? If not, we are probably better off defining our own fluent interfaces, instead of mixing the `AutoValue.Builder` with our custom definitions..

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+ * <ul>
+ *   <li>Selecting the correct {@link RowMapper} and {@link QueryProvider} for each {@link Range}.
+ *   <li>Adjusting {@code fetchSize} on a per-table basis.
+ *   <li>Reporting data lineage to the Dataflow service for each unique source table encountered.
+ * </ul>
+ *
+ * <p>By using a single {@code DoFn} for all tables, we maintain a constant-size Dataflow graph
+ * regardless of the number of tables in the migration.
+ */
+public class MultiTableReadFn<ParameterT, OutputT> extends DoFn<ParameterT, OutputT> {
```
manitgupta, line 62: Consider segregating DoFn and PTransforms into their own folders `transforms` and `dofn`. Currently everything is in `transforms`..

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+          }
+        }
+      } finally {
+        connectionLock.unlock();
+      }
+
+      reportLineage(element, connection, validSource, query, reportedLineages);
+    }
+    return connection;
+  }
```
manitgupta, line 141: I didn't understand this, can you explain the intuition behind having 1 connection per DoFn? Why do we need to manage this manually?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+   * <p>This method is thread-safe and ensures that lineage for a specific table is reported only
+   * once per worker bundle to minimize overhead.
+   *
+   * @param element the input element (e.g., a Range) containing information to derive the query.
+   * @param connection an active database connection.
+   * @param validSource the data source used to establish the connection.
+   * @param query the provider for the SQL query.
+   * @param reportedLineages a set used to track and deduplicate reported lineages.
+   * @throws Exception if an error occurs during lineage extraction or reporting.
+   */
```
manitgupta, line 181: This is interesting - can you show screenshots of how does this actually manifest in the Dataflow UI?
I have not really used the `Lineage` tab before so I'm curious to see how it improves the observability of the pipeline..

**Note** - I am not sure if it is was possible to break this up, but it would have been easier to review if we added/implemented the "lineage feature" after the basic implementation of graph size agnostic bulk migration was done..it would have simplified things..

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+      if (fqn != null) {
+        fqn.reportLineage(Lineage.getSources(), schemaWithTable);
+      }
+    }
+  }
+
+  /** Extract schema and table name a SELECT statement. Return null if fail to extract. */
+  private static final Pattern READ_STATEMENT_PATTERN =
+      Pattern.compile(
+          "SELECT\\s+.+?\\s+FROM\\s+(\\[?`?(?<schemaName>[^\\s\\[\\]`]+)\\]?`?\\.)?\\[?`?(?<tableName>[^\\s\\[\\]`]+)\\]?`?",
```
manitgupta, line 211: Usual comment on if this regex will accurately handle table names with special characters etc.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+          context.output(rowMapper.mapRow(resultSet));
+        }
+      }
+    }
+  }
+
+  @FinishBundle
+  public void finishBundle() throws Exception {
+    cleanUpConnection();
+  }
```
manitgupta, line 271: Why do we need to clean up connection per bundle? Won't this be expensive? 

Should we be keeping a connection alive for the duration of the DoFn?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+  @ProcessElement
+  public void processElement(ProcessContext context) throws Exception {
+    // Only acquire the connection if we need to perform a read.
+    ParameterT element = context.element();
+    TableIdentifier tableIdentifier = tableIdentifierFn.apply(element);
+    TableReadSpecification<OutputT> spec = tableReadSpecifications.get(tableIdentifier);
+    if (spec == null) {
+      throw new RuntimeException("TableReadSpecification not found for table: " + tableIdentifier);
+    }
+    Connection connection = getConnection(element);
```
manitgupta, line 251: Should we not get the connection in `@Setup`? Create the connection when the DoFn is deployed and use it for the lifecycle of the DoFn?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+  public void tearDown() throws Exception {
+    cleanUpConnection();
+  }
+
+  private void cleanUpConnection() throws Exception {
+    if (connection != null) {
+      try {
+        connection.close();
+      } finally {
+        connection = null;
```
manitgupta, line 283: If the connection does not actually close, this will still set it to `null`. Can this create connections in Zombie state?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/Reparallelize.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p><b>Timestamp Side Effect:</b>
+ *
+ * <p>Because this fusion break uses a side-input generated in the {@link
+ * org.apache.beam.sdk.transforms.windowing.GlobalWindow}, all output elements inadvertently inherit
+ * a timestamp of {@code TIMESTAMP_MAX_VALUE}. To correct this, this transform applies {@link
+ * TimeStampRow} as a final step to reset the timestamps back to the current system time, ensuring
+ * compatibility with downstream components like the Dead Letter Queue (DLQ).
+ *
+ * @param <T> The type of the element being processed.
+ */
```
manitgupta: This way of breaking fusion seems like a very hard set way to do it - 

Q1. What is our path to undoing fusion if we don't need it in the future (will we always need it?). 

Q2. Why do the standard GPK based transforms (Reshuffle, Redistribute) don't work here to break fusion? It seems like we have implemented a alternative shuffle transform here..

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/TimeStampRow.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * <p>In the multi-table read pipeline, certain upstream transforms (like {@link Reparallelize} or
+ * those within {@link ReadWithUniformPartitions}) consume side-inputs generated in the {@link
+ * org.apache.beam.sdk.transforms.windowing.GlobalWindow}. In Apache Beam, side-inputs in the Global
+ * Window are only considered "ready" at the end of the window ({@code TIMESTAMP_MAX_VALUE}).
+ * Consequently, any element produced by a transform consuming such a side-input inherits this "End
+ * of Time" timestamp (approx. year 294247).
+ *
+ * <p>This causes downstream failures, specifically in the Dead Letter Queue (DLQ) path. The DLQ
+ * writer attempts to assign a "current" timestamp ({@link Instant#now()}) to failed records. Since
```
manitgupta, line 35: Much of this complexity is stemming from having to use a side-input to break the fusion...
I am not sure if this is the ONLY way to do this - but in it's current state if seems overly complicated and hard to understand to me...

Let's discuss this a bit with @bharadwaj-aditya and maybe someone from Dataflow as well so that we all get confidence that THIS is the only solution to the problem you are trying to solve...

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Datalineage is a new feature introduced in Apache beam. Please refer to https://beamsummit.org/sessions/2024/data-lineage/ for more details.

Also, We have merged `FQDN` and related classes that handled most of the lineage work in the previous PR - so it's broken into a few pieces already :)

Here we are calling into the same code from the MultiTableReadTransform.

The approach Apache beam has taken is that in genearal a pipeline need not worry about the lineage aspects as long as it's using Apache Beam based IO modules (for example `JDBCIO.readAll`) that's taken care of automatically. Here sine we can not use the existing `ReadAll` we need to report lineage manually. 

A small note: I belive we will need to do the same in other templates for example RR template as well (As we don't use any beam-IO module there) but that's a separate discussion.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/Reparallelize.java
VardhanThigle: This is not enforced. The MultiTableReadAll calls ReParallelize just as the singleTable one does - based on the build time options.

This Class is same as what we have for the single table equivalent - https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L2561

Unfortunately that one is private. I have added a tracking bug to make these public and remove the clones here. https://b.corp.google.com/issues/489293836

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/TimeStampRow.java
VardhanThigle: The collation information is  also provided as a side input. `ReParalleize` is entirely borrowed from the current single table implentation

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: This is similar to the extractor used in existing Single TableCase. Will remove it once we make it public

https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L836

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: I agree. 
This repo started w/ segarating POJOS from transforms and doFns, with this change we have got many of them so it makes sense to segarate further.  Segregating just for this change will not be meaningful, we might want to segarate older ones as well for avoiding confusion. 

At the moment I am trying to contain number of changes. Will do this as a small refactor once all my PRs are merged.

This is tracked in b/490297752.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
VardhanThigle: I might lean on exploring further if we are seeing any issue with Autovalue?


Dataflow generally recommends AutoValue as a Java best practice. Here are 2 things it gives us
1. It easily gives us serializability and immutability - PTransform is generally not mutable in run time.
2. It provides implementation for hashCode which are compliant with what Beam needs. 

All the transforms in ApacheBeam repo are AutoValue Tansforms, including [ReadAll](https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L366) which we are replacing.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: This is quite similar to what we do in `JdbcIO.readAll`.

In [JdbcIO.readAll](https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1766) The first call to `processElement` opens the connection and FinishBundle closes it. This helps in not having a stray or almost zombie connection as a DoFN might not get elements for some time as per Dataflow's scaling considerations. So doing it at the level of a Bundle is a better tradeoff IMHO. 

Our Bundle here is a group of Ranges. So it's going to be a long running RangeRead and not a one-shot point read. So from that POV, we make a couple of network calls to open a connection for reading O(1K to 10K)  rows and with this tradeoff we are getting more fresher connections.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Please refer to the thread [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3441#discussion_r2894585346) for all connection related discussions.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Please refer to the thread [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3441#discussion_r2894585346) for all the connection related discussion.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Please refer to the comment [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3441#discussion_r2894585346) for all the connection related discussion.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: If the connection does not close, `close()` will throw an exception - we are not catching that, so there's no silent ignore.  

This will happen only in cases where we have lost access to the DB. 
Please refer to [Connection.close](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#close--)- in such a case the DoFn would always need to reopen the connection - so making it null makes sense.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/TimeStampRow.java
VardhanThigle: As discussed The ReParalleize is introduced as used in ApacheBeam's existing framework (to circumvent an internal bug) 
We are not introducing the complexity here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Also, this is similar to what existing Beam code does [here](https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1790),

Till now we have not seen any major performance or concurrancy issues with ReadAll, so I don't see a very strong reason to change it.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
```diff
@@ -0,0 +1,387 @@
...
+ * Consequently, any element produced by a transform consuming such a side-input inherits this "End
+ * of Time" timestamp (approx. year 294247). This can cause downstream failures, specifically in the
+ * Dead Letter Queue (DLQ) path. To address this, {@link TimeStampRow} is used to explicitly reset
+ * the timestamp of elements to the current system time.
+ *
+ * @param <ParameterT> the type of the input parameter elements.
+ * @param <OutputT> the type of the data read from the database.
+ */
+@AutoValue
+public abstract class MultiTableReadAll<ParameterT, OutputT>
```
bharadwaj-aditya, line 80: Is the goal here to cover all the functionality that JDBCIO.ReadAll provides ?

If yes, most of this builder should look similar to JDBCIO - apart from the additional functionality.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
bharadwaj-aditya: Seems like the pattern is being followed in JDBC IO as well - eg: https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L824

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+   *
+   * @param query the SQL query string.
+   * @return a {@link KV} containing (schema, table), or null if extraction fails.
+   */
+  @VisibleForTesting
+  protected static @Nullable KV<String, String> extractTableFromReadQuery(@Nullable String query) {
+    if (query == null) {
+      return null;
+    }
+    Matcher matchRead = READ_STATEMENT_PATTERN.matcher(query);
```
bharadwaj-aditya, line 225: This seems quite brittle to me. I'm not too sure if a single regex can encapsulate every way a table can be declared.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -0,0 +1,283 @@
...
+          Pattern.CASE_INSENSITIVE);
+
+  /**
+   * Extracts the schema and table name from a SELECT statement using regex.
+   *
+   * @param query the SQL query string.
+   * @return a {@link KV} containing (schema, table), or null if extraction fails.
+   */
+  @VisibleForTesting
+  protected static @Nullable KV<String, String> extractTableFromReadQuery(@Nullable String query) {
```
bharadwaj-aditya, line 221: It seems like a lot of code is copied from existing methods. Is there a way to reference those methods instead ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
bharadwaj-aditya: This is a common pattern with things like connections, i.e. do it in best effort mode. The most common reason for a connection to not close is either network issues or the source db not responding. In both those cases, timeouts will kick in and clean this stuff up.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/Reparallelize.java
```diff
@@ -0,0 +1,78 @@
...
+public class Reparallelize<T> extends PTransform<PCollection<T>, PCollection<T>> {
```
bharadwaj-aditya, line 50: same comment as above. Can we use existing classes instead of re-implementing all of this ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/AssignSyntheticKeyDoFn.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A {@link DoFn} that assigns a synthetic integer key to each {@link Range} based on its {@link
+ * com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range.TableIdentifier}.
+ *
+ * <p>This enables downstream transforms like {@link RangeCombiner} to process ranges in parallel,
+ * memory-safe batches (keyed-combinations) rather than attempting a global combination that could
+ * lead to Out-Of-Memory (OOM) errors when dealing with millions of ranges from thousands of tables.
+ */
+@AutoValue
+public abstract class AssignSyntheticKeyDoFn extends DoFn<Range, KV<Integer, Range>> {
```
bharadwaj-aditya, line 32: Do we need a DoFn for this ? Don't a simple utility do ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/InitialSplitRangeDoFn.java
```diff
@@ -17,56 +17,69 @@
...
+   * @param c process context.
    */
   @ProcessElement
   public void processElement(
       @Element Range input, OutputReceiver<ImmutableList<Range>> out, ProcessContext c) {
     // Note Searching for "RWUPT -" for ReadWithUniformPartition gives the most import logs for the
     // splitting process.
     logger.info(
-        "RWUPT - Began split process for table {} with initial range as {}", tableName(), input);
+        "RWUPT - Began split process for table {} with initial range as {}",
```
bharadwaj-aditya, line 62: the acronyms are hard to follow. Suggest using proper names here instead

bharadwaj-aditya: What is the number of times we expect this log ? Is there a possibility of getting throttled ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -468,54 +479,98 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+          TableSplitSpecification.builder()
+              .setTableIdentifier(tableIdentifier)
+              .setPartitionColumns(tableConfig.partitionColumns())
+              .setApproxRowCount(tableConfig.approxRowCount());
+      if (tableConfig.maxPartitions() != null) {
+        tableSplitSpecificationBuilder =
+            tableSplitSpecificationBuilder.setMaxPartitionsHint((long) tableConfig.maxPartitions());
+      }
+      if (config.splitStageCountHint() >= 0) {
+        tableSplitSpecificationBuilder =
```
bharadwaj-aditya, line 526: This seems like it will be set for every table. Not clear how a static value would make sense here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java
```diff
@@ -249,6 +254,9 @@ public boolean isMergable(Range other) {
       }
       return this.childRange().isMergable(other.childRange());
     } else {
+      if (!this.boundary().tableIdentifier().equals(other.boundary().tableIdentifier())) {
```
bharadwaj-aditya, line 257: Seems like i saw this change in another commit as well :)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
```diff
@@ -69,26 +82,66 @@ public PCollectionTuple expand(PBegin input) {
...
-                             * As needed, more complex reduce operation can be performed providing additional information to the pipelineController.
-                             * TODO(vardhanvthigle): Add a note on cost vs performance of the Count transform after benchmarking.
-                             */
-                            e.getValue()
-                                .apply("TableRowCounter." + e.getKey().getName(), Count.globally())
-                                .apply(
-                                    "TableCompletion." + e.getKey().getName(),
-                                    MapElements.via(new SourceTableReferenceWithCount(e.getKey()))))
+                        e -> {
+                          ImmutableList<SourceTableReference> tableReferences = e.getKey();
```
bharadwaj-aditya: readability and testing nit. This is nested a bunch of times. May make sense to extract into a method.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/ExtractTableNameFn.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.cloud.teleport.v2.source.reader.io.schema.SourceTableReference;
+import org.apache.beam.sdk.transforms.DoFn;
+
+/**
+ * DoFn to extract the table name from a {@link SourceRow} and delimit it to match the format used
+ * in {@link SourceTableReference}.
+ */
+class ExtractTableNameFn extends DoFn<SourceRow, String> {
+  @ProcessElement
+  public void processElement(@Element SourceRow row, OutputReceiver<String> out) {
```
bharadwaj-aditya, line 28: same question. Does this need to be a do fn ? Can it not be a util ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
VardhanThigle: That's right. The builder will change for taking in list of tables (for Example we take a list of TableIdentifier), similarly for RowMapper. Rest of the properties are same. This class started out as the existing code and then the new changes were updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: This is borrowed from the  Beam's [singleTableReader](https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L851). It's used for lineage reporting.

We are tracking removing this in b/489293836

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: They are private at the moment. We are tracking b/489293836 to remove the same.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/Reparallelize.java
VardhanThigle: Same reply :). They are private. We are tracking b/489293836 for the removal.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/AssignSyntheticKeyDoFn.java
VardhanThigle: We will need to implement a DoFn interface as the incoming element is PElement. Rest of the code is just similar to utility code.  It could be inlined where it's pApplied but we have experienced increased graph sizes with inlined DoFns, so kept it clean by adding an independent DoFns.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/InitialSplitRangeDoFn.java
VardhanThigle: `RWUPT` helps as a short key to search the logs, bigger words are difficult or prone to fat fingering there :).
The full class name is always part of the log, so that could be used to search as well if needed.
 
We are not adding this acronym new as a part of this PR.
The same key is used in the existing ReadWithUniformTransfom and this log is just adding the table to existing log.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: This is the flag for enabling uniformization. At the moment it's not a table level parameter it's a dataflow level parameter. 
Agree that We would like to make it table level like `fetchsize`, but that's currently outside the scope of current PR. I have a todo to make this table level with the future work on uniformization.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java
VardhanThigle: There might be similar change in Boundary. But there's no merge conflict here :).

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/ExtractTableNameFn.java
VardhanThigle: We need to implement the DoFN function when we reduce the produce rows into per-table counts for completion - Implementing the same code in a clean doFn avoids inlined doFns which we want to avoid (for increased graphSize at all costs)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/AssignSyntheticKeyDoFn.java
VardhanThigle: Fixed in current push. Thanks.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/transform/AccumulatingTableReader.java
VardhanThigle: Makes sense. Thanks for catching this. extracted the lambda a function

VardhanThigle:
> * Please add TODOs in the classes that had to be copied from beam due to their private nature for removal. Please make the Javadoc at the start of the class, crisp and clear on the source of the class (that should be the first thing, the explanation (if you want to keep it) can be added after that.

Thanks. The TODOs are added. Will share the b/ separately. 

> * Lets raise the PR to make them public and get the conversation going on making these `public` - Having clarity on that (if beam has not objections in making them public) can inform how we want to approach the `TODOs` - for example, if beam does not agree to change the visibility, we need to approach the copied code from a different mental model (temporary vs permanent)

Will do once we merge this in.

> * Since this is a large change, please get an LGTM for Aditya as well.

Sure.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java
```diff
@@ -56,8 +65,40 @@ public void setParameters(
...
+
     int parameterIdx = 1;
-    new RangePreparedStatementSetter(partitionCols.size())
-        .setRangeParameters(element.parentRange(), preparedStatement, parameterIdx);
+    for (PartitionColumn partitionColumn :
+        tableSplitSpecificationMap.get(element.tableIdentifier()).partitionColumns()) {
+
+      if (partitionColumn.columnName().equals(element.columnName())) {
+        // For the target column, we want to scan all values, so we disable the predicate.
+        preparedStatement.setObject(parameterIdx++, false); // Disabled
```
bharadwaj-aditya, line 82: why are there 5 parameters here. Can you link to the query or add some docs for this ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadAll.java
```diff
@@ -0,0 +1,387 @@
...
+      TypeDescriptor<OutputT> typeDesc = coder.getEncodedTypeDescriptor();
+      SchemaRegistry registry = input.getPipeline().getSchemaRegistry();
+      Schema schema = registry.getSchema(typeDesc);
+      output.setSchema(
+          schema,
+          typeDesc,
+          registry.getToRowFunction(typeDesc),
+          registry.getFromRowFunction(typeDesc));
+    } catch (NoSuchSchemaException e) {
+      // ignore
```
bharadwaj-aditya, line 347: is this fine to ignore ? If yes can add context as to why ? It is generally odd to ignore and not even log an exception

VardhanThigle: Thanks for the detailed review, @bharadwaj-aditya 
Yes, that's safe.


This ignore is intentional as Beam Schema registration is a best-effort enhancement for the PCollection.
 At this stage in the expand() method (line 347), the PCollection already has a valid Coder assigned (at line 335), ensuring the pipeline is fully functional. The NoSuchSchemaException simply indicates that the specific output type hasn't been registered in the Beam SchemaRegistry.  Importantly, we already log a descriptive INFO message during the inferCoder() phase (line 298) if the schema is missing. Logging it again here would be redundant and would create unnecessary log noise for users who are not utilizing Beam Schema
 
 **This is the same pattern followed by the Apache Beam's existing single table reader [here](https://github.com/apache/beam/blob/d27dc82c56882e38a1df43ec90706f14811d50fb/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1317)**
 
 I will add a more detailed explanation on this in the comments in one of my upcoming PRs.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java
VardhanThigle: This is refactored from the existing code [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/36935dc235a589d988a56da8aca7167a545103b4/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetter.java#L59)

We had to split the earlier parameter setter into 2 as in the previous world the table was known at construction time. In this world, the table is extracted from the Range during actual reads, for extracting boundaries though that does not work as there's no range when the first min-max boundary is extracted.

It's not entirely a new thing that the PR is adding, it's a multitable version of the same thing.


We have tested this in various ways. The query can be found in the unit test against a Derby instance on line 517 of v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetterTest.java in this PR.
Will add db-fiddle links in my upcoming PR

VardhanThigle:
> Seems ok based on the previous pass.
> 
> For future reference , Please keep PRs under 1000-1500 lines.

Thanks a ton for the detailed review @bharadwaj-aditya  and @manitgupta .  :)

As we have discussed, the graphsize bit is a complex task overall - We have split this into 2 milestons -
Milestone-1 variable tables in a single shard
Milestone-2 variable number of shards

The Milestone-1 has been broken into 4 PRs (as the description of this PR and states), for ease of review we also have all the 4 PRs together in #3399 

Among the 4 PRs of Milestone-1, this is the second one and it's hard to break this one as we are integrating into the refactoring and integrating into existing code and we don't want either of these 3 natural constraints:
1. The build or tests to break in existing flow should not breaks.
2. Each PR should go with high UT coverage - so we might not want to split code and UT (it's gets akward to review)
3. Avoiding artificial do and undo changes - like adding a fields and ignoring beam null checkers for 1 PR and undoing that once the field is populated in next.

PR #3442, opened 2026-03-05
Adding model classes for data generator

shreyakhajanchi:


v2/cdc-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorColumn.java
```diff
@@ -0,0 +1,80 @@
...
+  public abstract LogicalType logicalType();
+
+  /** Whether the column is nullable. */
+  public abstract boolean isNullable();
+
+  /** Whether the column is skipped. */
+  public abstract boolean isSkipped();
+
+  /** Whether the column is a generated column. */
+  public abstract boolean isGenerated();
```
manitgupta, line 39: Where is this used?

v2/data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorSchema.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public static Builder builder() {
+    return new AutoValue_DataGeneratorSchema.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+    public abstract Builder tables(Map<String, DataGeneratorTable> tables);
+
+    public abstract Builder tables(ImmutableMap<String, DataGeneratorTable> tables);
```
manitgupta: Why two of these? Map is the interface and will also accept `ImmutableMap` right?

v2/data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorTable.java
```diff
@@ -0,0 +1,108 @@
...
+  @Nullable
+  public abstract ImmutableList<String> childTables();
+
+  public static Builder builder() {
+    return new AutoValue_DataGeneratorTable.Builder()
+        .isRoot(true)
+        .insertQps(0)
+        .updateQps(0)
+        .deleteQps(0)
+        .recordsPerTick(1.0);
```
manitgupta: Why do you need these defaults? Wouldn't it be better if as part of building the `DataGeneratorTable` all of these configurations are immutably defined and stored in the object?

v2/pom.xml
```diff
@@ -701,6 +701,7 @@
         <module>bigquery-to-parquet</module>
         <module>cdc-parent</module>
         <module>common</module>
+        <module>data-generator</module>
```
manitgupta: Naming is very, very broad. Just naming it to something that tells us its related to Spanner in some shape or form?

v2/cdc-data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorColumn.java
shreyakhajanchi: it will be referred to in order to skip the generated columns during data generation

v2/data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorSchema.java
shreyakhajanchi: removed the `ImmutableMap` , it was an optimization so that autovalue doesn't create copy

v2/pom.xml
shreyakhajanchi: didn't want to keep naming related to spanner as we are generating data for other sinks, do you have a suggestion in mind?

v2/data-generator/src/test/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorTable.java
shreyakhajanchi: sure, will remove the defaults

v2/pom.xml
shreyakhajanchi: renamed it to cdc-data-generator

PR #3451, opened 2026-03-06
fix spotless error

derrickaw:
caused by https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3439

PR #3455, opened 2026-03-06
add gcs-spanner-dv path to spanner-pr.yaml workflow

derrickaw:
Continuation of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3451 and #3439


shunping:
What is this template? Should we add a README.md to it just like other folders?

derrickaw:
> What is this template? Should we add a README.md to it just like other folders?

Hi @shunping, So the issue is that a previously submitted PR didn't have any test done on it, which didn't catch a spotless issue that broke many workflows.  So fixed the spotless issue.  This PR is just to add the folder to the spanner-pr workflow.  I don't think they have a formal template yet. Thanks.

PR #3458, opened 2026-03-09
PubSub to ClickHouse Flex Template

BentsiLeviav:
### Description 

Adds a new Flex Template that reads JSON messages from a Pub/Sub subscription and writes them to a ClickHouse table using the ClickHouseIO connector.

###  Key features
  - Schema auto-detected at launch from the target ClickHouse table
  - Flexible windowing: time-based, count-based, or combined (whichever fires first)
  - Dual dead-letter routing: failed messages can go to a ClickHouse table, a Pub/Sub topic, or both simultaneously
  - Configurable write tuning: max insert block size, retries, deduplication, distributed sync, quorum

  Includes:
  - Unit tests covering the parsing DoFn, type handling (String, Float32, Float64, DateTime, Array), and DLQ routing
  - Integration tests covering the happy path, both DLQ destinations, numeric/DateTime columns, and time-window mode


v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/options/PubSubToClickHouseOptions.java
```diff
@@ -0,0 +1,233 @@
...
+  // -------------------------------------------------------------------------
+
+  @TemplateParameter.Long(
+      order = 11,
+      optional = true,
+      description = "Max ClickHouse insert block size",
+      helpText =
+          "Maximum number of rows per INSERT statement sent to ClickHouse. Defaults to 1,000,000.",
+      example = "10000")
+  @Default.Long(1000000L)
```
mzitnik: I will reduce it to `100,000` instead of `1,000,000`

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
```diff
@@ -0,0 +1,464 @@
...
+
+    LOG.info("Fetching ClickHouse schema for table: {}", options.getClickHouseTable());
+
+    TableSchema clickHouseSchema =
+        ClickHouseIO.getTableSchema(
+            options.getClickHouseUrl(),
+            options.getClickHouseDatabase(),
+            options.getClickHouseTable(),
+            props);
+    Schema beamSchema = TableSchema.getEquivalentSchema(clickHouseSchema);
```
mzitnik, line 205: maybe we need to write some where how to create this table

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
```diff
@@ -0,0 +1,464 @@
...
+                    stackTraceToString(e),
+                    DateTime.now().toString("yyyy-MM-dd HH:mm:ss"))
+                .build();
+
+        out.get(TRANSFORM_DEADLETTER_OUT).output(deadLetterRow);
+        rowsParseFailed.inc();
+      }
+    }
+
+    private static String stackTraceToString(Throwable t) {
```
mzitnik, line 469: We can also use it to avoid cyclic dependency, which can cause an endless loop.

```
private static String stackTraceToString(Throwable t) {
    StringWriter sw = new StringWriter();
    t.printStackTrace(new PrintWriter(sw));
    return sw.toString();
}
```

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/options/PubSubToClickHouseOptions.java
BentsiLeviav: I removed it completely, as it works today with the `BigQueryToClickHouse` template

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
BentsiLeviav: Added a show create table statement, with an indication to pay attention to setup variation for cluster setup/different engine usage.

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
BentsiLeviav: Good catch. Added your suggestion

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
```diff
@@ -0,0 +1,475 @@
...
+ *
+ * <p>When {@code --clickHouseDeadLetterTable} is set, the dead-letter table must exist before
+ * running the pipeline. Minimal example for a single-node deployment:
+ *
+ * <pre>{@code
+ * CREATE TABLE my_dead_letter (
+ *     raw_message   String,
+ *     error_message String,
+ *     stack_trace   String,
+ *     failed_at     String
```
mzitnik: What does this field contain? Should it be some timestamp?

BentsiLeviav: Changed to Datetime

v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/PubSubToClickHouse.java
```diff
@@ -0,0 +1,475 @@
...
+    description =
+        "The Pub/Sub to ClickHouse template is a streaming pipeline that reads JSON-encoded "
+            + "messages from a Pub/Sub subscription and writes them to a ClickHouse table. "
+            + "Messages that fail due to schema mismatch or malformed JSON are routed to a "
+            + "dead-letter destination: a ClickHouse table, a Pub/Sub topic, or both — "
+            + "determined implicitly by which of --clickHouseDeadLetterTable and "
+            + "--deadLetterTopic are provided. "
+            + "The target table and any ClickHouse dead-letter table must exist prior to execution.",
+    optionsClass = PubSubToClickHouseOptions.class,
+    flexContainerName = "pubsub-to-clickhouse",
```
Abacn: We now share containers for Templates within same maven module. Consider set both here and https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/bb033bccad18bdf289eedaa1b248bae708b60272/v2/googlecloud-to-clickhouse/src/main/java/com/google/cloud/teleport/v2/clickhouse/templates/BigQueryToClickHouse.java#L58 to "googlecloud-to-clickhouse" (or googlecloud-and-clickhouse if there are reverse templates planned)

BentsiLeviav: Thanks! Changed it to `googlecloud-to-clickhouse`

v2/googlecloud-to-clickhouse/src/test/java/com/google/cloud/teleport/v2/clickhouse/PubSubToClickHouseIT.java
```diff
@@ -0,0 +1,502 @@
...
+  private static final int BAD_MESSAGES_COUNT = 3;
+
+  private PubsubResourceManager pubsubResourceManager;
+  private ClickHouseResourceManager clickHouseResourceManager;
+
+  @Before
+  public void setup() throws IOException {
+    pubsubResourceManager =
+        PubsubResourceManager.builder(testName, PROJECT, credentialsProvider).build();
+    clickHouseResourceManager = ClickHouseResourceManager.builder(testId).build();
```
Abacn: in the past we experienced race condition when multiple test container spins up at the same time: #2634. SInce there are 6 tests here and they may run in parallel under Maven surefire plugins, consider put clickHouseResourceManager into `@BeforeClass` and teardown in `@AfterClass`?

BentsiLeviav: Thanks. moved it per your suggestion

PR #3467, opened 2026-03-10
feat(graph-size-per-shard): Parallelizing Schema Discovery for 5K Table Support

VardhanThigle:
# PR Description: Adding MultiTable Read Transform to optimize Graph Size

## Summary
This PR introduces the **Parallel Schema Discovery**  for `sourcedb-to-spanner` template. This is needed for supporting the scale of 5K Tables w/o pipeline timeout. With this PR on the default launcher VM we are able to discover 5000 tables in 3 mins (w/o parallelization it takes more than 15 mins).

This is (third) child of [pull/3399](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3399). Please refer to the parent PR to get a complete context on the changes involved.

## Key Changes
- **Parallel Schema Discovery:** Implemented parallelized discovery for table schemas and indexes in `SchemaDiscoveryImpl`. By partitioning the table list into batches (default 500) and using a fixed thread pool (4 threads), the pipeline significantly reduces initialization time for large-scale migrations (e.g., 5,000 tables in ~3 mins vs. 15+ mins).
- **Bulk Metadata Discovery (MySQL & PostgreSQL):** Refactored `MysqlDialectAdapter` and `PostgreSQLDialectAdapter` to support bulk discovery using SQL `IN` clauses. This transitions metadata fetching from $O(N_{tables})$ individual queries to $O(N_{tables} / 500)$ bulk queries, minimizing database round-trip overhead.
- **Memory Efficiency:** Integrated `setFetchSize(1000)` into metadata `PreparedStatements` to ensure that the Dataflow Launcher VM can process large result sets without encountering `OutOfMemoryError`.
- **MySQL 5.7 & 8.x Compatibility:** Enhanced `MysqlDialectAdapter` to dynamically detect the `PAD_ATTRIBUTE` column in `INFORMATION_SCHEMA.COLLATIONS`. It provides a safe fallback to `PAD SPACE` for MySQL 5.7, ensuring correct string partitioning across different database versions.
- **Reliable Thread Management:** Implemented robust exception unwrapping in `SchemaDiscoveryImpl` to preserve `SchemaDiscoveryException` types during parallel execution and ensured proper `ExecutorService` shutdown to prevent thread leaks in the launcher environment.
- **Cassandra Impact Analysis:** While `SchemaDiscoveryImpl` (the orchestrator) is updated to support parallel batches, the **Cassandra implementation remains functionally unchanged**. Unlike the JDBC adapters, `CassandraSchemaDiscovery` relies on the Datastax Java Driver's `Metadata` API, which fetches the entire keyspace metadata (including all tables) in a single underlying driver operation. Consequently, while Cassandra now benefits from the parallel orchestration layer, it was already performing a form of bulk discovery and is not affected by the SQL-specific `IN` clause refactoring.

## Todos
These items will be covered in followup activities
1. Monitoring/metrics for per-table ranges  produced and read.
2. Multi Shard support.

## Out of scope
These items are out of scope of the current work:

1. Checkpointed migrations.
2. Fixed graph size for multi level migrations.
3. Any architectural changes.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImpl.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImpl.java
@@ -20,10 +20,19 @@ import com.google.cloud.teleport.v2.source.reader.io.exception.RetriableSchemaDi
 import com.google.cloud.teleport.v2.source.reader.io.exception.SchemaDiscoveryException;
 import com.google.cloud.teleport.v2.source.reader.io.exception.SchemaDiscoveryRetriesExhaustedException;
 import com.google.cloud.teleport.v2.spanner.migrations.schema.SourceColumnType;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import org.apache.beam.sdk.util.BackOff;
 import org.apache.beam.sdk.util.FluentBackoff;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Default Implementation for {@link SchemaDiscovery}. {@link SchemaDiscoveryImpl} wraps an
@@ -34,8 +43,11 @@ import org.apache.beam.sdk.util.FluentBackoff;
  * RetriableSchemaDiscoveryException} for retriable errors.
  */
 public final class SchemaDiscoveryImpl implements SchemaDiscovery {
+  private static final Logger LOG = LoggerFactory.getLogger(SchemaDiscoveryImpl.class);
   private final RetriableSchemaDiscovery retriableSchemaDiscovery;
   private final FluentBackoff fluentBackoff;
+  private static final int BATCH_SIZE = 500;
+  private static final int THREAD_POOL_SIZE = 4;
 
   public SchemaDiscoveryImpl(
       RetriableSchemaDiscovery retriableSchemaDiscovery, FluentBackoff fluentBackoff) {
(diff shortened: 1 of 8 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImpl.java
```diff
@@ -77,16 +95,75 @@ public ImmutableMap<String, ImmutableMap<String, SourceColumnType>> discoverTabl
...
+                executor));
+      }
+
+      CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
+      for (CompletableFuture<ImmutableMap<String, ImmutableMap<String, SourceColumnType>>> future :
+          futures) {
+        result.putAll(future.join());
+      }
+    } catch (Exception e) {
+      convertException(e);
```
VardhanThigle, line 132: The exception and it's handling is tested in the UT. Please refer to the tests for the same.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImpl.java
```diff
@@ -98,11 +175,55 @@ public ImmutableMap<String, ImmutableList<SourceColumnIndexInfo>> discoverTableI
...
+                executor));
+      }
+
+      CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
+      for (CompletableFuture<ImmutableMap<String, ImmutableList<SourceColumnIndexInfo>>> future :
+          futures) {
+        result.putAll(future.join());
+      }
+    } catch (Exception e) {
+      convertException(e);
```
VardhanThigle, line 208: The exception and it's handling is tested in the UT. Please refer to the tests for the same.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SchemaDiscoveryImpl.java
```diff
@@ -77,16 +95,75 @@ public ImmutableMap<String, ImmutableMap<String, SourceColumnType>> discoverTabl
       DataSource dataSource,
       SourceSchemaReference sourceSchemaReference,
       ImmutableList<String> tables)
-      throws SchemaDiscoveryException {
-    return doRetries(
-        () ->
-            retriableSchemaDiscovery.discoverTableSchema(
-                dataSource, sourceSchemaReference, tables));
+      throws SchemaDiscoveryRetriesExhaustedException {
+    if (tables.isEmpty()) {
```
sm745052, line 99: nit: this "parallel db calls" logic looks very similar to what was added in `discoverTableIndexes` function... maybe we should make a function and reuse it ?

VardhanThigle: Agree, I had pondered on this a lot. There's a bit of syntatic difference which reduces the ROI.

For example the return types of index discovery is different, so the future will be different as well.  do-retries is anyways forked out common.

So what we are getting in common is just this part
```
ExecutorService executor = Executors.newFixedThreadPool(THREAD_POOL_SIZE);
   
```
We could probably make the loop common - but it will start handling abstract type for almost any type - there's no common interface between the two.
Same with the completion polling.

Due to the type differences, it made less sense to combine these 2 (just for the sake of combining)

PR #3473, opened 2026-03-11
feat(dv): main template implementation

manitgupta:
### TL;DR 
Added the main template class `GCSSpannerDV` for the GCS-Spanner Data Validation pipeline. This class acts as the entry point, defining the pipeline options and wiring together the reader, matching, and reporting transforms.
### What changed?
- **GCSSpannerDV**: New main class that orchestrates the data validation pipeline.
- Defined the `GCSSpannerDV.Options` interface to capture all necessary pipeline parameters, including project/Spanner specifics, instance/database IDs, GCS input paths, schema/table/column overrides, and BigQuery dataset configurations for reporting.
- Wired the end-to-end pipeline in the `run` method:
  1. **Schema Retrieval**: Reads Spanner Information Schema via `SpannerInformationSchemaProcessorTransform` as a side input.
  2. **Schema Mapping**: Initializes `SchemaMapperProviderFn` with session files and overrides.
  3. **Data Ingestion**: Reads and parses both source records from GCS (`SourceReaderTransform`) and Spanner records (`SpannerReaderTransform`).
  4. **Data Matching**: Pits the two datasets against each other through the `MatchRecordsTransform` to determine equivalence.
  5. **Validation Reporting**: Aggregates and reports the match/mismatch results using `ReportResultsTransform` to BigQuery.

### DIFF
--- a/v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
+++ b/v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
@@ -47,7 +47,7 @@ import org.apache.beam.sdk.values.PCollectionView;
 import org.joda.time.Instant;
 
 @Template(
-    name = "GCS_Spanner_DV",
+    name = "GCS_Spanner_Data_Validator",
     category = TemplateCategory.BATCH,
     displayName = "GCS Spanner Data Validation",
     description =
@@ -61,7 +61,7 @@ import org.joda.time.Instant;
     requirements = {
       "The GCS directory for AVRO files must exist before pipeline execution.",
       "The Spanner tables must exist before pipeline execution.",
-      "The Spanner tables must have a compatible schema."
+      "The Spanner tables must have a compatible schema (either directly or schema mapping)."
     })
 public class GCSSpannerDV {
 


v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
```diff
@@ -0,0 +1,286 @@
...
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.PCollectionView;
+import org.joda.time.Instant;
+
+@Template(
+    name = "GCS_Spanner_DV",
```
bharadwaj-aditya: DV is not a well understood acronym. Might make sense to just call it dataValidation or just GCS Spanner validation

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
```diff
@@ -0,0 +1,286 @@
...
+    optionsClass = GCSSpannerDV.Options.class,
+    flexContainerName = "gcs-spanner-dv",
+    documentation =
+        "https://cloud.google.com/dataflow/docs/guides/templates/provided/gcs-spanner-dv",
+    contactInformation = "https://cloud.google.com/support",
+    preview = true,
+    requirements = {
+      "The GCS directory for AVRO files must exist before pipeline execution.",
+      "The Spanner tables must exist before pipeline execution.",
+      "The Spanner tables must have a compatible schema."
```
bharadwaj-aditya: double checking on this bit. This will support using schema mapping right ? Which means the schema need not be exactly the same. The description here should reflect that in some form

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
```diff
@@ -0,0 +1,286 @@
...
+    void setColumnOverrides(String value);
+
+    @TemplateParameter.Text(
+        order = 11,
+        optional = false,
+        regexes = {"^[^ ;]*$"},
+        description = "BigQuery dataset for reporting",
+        helpText = "The BigQuery dataset ID where the validation results will be stored.",
+        example = "validation_report_dataset")
+    String getBigQueryDataset();
```
bharadwaj-aditya, line 200: Is it possible to generate this dataset without taking an input ? If yes, this should be marked as optional

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
manitgupta: Yes it does. Okay will update the text here.

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
manitgupta: The dataset is mandatory - BQIO only supports creating the tables in it in runtime but the dataset needs to be exist beforehand.

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
manitgupta: Okay will call it `GCS_Spanner_Data_Validator`

PR #3479, opened 2026-03-11
Prototype BigQuery CDC -> Anomaly detection python flex template.

claudevdm:


claudevdm:
@shunping can you please take a look at the anomaly detection parts?

tvalentyn:
@claudevdm would it make sense to break up this contribution into several commits that can be reviewed one-by-one?

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1053,37 +1053,53 @@ private void stageFlexPythonTemplate(
     String dockerfilePath = dockerfileContainer + "/Dockerfile";
     File dockerfile = new File(dockerfilePath);
     if (!dockerfile.exists()) {
-      List<String> filesToCopy = List.of(definition.getTemplateAnnotation().filesToCopy());
+      List<String> allFilesToCopy = List.of(definition.getTemplateAnnotation().filesToCopy());
```
Abacn, line 1056: In terms of idea of splitting the PR, changes to existing file/infra can go in first while reviewing new templates is still in progress.

I see some fixes to stageFlexPythonTemplate (e.g. honor directoriesToCopy). As you may have noticed we currently do not build or release any Python templates. Existing ones have been commented out

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/395237d33c35a13189dc73d82bb258e4246cce58/python/src/main/java/com/google/cloud/teleport/templates/python/StreamingLLM.java#L26

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/395237d33c35a13189dc73d82bb258e4246cce58/python/src/main/java/com/google/cloud/teleport/templates/python/WordCountPython.java#L26

as suggested in comment, they might never worked before. Wondering we can just get rid of these dead code (or revive them after staging being fixed, not in scope of this PR though)

claudevdm:
I will think about how to split. The CDC/IO source is already reviewed in https://github.com/apache/beam/pull/37724 so reviewers can ignore that part.

I forked it here until it rolls out in a beam release

PR #3480, opened 2026-03-11
fix pipeline-yaml flex container name

derrickaw:
1. Fix pipeline-yaml flex container name - issue is that the yaml blueprint is the source of truth, but the container name was only changed previously in the generated template files.
2. Regenerate template files based on some changes


tarun-google (approved):
LGTM

PR #3481, opened 2026-03-11
Optimize Kafka sink producer configs for improved batching efficiency

kir-titievsky-google:
Sets linger.ms=100 and batch.size=64000 to improve batching efficiency for high throughput Kafka streaming. Fixes https://buganizer.corp.google.com/issues/491137242


v2/streaming-data-generator/pom.xml
```diff
@@ -99,6 +99,13 @@
       <version>${snakeyaml.version}</version>
     </dependency>
 
+    <!-- Conscrypt Provider for Java 17 Dataflow Containers -->
+    <dependency>
+      <groupId>org.conscrypt</groupId>
```
an2x: Why is the conscrypt dependency being added?
1) Seems to be unrelated to the main change.
2) We had some problems with conscrypt library in the past and it required some special treatment (e.g. it can't be shaded into a single jar when a template is built and has to be separate).

v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/transforms/StreamingDataGeneratorWriteToKafka.java
```diff
@@ -187,7 +187,11 @@ public PDone expand(PCollection<byte[]> generatedMessages) {
           KafkaIO.<Void, String>write()
               .withBootstrapServers(getPipelineOptions().getBootstrapServer())
               .withTopic(getPipelineOptions().getKafkaTopic())
-              .withValueSerializer(StringSerializer.class);
+              .withValueSerializer(StringSerializer.class)
+              .withProducerConfigUpdates(
+                  java.util.Map.of(
```
an2x: nit (code style): import the `java.util.Map` class and use just `Map.of()` here.

v2/streaming-data-generator/pom.xml
kir-titievsky-google: Following directions fully helped me confirm this is unnecessary :)

v2/streaming-data-generator/src/main/java/com/google/cloud/teleport/v2/transforms/StreamingDataGeneratorWriteToKafka.java
kir-titievsky-google: fixed.

kir-titievsky-google:
The latest commit addresses you feedback. Thank you!

PR #3490, opened 2026-03-13
[SourceDbToSpanner] Handle NaN and Infinity values gracefully in DLQ in Bulk migration template

darshan-sj:
This PR fixes NaN and Inifinity handling issue in the Dead Letter Queue (DLQ) for Bulk migration template.

#### Problem
PostgreSQL and Cloud Spanner support NaN (Not a Number) and Infinity values for floating-point data types. However, the org.json.JSONObject implementation used in the DLQ writer strictly adheres to the JSON standard, which does not support these values. Attempting to put a NaN or Infinity value into a JSONObject throws a JSONException, causing the DLQ serialization to fail and lead to data loss.

#### Solution
Implemented a try-catch approach in DeadLetterQueue.putValueToJson. If json.put fails (e.g., due to NaN/Infinity) or other serialization issues), the code now catches the exception and falls back to encoding the value as a string. This ensures that the DLQ event is successfully written and the pipeline continues to run, preserving the error event in DLQ.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
@@ -371,6 +371,9 @@ public class DeadLetterQueue implements Serializable {
         // Postgres DB and Spanner contain NaN and Infinity values for Numbers.
         // When we get here, we don't have any other option than to encode the value as string and
         // write to json in DLQ file.
+        LOG.warn(
+            "Failure while creating DLQ event JSON. Failed to put a column value in original type, falling back to string",
+            e);
         json.put(key, value.toString());
       }
     } else {


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
```diff
@@ -363,7 +363,16 @@ private void putValueToJson(JSONObject json, String key, Object value) {
...
     } else if (value instanceof Number || value instanceof java.util.Collection) {
-      json.put(key, value);
+      try {
+        json.put(key, value);
+      } catch (Exception e) {
+        // Fallback to string if putting as-is fails.
+        // json.put fails if NaN and Infinity are passed.
+        // Postgres DB and Spanner contain NaN and Infinity values for Numbers.
+        // When we get here, we don't have any other option than to encode the value as string and
+        // write to json in DLQ file.
```
VardhanThigle, line 373: would it make sense to add a more targetted fix for NAN and the like?

Either we can catch the exact exception - or check if the input falls into one of those floating point values.

darshan-sj: Added a log as discussed.

VardhanThigle (approved):

darshan-sj:
This change is merged as part of https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3491

PR #3144, opened 2025-12-22
Add integration tests to bulk spanner migration for checking PostgreSQL data type support

nmemond:
This adds some missing data type mappings to the existing data type integration test, and also adds a test for a migration to a postgres dialect spanner instance.

Note that some of the type mappings fail to migrate as expected. The checks for those are still included for completeness, but they're commented out to avoid failing the tests.

PR #3491, opened 2026-03-13
[SourceDbToSpanner] Bulk template DLQ NaN fix + PG all datatypes test

darshan-sj:
This is combination of 2 Changes 

1. https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3490 
2. https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3144

### 1. This PR fixes NaN and Inifinity handling issue in the Dead Letter Queue (DLQ) for Bulk migration template.

Problem
PostgreSQL and Cloud Spanner support NaN (Not a Number) and Infinity values for floating-point data types. However, the org.json.JSONObject implementation used in the DLQ writer strictly adheres to the JSON standard, which does not support these values. Attempting to put a NaN or Infinity value into a JSONObject throws a JSONException, causing the DLQ serialization to fail and lead to data loss.

Solution
Implemented a try-catch approach in DeadLetterQueue.putValueToJson. If json.put fails (e.g., due to NaN/Infinity) or other serialization issues), the code now catches the exception and falls back to encoding the value as a string. This ensures that the DLQ event is successfully written and the pipeline continues to run, preserving the error event in DLQ.

### 2. This adds some missing data type mappings to the existing data type integration test, and also adds a test for a migration to a postgres dialect spanner instance.

Note that some of the type mappings fail to migrate as expected. The checks for those are still included for completeness, but they're commented out to avoid failing the tests.
https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3144

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/writer/DeadLetterQueue.java
@@ -371,6 +371,9 @@ public class DeadLetterQueue implements Serializable {
         // Postgres DB and Spanner contain NaN and Infinity values for Numbers.
         // When we get here, we don't have any other option than to encode the value as string and
         // write to json in DLQ file.
+        LOG.warn(
+            "Failure while creating DLQ event JSON. Failed to put a column value in original type, falling back to string",
+            e);
         json.put(key, value.toString());
       }
     } else {


v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDataTypesIT.java
```diff
@@ -152,60 +174,168 @@ public void allTypesTest() throws Exception {
...
+        "bigserial_to_string", createRows("-9223372036854775808", "9223372036854775807", "42"));
+    // bit is commented out to avoid failing the test case; returned data is bits "00110000" and
+    // "00110001" (base64 strings "MA==" and "MQ==")
+    // result.put("bit", createRows("AA==", "gA==", "NULL"));
+    // bit_to_string is commented out to avoid failing the test case; returned data is the literal
+    // string "java.nio.HeapByteBuffer[pos=0 lim=32 cap=32]"
+    // result.put("bit_to_string", createRows("0", "1", "NULL"));
+    // bit_varying is commented out to avoid failing the test case; returned data is bits "00110000
+    // 00110001 00110000 00110001" (base64 string "MDEwMQ==")
+    // result.put("bit_varying", createRows("UA==", "NULL"));
```
rohitwali, line 191: Are these data types expected to fail?

VardhanThigle: We will need to fix these types mappings in the PG part.
Are they being tracked in the report?

rohitwali: if they are supported for migration to Spanner, they should be part of the report. question I would be asking is, if they were never supported or is this a regression.

darshan-sj: Synced offline on this. Resolving this conversation.

PR #3509, opened 2026-03-17
Add CRC32C checksum based validation in GCS Avro to Cloud Spanner pipeline

adityatulasi-google:


v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
```diff
@@ -923,32 +923,72 @@ private void validateGcsFiles(ProcessContext c, String table, TableManifest mani
...
+        if (!Strings.isNullOrEmpty(expectedHash)) {
+            String actualHash = checksums.get(i);
+            Verify.verify(
+                expectedHash.equals(actualHash),
+                "Inconsistent file: %s expected hash %s actual hash %s",
+                fileName,
+                expectedHash,
+                actualHash);
+        } else {
+            String expectedCrc32c = file.getCrc32C();
```
bharadwaj-aditya: this check is very similar to the md5 check, moving the checks to the same level would make more sense.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
```diff
@@ -923,32 +923,72 @@ private void validateGcsFiles(ProcessContext c, String table, TableManifest mani
...
-        String actualHash = FileChecksum.getLocalFileChecksum(filePath);
         String expectedHash = file.getMd5();
-        Verify.verify(
-            expectedHash.equals(actualHash),
-            "Inconsistent file: %s expected hash %s actual hash %s",
-            filePath,
-            expectedHash,
-            actualHash);
+        
+        if (!Strings.isNullOrEmpty(expectedHash)) {
```
bharadwaj-aditya, line 974: seems like the same logic is repeated twice. Can we move to a method ?

bharadwaj-aditya (changes requested):
There should be changes to the export pipeline to provide this as well.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
adityatulasi-google: I'm guessing that you're suggesting the following change:
```
String expectedHash = file.getMd5();
String expectedCrc32c = file.getCrc32C();
if (!StringsisNullOrEmpty(expectedHash)) {
...
} else if (!Strings.isNullOrEmpty(expectedCrc32c)) {
...
} else {
...
}
```

Is that right? If yes, then I did not want to retrieve CRC32C checksum unless it was required because it will be an additional call to GCS and can cost the customer some money. Given that this will be required in a very specific case, there's no need for making this call in every run. Does that justify this structure?

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
adityatulasi-google: Well, the answer is similar to the above. I don't want to retrieve CRC32C for GCS object unless I know I will need it. Without that, making a helper function isn't exactly simple/clean. I can probably pass a lambda which retrieves CRC32C but that's a little clumsy. I'd rather duplicate these checks. WDYT?

adityatulasi-google:
> There should be changes to the export pipeline to provide this as well.

> There should be changes to the export pipeline to provide this as well.

It would not make sense to export CRC32C checksum for all exports. Should I introduce a parameter like below and use it?

```
    @TemplateParameter.Boolean(
        order = 14,
        groupName = "Source",
        optional = true,
        description = "Use CRC32C checksum in manifests",
        helpText = "Use CRC32C checksum instead of MD5 checksum for Avro files in the created manifests.")
    @Default.Boolean(false)
    ValueProvider<Boolean> useCrc32cChecksum();
```

anowardear062-svg:
Support and proses earnings

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
bharadwaj-aditya: makes sense. Please add a comment here in the code so that someone doesn't come and refactor it without that context.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
bharadwaj-aditya: i would guess taking a TableManigest.file object and performing this check would work without a lambda. Something i'm missing here ?

adityatulasi-google: Taking file as input is fine. For expected checksums, I'm not initializing the list of CRC32C checksums in GCS case unless I know I will need it. The validation function will need both MD5 and CRC32C for expected checksums too and I cannot pass both of them without knowing which one I'll need. Maybe I can check if the list if empty and send NULL but then the helper function will have to handle that case too which is unnecessary for Local file case. Overall, I still feel like duplicating for readability is better here. 

PLMK if I'm missing some other way of cleanly doing this.

bharadwaj-aditya: ok. Fine to keep as is.

bharadwaj-aditya:
> > There should be changes to the export pipeline to provide this as well.
> 
> > There should be changes to the export pipeline to provide this as well.
> 
> It would not make sense to export CRC32C checksum for all exports. Should I introduce a parameter like below and use it?
> 
> ```
>     @TemplateParameter.Boolean(
>         order = 14,
>         groupName = "Source",
>         optional = true,
>         description = "Use CRC32C checksum in manifests",
>         helpText = "Use CRC32C checksum instead of MD5 checksum for Avro files in the created manifests.")
>     @Default.Boolean(false)
>     ValueProvider<Boolean> useCrc32cChecksum();
> ```

Please make this an ENUM input with default as MDF

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
adityatulasi-google: Done.

v1/src/main/java/com/google/cloud/teleport/spanner/ImportTransform.java
adityatulasi-google: Ack. Resolving.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportPipeline.java
```diff
@@ -231,6 +231,27 @@ public interface ExportPipelineOptions extends PipelineOptions {
...
+        groupName = "Source",
+        enumOptions = {
+          @TemplateEnumOption("MD5"),
+          @TemplateEnumOption("CRC32C")
+        },
+        optional = true,
+        description = "Checksum algorithm",
+        helpText =
+            "The checksum algorithm to use for the exported files. Possible values are MD5 and CRC32C. The default value is MD5.")
+    ValueProvider<ChecksumAlgorithm> getChecksumAlgorithm();
```
bharadwaj-aditya, line 249: can you mark the default in the template enum directly ?

adityatulasi-google: It's not supported in v1.

adityatulasi-google: Discussed offline. Does not work for v1 templates.

v1/src/main/java/com/google/cloud/teleport/spanner/ExportTransform.java
```diff
@@ -1110,14 +1153,30 @@ private TableManifest buildGcsManifest(ProcessContext c, Iterable<GcsPath> files
...
+        checksums = FileChecksum.getGcsFileChecksums(gcsUtil, gcsPaths);
+      } else {
+        checksums = FileChecksum.getGcsFileChecksumsCrc32c(gcsUtil, gcsPaths);
+      }
+
       for (int i = 0; i < gcsPaths.size(); i++) {
         String fileName = gcsPaths.get(i).getFileName().getObject();
         String hash = checksums.get(i);
-        result.addFilesBuilder().setName(fileName).setMd5(hash);
+        if (algorithm == ExportPipeline.ExportPipelineOptions.ChecksumAlgorithm.MD5) {
```
bharadwaj-aditya, line 1167: if nothing is set in the algorithm, it should default to MD5, not crc

adityatulasi-google: The helper method make sure default value is MD5.

bharadwaj-aditya (approved):
it seems like the job does not work with incorrect parameters - so marking LGTM

PR #3533, opened 2026-03-19
Fix Python Flex Template staging

claudevdm:
support setup.py, directories, and split requirements
Requirements files

Python templates now use two requirements files:

- requirements_all.txt: full pip-compile lockfile. Used at Docker build time to install all deps into the template launcher image. This is the default and is required.
- requirements.txt (optional): extra non-Beam deps to stage to Dataflow workers via FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE. Only needed when the template has deps not already on the Beam SDK worker image.

FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE is no longer set unconditionally. The old behavior caused the Beam stager to run pip download --platform manylinux2014_x86_64 for every package, which hangs on packages like numpy 2.4+ that no longer publish manylinux2014 wheels.

setup.py and directory support:
Templates can now use setup.py with a package directory (e.g. src/). Both are auto-detected during staging:

- setup.py sets FLEX_TEMPLATE_PYTHON_SETUP_FILE for worker code staging, runs pip install . at build time
- Directories in @Template(filesToCopy = {...}) → COPY'd into the image

Other fixes
- Fixed pip install -U -r --require-hashes argument order (-r must precede the filename)
- Fixed empty entryPoint annotation default ({""}) not being detected
- Fixed wrong base directory passed to DockerfileGenerator

### DIFF
--- a/plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
+++ b/plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
@@ -214,6 +214,7 @@ public class DockerfileGenerator {
       this.parameters.put("setupFileEnv", "");
       this.parameters.put("setupInstall", "");
       this.parameters.put("workerRequirementsEnv", "");
+      this.parameters.put("workerRequirementsCache", "");
       this.parameters.put("commandSpec", "");
     }
 
@@ -342,8 +343,8 @@ public class DockerfileGenerator {
     }
 
     /**
-     * Sets the requirements file used for {@code pip install} and {@code pip download} at Docker
-     * build time. Defaults to {@code requirements_all.txt}.
+     * Sets the requirements file used for {@code pip install} at Docker build time. Defaults to
+     * {@code requirements_all.txt}.
      *
      * @param requirementsFile the requirements filename (e.g. "requirements_all.txt").
      * @return this {@link Builder}.
@@ -385,6 +386,10 @@ public class DockerfileGenerator {
       addParameter(
           "workerRequirementsEnv",
           "ENV FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE=\"" + requirementsFile + "\"");
+      addParameter(
+          "workerRequirementsCache",
+          "RUN pip download --no-cache-dir --dest /tmp/dataflow-requirements-cache -r "
+              + requirementsFile);
       return this;
     }
 
--- a/plugins/core-plugin/src/main/resources/Dockerfile-template-python
+++ b/plugins/core-plugin/src/main/resources/Dockerfile-template-python
@@ -17,6 +17,6 @@ ${airlockConfig}
 
 RUN pip install -U --require-hashes -r ${requirementsFile}
 ${setupInstall}
-RUN pip download --require-hashes --no-cache-dir --dest /tmp/dataflow-requirements-cache -r ${requirementsFile}
+${workerRequirementsCache}
 
 ENTRYPOINT ${entryPoint}


plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
```diff
@@ -210,6 +210,8 @@ public Builder(
 
       this.parameters.put("filesToCopy", "");
       this.parameters.put("directoriesToCopy", "");
+      this.parameters.put("setupFileEnv", "");
+      this.parameters.put("setupInstall", "");
```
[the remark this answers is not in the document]
claudevdm: Not how stager works.

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
```diff
@@ -337,6 +339,25 @@ public Builder setDirectoriesToCopy(Set<String> directoriesToCopy) {
...
+     */
+    public Builder setSetupFile(String setupFile) {
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(setupFile));
+      String workDir =
+          (String) this.parameters.getOrDefault("workingDirectory", DEFAULT_WORKING_DIRECTORY);
+      addParameter(
+          "setupFileEnv",
+          "ENV FLEX_TEMPLATE_PYTHON_SETUP_FILE=\"" + workDir + "/" + setupFile + "\"");
+      addParameter("setupInstall", "RUN pip install --no-cache-dir .");
+      return this;
```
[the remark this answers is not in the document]
claudevdm: Not how stager works.

plugins/core-plugin/src/main/resources/Dockerfile-template-python
```diff
@@ -3,17 +3,27 @@ FROM ${basePythonContainerImage}
...
 RUN if ! [ -f requirements.txt ] ; then >&2 echo "error: no requirements.txt file found" && exit 1 ; fi
 
 # Set up custom PyPi repository, if applicable
 ${airlockConfig}
 
-RUN pip install -U -r --require-hashes $FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE
-RUN pip download --require-hashes --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE
+RUN pip install -U --require-hashes -r $REQUIREMENTS_FILE
+${setupInstall}
+RUN pip download --require-hashes --no-cache-dir --dest /tmp/dataflow-requirements-cache -r $REQUIREMENTS_FILE
```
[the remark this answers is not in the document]
claudevdm: Not how stager works.

plugins/core-plugin/src/main/resources/Dockerfile-template-python
```diff
@@ -3,17 +3,27 @@ FROM ${basePythonContainerImage}
 ARG WORKDIR=${workingDirectory}
 RUN mkdir -p $WORKDIR
 ${filesToCopy}
+${directoriesToCopy}
 WORKDIR $WORKDIR
 
-ENV FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE=requirements.txt
+# Do NOT use ENV FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE here. All deps are preinstalled
```
tvalentyn: how will template launcher know to supply `--requirements_file` when we launch the pipeline ?

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
```diff
@@ -1053,37 +1053,53 @@ private void stageFlexPythonTemplate(
...
+      Set<String> directoriesToCopy = new HashSet<>();
+      for (String f : allFilesToCopy) {
+        File source = new File(dockerfileContainer + "/" + f);
+        if (source.isDirectory()) {
+          directoriesToCopy.add(f);
+        } else {
+          filesToCopy.add(f);
+        }
+      }
       if (filesToCopy.isEmpty()) {
```
tvalentyn: do we still need this given line 1057?

plugins/core-plugin/src/main/resources/Dockerfile-template-python
claudevdm: For templates with setup.py, FLEX_TEMPLATE_PYTHON_SETUP_FILE handles worker code distribution. For templates without setup.py that need non-Beam deps on workers, a future change would use an extra_requirements.txt containing only the additional packages.

Setting FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE causes the Beam stager to run pip download --platform manylinux2014_x86_64 for every package in the requirements file ([stager.py L767-798](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/portability/stager.py#L767-L798)). This codepath is broken for packages like numpy 2.4.x, which no longer publish manylinux2014-tagged wheels on PyPI (numpy now publishes manylinux_2_17 / manylinux_2_27+ wheels instead). When pip download can't find a wheel matching manylinux2014, pip downloads the source tarball and attempts to compile from source on the launcher VM which hangs indefinitely.

Reproduced inside the built container image:

```
$ pip download --dest /tmp/cache --no-deps \
    --implementation cp --abi cp311 --platform manylinux2014_x86_64 \
    "numpy==2.4.2"

Downloading numpy-2.4.2.tar.gz (20.7 MB)
Installing build dependencies: started
Installing build dependencies: finished with status 'done'
Preparing metadata (pyproject.toml): started
<hangs indefinitely>
The pre-populated cache at /tmp/dataflow-requirements-cache doesn't help either the cache contains numpy-2.4.2-cp311-cp311-manylinux_2_27_x86_64.whl, but pip's --platform manylinux2014_x86_64 filter doesn't match the manylinux_2_27 tag.
```

The Beam SDK worker image comes with apache-beam[gcp] and all transitive deps (numpy, grpcio, google-cloud-bigquery-storage, etc.) pre-installed. These are the same packages that make up the bulk of requirements.txt.

FLEX_TEMPLATE_PYTHON_SETUP_FILE triggers the Beam stager to build an sdist tarball of the custom package source code and stage the tarball to workers. The setup_file codepath does no dependency resolution  the stager just tars up the source. Workers pip install the tarball, and install_requires is satisfied from packages already on the Beam SDK worker image.

plugins/templates-maven-plugin/src/main/java/com/google/cloud/teleport/plugin/maven/TemplatesStageMojo.java
claudevdm: Removed

plugins/core-plugin/src/main/resources/Dockerfile-template-python
tvalentyn: > causes the Beam stager to run pip download --platform manylinux2014_x86_64 for every package in the requirements file ([stager.py L767-798](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/portability/stager.py#L767-L798)).

We should probably patch beam stager code then. The intent is: download those wheels that will be used in Beam's containers.

we don't need to list every package, including beam or beam deps in requirements.txt when passing --requirements_file as part of job submission, we can treat this as extra_requirements.txt that you are suggesting.

> a future change would use an extra_requirements.txt containing only the additional packages.

I think this should be a part pipeline package requirements (e.g. install_requires in setup.py) , or it should be  in requirements.txt itself.

tvalentyn: if you use requirements.txt as a snapshot for environment you want to have when building the template launcher image but not the worker container image, then maybe you can call it requirements_all.txt or requirements_launcher.txt and have requirements.txt for dataflow workers (minimum necessary deps) to stage and resolve at runtime.

tvalentyn: but to have a requirements.txt as part of the template and not use FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE sounds confusing to since it becomes unclear what is the purpose of requirements.txt.

We can think from this standpoint:

if i am authoring a new Python template that needs extra dependencies, what is the guidance of where I must declare these deps? Only top-level or also transitive? How do we ensure that required packages will be staged to workers to ensure template works in private IP scenario.

IMO, in current setup, we should only declare only additionally-needed dependencies, so extra dependencies and their transitives, and have that in requirements.txt, and we pass that in --requirements_file to DF Workers (via setting FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE variable)

In ideal setup, I'd use the same docker container for template submission and runtime, install everything at template build time, and don't set any variables like FLEX_TEMPLATE_PYTHON_REQUIREMENTS_FILE

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
```diff
@@ -337,6 +341,53 @@ public Builder setDirectoriesToCopy(Set<String> directoriesToCopy) {
       return addStringParameter("directoriesToCopy", directories.toString());
     }
 
+    /**
+     * Sets the requirements file used for {@code pip install} and {@code pip download} at Docker
```
tvalentyn: I think pip download is necessary only for the SDK container deps?

tvalentyn: in other words, for the requirements.txt but not for the requirements_all.txt

plugins/core-plugin/src/main/java/com/google/cloud/teleport/plugin/DockerfileGenerator.java
```diff
@@ -337,6 +341,53 @@ public Builder setDirectoriesToCopy(Set<String> directoriesToCopy) {
...
+     * build time. Defaults to {@code requirements_all.txt}.
```
tvalentyn: consider adding when to use it. perhaps sth like:

Use to store the snapshot of all dependencies of the flex template launcher image (such as `pip freeze` output) to create reproducible builds.

PR #3545, opened 2026-03-23
Add pipeline option for Spanner emulator support in SpannerChangeStreamsToPubSub

hunsche:
### Description
This pull request adds support for the Spanner emulator in the `SpannerChangeStreamsToPubSub` template by introducing a new pipeline option `useSpannerEmulatorHost`.

Following review feedback, we avoided using environment variables as they are not easily supported in Google-provided templates at runtime. Instead, we added a boolean option `useSpannerEmulatorHost`.

When `useSpannerEmulatorHost` is set to `true`, the pipeline configures `SpannerConfig` with `.withEmulatorHost(options.getSpannerHost())`. This allows users to point to a local Spanner emulator (e.g., `http://localhost:9010`) for debugging and testing purposes without SSL verification issues.

Additionally, added a unit test `testSpannerEmulatorHostOptions` to cover the new code paths and ensure coverage requirements are met.

### Changes
- Added `useSpannerEmulatorHost` to `SpannerChangeStreamsToPubSubOptions`.
- Updated `SpannerChangeStreamsToPubSub.java` to use `.withEmulatorHost()` when the flag is true.
- Added a unit test in `SpannerChangeStreamsToPubSubTest.java` to cover the new flag.

### Verification
- Verified by adding a unit test that covers the new logic.

### DIFF
--- a/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSubTest.java
+++ b/v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSubTest.java
@@ -225,6 +225,25 @@ public final class SpannerChangeStreamsToPubSubTest extends SpannerTestHelper {
     p.run();
   }
 
+  @Test(expected = Exception.class)
+  @Category(IntegrationTest.class)
+  public void testSpannerEmulatorHostOptions() {
+    SpannerChangeStreamsToPubSubOptions options =
+        PipelineOptionsFactory.create().as(SpannerChangeStreamsToPubSubOptions.class);
+    options.setSpannerProjectId(TEST_PROJECT);
+    options.setSpannerInstanceId(TEST_INSTANCE);
+    options.setSpannerDatabase(TEST_TABLE);
+    options.setSpannerMetadataInstanceId(TEST_INSTANCE);
+    options.setSpannerMetadataDatabase(TEST_TABLE);
+    options.setSpannerChangeStreamName(TEST_CHANGE_STREAM);
+    options.setPubsubTopic(outputTopicName);
+    options.setOutputDataFormat("JSON");
+
+    options.setUseSpannerEmulatorHost(true);
+    options.setSpannerHost("http://localhost:12345");
+    run(options);
+  }
+
   @Test
   @Category(IntegrationTest.class)
   // This test can only be run locally with the following command:


v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSub.java
```diff
@@ -158,10 +158,17 @@ public static PipelineResult run(SpannerChangeStreamsToPubSubOptions options) {
     final RpcPriority rpcPriority = options.getRpcPriority();
     SpannerConfig spannerConfig =
         SpannerConfig.create()
-            .withHost(ValueProvider.StaticValueProvider.of(options.getSpannerHost()))
             .withProjectId(spannerProjectId)
             .withInstanceId(instanceId)
             .withDatabaseId(databaseId);
+    if (System.getenv("SPANNER_EMULATOR_HOST") != null) {
+      spannerConfig =
+          spannerConfig.withEmulatorHost(
```
bharadwaj-aditya, line 167: These should typically be passed as pipeline options.

Abacn: Took a look again, this won't be effective for Google-provided Templates. Basically  we do not expose interface to add environment variable for template launcher after Templates built. Users would need to build their custom container, either by modified command spec json or add a "ENV ..." in their Dockerfile.

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
```diff
@@ -239,4 +239,15 @@ public interface SpannerChangeStreamsToPubSubOptions extends DataflowPipelineOpt
   String getOutputMessageMetadata();
 
   void setOutputMessageMetadata(String value);
+
+  @TemplateParameter.Boolean(
```
Abacn, line 243: Currently we can use `--additionalPipelineOption=spannerHost=...` to set emulator Host. I understand this is a debugging option not meant to be submitted to Dataflow? (as local enumator host won't work on a remote machine)

hunsche: You are correct that this is primarily meant for local debugging and integration testing (e.g., executing the template locally via `DirectRunner`), and not for remote Dataflow workers.

Regarding `--additionalPipelineOption=spannerHost=...`, simply overriding the host string is unfortunately not enough. If we pass an emulator host (like `http://localhost:9010`) to the standard `SpannerConfig.withHost(...)` method, the underlying Spanner client still attempts to negotiate SSL and authenticate using Google credentials, which causes the connection to the emulator to immediately fail. 
 
To successfully connect to the emulator, the Beam SpannerIO connector requires us to explicitly call `.withEmulatorHost(...)` instead, which tells the client to disable SSL and credentials. The boolean option `useSpannerEmulatorHost` was introduced precisely to trigger this specific builder method during initialization!

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSub.java
hunsche: Thank you both for the review and the clarification regarding how environment variables behave inside Google-provided templates! That makes perfect sense.
 
I have just pushed an update that completely removes the `System.getenv("SPANNER_EMULATOR_HOST")` check. Instead, following your feedback, I introduced a new boolean pipeline option called `useSpannerEmulatorHost`. 
 
When this option is enabled, the pipeline explicitly initializes `SpannerConfig` using `.withEmulatorHost(options.getSpannerHost())`, disabling SSL/auth so it can natively connect to the local emulator for testing. Otherwise, it gracefully falls back to the default `.withHost()` behavior.
 
PTAL when you have a chance!

v2/googlecloud-to-googlecloud/src/main/java/com/google/cloud/teleport/v2/options/SpannerChangeStreamsToPubSubOptions.java
Abacn: Thanks for the context. This LGTM

v2/googlecloud-to-googlecloud/src/test/java/com/google/cloud/teleport/v2/templates/SpannerChangeStreamsToPubSubTest.java
```diff
@@ -225,6 +225,25 @@ public void testFileFormatFactoryInvalid() {
     p.run();
   }
 
+  @Test(expected = Exception.class)
+  @Category(IntegrationTest.class)
```
Abacn, line 229: ~The unit tests in this test class somehow got annotated with `@Category(IntegrationTest.class)` but it's no-op. Consider clean up.~ I see, it's under spanner namespace and used by Spanner tests.

Abacn:
Going to merge (has to click bypass Waiting on code owner review) by tomorrow if no other review requests cc: @bharadwaj-aditya

hunsche:
Hi @Abacn, @bharadwaj-aditya! 👋

Just a quick follow-up on this PR. 

I have updated the PR description and title to accurately reflect the final implementation we agreed on (using the `useSpannerEmulatorHost` pipeline option instead of the environment variable, and adding the unit tests to fix the code coverage).

Since the PR was already approved by @Abacn a few days ago, could you please take a final look or let me know if there is anything else needed from my side to get this merged?

Thanks again for all the guidance during the reviews!

PR #3557, opened 2026-03-25
Upgrade to Beam 2.72.0

Abacn:


Abacn:
Now the error becomes

```
Caused by: com.clickhouse.client.api.ServerException: Code: 516. DB::Exception: default: Authentication failed:
 password is incorrect, or there is no user with such name. 
If you have installed ClickHouse and forgot password you can reset it in the configuration file. 
The password for default user is typically located at /etc/clickhouse-server/users.d/default-password.xml 
and deleting this file will reset the password. See also /etc/clickhouse-server/users.xml on the server 
where ClickHouse is installed. . (AUTHENTICATION_FAILED) (version 23.8.16.16 (official build)) 
...
Caused by: java.lang.RuntimeException: Failed to get table schema for table: test_big_query_to_click_house_20260326030950490
 at org.apache.beam.sdk.io.clickhouse.ClickHouseIO.getTableSchema(ClickHouseIO.java:765)
 at org.apache.beam.sdk.io.clickhouse.ClickHouseIO.getTableSchema(ClickHouseIO.java:699)
 at com.google.cloud.teleport.v2.clickhouse.templates.BigQueryToClickHouse.run(BigQueryToClickHouse.java:109)
 ... 1 more
```

We may need to do similar changes like here:

https://github.com/apache/beam/pull/37611/changes#diff-2801abf26a3a9e2cd1f4278738986ea658907f7341cd0db9faa0e265eeb01eecR249

Abacn:
clickhouse tests passed in the latest run

Abacn:
Build failure due to cannot find aiohttp 3.13.5 in internal pypi mirror

Abacn:
R: @apanich @tvalentyn if anyone still around

tvalentyn:
LGTM

tvalentyn (approved):

python/src/main/python/streaming-llm/base_requirements.txt
```diff
@@ -1,4 +1,4 @@
-apache-beam[gcp]==2.71.0
+apache-beam[gcp]
```
damccorm, line 1: I think that this doesn't work well. When there are several requirements, pip doesn't always resolve to the latest beam. For example, https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/e31b27975b7f197f5dc3cfcd15bed3f20a322c45/python/src/main/python/streaming-llm/requirements.txt#L176 is on an old beam. This doesn't a matter much for this example template, but would be easy to miss in other instances.

I think we should add back version numbers here, proposing this in https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4203

Abacn: This happens when your `__build__` dir isn't fresh. Adding version is indeed safer, though need to change it every time on a version bump or rc validation

damccorm: I don't think it is just that. We're getting this even when getting PRs generated by GitHub actions and I reproed on a fresh GitHub codespace - neither of these have preexisting `__build__` directories

damccorm: Also note that it only failed to update streaming-llm in those passes, not the other generated requirements. This suggests that this can be caused by including certain dependencies (for now these are ml deps, but I don't think there's any guarantee that the same thing couldn't happen for other dependencies)

damccorm: My guess is that pip resolves some of the other dependencies first, then picks a beam version which satisfies that; if the dependency (or one of its dependencies) is outside Beam's constraints, then we'll see this issue

Abacn: Yeah it's possible. It's noticed some Beam's optional dependency started to depend on Beam: https://github.com/apache/beam/pull/39884

PR #3646, opened 2026-04-08
add defaults for retries - elasticsearch

derrickaw:
1. Current templates like BQ to ES will exit execution after one simple failure trying to write to ES.
2. Customers have ran into transient errors where the pipeline fails but on subsequent retry it works fine.
3. Adding retry defaults to ES writeoptions like PubSubToClickHouseOptions and BigtableChangeStreamsToPubSubOptions does.


derrickaw:
Hi @Abacn, can you review this at your convenience? thanks

Abacn (approved):

PR #3658, opened 2026-04-09
fix(Spanner): Added read-only validation for Mysql shards. This does not validate the user authority.

pratickchokhani:


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/JdbcConnectionHelper.java
```diff
@@ -84,10 +80,24 @@ public synchronized void init(ConnectionHelperRequest connectionHelperRequest) {
         config.addDataSourceProperty(key, value);
       }
       HikariDataSource ds = new HikariDataSource(config);
+      validateMysqlHikariConnection(ds);
       connectionPoolMap.put(sourceConnectionUrl + "/" + shard.getUserName(), ds);
```
manitgupta, line 91: Is this class specific to MySQL? How does this work for PG and other JDBC sources?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/JdbcConnectionHelper.java
```diff
@@ -84,10 +80,24 @@ public synchronized void init(ConnectionHelperRequest connectionHelperRequest) {
...
     }
   }
 
+  private void validateMysqlHikariConnection(HikariDataSource ds) {
+    try (Statement stmt = ds.getConnection().createStatement();
+        ResultSet rs = stmt.executeQuery("SELECT @@global.read_only")) {
+
+      if (rs.next() && rs.getBoolean(1)) {
+        // Node is read-only. Close connection and throw exception.
+        throw new SQLException("Connection rejected: MySQL Shard is in read-only mode.");
```
manitgupta, line 101: Is it possible to add some helpful debugging data about the shard which was in read-only mode? The current error does not tell the user **where** the error is occurring?

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/connection/JdbcConnectionHelperTest.java
```diff
@@ -136,4 +149,87 @@ public void testInitConnectionPool() {
       }
     }
   }
+
+  @Test
+  public void testInit_readOnlyShard_throwsException() throws SQLException {
```
manitgupta, line 155: Nit: Lets use the same naming convention as the existing tests. (camelCase?)

PR #3665, opened 2026-04-10
fix neo4j upgrade issues and removed redundant property

derrickaw:
1. Fixes #3662
2. Removes redundant neo4j property


Abacn (approved):
Thanks!

PR #3667, opened 2026-04-10
Adding IT for PG as a source in reverse replication

shreyakhajanchi:


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
```diff
@@ -230,7 +242,11 @@ public PipelineLauncher.LaunchInfo launchDataflowJob(
             put(
```
sm745052, line 242: can't we use both constants (`MYSQL_SOURCE_TYPE` and `SOURCE_POSTGRESQL`) from either

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/constants/Constants.java

or

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/constants/Constants.java

sm745052 (approved):
LGTM. I am assuming that the IT failure in JAVA PR is unrelated.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
shreyakhajanchi: We can but I didn't want to change code pertaining to MySQL so followed the existing pattern

sm745052: Then we should be importing `POSTGRES_SOURCE_TYPE` from v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/constants/Constants.java

looks kinda odd . Anyways consider this a nit.

PR #3669, opened 2026-04-10
Astra to BQ - exclude vulnerable dependencies

stankiewicz:
datastax astra io had vulnerable logback dependencies.
PR excludes them.


damccorm (approved):
Thanks!

PR #3685, opened 2026-04-14
POJOs for MultiShard TableReader.

VardhanThigle:
#  Foundations for Multi-Shard Routing (POJOs)
This is the first child of #3684 

## Design Decision
This PR introduces the foundational data models and services required to support multi-shard JDBC migrations. 

### Key Components:
- **`JdbcIoWrapperConfigGroup`**: A container that aggregates multiple `JdbcIOWrapperConfig` instances. It ensures that all shards in a group share a uniform `SQLDialect`, simplifying downstream transform logic.
- **`DataSourceProvider` Interface**: Defines the contract for worker-side connection routing.
- **`DataSourceManager` / `DataSourceManagerImpl`**: A centralized service for managing and caching connection sources across multiple physical shards.

### Rationale:
To achieve a constant graph size, we must move connection details from pipeline construction time (PTransform fields) to element processing time (DoFn logic). These POJOs provide the structure for this dynamic routing.

---

## Why it's Safe (Concurrency & Resource Leaks)
- **Thread-Safe Caching**: `DataSourceManagerImpl` implements a thread-safe cache for `DataSource` instances using `ReentrantLock` and a double-checked locking pattern. This ensures that even when multiple Beam bundles execute concurrently on the same worker, each physical shard connection pool is initialized exactly once.
- **Resource Management**: The `DataSourceManager` provides a `closeAll()` method. This is specifically designed to be called during `DoFn.teardown()` or `finishBundle()` to ensure that all opened connection pools are gracefully released, preventing resource leaks on Dataflow workers.
- **Serialization Robustness**: Fields within `DataSourceManagerImpl` are marked `transient` where appropriate, and the class includes null-safety checks to handle lifecycle transitions in the Dataflow environment.

## Testing
The added tests verify:
- Correct aggregation of shard configurations.
- Thread-safe lazy initialization of data sources.
- Proper handling of unknown shard IDs.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/DataSourceManagerImpl.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/DataSourceManagerImpl.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms;
+
+import com.google.auto.value.AutoValue;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.DataSourceProvider;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Concrete implementation of {@link DataSourceManager} using {@link AutoValue}.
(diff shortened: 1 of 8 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/DataSourceManagerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+   * a multi-threaded execution environment within the worker. It uses double-checked locking
+   * pattern with {@link #dataSourceLock}.
+   *
+   * @param dataSourceId reference to the datasource.
+   * @return The initialized {@link DataSource} for the shard.
+   */
+  @Override
+  public DataSource getDatasource(String dataSourceId) {
+
+    DataSource dataSource = this.dataSources.get(dataSourceId);
```
sm745052, line 90: Is this ThreadSafe?

take this example
datasource contains `12 keys / shards(1 to 12)` (Maintaining internal array size = 16)

there are 2 threads:
T1: working on shard 7
T2 working on shard 13

T2 finds null -> acquires lock -> updates the hash map -> hashmap goes through compaction (increases internal array size to 32) ->  T1 tries to get its value. The value might be corrupted


https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html
"""
Note that this implementation is not synchronized. If multiple threads access a hash map concurrently, and at least one of the threads modifies the map structurally, it MUST be synchronized externally. (A structural modification is any operation that adds or deletes one or more mappings
"""

VardhanThigle: We are using double checked locking here. so if you get null, you take a lock and then check again if it's null or not with the lock acquired.

Let's sync offline to discuss more.


Also currently this is used only in context of a given bundle.

VardhanThigle: Moved to ConcurrentHashMaps.

sm745052 (dismissed):
LGTM

sm745052 (approved):
LGTM

PR #3686, opened 2026-04-14
Changes in Reader Interface to support multi-shard reader

VardhanThigle:
# Core Interface Refactor for Multi-Transform Routing.

This is the second child PR of #3684 

## Design Decision
This PR refactors the core `Reader` and `IoWrapper` interfaces to support mapping groups of tables to specific reader transforms. 

### Key Changes:
- **`IoWrapper.getTableReaders()`**: Refactored to return an `ImmutableMap<ImmutableList<SourceTableReference>, PTransform<PBegin, PCollection<SourceRow>>>`. 
- **Mapping Logic**: Instead of a simple list of transforms, the wrapper now explicitly states which tables are handled by which transform. This is essential for the `PipelineController` to correctly iterate and apply constant-size multi-table transforms.

### Rationale:
In a multi-shard world, we need a way for the IO layer (e.g., `JdbcIoWrapper`) to signal to the orchestration layer (e.g., `PipelineController`) that a set of tables from multiple shards can be efficiently processed by a single, consolidated transform. This interface change enables that signaling.

---

## Why it's Safe (Generics & Backward Compatibility)
- **Generics Preservation**: The `SourceRow` generics are strictly enforced in the new map structure. This prevents type erasure issues that can occur in Beam pipelines when processing heterogeneous data streams.
- **Minimal Surface Change**: While the method signature changed, the underlying data models for `SourceTableReference` and `SourceRow` remain backward compatible with existing single-source logic.
- **Cassandra Compatibility**: The `CassandraIoWrapper` implementation was updated to return a single-entry map, maintaining its existing behavior while complying with the new interface.

---

The added tests confirm that the map-based return type is correctly populated and that table-to-transform routing is accurate.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/Reader.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/Reader.java
@@ -17,6 +17,7 @@ package com.google.cloud.teleport.v2.source.reader;
 
 import com.google.cloud.teleport.v2.source.reader.io.schema.SourceSchema;
 import com.google.cloud.teleport.v2.source.reader.io.transform.ReaderTransform;
+import com.google.common.collect.ImmutableList;
 
 /**
  * The Reader interfaces with the source-db allowing the pipeline to read from various data sources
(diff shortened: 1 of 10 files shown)


sm745052 (approved):
LGTM

PR #3697, opened 2026-04-15
Graph size multi shard table identifier

VardhanThigle:
# Propogating DataSourceID as a part of Table Identifier for Multi-Shard Reading
This is the third child of #3684 .

## Design Decision
This PR adds `dataSourceId` to the tableIdentifer.
As a part of Milestone-1, a `Range` now has the `TableIdentifier` which helps move the table level details from `Ptransform` construction to the data layer. As was discussed in the same Milestone, the tableIdentifier was design to be extensible by allowing the DataSourceID to be contained in it. This helps in allowing the reader to correctly select the datasource that the given Range is associated and also eliminates the need to actually embed (and shuffle) the entire dataSource as a part of the range. The DataSourceProvider checked in as a part of #3685 will be provided to the Reader at construction time and that map coupled with the ID provided by the range as a part of this PR will help the Reader connect to the right source.

### Key Changes:
- **`TableIdentifier` Enhancement**: Added a `dataSourceId()` field to correctly distinguish tables with the same name that exist on different physical shards.
- **`JdbcSchemaReference` Integration**: Propagated the `dataSourceId` into the schema reference layer.
- **Unique ID Generation**: Implemented a robust ID generation mechanism in `JdbcIOWrapperConfig` to ensure every shard configuration has a unique identifier throughout the pipeline lifecycle.

### Rationale:
In a multi-shard environment, a table name (e.g., "Users") is no longer unique. We must qualify every table with its source database instance to ensure that Dataflow's grouping and splitting logic routes data to the correct destination.

---

## Why it's Safe (Identity Collision Prevention)
- **Global Uniqueness**: By combining `shardID` and `tableName` into a single identifier, we prevent collisions that would previously have caused Dataflow to aggregate data from different physical sources into a single logical partition.
- **Deterministic Hashing**: Updated `equals()` and `hashCode()` in `TableIdentifier` to include the `dataSourceId`. This is critical for Beam's `GroupByKey` operations, ensuring that ranges from different shards are never incorrectly merged.
- **Nullable Robustness**: Implemented strict null-checks and added `@Nullable` annotations where appropriate to ensure the system handles legacy configurations that may not have an explicit `dataSourceId`.

---

## Tests
The added tests verify that:
- Tables with the same name but different shard IDs are treated as distinct entities.
- Hash collisions are avoided.
- Propagation through the configuration layer is consistent.

--
## A quick note on PR size
This PR is divided into 2 commits - the first one just adds the identifier.
The second one ensures that the builder is correctly initialized in existing UTs - given the spotless requirement, this appears to have gotten into many lines (though in essentiality it's not)

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -311,11 +311,13 @@ public final class JdbcIoWrapper implements IoWrapper {
     return tableConfigsBuilder.build();
   }
 
-  private static TableConfig getTableConfig(
+  @VisibleForTesting
+  protected static TableConfig getTableConfig(
       String tableName,
       JdbcIOWrapperConfig config,
       ImmutableMap<String, ImmutableList<SourceColumnIndexInfo>> indexInfo) {
-    TableConfig.Builder tableConfigBuilder = TableConfig.builder(tableName);
+    TableConfig.Builder tableConfigBuilder =
+        TableConfig.builder(tableName).setDataSourceId(config.id());
     if (config.maxPartitions() != null && config.maxPartitions() != 0) {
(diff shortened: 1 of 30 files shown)


aasthabharill (approved):
LGTM

PR #3706, opened 2026-04-17
Graph size multi shard support -  changes in JdbcIOWrapper

VardhanThigle:
# JDBC Engine Support for Multi-Source Readers
This is the fourth child of #3684 . Here we refactor JdbcIO to allow multi sharded read. This will be followed by changes in Pipeline Controller and  ReadWithUniformPartitions Changes. Please look at #3684  for all the details.

## Design Decision
This PR implements the core multi-shard logic within the `JdbcIoWrapper`. 

### Key Changes:
- **Parallel Schema Discovery**: Refactored the discovery phase to process all shards in a `JdbcIoWrapperConfigGroup` in parallel using `parallelStream()`.
- **Bulk Transform Building**: The wrapper now aggregates all tables across all shards into a consolidated list and instantiates a single `ReadWithUniformPartitions` instance.
- **Fail-Fast Initialization**: Implemented a strict error handling policy where failure to discover a single shard results in immediate job failure.

### Rationale:
To support thousands of tables across hundreds of shards, sequential schema discovery was no longer feasible. By parallelizing this phase and consolidating the resulting transforms, we achieve both rapid job startup and a constant-sized Dataflow graph.

---

## Why it's Safe (Concurrency & Error Isolation)
- **Parallel Discovery Performance**: We used standard Java parallel streams for cross-shard discovery. Since this happens during job submission (on the client/launcher), it significantly reduces job setup time without impacting worker resources.
- **Fail-Fast Integrity**: We chose a fail-fast approach because a multi-shard migration is only as reliable as its weakest shard. Failing early prevents the template from starting a "partial" migration which could lead to difficult-to-detect data gaps in Spanner.
- **Retry Mechanism**: While the discovery is fail-fast, individual shard operations are wrapped in exponential backoff retries (via `FluentBackoff`), ensuring that transient network issues don't trigger unnecessary job failures.

---

## How to Verify
The added tests simulate multi-shard configurations and verify that parallel discovery produces the correct aggregated schema and a consolidated reader transform.


build.log
shreyakhajanchi, line 1: Is this file intended to be checked in?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -64,6 +78,7 @@
  */
 public final class JdbcIoWrapper implements IoWrapper {
   private static final Logger LOG = LoggerFactory.getLogger(JdbcIoWrapper.class);
+  private static final int SOURCE_DISCOVERY_PARALLELISM = 4;
```
shreyakhajanchi, line 83: nit: can you please add a comment here on why we are setting this to 4?

build.log
VardhanThigle: Thanks. I will remove it and add it to repo level git ignore.

It gets created if you use maven cli to generate code coverage.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: Sure. 4 works well with the number of thread we get on default launcher.

build.log
VardhanThigle: Its removed from the recent patch.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: Added the comment.

shreyakhajanchi (approved):

PR #3711, opened 2026-04-20
Graph size multi shard - Refactoring Read With Uniform Partitions

VardhanThigle:
# Read with uniform Partitions Refactor for Multi-Shard Dynamic Routing
This is the sixth child of #3684. Here we refactor Read With Uniform Partitions for supporting multi shard read.

## Design Decision
This PR refactors the core `ReadWithUniformPartitions` (RWUP) transform to support dynamic routing across multiple physical shards. 

### Key Changes:
- **`DataSourceProvider` Injection**: Replaced the single `SerializableFunction` with a more robust `DataSourceProvider`.
- **Dynamic Connection Routing**: Internal `DoFn`s (`RangeCountDoFn`, `RangeBoundaryDoFn`) now extract the `dataSourceId` from each processing element and use the provider to acquire the correct physical connection.
- **Worker-Local Caching**: Implemented a caching mechanism within the `DoFn`s to ensure that a single worker only opens connection pools for the specific shards it is actively processing.

### Rationale:
To process a heterogeneous stream of ranges from different physical databases, the partitioning engine must be able to switch between data sources on-the-fly. This refactor enables that flexibility while maintaining the performance of the uniform splitting algorithm.

---

## Why it's Safe (Serialization & Connection Leaks)
- **Tested Serialization**: The `DataSourceProvider` implementations are explicitly tested for Java serialization. This ensures they can be safely distributed to Dataflow workers without triggering `NotSerializableException`.
- **Resource Leak Prevention**: Added explicit `teardown()` logic to all refactored `DoFn`s. Each `DoFn` calls `DataSourceManager.closeAll()` to ensure that all connection pools opened during the bundle execution are properly shut down.
- **Thread Safety**: Connection routing is protected by the locking strategy implemented in the `DataSourceManager`, ensuring safe concurrent access within the Dataflow worker process.

---

## How to Verify
The added tests use multiple in-memory Derby shards to verify that RWUP correctly routes requests and aggregates counts across different data sources.

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
@@ -49,6 +49,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
@@ -77,10 +78,13 @@ import org.slf4j.LoggerFactory;
  *  and maxPartitionHints (auto inferred) to the pipeline controller helping a better sequencing of tables.
  */
 public final class JdbcIoWrapper implements IoWrapper {
+
   private static final Logger LOG = LoggerFactory.getLogger(JdbcIoWrapper.class);
   // We parallelize to 4 threads to work well with the default launcher machine we get.
   // With this for 1024 shards test we have the launcher complete in 4 minutes in the load test.
   private static final int SOURCE_DISCOVERY_PARALLELISM = 4;
+  // Number of shards per log emitted at the end of discovery.
+  public static final int MAX_SHARDS_PER_LOG = 50;
 
   private final ImmutableMap<
           ImmutableList<SourceTableReference>, PTransform<PBegin, PCollection<SourceRow>>>
@@ -334,6 +338,12 @@ public final class JdbcIoWrapper implements IoWrapper {
 
   private static int getFetchSize(
       JdbcIOWrapperConfig config, TableConfig tableConfig, SourceTableSchema sourceTableSchema) {
+    if (tableConfig.fetchSize() != null) {
+      return tableConfig.fetchSize();
+    }
+    if (config.maxFetchSize() != null) {
+      return config.maxFetchSize();
+    }
     return FetchSizeCalculator.getFetchSize(
         tableConfig,
         sourceTableSchema.estimatedRowSize(),
@@ -640,42 +650,50 @@ public final class JdbcIoWrapper implements IoWrapper {
 
     /* Todo  in subsequent PR for multishard graphsize support, pass this to table reader. */
     DataSourceProvider dataSourceProvider = getDataSourceProvider(perSourceDiscoveries);
-    for (PerSourceDiscovery perSourceDiscovery : perSourceDiscoveries) {
-      ImmutableList.Builder<SourceTableReference> tableReferencesBuilder = ImmutableList.builder();
-      ImmutableList.Builder<TableSplitSpecification> splitSpecsBuilder = ImmutableList.builder();
-      ImmutableMap.Builder<TableIdentifier, TableReadSpecification<SourceRow>> readSpecsBuilder =
-          ImmutableMap.builder();
-      JdbcIOWrapperConfig config = perSourceDiscovery.config();
-      DataSourceConfiguration dataSourceConfiguration =
-          perSourceDiscovery.dataSourceConfiguration();
-      ImmutableList<TableConfig> tableConfigs = perSourceDiscovery.tableConfigs();
+    ImmutableList.Builder<SourceTableReference> tableReferencesBuilder = ImmutableList.builder();
+    ImmutableList.Builder<TableSplitSpecification> splitSpecsBuilder = ImmutableList.builder();
+    ImmutableMap.Builder<TableIdentifier, TableReadSpecification<SourceRow>> readSpecsBuilder =
+        ImmutableMap.builder();
+    accumulateSpecs(
+        perSourceDiscoveries, tableReferencesBuilder, splitSpecsBuilder, readSpecsBuilder);
 
-      if (!config.readWithUniformPartitionsFeatureEnabled() || tableConfigs.isEmpty()) {
-        continue;
-      }
-      accumulateSpecs(
-          perSourceDiscovery, tableReferencesBuilder, splitSpecsBuilder, readSpecsBuilder);
-
-      ReadWithUniformPartitions<SourceRow> readWithUniformPartitions =
-          ReadWithUniformPartitions.<SourceRow>builder()
-              .setTableSplitSpecifications(splitSpecsBuilder.build())
-              .setTableReadSpecifications(readSpecsBuilder.build())
-              .setDataSourceProviderFn(
-                  JdbcIO.PoolableDataSourceProvider.of(dataSourceConfiguration))
-              .setDbAdapter(dialectAdapter)
-              .setWaitOn(waitOn)
-              .setDbParallelizationForSplitProcess(dbParallelizationForSplitProcess)
-              .setDbParallelizationForReads(dbParallelizationForReads)
-              .setAdditionalOperationsOnRanges(additionalOperationsOnRanges)
-              .build();
-
-      LOG.info(
-          "Configured Multi-Table ReadWithUniformPartitions {} for tables {} with config {}",
-          readWithUniformPartitions,
-          tableConfigs.stream().map(TableConfig::tableName).collect(Collectors.toList()),
-          config);
-      tableReadersBuilder.put(tableReferencesBuilder.build(), readWithUniformPartitions);
+    ImmutableList<SourceTableReference> tableReferences = tableReferencesBuilder.build();
+    if (tableReferences.isEmpty()) {
+      return ImmutableMap.of();
     }
+
+    ReadWithUniformPartitions<SourceRow> readWithUniformPartitions =
+        ReadWithUniformPartitions.<SourceRow>builder()
+            .setTableSplitSpecifications(splitSpecsBuilder.build())
+            .setTableReadSpecifications(readSpecsBuilder.build())
+            .setDataSourceProvider(dataSourceProvider)
+            .setDbAdapter(dialectAdapter)
+            .setWaitOn(waitOn)
+            .setDbParallelizationForSplitProcess(dbParallelizationForSplitProcess)
+            .setDbParallelizationForReads(dbParallelizationForReads)
+            .setAdditionalOperationsOnRanges(additionalOperationsOnRanges)
+            .build();
+
+    // We batch the partitions in groups of MAX_SHARDS_PER_LOG batches for
+    // enhanced log readability and debuggability at the same time containing
+    // the number of logs emitted.
+    Lists.partition(perSourceDiscoveries, MAX_SHARDS_PER_LOG)
+        .forEach(
+            batch ->
+                LOG.info(
+                    "Configured Multi-Table ReadWithUniformPartitions for sources batch: {}",
+                    batch.stream()
+                        .map(
+                            d ->
+                                "id="
+                                    + d.config().id()
+                                    + ":"
+                                    + "shard_id="
+                                    + d.config().shardID()
+                                    + ":'"
+                                    + d.sourceSchema().schemaReference().jdbc().toString())
+                        .collect(Collectors.joining(","))));
+    tableReadersBuilder.put(tableReferences, readWithUniformPartitions);
     return tableReadersBuilder.build();
   }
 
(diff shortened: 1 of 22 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -111,31 +119,32 @@ public void setup() throws Exception {
...
           // PostgreSQL requires autocommit to be disabled to enable cursor streaming
           // see https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
           // This option is configurable as Informix will error
           // if calling setAutoCommit on a non-logged database
           if (disableAutoCommit) {
-            LOG.info("Autocommit has been disabled");
+            LOG.info("Autocommit has been disabled for shard {}", dataSourceId);
             connection.setAutoCommit(false);
           }
+          this.connections.put(dataSourceId, connection);
```
sm745052: i think we should put this line just after getConnection() . that would ensure if `connection.setAutoCommit(false);` throws error, the created connection gets cleared up by cleanUpConnection

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -659,42 +660,47 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+            .setTableReadSpecifications(readSpecsBuilder.build())
+            .setDataSourceProvider(dataSourceProvider)
+            .setDbAdapter(dialectAdapter)
+            .setWaitOn(waitOn)
+            .setDbParallelizationForSplitProcess(dbParallelizationForSplitProcess)
+            .setDbParallelizationForReads(dbParallelizationForReads)
+            .setAdditionalOperationsOnRanges(additionalOperationsOnRanges)
+            .build();
+
+    Lists.partition(perSourceDiscoveries, 50)
```
rohitwali, line 687: curious: why 50?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Makes sense, `setAutoCommit` can technically throw exceptions.
Moved it up.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
```diff
@@ -38,55 +39,54 @@
...
-    dataSource = dataSourceProviderFn.apply(null);
     this.tableSplitSpecificationMap =
         this.tableSplitSpecifications.stream()
             .collect(
                 Collectors.toMap(TableSplitSpecification::tableIdentifier, Function.identity()));
   }
 
-  private Connection acquireConnection() throws SQLException {
-    return checkStateNotNull(this.dataSource).getConnection();
+  @StartBundle
```
rohitwali, line 86: curious: whats the significance of StartBundle? why wouldnt I put this in @Setup instead?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -720,58 +726,71 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
-        tableReadSpecificationBuilder =
-            tableReadSpecificationBuilder.setFetchSize(config.maxFetchSize());
+    for (PerSourceDiscovery perSourceDiscovery : perSourceDiscoveries) {
+      JdbcIOWrapperConfig config = perSourceDiscovery.config();
+      SourceSchemaReference sourceSchemaReference =
+          perSourceDiscovery.sourceSchema().schemaReference();
+      ImmutableList<TableConfig> tableConfigs = perSourceDiscovery.tableConfigs();
+      for (TableConfig tableConfig : tableConfigs) {
+
+        if (!config.readWithUniformPartitionsFeatureEnabled() || tableConfigs.isEmpty()) {
```
rohitwali, line 741: nit: `config.readWithUniformPartitionsFeatureEnabled()` check be moved out of the loop?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -720,58 +726,71 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+        if (config.splitStageCountHint() >= 0) {
+          tableSplitSpecificationBuilder =
+              tableSplitSpecificationBuilder.setSplitStagesCount(
+                  (long) config.splitStageCountHint());
+        }
+        splitSpecsBuilder.add(tableSplitSpecificationBuilder.build());
+
+        TableReadSpecification.Builder<SourceRow> tableReadSpecificationBuilder =
+            TableReadSpecification.<SourceRow>builder()
+                .setFetchSize(fetchSize)
```
rohitwali, line 768: do we need this now?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -720,58 +726,71 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+            TableSplitSpecification.builder()
+                .setTableIdentifier(tableIdentifier)
+                .setPartitionColumns(tableConfig.partitionColumns())
+                .setApproxRowCount(tableConfig.approxRowCount());
+        if (tableConfig.maxPartitions() != null) {
+          tableSplitSpecificationBuilder =
+              tableSplitSpecificationBuilder.setMaxPartitionsHint(
+                  (long) tableConfig.maxPartitions());
+        }
+        if (config.splitStageCountHint() >= 0) {
```
rohitwali, line 759: what does split count = 0 mean?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
```diff
@@ -720,58 +726,71 @@ private static PTransform<PBegin, PCollection<SourceRow>> getJdbcIO(
...
+                .setTableIdentifier(tableIdentifier)
+                .setRowMapper(
+                    new JdbcSourceRowMapper(
+                        config.valueMappingsProvider(),
+                        sourceSchemaReference,
+                        sourceTableSchema,
+                        config.shardID()));
+        if (config.maxFetchSize() != null) {
+          tableReadSpecificationBuilder =
+              tableReadSpecificationBuilder.setFetchSize(config.maxFetchSize());
```
rohitwali, line 778: why override here? care to leave a comment as well?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
rohitwali: also, do you mean tableConfig.isEmpty? (minus `s`)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
rohitwali, line 277: should we be worried about the connection churn that this will cause?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -275,13 +284,30 @@ public void tearDown() throws Exception {
...
+          try {
+            conn.close();
+          } catch (SQLException e) {
+            LOG.warn("Failed to close connection", e);
+          }
+        }
       }
+      connections.clear();
+    } finally {
+      connectionLock.unlock();
```
rohitwali, line 307: should we unlock after making connections as null?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -275,13 +284,30 @@ public void tearDown() throws Exception {
...
+      dataSourceManager.closeAll();
+      this.connections = null;
+      this.dataSourceManager = null;
```
rohitwali, line 310: curious: whats the impact of not doing this?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/CollationMapperTransform.java
```diff
@@ -44,10 +44,10 @@ public abstract class CollationMapperTransform
     implements Serializable {
 
   /** List of {@link CollationReference} to discover the mapping for. */
-  public abstract ImmutableList<CollationReference> collationReferences();
+  public abstract ImmutableList<KV<String, CollationReference>> collationReferences();
```
rohitwali, line 47: noob question: Is it possible that the shards in a pipeline may have diff collations?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -111,31 +119,32 @@ public void setup() throws Exception {
...
           // see https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
           // This option is configurable as Informix will error
           // if calling setAutoCommit on a non-logged database
           if (disableAutoCommit) {
-            LOG.info("Autocommit has been disabled");
+            LOG.info("Autocommit has been disabled for shard {}", dataSourceId);
             connection.setAutoCommit(false);
           }
+
+          reportLineage(element, connection, dataSource, query, reportedLineages);
```
rohitwali, line 143: should this be outside of try finally? i can see that this might increase the API usage, but at the current place, it would appear to limit our observability. am I missing something?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
```diff
@@ -110,8 +110,8 @@ public void processElement(
                 .map(pc -> pc.columnName())
                 .collect(ImmutableList.toImmutableList()),
             input.columnName());
-
-    try (Connection conn = acquireConnection()) {
+    DataSource dataSource = dataSourceManager.getDatasource(input.tableIdentifier().dataSourceId());
```
rohitwali, line 113: so here and in Range Count Do Fn we are using try with resources, where as in mutli-table reader, we are caching the connections, is there a reason for this?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapperTest.java
```diff
@@ -1368,6 +1376,41 @@ public void testAccumulateSpecs_Empty() {
    * Tests that {@link JdbcIoWrapper#accumulateSpecs} does not generate split specifications when
    * the uniform partitions feature is disabled.
    */
+  @Test
+  public void testAccumulateSpecs_FeatureDisabled() {
```
rohitwali, line 1380: would it make sense to add a test case where one of the shards is not reachable? not sure if its already covered elsewhere

VardhanThigle: If any shard is not reachable the pipeline will fail, this is similar to the older behavior (not affected by graph size). A transient unavailability is handled by dataflow via 5 retries, but beyond that pipeline fails.

Now that we have forked away from the beam's connecter we could tweak this slightly but that will need a slightly deeper design. I will sync up with you separately about this and take it up as a todo.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: We report lineage only when we open the connection for the first time. If the connection opening fails, it might not make sense to report the lineage, this is inline with the [JdbcIO](https://github.com/apache/beam/blob/25370f50b39f2703ee1a15780b76fa52464fc581/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1742) behavior always.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/CollationMapperTransform.java
VardhanThigle: Thanks. Good questions actually :)
The schemas could have different collation and we handle that - for exmaple one table could use utf8mb4 and other utf8mb3 - that's perfectly handled.

Could it be possible that `utf8mb4` has a slightly different implementation across shards - highly unlikely. In the current scheme we assume that all shards have same implementation of a given collation and make the scale work to our advantage (we parallelize detection of different collations on different shards if they are available) - having multiple shards helps us get better parallelism by utilizing them in parallel to get the ordering of all collation that we need for this migration. Apart from performance considerations, maintaining a separate map of the collation order for 1024 shards will be quite a load worker memory.

I will update the readme note to reflect this.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: This is just for debugability and logging.

As of now logs are the only way for us to debug the details and  we can print all the 1024 shards that we have initialized in the same log, with a bit of trial and error on my load test runs, found out 50 is a good number  thats not making one log too big and not leading to too many logs as well.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: At the level of reader we support switching the the upstream beam jdbcIO just in case needed for quick experiments or any unforeseen issue. That was supported at table level, so we are continuing to retain that here.

I had to refactor the code a bit and this check moved inside. Thanks. it belongs to the outer loop. Will move it.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: `0` means the ReadWithUniformPartition will autoinfer  the hieght of the tree to `log2(ApproxRowCount)`

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: yes, the `fetchsize` is autoinfered by JdbcIO based on the Sql dialect, schema  etc and pased to the `ReadWithUniformPartiton`.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: This is for handling user configured override. Will add a comment.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
```diff
@@ -208,7 +217,7 @@ protected static <ParameterT> void reportLineage(
   // https://github.com/apache/beam/blob/676c998dec78e878d54ad21cde46f91cc9a598b7/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L836
   private static final Pattern READ_STATEMENT_PATTERN =
       Pattern.compile(
-          "SELECT\\s+.+?\\s+FROM\\s+(\\[?`?(?<schemaName>[^\\s\\[\\]`]+)\\]?`?\\.)?\\[?`?(?<tableName>[^\\s\\[\\]`]+)\\]?`?",
+          "SELECT\\s+.+?\\s+FROM\\s+(\\[?`?(?P<schemaName>[^\\s\\[\\]`]+)\\]?`?\\.)?\\[?`?(?P<tableName>[^\\s\\[\\]`]+)\\]?`?",
```
sm745052, line 220: why did we change this ?

sm745052: was there a regex library change ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
sm745052: +1

sm745052 (commented):
Mostly LGTM .. some minor comments

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
rohitwali: ok. do you want to extract this into a constant with comment/guidance on how to revisit this if required?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
VardhanThigle: The connections are managed at the bundle level. In typical large-scale migrations, Dataflow processes large bundles of ranges, so the overhead of opening/closing connections per bundle is amortized across many elements.  Also a given Range itself is a long range read. 

Even in the upstream JdbcIO, the connections are managed at bundle level - [ref](https://github.com/apache/beam/blob/master/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1782)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java
VardhanThigle: Thanks. I have moved it in the recent patch.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/CollationMapperTransform.java
VardhanThigle: The readme is updated in the recent patch.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/RangeBoundaryDoFn.java
VardhanThigle: The counts are expected to be shortlived - unlike the reads which are long running reads.
The pattern followed here is same as the one followed in upstream JdbcIOWrapper.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: We have corrected this from the previous table-wide changes.
As we use the google standard [re2](https://github.com/google/re2/wiki/Syntax), it required `P` for the named capture group.  This was caught while adding UTs hence fixed. 

Rest we are making the pattern allow quotes and backticks in table names.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: This is the cleanup at finish bundle level, start bundle will re-initialize the connections. This is same as the existing pattern - [ref](https://github.com/apache/beam/blob/0cdd1b49f3b5ea61ad4efc5ee22131c8d951044b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L1782)

sm745052 (approved):

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Makes sense. Reordered the sequence in recent push.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/transforms/MultiTableReadFn.java
VardhanThigle: Discussed ofline on the lifetime of connection from startBundle to FInishBundle.

rohitwali (approved):
thanks for the changes. LGTM!

PR #3712, opened 2026-04-20
Adding MySQL data writer in data generator

shreyakhajanchi:
# PR Description: Introduce MySqlDataWriter and Refactor Connection Helpers

## Goal
This PR introduces the `MySqlDataWriter` and related infrastructure in the `cdc-data-generator` module to support writing generated data to MySQL. Additionally, it refactors database connection helper classes by moving them from `spanner-to-sourcedb` to a more common location (`spanner-common`) to enable code sharing.

## Proposed Changes

*   **`MySqlDataWriter.java`**: Added implementation to handle writing batch rows to MySQL shards, supporting dynamic pool size configuration.
*   **`DataWriter.java`**: Defined the interface for data writers, including the parameterized `write` method.
*   **`MySqlDataWriterTest.java`**: Added comprehensive unit tests covering statement parameter setting, batch execution, and pool initialization.
*   **`Constants.java`**: Added relevant constants for the data generator.


v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/sink/DataWriter.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param operation one of {@link
+   *     com.google.cloud.teleport.v2.templates.utils.Constants#MUTATION_INSERT}, {@code
+   *     MUTATION_UPDATE}, or {@code MUTATION_DELETE}.
+   * @param maxShardConnections maximum number of connections per shard.
+   */
+  void write(
+      List<Row> rows,
+      DataGeneratorTable table,
+      String shardId,
+      String operation,
```
VardhanThigle: It might be more idiomatic to have operation as an `Enum` instead of a `String`.  It will avoid the `Constants.<Operation>.equalsIgnoreCase(operation)`  and you can directly switch case. Crucially, it's more type-safe as an enum can take only well defined values.

From a slightly different perspective, since there are only 3 operations most of the databases need to support , and all do support these for sure, it might make sense for  `DataWriter`
to have `Insert`, `Update` and `Delete` methods.  This removes the switch case from implementation of each dialect and the code remains more readable for most of the cases.


For extensibility to other dbs like Cassandra (which have LWT for example and we might want to use them when Spanner supports them for example), we can probably have specialOperations method that takes an enum. But we don't need to add that right away.


I am fine with any of the approaches, either have one common API with an Enum or 3 well defined methods. I might lean more on the later than the former though.

VardhanThigle (commented):
Looks good to me overall. One open point to discsuss.

VardhanThigle (approved):

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/sink/DataWriter.java
shreyakhajanchi: Makes sense, updated

PR #3713, opened 2026-04-20
Adding 1024 shard (32 physical X 32 logical) load test for multi shard graph size optimization support.

VardhanThigle:
# Verification of 1024 shard support
This is the 7th child of #3684 . This PR introduces the integration tests and load test infrastructure required to validate the multi-shard graph size optimization at scale.   Currently the tests are verified manually and added in disabled mode as we are waiting the release.

### Key Components:
- **`CloudSqlShardOrchestrator`**: A test utility that automates the provisioning and configuration of multi-shard environments in Cloud SQL. This just orchestrates the shards and does not orchestrate data at scale in those shards. Also currently it's emmiting only bulk related shard configuration as that's the immediate use case but the class can be reparented and extended as needed.
- **1,024-Shard Load Test**: A high-scale test case that migrates **5,120 tables** distributed across **1,024 logical shards** in a single job.
- **Dialect Coverage**: Separate test cases for massive-scale MySQL and PostgreSQL migrations.

### Rationale:
The theoretical O(1) graph complexity must be backed by empirical evidence. This PR provides the tools to prove that the template can handle real-world sharded database fleets without degradation.

---

## Why it's Safe (Empirical Proof of Stability)
- **O(1) Graph Complexity Verified**: During the 1,024-shard test, the Dataflow graph remained small and constant in size. This confirms that the decoupling logic is fully effective at massive scales.
- **Resource Management Proof**: Verified that workers successfully managed connection routing to 32 distinct physical MySQL/PG instances without hitting connection limits or worker memory exhaustion.
- **Data Integrity**: The load test verifies that data from all 1,024 shards is correctly consolidated into Spanner with the appropriate `migration_shard_id` metadata.

---

## How to Verify
Execute the massive-scale load tests:
```bash
mvn test -pl v2/sourcedb-to-spanner -Dtest=MySQLMultiSharded1024ShardsLT,PostgreSQLMultiSharded1024ShardsLT
```
*(Requires a dedicated GCP project with high quota for Cloud SQL and Spanner instances)*

---

### Load Test Summary:
| Metric | Result |
| :--- | :--- |
| **Physical Shards** | 32 |
| **Logical Shards** | 1,024 |
| **Total Tables** | 5,120 |
| **Graph Complexity** | Constant (O(1)) |
| **Job Result** | SUCCESS |

### DIFF
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/CloudSqlShardOrchestrator.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/CloudSqlShardOrchestrator.java
@@ -1,0 +1,512 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates.loadtesting;
+
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.services.sqladmin.SQLAdmin;
+import com.google.api.services.sqladmin.model.DatabaseInstance;
+import com.google.api.services.sqladmin.model.IpMapping;
+import com.google.api.services.sqladmin.model.Operation;
+import com.google.api.services.sqladmin.model.Settings;
+import com.google.api.services.sqladmin.model.User;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.teleport.v2.source.reader.io.jdbc.iowrapper.config.SQLDialect;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import org.apache.beam.it.gcp.artifacts.GcsArtifact;
+import org.apache.beam.it.gcp.cloudsql.CloudMySQLResourceManager;
+import org.apache.beam.it.gcp.cloudsql.CloudPostgresResourceManager;
+import org.apache.beam.it.gcp.cloudsql.CloudSqlResourceManager;
+import org.apache.beam.it.gcp.storage.GcsResourceManager;
+import org.json.JSONArray;
+import org.json.JSONObject;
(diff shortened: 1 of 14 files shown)


v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/CloudSqlShardOrchestrator.java
```diff
@@ -0,0 +1,492 @@
+/*
```
pratickchokhani, line 1: Should this be moved to ID google-cloud-platform? 
1. This can be use full for other templates as well.
2. With this being here, we are adding test for a test.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQL5KTablesLT.java
```diff
@@ -117,14 +117,6 @@ public void mySQLToSpannerBulkFiveThousandTablesTest() throws Exception {
     columns.put("id", "BIGINT UNSIGNED");
     JDBCResourceManager.JDBCSchema schema = new JDBCResourceManager.JDBCSchema(columns, "id");
 
-    // OPTIMIZE MYSQL:
-    // We disable synchronous flushing and binary logging to speed up the creation and
-    // population of 5,000 tables on the source database.
-    try (Connection jdbcConnection = getJdbcConnection(mySQLResourceManager);
-        PreparedStatement pstmt =
```
pratickchokhani, line 122: Not clear why are we removing this?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/CloudSqlShardOrchestrator.java
VardhanThigle: We not are adding UT for the actual Load test,  We are adding UT for the part that's helping us orchestrate multisharded load tests. The logic has it's own complexity, hence it makes sense to UT the same. 

This is giving a shard file for Bulk, so it's currently in bulk folder. We can reparent it  to a common place once we have a common sharding file across all our templates.

It is intended to be used by other tests that need static resources in future, though right now bulk is the only consumer.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQL5KTablesLT.java
VardhanThigle: Thanks. This is an unclean merge with my other PRs.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/CloudSqlShardOrchestrator.java
```diff
@@ -0,0 +1,492 @@
...
+    if (privateIp == null) {
+      throw new RuntimeException("Instance " + instanceName + " does not have a private IP.");
+    }
+    return privateIp;
+  }
+
+  protected void createPhysicalInstance(String instanceName)
+      throws IOException, InterruptedException {
+    String databaseVersion = dbType == SQLDialect.MYSQL ? "MYSQL_8_0" : "POSTGRES_14";
+    String tier = dbType == SQLDialect.MYSQL ? "db-n1-standard-2" : "db-custom-2-7680";
```
sm745052: can we parameterize these given that we have mysql 5.7 support coming as well

sm745052: not sure if we intend to add LTs there but having this might help in future

VardhanThigle: The version part is parametrized. 
The dialect to tier part is CluodSQL nuance, I don't think it depends on the version, so kept it same for now.

sm745052 (approved):
LGTM

PR #3719, opened 2026-04-21
fix multiple gcs reads for constructing schema mapper

sm745052:
fixes b/504540481

### DIFF
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
@@ -36,6 +36,9 @@ import com.google.cloud.teleport.v2.cdc.dlq.StringDeadLetterQueueSanitizer;
 import com.google.cloud.teleport.v2.coders.FailsafeElementCoder;
 import com.google.cloud.teleport.v2.common.UncaughtExceptionLogger;
 import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.Schema;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaFileOverride;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SchemaFileOverridesParser;
 import com.google.cloud.teleport.v2.spanner.migrations.shard.CassandraShard;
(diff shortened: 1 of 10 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/SchemaMapperUtils.java
```diff
@@ -51,15 +51,15 @@ public static ISchemaMapper getSchemaMapper(
 
     if (overrideTypesCount > 1) {
       throw new IllegalArgumentException(
-          "Only one type of schema override can be specified. Please use only one of: sessionFilePath, "
-              + "schemaOverridesFilePath, or tableOverrides/columnOverrides.");
+          "Only one type of schema override can be specified. Please pass only one of: schema, "
```
aasthabharill: For logging the earlier message seems clearer as it mentions the template parameters that the user would have passed. "schema" is ambiguous and doesn't point the user to session file IMO

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDb.java
```diff
@@ -795,13 +812,10 @@ public static PipelineResult run(Options options) {
                             options.getShardingCustomParameters(),
                             options.getMaxShardConnections() * shards.size(),
                             options.getSourceType(),
-                            options.getSessionFilePath(),
-                            options.getSchemaOverridesFilePath(),
+                            schema,
```
aasthabharill: for massive schemas with a large number of tables, won't this object become huge? Will that result in higher memory usage on the workers or increase size of the pipeline graph?

sm745052: added it in setup

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/SchemaMapperUtils.java
sm745052: ack

manitgupta (approved):
We discussed a bunch around class serialization getting bloated up if we construct the schemaMapper during compile time. Some tests were done which said that session file can potentially be in GBs for worst cases.

PR #3732, opened 2026-04-24
Add AWS dependencies to the YAML template

chamikaramj:
This will be useful for YAML and Dataflow Job builder when executing pipelines that connect to AWS (for example, reading from S3).


derrickaw (approved):

chamikaramj:
Thanks!

PR #3733, opened 2026-04-24
Adds a package that includes utility transforms for Job builder

chamikaramj:
This will include utility transforms that wii be used by Job builder via a Python provider. Adding them here (instead of directly in Job builder makes such transforms better testable).

Also adds a Github workflow for pushing a package that includes such transforms to GCS.

### DIFF
--- a/.github/workflows/upload-python-package.yml
+++ b/.github/workflows/upload-python-package.yml
@@ -40,6 +40,14 @@ jobs:
           python -m pip install --upgrade pip
           pip install poetry
 
+      - name: Install project dependencies
+        working-directory: python/src/main/python/job-builder-util-transforms
+        run: poetry install
+
+      - name: Run Tests
+        working-directory: python/src/main/python/job-builder-util-transforms
+        run: poetry run pytest ../../../test/python/job-builder-util-transforms/
+
       - name: Build Package
         working-directory: python/src/main/python/job-builder-util-transforms
         run: poetry build
--- a/.github/workflows/upload-python-package.yml
+++ b/.github/workflows/upload-python-package.yml
@@ -1,0 +1,51 @@
+# ====================================================================================
+# Purpose:
+#   This workflow builds and uploads the job-builder-util-transforms Python package to Google Cloud Storage (GCS).
+#   It is designed to distribute the utility transforms package for use in Dataflow jobs.
+#   The workflow:
+#     1. Sets up Python and installs build dependencies.
+#     2. Builds the Python package (wheel and sdist).
+#     3. Uploads the built artifacts to a specified GCS bucket.
+#
+# How to Run:
+#   This workflow is triggered manually via GitHub Actions `workflow_dispatch`.
+#   Required Inputs:
+#     - bucket_path: The GCS bucket destination (e.g., gs://my-bucket/path).
+# ====================================================================================
+
+name: Build and Upload Python Package for Job Builder
+
+on:
+  workflow_dispatch:
+    inputs:
+      bucket_path:
+        description: 'GCS bucket path (e.g., gs://my-bucket/path)'
+        required: true
+
+jobs:
+  build-and-upload:
+    runs-on: [self-hosted, release]
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.11'
+
+      - name: Install build dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
+
+      - name: Build Package
+        working-directory: python/src/main/python/job-builder-util-transforms
+        run: poetry build
+
+      - name: Upload to GCS
+        env:
+          BUCKET_PATH: ${{ github.event.inputs.bucket_path }}
+        run: |
+          gcloud storage cp python/src/main/python/job-builder-util-transforms/dist/* "$BUCKET_PATH"
--- a/python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
+++ b/python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
@@ -1,0 +1,89 @@
+"""Module containing transforms to copy files to Google Cloud Storage (GCS)."""
+
+import apache_beam as beam
+from apache_beam import Row
+from apache_beam.transforms import DoFn
+from apache_beam.transforms import ParDo
+from apache_beam.transforms import PTransform
+from apache_beam.io.filesystems import FileSystems
+from apache_beam.typehints.row_type import RowTypeConstraint
+import shutil
+import secrets
+
+class CopyFilesToGCSDoFn(DoFn):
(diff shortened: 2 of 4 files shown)


.github/workflows/upload-python-package.yml
```diff
@@ -0,0 +1,51 @@
...
+          python-version: '3.11'
+
+      - name: Install build dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
+
+      - name: Build Package
+        working-directory: python/src/main/python/job-builder-util-transforms
+        run: poetry build
```
damccorm, line 53: Should we be running tests in this workflow as well?

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    Args:
+      job_level_gcs_prefix: The job-level GCS prefix to use for copying files.
+    """
+    self.job_level_gcs_prefix = job_level_gcs_prefix
+
+  def start_bundle(self):
+    """Starts a bundle. Generates a unique bundle-level prefix if a job-level prefix is set."""
+    # Generating a random string to be added to the path during bundle startup
+    # so that we do not write duplicate data due to bundle retries by the runner.
```
damccorm: Doesn't start_bundle get called again when a retry happens?

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
```diff
@@ -0,0 +1,89 @@
...
+from apache_beam import Row
+from apache_beam.transforms import DoFn
+from apache_beam.transforms import ParDo
+from apache_beam.transforms import PTransform
+from apache_beam.io.filesystems import FileSystems
+from apache_beam.typehints.row_type import RowTypeConstraint
+import shutil
+import secrets
+
+class CopyFilesToGCSDoFn(DoFn):
```
damccorm, line 13: Is there a reason we're not just putting this in Beam?

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
```diff
@@ -0,0 +1,89 @@
...
+      return
+
+    if not self.bundle_level_gcs_prefix:
+      raise ValueError(f"Encountered non-GCS file {input_file.path} but no gcs_file_path was provided to copy to.")
+
+    file_name = input_file.path.split('/')[-1]
+    gcs_file_path = self.bundle_level_gcs_prefix + file_name
+    with FileSystems.open(input_file.path, 'application/octet-stream') as s3_file:
+      with FileSystems.create(gcs_file_path, 'application/octet-stream') as gcs_file:
+        shutil.copyfileobj(s3_file, gcs_file, length=CopyFilesToGCSDoFn.COPY_BUFFER_SIZE)
```
damccorm:
```suggestion
    with FileSystems.open(input_file.path, 'application/octet-stream') as src_file:
      with FileSystems.create(gcs_file_path, 'application/octet-stream') as gcs_file:
        shutil.copyfileobj(src_file, gcs_file, length=CopyFilesToGCSDoFn.COPY_BUFFER_SIZE)
```

I think this should work for arbitrary filesystems supported by beam, not just s3

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
damccorm: Also, can we just use https://beam.apache.org/releases/pydoc/current/apache_beam.io.fileio.html directly?

.github/workflows/upload-python-package.yml
chamikaramj: Done.

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: These classes are expected to be used by Job builder so better to keep them in the Templates repo I think.

Also, this makes is faster to deploy and release fixes. We can remove this after @derrickaw's work on moving Job builder blueprints to the templates repo.

chamikaramj: I don't think we can use fileio directly for this utility. This follows a `fileio.MatchAll()` glob expansion in the blueprint.

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: This is intentional. We want to generate a new path for bundle retries and only consider such retries in subsequent steps. Updated the docs.

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: Done.

chamikaramj (commented):
Thanks. PTAL.

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
damccorm: Can't we then use ReadMatches? You could keep the filter step for getting rid of gcs files if you wanted I guess.

Then we could use WriteToFiles

The thing I'm worried about is that we're duplicating some functionality here, but doing it in a less robust way (e.g. that won't leave orphaned files)

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
damccorm: Won't this leave orphaned files?

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
damccorm: > These classes are expected to be used by Job builder so better to keep them in the Templates repo I think.

What is special about them such that they're not generally useful?

> We can remove this after @derrickaw's work on moving Job builder blueprints to the templates repo.

Is this because at that point, this would then become part of the blueprints themselves?

damccorm (approved):
Thanks

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: Discussed offline: we will get this in for now and will explore using/adjusting existing Beam utilities in the future.

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: Discussed offline: orphaned files can be addressed via Iceberg cleanup utilities since Iceberg is the primary use-case for this transform.

chamikaramj (commented):
Thanks!

python/src/main/python/job-builder-util-transforms/copy_files_to_gcs.py
chamikaramj: Created https://github.com/apache/beam/issues/38312 for tracking this.

PR #3738, opened 2026-04-28
Adding generate primary key transform in data generator

shreyakhajanchi:
This PR introduces synthetic primary key generation for the CDC Data Generator.

### Primary Key Generation
- **`GeneratePrimaryKey.java` & `GeneratePrimaryKeyFn.java`**: Introduced a new `PTransform` and associated `DoFn` to generate synthetic primary-key `Row`s for tables. The generated rows include the table's PK columns and a synthesised logical shard ID.
- **`GeneratePrimaryKeyTest.java`**: Added corresponding unit and pipeline tests to verify the behavior of primary key generation, including composite keys, shard ID distribution, and schema isolation.
### Data Generator Utilities
- **`DataGeneratorUtils.java`**: Added a utility class to centralize value synthesis and Beam type mapping. It supports generating random values for various supported logical types (STRING, INT64, FLOAT64, BOOLEAN, BYTES, DATE, TIMESTAMP, NUMERIC, JSON, UUID) using Java Faker and `ThreadLocalRandom`.
- **`DataGeneratorUtilsTest.java`**: Added a new test file to provide comprehensive unit tests for `DataGeneratorUtils`, covering value generation for all supported types, type mapping, and edge cases like string length clamping and numeric precision.

### DIFF
--- a/v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
+++ b/v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
@@ -47,11 +47,8 @@ import org.slf4j.LoggerFactory;
  * writers are reused here.
  *
  * <p>Randomness: {@link Faker} is constructed with its default no-arg constructor (which seeds
- * itself from {@code System.nanoTime()} plus a per-instance counter, not from {@code
- * /dev/urandom}), and shard-id selection uses {@link ThreadLocalRandom#current()} to avoid
- * contention under high fan-out. Duplicate-PK errors observed during earlier development were
- * caused by Beam retrying already-applied inserts, not by seed collisions across workers; the sink
- * writers now use upsert semantics, which absorbs retries cleanly.
+ * itself from {@code System.nanoTime()} plus a per-instance counter and shard-id selection uses
+ * {@link ThreadLocalRandom#current()} to avoid contention under high fan-out.
  *
  * <p>Output: {@code KV<tableName, pkRow>} so downstream transforms can shuffle by table while
  * keeping the row schema per-table.
@@ -103,8 +100,10 @@ public class GeneratePrimaryKeyFn extends DoFn<DataGeneratorTable, KV<String, Ro
     List<DataGeneratorColumn> pkColumns = primaryKeyColumns(table);
 
     if (pkColumns.isEmpty()) {
-      // A table without a PK cannot have a unique row key synthesised for it. This is a
-      // schema-definition problem, so log + skip rather than emit an empty Row (which would
+      // A table without a PK cannot have a unique row key synthesised for it. This is
+      // a
+      // schema-definition problem, so log + skip rather than emit an empty Row (which
+      // would
       // corrupt downstream joins on the PK).
       LOG.warn(
           "Table {} has no primary-key columns, or its PK list references unknown columns — "
@@ -159,9 +158,8 @@ public class GeneratePrimaryKeyFn extends DoFn<DataGeneratorTable, KV<String, Ro
   }
 
   /**
-   * Select a shard id for this row. Prefer a configured logical shard id (MySQL with shardFile),
-   * otherwise synthesise a placeholder {@code shard<N>}. Bounded at maxShards >= 1 so {@code
-   * nextInt} never throws.
+   * Select a shard id for this row. Prefer a configured logical shard id, otherwise synthesise a
+   * placeholder {@code shard<N>}. Bounded at maxShards >= 1 so {@code nextInt} never throws.
    */
   private String pickShardId() {
     ThreadLocalRandom rng = ThreadLocalRandom.current();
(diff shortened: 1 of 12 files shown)


anowardear062-svg (approved):

anowardear062-svg (approved):

anowardear062-svg (approved):

anowardear062-svg (approved):

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
```diff
@@ -0,0 +1,169 @@
...
+      @Element DataGeneratorTable table, OutputReceiver<KV<String, Row>> out) {
+    List<DataGeneratorColumn> pkColumns = primaryKeyColumns(table);
+
+    if (pkColumns.isEmpty()) {
+      // A table without a PK cannot have a unique row key synthesised for it. This is
+      // a
+      // schema-definition problem, so log + skip rather than emit an empty Row (which
+      // would
+      // corrupt downstream joins on the PK).
+      LOG.warn(
```
VardhanThigle: [nit] If we are not supporting tables w/ primary keys, should this be an error instead - also do we want to assert and fail the pipeline (if that's something we don't support)?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /** Resolve the PK column list for a table. */
+  @VisibleForTesting
+  public static List<DataGeneratorColumn> primaryKeyColumns(DataGeneratorTable table) {
+    if (table.primaryKeys() == null || table.primaryKeys().isEmpty()) {
+      return new ArrayList<>();
+    }
+    Map<String, DataGeneratorColumn> byName = new HashMap<>();
+    for (DataGeneratorColumn col : table.columns()) {
```
VardhanThigle: /*might not need an immediate code change*/
Looking at few of the exsisting uses of `table.columns()` here, for example [this one](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/78ef89e8ae1a7d9fadb03bc7d1d6750c976f883c/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SourceTable.java#L81), does it make sense to make this interface a map itself? 

This could be a todo, and doesn't block these code changes at the moment.

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
```diff
@@ -0,0 +1,169 @@
...
+    out.output(KV.of(table.name(), rowBuilder.build()));
+  }
+
+  /** Resolve the PK column list for a table. */
+  @VisibleForTesting
+  public static List<DataGeneratorColumn> primaryKeyColumns(DataGeneratorTable table) {
+    if (table.primaryKeys() == null || table.primaryKeys().isEmpty()) {
+      return new ArrayList<>();
+    }
+    Map<String, DataGeneratorColumn> byName = new HashMap<>();
```
VardhanThigle: [nit] Consider using ImmutableMap  ([go/java-practices/choosing-collections](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/choosing-collections.md?cl=head))

Also, it might make sense to use Java Streams here.
Something on the lines of
 ```java
ImmutableMap<String, DataGeneratorColumn> byName = table.columns().stream()
    .collect(ImmutableMap.toImmutableMap(
        DataGeneratorColumn::name,   // Key mapper
        col -> col                   // Value mapper
    ))
```

Similarly, the bottom half of this function where we create an arrayList might benefit with streams as well.

File: v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/utils/DataGeneratorUtils.java [hunk 4]
```diff
@@ -0,0 +1,210 @@
...
+   * the same logical type, so the value can be added directly to a {@code Row.Builder}.
+   */
+  public static Object generateValue(DataGeneratorColumn column, Faker faker) {
+    LogicalType type = column.logicalType();
+    Long size = column.size();
+
+    switch (type) {
+      case STRING:
+        return faker.lorem().characters(limitStringLength(size));
+      case JSON:
```
VardhanThigle, line 64: Could you add more details on the schema of this Json, I am not having full context of this one.

VardhanThigle (commented):
Looks good over all. A few nits and one request for clarification.

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/utils/DataGeneratorUtils.java
shreyakhajanchi: As discussed offline, will handle json schema in the schema config override handling

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/GeneratePrimaryKeyFn.java
shreyakhajanchi: added a todo, will address later

VardhanThigle (approved):

PR #3746, opened 2026-04-30
Apply schema config overrides for data generator

shreyakhajanchi:
### Description
This Pull Request introduces a schema configuration override system and an enhanced schema dependency mapping implementation for the distributed Data Generator pipeline (`v2/cdc-data-generator`). This enables custom data generation rates (QPS limits), Java Faker expressions integration for column-level data synthesis, and data dependency mapping for multi-table pipelines.

#### 1. Schema Configuration & Overrides Processing
* **`SchemaConfig.java`**: Introduced a serialized POJO configuration model representing table-level limits, column skipping rules, standard faker expressions, and foreign key requirements.
* **`ApplyOverridesFn.java`**: Created a `DoFn` to apply table-level overrides (QPS write limits), custom Faker column generation formulas, skip constraints, and foreign key merges.

#### 2. Discovered Schema Mapping (MySQL & Spanner)
* **Primary Key Detection**: Upgraded `MySqlSchemaFetcher` and `SpannerSchemaFetcher` metadata scanning to fetch and map `isPrimaryKey` metadata for tables, ensuring fields required for stateful Dataflow lifecycle events cannot be skipped.
* **Data Model Enhancement**: Expanded the `DataGeneratorColumn` and `DataGeneratorTable` models to support tracking primary keys, builder defaults, and faker generation rules cleanly.

#### 3. Dependency Mapping & Pipeline Refactoring
* **`BuildSchemaDagFn.java`**: Introduced a transform step to build a schema Dependency DAG from foreign key relationships, automatically resolving dependencies and identifying root tables in the generator workflow.
* **`SchemaLoader.java`**: Refactored the schema-loading Side Input pipeline to run DB fetching, override applications, and DAG dependency mapping in a sequential flow.
* **`DataGeneratorUtils.java`**: Added utility helper functions to parse and safe-convert evaluated Faker expressions to correct pipeline Beam logical types.

#### 4. Testing & Cleanup
* Introduced extensive unit test suites leveraging Mockito to achieve high test reliability (>90% line/branch coverage) for all newly added transforms and utilities (`ApplyOverridesFnTest`, `BuildSchemaDagFnTest`, `DataGeneratorUtilsTest`).


v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
```diff
@@ -0,0 +1,197 @@
...
+          }
+          if (tableConfig.getForeignKeys() != null) {
+            tableBuilder.foreignKeys(mergeForeignKeys(existingTable, tableConfig.getForeignKeys()));
+          }
+          tableMap.put(tableName, tableBuilder.build());
+        }
+      }
+    }
+
+    receiver.output(DataGeneratorSchema.builder().tables(ImmutableMap.copyOf(tableMap)).build());
```
manitgupta: Is there a cleaner, functional way to write this? Seems like we could have used something `getOrElse()` sort of thing. Essentially we will use the override if it is exists `orElse` use the default?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+    return ImmutableList.copyOf(updatedColumns);
+  }
+
+  private ImmutableList<DataGeneratorForeignKey> mergeForeignKeys(
+      DataGeneratorTable existingTable, List<SchemaConfig.ForeignKeyConfig> fkConfigs) {
+    LinkedHashMap<String, DataGeneratorForeignKey> mergedFksByName = new LinkedHashMap<>();
+    for (DataGeneratorForeignKey fk : existingTable.foreignKeys()) {
+      mergedFksByName.put(fk.name(), fk);
+    }
```
manitgupta, line 144: Can you add a comment where why `LinkedHashMap` was used?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/BuildSchemaDagFn.java
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory;
+
+/** DoFn to build the schema DAG and log root tables. */
+public class BuildSchemaDagFn extends DoFn<DataGeneratorSchema, DataGeneratorSchema> {
+  private static final Logger LOG = LoggerFactory.getLogger(BuildSchemaDagFn.class);
+
+  @ProcessElement
+  public void processElement(
+      @Element DataGeneratorSchema schema, OutputReceiver<DataGeneratorSchema> receiver) {
+    DataGeneratorSchema processedSchema = SchemaUtils.setSchemaDAG(schema);
```
manitgupta: `setSchemaDAG` doesn't seem like an appropriate name. This looks like a static method creating the DAG so maybe `generateSchemaDAG`?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorTable.java
```diff
@@ -70,7 +70,12 @@ public abstract class DataGeneratorTable implements Serializable {
   public abstract Builder toBuilder();
 
   public static Builder builder() {
-    return new AutoValue_DataGeneratorTable.Builder();
+    return new AutoValue_DataGeneratorTable.Builder()
+        .isRoot(true)
+        .insertQps(1000)
+        .updateQps(0)
+        .deleteQps(0)
+        .recordsPerTick(1.0);
```
manitgupta: Why do we have these defaults? Is this dangerous?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/model/SchemaConfig.java
```diff
@@ -0,0 +1,142 @@
...
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates.model;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+/** Configuration class for schema overrides, used for deserializing HOCON/JSON config files. */
+public class SchemaConfig implements Serializable {
```
manitgupta, line 23: 1. Want to use `@AutoValue` ?
2. Lets try to avoid using Serializable (and its coder) where possible.

manitgupta (commented):
I saw `Serializable` being used in a number of places. Please refactor to use better coders (in this PR or as a follow up). Let's try to avoid using Serializable as a rule without going into too many details of whether its a data class or not -- not having it is always better than having it and reasoning that it won't cause any harm (then we don't need to have that discussion!!)

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/model/DataGeneratorTable.java
shreyakhajanchi: Ideally the code should never need them because the values are derived from global defaults first which are the pipeline options , but in case they can cause issues removed them

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
shreyakhajanchi: added

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/model/SchemaConfig.java
shreyakhajanchi: Modified the schema file processing to get rid of Serializable, for other classes will create a follow up PR

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
```diff
@@ -0,0 +1,227 @@
...
+            new InputStreamReader(Channels.newInputStream(channel), StandardCharsets.UTF_8)) {
+          String content = CharStreams.toString(reader);
+          ObjectMapper mapper = new ObjectMapper(new HoconFactory());
+          this.schemaConfig = mapper.readValue(content, SchemaConfig.class);
+        }
+      } catch (IOException e) {
+        throw new RuntimeException("Failed to read schema config from " + schemaConfigPath, e);
+      }
+    }
+  }
```
manitgupta: Why do we want to load this static schema configuration file on every @Setup of this DoFn? Can we not pass it into it?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
```diff
@@ -0,0 +1,227 @@
...
+
+  @ProcessElement
+  public void processElement(
+      @Element DataGeneratorSchema schema, OutputReceiver<DataGeneratorSchema> receiver) {
+    Map<String, SchemaConfig.TableConfig> configTables =
+        (schemaConfig != null && schemaConfig.getTables() != null)
+            ? schemaConfig.getTables()
+            : ImmutableMap.of();
```
manitgupta: Can we not do these checks at the time of creation of the `schemaConfig` objects instead of potentially everytime we try to read from it?

Simply put - a `null`  schemaConfig or a schemaConfig with `null` tables should simply never be constructed and passed through to other places?

v2/cdc-data-generator/pom.xml
```diff
@@ -80,6 +80,21 @@
...
+            <version>1.1.0</version>
+        </dependency>
+
+
+        <!-- Java Faker for data generation -->
+        <dependency>
+            <groupId>com.github.javafaker</groupId>
+            <artifactId>javafaker</artifactId>
+            <version>1.0.2</version>
+        </dependency>
```
manitgupta: I googled this a little bit, I see that DataFaker is a modern successor of JavaFaker. Please evaluate if we should be switching to that.

shreyakhajanchi: yes, i had missed it, done

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
shreyakhajanchi: done

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/ApplyOverridesFn.java
shreyakhajanchi: For now I have restored to passing config and using Serializable and created a bug to remove Serializable across files in the pipeline.

manitgupta (approved):

PR #3315, opened 2026-02-11
Merge MongoDb and Jms modules

Abacn:
#2279 - moved 5 modules into 2, 7 containers into 3

PR #3752, opened 2026-05-01
Create mongodb-to-mongodb template

michaeltle-goog:


v2/pom.xml
```diff
@@ -731,6 +731,7 @@
         <module>kinesis-to-pubsub</module>
         <module>kudu-to-bigquery</module>
         <module>managed-io-to-managed-io</module>
+        <module>mongodb-to-mongodb</module>
```
Abacn, line 734: There is effort to consolidate templates by its source/sink dependency types. In particular MongoDB and Googlecloud templates has put into `v2/mongodb-and-googlecloud` module and share the template image: #3315. We can use the existing module.

michaeltle-goog: Thanks for the heads up! I'll move it once the impl is complete

michaeltle-goog: We're going to move this under a shared `firestore` directory instead so that the Firestore team can own the code. For now, we intend to submit this code to unblock a customer, then I'll send a follow up PR to fix the grouping.

ninjaAB-5 (approved):

PR #3716, opened 2026-04-20
[DO NOT MERGE] Validate beam 273 rc1

Amar3tto:

PR #3760, opened 2026-05-05
Update target Beam version to 2.73.0

jrmccluskey:
Updates target Beam version to 2.73.0, updates Python requirements files with new dependency versions and corresponding hashes.

### DIFF
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
@@ -143,7 +143,7 @@ public class DataCastingUtils {
                 case NanosDuration.IDENTIFIER:
                   castVals.add(asDuration(objVal));
                   break;
-                case org.apache.beam.sdk.schemas.logicaltypes.Date.IDENTIFIER:
+                case "beam:logical_type:date:v1":
                   castVals.add(asDate(objVal));
                   break;
                 case org.apache.beam.sdk.schemas.logicaltypes.DateTime.IDENTIFIER:
--- a/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
+++ b/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
@@ -77,7 +77,7 @@ public class DataCastingUtils {
   private static final Logger LOG = LoggerFactory.getLogger(DataCastingUtils.class);
 
   // Evaluate the Date IDENTIFIER here so the switch statement doesn't complain.
-  private static final String DATE_LOGICAL_TYPE = org.apache.beam.sdk.schemas.logicaltypes.Date.IDENTIFIER;
+  private static final String DATE_LOGICAL_TYPE = org.apache.beam.sdk.schemas.logicaltypes.Date.getIdentifier();
 
   public static List<Object> sourceTextToTargetObjects(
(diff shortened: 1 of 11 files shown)


v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
```diff
@@ -143,7 +143,7 @@ public static List<Object> sourceTextToTargetObjects(
                 case NanosDuration.IDENTIFIER:
                   castVals.add(asDuration(objVal));
                   break;
-                case org.apache.beam.sdk.schemas.logicaltypes.Date.IDENTIFIER:
+                case "beam:logical_type:date:v1":
```
jrmccluskey: This was the fix in the validation PR #3716. I am unfamiliar with schema changes but I would probably like to avoid just using a hard-coded string for this.

jrmccluskey: Ah, I see what happened, `org.apache.beam.sdk/.schemas.logicaltypes.Date.IDENTIFIER`changed from a static string to an expression in https://github.com/apache/beam/pull/38077

The relevant error here is:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.15.0:compile (default-compile) on project googlecloud-to-neo4j: Compilation failure
Error:  /home/runner/actions-runner/_work/DataflowTemplates/DataflowTemplates/v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java:[146,67] constant string expression required
Error:  -> [Help 1]
```

jrmccluskey:
@gemini-code-assist review

v2/googlecloud-to-neo4j/src/main/java/com/google/cloud/teleport/v2/neo4j/utils/DataCastingUtils.java
jrmccluskey: CC @ahmedabu98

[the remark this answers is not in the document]
damccorm: These seem equivalent - https://github.com/apache/beam/blob/6fc8fdebb2349d9efbb1b562407e08d787d01d8f/model/pipeline/src/main/proto/org/apache/beam/model/pipeline/v1/schema.proto#L183

But I see https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3716/changes#diff-1d27ca871b3250036e5dd6775e438188f93bc827c59bf2c4ed4c627f84a236db made this change

@Amar3tto @Abacn why was this change in the validation PR?

jrmccluskey: @damccorm this is the context for the problem with org.apache.beam.sdk.schemas.logicaltypes.Date.IDENTIFIER

Replacing the switch statement with an if-else change avoids that breakage though, which is the solution in the current PR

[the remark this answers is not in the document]
damccorm: I see - https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3760#discussion_r3196566453 explains this. Thanks Jack!

damccorm: SGTM, this approach is more robust

damccorm (approved):
Thanks!

PR #3773, opened 2026-05-08
feat(spanner-to-sourcedb): add Spanner-to-Spanner reverse replication support

srozsnyai:
Extends the spanner-to-sourcedb reverse replication template to support Cloud Spanner as a target database.

Change stream events from a source Spanner instance are converted to mutations and written to a target Spanner database, coordinated through the existing shadow-table mechanism that prevents duplicate and out-of-order writes.

Because calling any Spanner write API inside an active `readWriteTransaction` causes a nested-transaction error, the target write is deferred: the mutation is generated inside the shadow-table transaction (where ordering and filtering decisions are made), then committed via `writeAtLeastOnce()` after the shadow-table transaction completes.


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -328,6 +354,13 @@ public void processElement(ProcessContext c) {
                           }
                           return isRecordWritten.get();
                         });
+
+        // Apply the deferred Spanner target write now that the shadow-table TX has committed.
+        if (pendingSpannerWrite.get() != null) {
+          IDao sourceDao = sourceProcessor.getSourceDao(shardId);
+          sourceDao.write(pendingSpannerWrite.get(), null);
+        }
```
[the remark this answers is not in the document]
srozsnyai: shadowTransaction.buffer(mutation) only works when the mutation targets the same database as the transaction. The shadow table currently lives in a separate metadata database, so buffer() cannot span them.

writeAtLeastOnce inside readWriteTransaction throws INTERNAL: Nested transactions are not supported for every record, regardless of which DatabaseClient instance is used. The same error reproduces with the Spanner JDBC driver, so JDBC is not an escape hatch. MySQL/Postgres can use this pattern only because their JDBC writes target a non-Spanner engine.

The residual risk is narrow in practice: the client retries ABORTED / UNAVAILABLE / DEADLINE_EXCEEDED internally, errors that escape are typically permanent and operator-visible, and Spanner mutations are idempotent.

IMHO the only real fix is colocating the reverse-replication metadata in the target Spanner database so tx.buffer() becomes available. That is an architectural change and will be addressed as a follow-up once there is team consensus, not as a side effect of this PR.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,424 @@
...
+      if (targetCol.isGenerated()) {
+        continue;
+      }
+
+      String targetColName = targetCol.name();
+
+      if (request.getCustomTransformationResponse() != null
+          && request.getCustomTransformationResponse().containsKey(targetColName)) {
+        Object customVal = request.getCustomTransformationResponse().get(targetColName);
+        builder.set(targetColName).to(customVal == null ? (String) null : customVal.toString());
```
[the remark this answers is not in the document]
srozsnyai: Fixed in 7a5274c14. Custom-transformation values are now routed through `setCustomColumnValue` / `appendCustomKeyComponent`, which dispatch on `Column.type().getCode()` and call the matching typed `.to(...)` / `Key.Builder.append(...)` overload. `null` custom values emit a properly typed `NULL` via `setNullValue`. Added tests.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Manages {@link DatabaseClient} connections to target Cloud Spanner databases.
+ */
+public class SpannerConnectionHelper implements IConnectionHelper<DatabaseClient> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerConnectionHelper.class);
+
+  private static Map<String, DatabaseClient> clientMap = new ConcurrentHashMap<>();
+  private static Spanner spannerService;
```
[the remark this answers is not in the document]
srozsnyai: Same idempotent-init pattern as JdbcConnectionHelper and CassandraConnectionHelper — all shards are processed in a single init() call, so the early return doesn't skip later shards.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,424 @@
...
+        builder.set(targetColName).to((Timestamp) null);
+        break;
+      case NUMERIC:
+        builder.set(targetColName).to((BigDecimal) null);
+        break;
+      case JSON:
+        builder.set(targetColName).to(Value.json(null));
+        break;
+      case ARRAY:
+        builder.set(targetColName).to(Value.stringArray(null));
```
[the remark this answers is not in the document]
srozsnyai: Fixed in `a00dd66b5`. setNullValue now dispatches on
type.getArrayElementType().getCode() and emits the matching
Value.*Array((Iterable<X>) null) so NULL ARRAY values carry the correct
element type. Added tests.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SpannerShardFileReader.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Reads a JSON array of Spanner shard configurations from GCS and returns a list of {@link
+ * SpannerShard} instances.
+ *
+ * <p>Each entry in the JSON file must contain the following fields:
+ *
+ * <pre>
+ * [
+ *   {
+ *     "logicalShardId": "shard1",
```
shreyakhajanchi: why do we need shard id ,in case of spanner as it will always be a single shard migration?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/SpannerShard.java
```diff
@@ -0,0 +1,75 @@
...
+public class SpannerShard extends Shard {
+
+  private final String projectId;
+
+  public SpannerShard(
+      String logicalShardId, String projectId, String instanceId, String databaseId) {
+    super();
+    this.projectId = projectId;
+    setLogicalShardId(logicalShardId);
+    setNamespace(instanceId);
```
shreyakhajanchi: we should use instance id only because here namespace refers to schema in postgresql and we should use it for the same in case of spanner .

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
```diff
@@ -0,0 +1,170 @@
...
+   */
+  static String spannerTypeToString(Type type) {
+    switch (type.getCode()) {
+      case BOOL:
+        return "BOOL";
+      case INT64:
+        return "INT64";
+      case FLOAT32:
+        return "FLOAT32";
+      case FLOAT64:
```
shreyakhajanchi: can we simplify this switch case to something like:
```
static String spannerTypeToString(Type type) {
    switch (type.getCode()) {
      case ARRAY:
        return "ARRAY<" + spannerTypeToString(type.getArrayElementType()) + ">";
      case PG_ARRAY:
        return "PG_ARRAY<" + spannerTypeToString(type.getArrayElementType()) + ">";
      default:
        return type.getCode().name();
    }
}
```

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGeneratorTest.java
```diff
@@ -0,0 +1,1139 @@
...
+  @Test
+  public void customTransformationTimestampIsBoundAsTimestamp() throws Exception {
+    Ddl ddl = buildDdlWithSingleNonPkCol("Ts", Type.timestamp());
+    SourceSchema schema = buildSchemaWithSingleNonPkCol("Ts", "TIMESTAMP");
+    ISchemaMapper mapper = buildMapperForSingleColTable(schema);
+
+    JSONObject newValues = new JSONObject("{\"Ts\":\"2024-01-15T10:30:00Z\"}");
+    JSONObject keyValues = new JSONObject("{\"Id\":\"1\"}");
+
+    java.util.Map<String, Object> custom = new java.util.HashMap<>();
```
shreyakhajanchi: can we import the classes wherever there is no conflict for better readability rather than using fully classified class paths?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,624 @@
...
+      throw new InvalidDMLGenerationException("Spanner DDL must not be null.");
+    }
+    if (sourceSchema == null) {
+      throw new InvalidDMLGenerationException("SourceSchema must not be null.");
+    }
+
+    Table spannerTable = spannerDdl.table(spannerTableName);
+    if (spannerTable == null) {
+      throw new InvalidDMLGenerationException(
+          "Spanner table '" + spannerTableName + "' not found in DDL.");
```
shreyakhajanchi: since both source and sinks are spanner, it might not be clear where the issue is if we just mention spanner in the log

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,624 @@
...
+      return buildDeleteMutation(spannerTable, targetTable, schemaMapper, request, targetTableName);
+    } else {
+      throw new InvalidDMLGenerationException(
+          "Unsupported modType '" + modType + "' for table " + spannerTableName);
+    }
+  }
+
+  private static DMLGeneratorResponse buildUpsertMutation(
+      Table spannerTable,
+      SourceTable targetTable,
```
shreyakhajanchi: the naming convention is very confusing , can we make it easier to understand? For instance, `SourceTable targetTable` looks like an oxymoron and when we say `spannerTable` its not clear whether this the one from which data is being read from or written to

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,624 @@
...
+      case JSON:
+        builder.set(targetColName).to(Value.jsonArray(null));
+        break;
+      default:
+        builder.set(targetColName).to(Value.stringArray((Iterable<String>) null));
+    }
+  }
+
+  /** Appends a single primary-key component to the Key builder. */
+  private static void appendKeyComponent(
```
shreyakhajanchi, line 321: seems like we can reuse `appendCustomKeyComponent` here, do you see any concerns?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
```diff
@@ -0,0 +1,624 @@
...
+      } else {
+        appendKeyComponent(keyBuilder, sourceCol, valuesJson, sourceColName);
+      }
+    }
+
+    Mutation mutation = Mutation.delete(targetTableName, keyBuilder.build());
+    return new SpannerMutationResponse(mutation);
+  }
+
+  private static void setColumnValue(
```
shreyakhajanchi, line 226: there is a lot of overlap between this method and `setCustomColumnValue`, can we do this instead?
```
  private static void setColumnValue(
      Mutation.WriteBuilder builder, String targetColName, Column col, JSONObject valuesJson) {
    String sourceColName = col.name();
    Type type = col.type();

    if (type.getCode() == Type.Code.ARRAY) {
      builder.set(targetColName).to(
          buildArrayValue(type.getArrayElementType(), valuesJson.getJSONArray(sourceColName)));
      return;
    }

    // Defer all scalar type parsing to the custom handler, removing the giant switch statement
    setCustomColumnValue(builder, targetColName, col, valuesJson.get(sourceColName));
  }
```

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
```diff
@@ -48,6 +49,77 @@ public class InputRecordProcessor {
...
+            .setSchemaMapper(schemaMapper)
+            .setCustomTransformationResponse(customTransformationResponse)
+            .setCommitTimestamp(spannerRecord.getCommitTimestamp())
+            .setDdl(ddl)
+            .setSourceSchema(sourceSchema)
+            .build();
+
+    DMLGeneratorResponse dmlGeneratorResponse = dmlGenerator.getDMLStatement(dmlGeneratorRequest);
+
+    if (!SOURCE_SPANNER.equals(source) && dmlGeneratorResponse.getDmlStatement().isEmpty()) {
```
shreyakhajanchi: why the if clause for spanner?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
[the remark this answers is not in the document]
shreyakhajanchi: During the tests for EA migrations we had encountered data loss issues because of the incorrect transaction chaining similar to what has been added now for spanner, this would result in data inconsistencies , therefore , we should err on the side of being safe and use metadata db same as the target db in case of replication to spanner.
Since Spanner targets will typically only be 1 shard, we can just enforce that metadataDatabase == targetDatabase for Spanner sinks via a pipeline validation check. If we enforce colocation, we just need to pass the shadowTransaction into the IDao.write() method. SpannerTargetDao can then simply call transaction.buffer(), while JdbcDao safely ignores it. This avoids data loss without requiring a massive architectural overhaul.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGeneratorTest.java
```diff
@@ -0,0 +1,1139 @@
+/*
```
shreyakhajanchi, line 1: Can we please ensure all supported Spanner datatypes (e.g. TIMESTAMP, NUMERIC, BYTES) are explicitly covered in SpannerDMLGeneratorTest? It would help us cover the edge cases

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
```diff
@@ -0,0 +1,170 @@
...
+      InformationSchemaScanner scanner = new InformationSchemaScanner(txn);
+      Ddl ddl = scanner.scan();
+      LOG.info("Scanned Spanner schema for database '{}'", spannerConfig.getDatabaseId().get());
+      return convertDdlToSourceSchema(ddl);
+    } finally {
+      accessor.close();
+    }
+  }
+
+  private SourceSchema convertDdlToSourceSchema(Ddl ddl) {
```
shreyakhajanchi: can we add UTs for this class?

shreyakhajanchi (commented):
I have done one round of review of the PR, the PR is big so it would be better if we can break it down for more thorough and better review. I would also request if we can add all datatypes IT for both the dialects (similar to https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToMySqlDataTypesIT.java and https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToMySqlDataTypesPGDialectIT.java) as a follow up, it would help solidify the confidence on the support.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SpannerShardFileReader.java
srozsnyai: I've removed logicalShardId to simplify the user configuration.

What changed:
* User Config & Constructor: Removed `logicalShardId` from the shard JSON schema and the `SpannerShard` constructor.
* Internal Defaulting: The pipeline now sets this internally to a default value ("single_shard") to satisfy the shared architecture without exposing it to the user.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/SpannerShard.java
srozsnyai: I've decoupled instanceId from the base class namespace field. 

What changed:
* Added a dedicated instanceId field to SpannerShard.
* Updated the constructor and getters to use this new field instead of overloading namespace.
* Updated the corresponding unit tests.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
srozsnyai: I've simplified the spannerTypeToString method to rely on type.getCode().name() as the default fallback, leaving only the explicit logic needed for ARRAY and PG_ARRAY types.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGeneratorTest.java
srozsnyai: I've cleaned up all the inline fully qualified class paths

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
srozsnyai: I've clarified the exception messages in SpannerDMLGenerator to explicitly specify "Source Spanner" (e.g. "Source Spanner table 'X' not found in source DDL") so it is immediately obvious where the schema mismatch is.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
srozsnyai: Refactored the parameter and variable names in SpannerDMLGenerator to make the data flow much clearer:
 * Table spannerTable -> Table sourceSpannerTable (explicitly representing where we are reading CDC events from)
 * SourceTable targetTable -> SourceTable targetSpannerTable (explicitly representing the target database we write to, resolving the oxymoron)
 * spannerTableName -> sourceTableName

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
srozsnyai: I've refactored both appendKeyComponent and setColumnValue to eliminate their duplicated type-switching blocks and delegate directly to their custom counterparts:
* appendKeyComponent: Extracted the JSON value as a generic Object and delegated directly to appendCustomKeyComponent
* setColumnValue: Preserved the array-building logic for JSON Arrays, but delegated all other types directly through setCustomColumnValue

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGenerator.java
srozsnyai: It now extracts scalar types using valuesJson.get(sourceColName) and delegates them to setCustomColumnValue, keeping only the native array-building logic handled explicitly. 

This eliminates the redundant large switch statement.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
srozsnyai: I've refactored the design to use polymorphism instead of hardcoded dialect checks:
1. DMLGeneratorResponse (Base): Added a polymorphic isEmpty() method that checks if the SQL string is empty.
2. SpannerMutationResponse (Subclass): Overrode isEmpty() to check if the underlying Mutation object is null.
3. InputRecordProcessor: Removed the !SOURCE_SPANNER check entirely and simplified the validation to if (dmlGeneratorResponse.isEmpty()).

This was a leaky abstraction where InputRecordProcessor was forced to know Spanner-specific details (i.e. that Spanner generates Mutation objects instead of SQL strings, meaning its DML string is always empty).

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/dbutils/dml/SpannerDMLGeneratorTest.java
srozsnyai: I've updated the SpannerDMLGeneratorTest and added supported datatypes.

I've added explicit assertEquals assertions for the following types to ensure they are correctly parsed from JSON and bound to the mutation

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
srozsnyai: I've added a comprehensive unit test suite in SpannerInformationSchemaScannerTest to cover the schema scanning and conversion logic.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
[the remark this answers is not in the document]
srozsnyai: I have implemented your suggested solution to resolve this without nesting transactions and ensuring absolute atomicity. 

1. Pipeline Validation Check: Added a strict check in SpannerToSourceDb for Spanner sinks that enforces metadataDatabase == targetDatabase. It throws an IllegalArgumentException on startup if they are not colocated.
2. IDao Interface Overload: Added a default write(T statement, TransactionalCheck check, Object transactionContext) overload which JDBC and Cassandra ignore.
3. SpannerTargetDao Buffering: Implemented the overload in SpannerTargetDao to cast the context to Spanner's TransactionContext and call transactionContext.buffer(mutation) instead of writeAtLeastOnce.
4. SourceWriterFn Cleanup: Removed pendingSpannerWrite entirely. It now writes to the target database atomically inside the shadow table transaction context block and the deferred execution block after the commit has been deleted.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SpannerShardFileReader.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        shards.add(new SpannerShard(projectId, instanceId, databaseId));
+        LOG.info(
+            "Loaded SpannerShard: project={}, instance={}, database={}",
+            projectId,
+            instanceId,
+            databaseId);
+      }
+
+      shards.sort(Comparator.comparing(Shard::getLogicalShardId));
```
bharadwaj-aditya: For Spanner this should be limited to 1 shard.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
```diff
@@ -0,0 +1,102 @@
...
+      LOG.info("Spanner connection pool is already initialized.");
+      return;
+    }
+
+    List<Shard> shards = connectionHelperRequest.getShards();
+    String projectId = ((SpannerShard) shards.get(0)).getProjectId();
+    spannerService = SpannerOptions.newBuilder().setProjectId(projectId).build().getService();
+
+    for (Shard shard : shards) {
+      if (!(shard instanceof SpannerShard)) {
```
bharadwaj-aditya, line 49: Is there any use case where there will be multiple Spanner shards ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
```diff
@@ -0,0 +1,102 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Manages {@link DatabaseClient} connections to target Cloud Spanner databases. */
+public class SpannerConnectionHelper implements IConnectionHelper<DatabaseClient> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerConnectionHelper.class);
+
+  private static Map<String, DatabaseClient> clientMap = new ConcurrentHashMap<>();
+  private static Spanner spannerService;
```
bharadwaj-aditya: Using a spanner accessor here is a more common pattern - This allows for multiple spanner connections as well. Ref implementation - https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/07dbfbe30d8b0329f94528477a91872d7b139253/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/SpannerTransactionWriterDoFn.java#L451

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/IDao.java
```diff
@@ -27,4 +27,19 @@ public interface IDao<T> {
...
+   * Executes a given write statement against the data source under a specific transaction context.
+   * If the transaction context is null, it delegates to {@link #write(Object, TransactionalCheck)}.
+   *
+   * @param statement Query statement.
+   * @param transactionalCheck Callback function which will be executed and checked before
+   *     committing the transaction.
+   * @param transactionContext The transaction context to execute the write under.
+   * @throws Exception If the write could not be successfully executed.
+   */
+  default void write(T statement, TransactionalCheck transactionalCheck, Object transactionContext)
```
bharadwaj-aditya, line 41: Is this method being used anywhere ? Can it be removed ?

bharadwaj-aditya (changes requested):

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/utils/SpannerShardFileReader.java
srozsnyai: In addition to the previous single shard constraint, I have enforced a constraint at the configuration reader level to limit to 1 shard. 

I've made the following updates to address this:
1. Removed Sort: Deleted the redundant .sort() call on the shards list from SpannerShardFileReader
2. Enforced Single-Shard Validation: Added a strict check directly after JSON parsing in SpannerShardFileReader that throws an IllegalArgumentException if the configuration contains more or fewer than exactly 1 shard
3. Added tests to cover this

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
srozsnyai: The `for (Shard shard : shards)` loop only exists to satisfy the generic IConnectionHelper interface shared across multi-sharded destinations.

I have now added a strict startup validation check in SpannerShardFileReader that explicitly throws an IllegalArgumentException if more than 1 shard is specified in the configuration file.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
srozsnyai: I've refactored SpannerConnectionHelper to use the SpannerAccessor pattern. 

Even though I'm enforcing now a single-shard constraint for Spanner-to-Spanner replication, this refactor aligns the implementation with the project's standards. 

* Replaced raw Spanner service management with SpannerAccessor and SpannerConfig
* Added an internal map to track accessors and ensure all connections are properly closed during shutdown
* Updated SpannerConnectionHelperTest to verify the new integration logic

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/IDao.java
srozsnyai: This method is critical - It is used in SourceWriterFn.java to pass the active shadowTransaction context down to the SpannerTargetDao.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Manages {@link DatabaseClient} connections to target Cloud Spanner databases. */
+public class SpannerConnectionHelper implements IConnectionHelper<DatabaseClient> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerConnectionHelper.class);
+
+  private static Map<String, DatabaseClient> clientMap = new ConcurrentHashMap<>();
```
bharadwaj-aditya: Spanner accessor wraps a database client, you don't need 2 maps for this.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
```diff
@@ -63,63 +134,22 @@ public static boolean processRecord(
...
-              .setCustomTransformationResponse(customTransformationResponse)
-              .setCommitTimestamp(spannerRecord.getCommitTimestamp())
-              .setDdl(ddl)
-              .setSourceSchema(sourceSchema)
-              .build();
-
-      DMLGeneratorResponse dmlGeneratorResponse = dmlGenerator.getDMLStatement(dmlGeneratorRequest);
-      if (dmlGeneratorResponse.getDmlStatement().isEmpty()) {
-        throw new InvalidDMLGenerationException("DML statement is empty for table: " + tableName);
+      if (dmlGeneratorResponse == null) {
```
bharadwaj-aditya, line 152: This check is unnecessary. If the DML generator response is not populated, the generateDML Response method will throw an exception. Please update the javadocs of that method to clarify this behaviour.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
```diff
@@ -63,63 +134,22 @@ public static boolean processRecord(
...
+        return true; // filtered
       }
-      // TODO we need to handle it as proper Interface Level as of now we have handle Prepared
```
bharadwaj-aditya, line 113: These comments have been removed. Have the TODOs been handled in the PR ? If not, please bring these back.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -266,54 +266,75 @@ public void processElement(ProcessContext c) {
                                   && ((shadowTableRecord
                                               .getProcessedCommitTimestamp()
                                               .compareTo(spannerRec.getCommitTimestamp())
-                                          > 0) // either the source already has record with greater
```
bharadwaj-aditya, line 269: please retain these comments for ease of understanding.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
```diff
@@ -266,54 +266,75 @@ public void processElement(ProcessContext c) {
...
+                                      source);
+                              if (response == null) {
+                                outputWithTag(
+                                    c,
+                                    Constants.FILTERED_TAG,
+                                    Constants.FILTERED_TAG_MESSAGE,
+                                    spannerRec);
+                              } else {
+                                IDao sourceDao = sourceProcessor.getSourceDao(shardId);
+                                sourceDao.write(response, null, shadowTransaction);
```
bharadwaj-aditya, line 304: The shadow transaction will be on the metadataDatabase. The main transaction will be on the target database.

This logic is relevant if both the databases are the same. If the shadow table and the target spanner are on different databases, this will not apply.

There should be a fork in the flow here to handle both flows.

bharadwaj-aditya: Refer to the SpannerTransactionWriterDoFn->processCrossDatabaseTransaction for an implementation where these are different databases.

bharadwaj-aditya (commented):
Few things here
1. The multi db transaction flow is not implemented. That will be a cause for concern
2. Comments and todos have been removed in some places. Please add them
3. We need to have integration tests for this flow. Please add them. (We can take those up in a follow up PR)

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/connection/SpannerConnectionHelper.java
srozsnyai: I have removed the redundant clientMap and refactored the class to rely solely on the accessorMap.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
srozsnyai: Moved comment back - I think I accidentally removed them during a refactoring.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/processor/InputRecordProcessor.java
srozsnyai: I've updated the Javadoc for generateDMLResponse to clarify that it explicitly throws an InvalidDMLGenerationException if the DML is not populated/empty.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/transforms/SourceWriterFn.java
srozsnyai: Accidentally removed during a previous refactor. I have restored the comments.

bharadwaj-aditya (approved):
With a check to ensure that the source and target spanner are the same, this should work. Based on that limitation, marking this as LGTM.

Please add integration tests to this as a fast follow

PR #3775, opened 2026-05-08
Run specified Load Test in Github Workflow Runner

aasthabharill:
#### **Summary**
Passes the `specific_test` workflow dispatch input parameter to the `run-load-tests` runner step. 

#### **Why is this change necessary?**
Previously, the workflow dispatch input `specific_test` was only being passed to the `run-load-test-observer` command (inside the `observe_load_tests` job) but was omitted from the main `run-load-tests` command (inside the `load_tests` job). 

Because these are separate jobs and commands, the main test runner was executing with no `--test` filter when triggered manually. This caused the CI pipeline to build and run **every** Spanner load test under the target modules instead of isolating the execution to the single test specified by the user. 

#### **Impact**
* Allows developers to trigger a single, specific load test via GitHub workflow dispatch instead of running the entire Spanner load test suite.
* Reduces CI resource consumption and test execution time when debugging or verifying individual load tests.

### DIFF
--- a/.github/workflows/spanner-load-tests.yml
+++ b/.github/workflows/spanner-load-tests.yml
@@ -57,6 +57,7 @@ jobs:
         --lt-export-project="cloud-teleport-testing" \
         --lt-export-dataset="performance_tests" \
         --lt-export-table="template_performance_metrics" \
+        --test="${{ github.event.inputs.specific_test }}"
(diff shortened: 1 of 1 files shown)


darshan-sj (approved):
LGTM

PR #3780, opened 2026-05-11
Add anomaly to webhook option and alert throtlling.

claudevdm:


python/src/main/java/com/google/cloud/teleport/templates/python/BigQueryAnomalyDetection.java
```diff
@@ -193,4 +197,41 @@ public interface BigQueryAnomalyDetection {
...
+      description = "Alert Cooldown (seconds)",
+      helpText =
+          "Session-window gap for debouncing alerts to external systems "
+              + "(Pub/Sub, webhook). Per anomaly key, the first anomaly fires "
+              + "immediately; subsequent anomalies are suppressed (logged as "
+              + "\"still active\") until a gap of at least this many seconds "
+              + "passes between consecutive anomalies. Continuous anomalies "
+              + "extend the active-alert window. The BigQuery sink table is "
+              + "unaffected and records every anomaly. Set to 0 to disable "
+              + "rate limiting. Default: 600 (10 minutes).")
```
damccorm, line 241: If we're defaulting the cooldown to on, then I think we should probably mention it in the top-level description. Otherwise I'd imagine this would be surprising default behavior.

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -604,7 +976,20 @@ def _add_argparse_args(cls, parser):
         '--topic',
         default=None,
         help='Pub/Sub topic for anomaly results. '
-        'Full path: projects/<project>/topics/<topic>.')
+        'Full path: projects/<project>/topics/<topic>. '
+        'Optional: at least one of --topic or --webhook_spec must be set.')
+    parser.add_argument(
+        '--webhook_spec',
+        default=None,
+        help='JSON object configuring a REST webhook for anomaly results. '
```
damccorm, line 624: I think an example of how this could be configured would help here

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -1187,32 +1740,64 @@ def _strip_offset_key(element, _keyed=has_group_by):
...
-              message_metadata=message_metadata))
-      | 'WriteToPubSub' >> WriteToPubSub(topic=topic_path))
+  # Apply alert rate limiting before the external-alert sinks (Pub/Sub,
+  # webhook). Session-window semantics: per anomaly key, the first
+  # anomaly fires and subsequent anomalies within the cooldown gap of
+  # the previous event are suppressed. Continuous anomalies extend the
+  # active-alert window. The BigQuery sink branch below intentionally
+  # consumes the un-rate-limited `anomalies` so the analytical record
+  # retains every detection.
+  if options.alert_cooldown_seconds > 0:
```
damccorm: Instead of forking the template here, should we do the forking inside of the DoFn (no-op if this condition is true)? This would make pipeline update simpler (and I imagine this is something people might want to update).

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -524,6 +643,259 @@ def process(self, element):
...
+  return (
+      example.window_start.micros,
+      example.window_end.micros,
+      prediction.model_id or '',
+  )
+
+
+# Sentinel key used to give unkeyed pipelines (no metric_spec.group_by) a
+# single bag-state lane for stateful DoFns. The string is deliberately
+# unlikely to collide with a real user-supplied key. Both _RateLimitAlerts
```
damccorm: "The string is deliberately unlikely to collide with a real user-supplied key."

Why does this matter? Won't it just be applied in cases where no key is supplied?

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -524,6 +643,259 @@ def process(self, element):
...
+    return element[1]
+  return element
+
+
+class _RateLimitAlerts(beam.DoFn):
+  """Stateful DoFn that debounces alerts using session-window semantics.
+
+  Only **outliers** (``predictions[0].label == 1``) participate in the
+  rate-limit logic; normal (label 0) and warmup (label -2) results pass
+  through untouched and never affect state. The downstream alert sinks
```
damccorm: Why are we even sending non-anomalies through? Shouldn't we just be doing this rate limiting right before the anomaly write step?

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -524,6 +643,259 @@ def process(self, element):
...
+  event. An outlier is emitted downstream (i.e. allowed to alert) iff:
+
+    - it's the first outlier for this key, OR
+    - ``window_end - last_event_micros >= cooldown_micros``.
+
+  ``last_event_micros`` is updated on **every outlier** regardless of
+  whether it fired, so a continuous stream of outliers keeps extending
+  the active-alert window. Alerts fire only on genuine gaps in the
+  outlier stream, which matches how a Beam ``SessionWindow`` with the
+  same gap would coalesce events.
```
damccorm: I'm a bit worried about this behavior for sufficiently high volume pipelines. For example, if my input rate is even 1000 elements per minute, and my anomaly rate is even 0.1%, that translates to 1 anomaly per minute. With the default behavior here, we'd probably fire one anomaly at the start and never fire again.

Arguably, the user should try to configure their alerting to be less sensitive in these cases, but it would probably take them a long time to even realize they have a problem.

One partial fix would be to have a hard cutoff (e.g. if your window is set to 10 minutes, and 30 minutes have passed with continuous anomalies, fire again). Alternately, we could fire a different message (e.g. tell an agent "this pipeline has been continuously firing anomalies for 30 minutes, please do something!).

claudevdm: Ack. Will do in followup as discussed

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
```diff
@@ -758,45 +785,11 @@ def _expand_window_size(d):
...
-  For statistical detectors, the ``type`` field must match a registered
-  @specifiable detector class (e.g. ZScore, IQR, RobustZScore).
-
-  ``features`` is automatically set to ``['value']`` to match the output of
-  ``ComputeMetric``. Any user-supplied ``features`` is overwritten.
-
-  Returns:
-    For statistical detectors: an instantiated AnomalyDetector.
-    For Threshold: a ``_ThresholdAlert`` DoFn instance.
+  """Parses ``--detector_spec`` JSON into a detector instance.
```
damccorm, line 792: There's a bunch of big non-functional/comment changes like this or ordering changes which create big diffs.

This is fine for now, but in general if you're going to make these changes:

1) It is probably better to do this in a separate PR
2) This is extra true when you're responding to comments - moving things around like this where they're unrelated to the change makes it quite hard to reorient my review when I want to just be focused on the things that changed since the last review, making it harder to see how comments were addressed

damccorm: (this is extra true because the commit that included the big comments also has some minor functional code changes, as well as some code reordering).

damccorm (approved):
Overall, this LGTM assuming checks go green

python/src/main/python/bigquery-anomaly-detection/src/bqmonitor/pipeline.py
claudevdm: Ack sorry about that.

PR #3795, opened 2026-05-11
Adding data generator engine in cdc data generator

shreyakhajanchi:


manitgupta (commented):
Did a peer review for this. Two asks - 

1. The BatchAndWrite algorithm is complex. Lets write a detailed comment somewhere that explains how does the timeTable generation logic works and how does Beam and onTimer guarantee sequential execution.
2. An AI-based review identified some concerns on nullity of insertTopoOrder and FK constraint violation, validate that.

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/DataGeneratorEngine.java
```diff
@@ -0,0 +1,622 @@
...
+ *       <ul>
+ *         <li>At T0+15s: The scheduled {@code DELETE} timers fire for both Parent and Child tables.
+ *         <li>{@link MutationBatcher} buffers the deletion rows. When flushed at bundle completion,
+ *             {@code flushDeletesInReverseTopoOrder} iterates in reverse topological order (Child
+ *             before Parent).
+ *         <li>All 3 Child {@code DELETE} mutations are flushed to the sink first, followed by the
+ *             Parent {@code DELETE} mutation, guaranteeing zero foreign key constraint violations.
+ *       </ul>
+ * </ol>
+ */
```
manitgupta, line 115: Beautifully written. Thanks for the rigor!

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/DataGeneratorEngine.java
```diff
@@ -0,0 +1,622 @@
...
+public class DataGeneratorEngine {
+  private static final Logger LOG = LoggerFactory.getLogger(DataGeneratorEngine.class);
+
+  private final Integer updateInterval;
+  private final Integer deleteInterval;
+  private final Faker faker;
```
manitgupta, line 121: I assuming this will be a heavy object since it produces the random data, can this be `transient`? Do we need to serialise it?

manitgupta (commented):
Looks good to me. Minor comment on `faker` serialization.

manitgupta (approved):

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/dofn/DataGeneratorEngine.java
shreyakhajanchi: As discussed, the entire DataGeneratorEngine object would be transient so adding inside it is not required

PR #3796, opened 2026-05-12
fix(sourcedb-to-spanner): Add support for partitioning PostgreSQL UUID data type

aasthabharill:
https://b.corp.google.com/issues/512087945

## The Issue
During a bulk migration from PostgreSQL to Cloud Spanner using the `sourcedb-to-spanner` template, the pipeline launch failed with a `SuitableIndexNotFoundException` (thrown in `JdbcIoWrapper` / `PipelineController`) when a table’s primary key or unique index was of `UUID` data type.

### 1. Map UUID Columns to a Virtual `"UUID"` Collation
* **Change:** In `PostgreSQLDialectAdapter.discoverTableIndexes`, if the `typeName` of a column is `"uuid"`, we assign `"UUID"` as its collation reference.
* **Why:** `CollationMapper.fromDB` expects a virtual `"UUID"` collation tag to trigger the static hexadecimal base-16 mapper (`buildStaticUuidMapper`). By assigning `"UUID"` during discovery, the splitter bypasses executing a database query to fetch collation rankings (which would fail or be extremely slow for a native UUID type that has no physical collation).

### 2. Configure Virtual Type Length to `32` for UUID Columns
* **Change:** In `PostgreSQLDialectAdapter.discoverTableIndexes`, if `typeLength` is `null` and `typeName` is `"uuid"`, we set `typeLength = 32`.
* **Why:** While a standard canonical UUID is 36 characters long (including hyphens), `CollationMapper` strips the hyphens out during mapping, leaving exactly 32 hexadecimal characters. Overriding the discovered length to `32` ensures that no additional padding (virtual zero-rank characters) is appended during range partitioning calculations, ensuring a clean 1-to-1 mapping and unmapping.

### 3. Register State-Based Query and Parameter Cast Wrappers for UUID
* **Change:** In `PostgreSQLDialectAdapter.discoverTableIndexes`, if `typeName` is `"uuid"`, we register explicit SQL cast statements in `columnCastWrappers` and `columnParameterCastWrappers` maps.
* **Why:**
  - **`columnCastWrappers` (`CAST(%s AS TEXT)`):** Used in `getBoundaryQuery` to query `MIN(CAST(col AS TEXT))` and `MAX(CAST(col AS TEXT))`. This is necessary to retrieve the UUID boundaries safely as standard text strings compatible with JDBC. UUID doesnt have a MIN or MAX. 
  - **`columnParameterCastWrappers` (`CAST(? AS uuid)`):** Used in `getReadQuery` and `getCountQuery` to bind parameter boundary placeholders as `col >= CAST(? AS uuid)`. This is necessary because PostgreSQL does not support implicit comparison of standard JDBC string parameter bindings against native `uuid` column types.

### 4. Verify Changes with Unit & Integration Tests
* **Collation Mapper Test:** Added `testUuidCollationMapper` in [CollationMapperTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/stringmapper/CollationMapperTest.java) to verify that canonical UUID strings are mapped to 128-bit BigIntegers and unmapped back with correct formatting and hyphen insertion.
* **Dialect Adapter Test:** Added `testDiscoverTableIndexesWithUuid` in [PostgreSQLDialectAdapterTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapterTest.java) verifying index discovery mappings, boundary query wrapping, and read/count query parameter bindings.
* **Integration Test expected data logic:** Updated `getExpectedData` in [PostgreSQLWithUniformizationIT.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLWithUniformizationIT.java) to support assertions for tables with non-integer primary keys (`uuid_pk`).


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -530,7 +532,10 @@ private String addWhereClause(String query, ImmutableList<String> partitionColum
    * Ref <a
    * href="https://www.postgresql.org/docs/16/catalog-pg-type.html#CATALOG-TYPCATEGORY-TABLE"></a>.
    */
-  private SourceColumnIndexInfo.IndexType indexTypeFrom(String typeCategory) {
+  private SourceColumnIndexInfo.IndexType indexTypeFrom(String typeCategory, String typeName) {
+    if ("uuid".equalsIgnoreCase(typeName)) {
+      return SourceColumnIndexInfo.IndexType.STRING;
```
VardhanThigle: This has the potential to explode for User-defined and Core extension types in future.

Would it be possible to add a case for `U` and call a new method that handles all user defined and core extension types? This function can switch case over  `typeName.toUpperCase()` ?

This is just from the point of view of future proofing the code base for readability.

VardhanThigle (dismissed):
Looks good overall, a small open point.
Could be considered nit accroding to the turn around time needed for this issue.

VardhanThigle (dismissed):

PR #3815, opened 2026-05-15
[sourcedb-to-spanner] Add Support for UUID-based Partitioning

aasthabharill:
https://b.corp.google.com/issues/512087945

## 1. Feature Request
**Goal**: Enable uniform range partitioning on `UUID` primary keys or unique index columns within the Uniform Splitter framework for bulk data migration from PostgreSQL to Cloud Spanner.

## 2. TL;DR of Changes
- **Map to BINARY Index**: Discovered PostgreSQL UUID columns are mapped to [IndexType.BINARY](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceColumnIndexInfo.java#L25) (`byte[].class`) while retaining the original database column type name (`"uuid"`).
- **Robust Wire Extraction**: Safely extracts exactly 16 raw binary bytes from PostgreSQL boundary query `ResultSet` objects regardless of driver wire mode (handling native `java.util.UUID`, `String`, or `PGobject`).
- **128-bit Midpoint Splitting**: Implements exact 128-bit unsigned midpoint integer math on byte arrays. Reconstructs fixed 16-byte arrays by truncating leading two's complement sign bytes (`0x00`) or left-padding leading zero bytes for small numbers to guarantee precise database B-tree byte alignment.
- **MIN/MAX for UUID**: PostgreSQL doesn't support MIN() or MAX() for UUID datatypes. To use the UUID index and get the boundary values, we used ORDER BY and LIMIT for the UUID case. 
- **Native JDBC Parameter Binding**: Automatically converts 16-byte raw binary arrays back to native `java.util.UUID` instances prior to setting parameters on JDBC `PreparedStatement` queries, eliminating PostgreSQL type mismatch errors (`BYTEA` vs `UUID`).
- **Value Equality for Byte Arrays**: Overrode array equality checks in boundary and range classes so that distinct `byte[]` instances containing identical bytes evaluate as equal in value, fixing a pre-existing reference equality bug (`Objects.equal`) and ensuring flawless range merging, splitting, and classification.

---

## 3. Partitioning Workflow and Step-by-Step Lifecycle Support for UUID

The partitioning lifecycle is broken down into six distinct stages across the Uniform Splitter framework:

### Stage 1: Schema & Index Discovery
- **Key Classes / Files**: [PostgreSQLDialectAdapter.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java), [SourceColumnIndexInfo.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceColumnIndexInfo.java), [MysqlDialectAdapter.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java)
- **What Changed**: 
  - Added `columnTypeName()` to [SourceColumnIndexInfo](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceColumnIndexInfo.java#L98-L100) to explicitly preserve the original database type name (e.g., `"uuid"`).
  - In [PostgreSQLDialectAdapter#discoverTableIndexes](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java#L351-L375), when discovering table indexes from `information_schema` / catalogs, if `typeName` is `"uuid"`, it is mapped to `IndexType.BINARY` (which maps to `byte[].class`).
  - Discovered UUID column keys (`ColumnKey(tableName, columnName)`) are tracked in a thread-safe concurrent set (`uuidColumnKeys`).

### Stage 2: Table Configuration & Partition Column Inference
- **Key Classes / Files**: [PartitionColumn.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java), [JdbcIoWrapper.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java)
- **What Changed**:
  - Added `columnTypeName()` to [PartitionColumn](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java#L64-L66) and propagated it from `SourceColumnIndexInfo` during [JdbcIoWrapper#getTableConfig](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapper.java#L553).
  - Because `IndexType.BINARY` is already contained within `supportedIndexTypes`, UUID columns are automatically selected as valid partitioning candidates.

### Stage 3: Boundary Query Extraction (`MIN` & `MAX` Retrieval)
- **Key Classes / Files**: [PostgreSQLDialectAdapter.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java), [BoundaryExtractorFactory.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java)
- **What Changed**:
  - In PostgreSQLDialectAdapter#getBoundaryQuery: In standard PostgreSQL catalogs (pg_catalog), built-in aggregate functions like MIN() and MAX() are not defined for UUID. If we were to cast this to TEXT to get the boundary, it would work but the query wouldn't use the indexed B-tree causing performance issues. We thus query the UUID directly to use the index, by querying using ORDER BY and LIMIT 1. 
  - In [BoundaryExtractorFactory#fromBinary](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java#L179-L181), if `columnTypeName()` is `"uuid"`, extraction delegates to `extractUuidBytes(ResultSet rs, int colIndex)`.
- **Edge Case Covered**: Standard `rs.getBytes()` on a PostgreSQL UUID column returns 36 ASCII string bytes (`"xxxxxxxx-xxxx..."`) instead of 16 raw binary bytes. `extractUuidBytes` safely intercepts native `java.util.UUID` objects, strings, or `PGobject` instances, extracts the most and least significant 64-bit integers, and serializes exactly 16 raw binary bytes into memory.

### Stage 4: In-Memory Midpoint Range Splitting
- **Key Classes / Files**: [BoundarySplitterFactory.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java)
- **What Changed**:
  - Updated [splitBytes](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java#L237-L264) to accept `PartitionColumn`. If a UUID range has open bounds (`start` or `end` is `null`), it defaults them to 16 bytes of `0x00` (minimum) and `0xFF` (maximum).
  - Reconstructs exact 16-byte arrays from `BigInteger` midpoint calculations via `toFixedLengthByteArray`.
- **Edge Case Covered**: Java's `BigInteger` maintains numbers in signed two's complement. When a 128-bit unsigned integer has its most significant bit set to `1` (e.g., UUIDs starting with hex `8` through `f`), `BigInteger.toByteArray()` prepends a leading `0x00` sign byte to prevent negative interpretation (returning 17 bytes). Conversely, for small numbers with leading zero bytes, `BigInteger` strips them. In database B-tree indexes, leading zeros are highly significant. `toFixedLengthByteArray` truncates leading sign bytes and left-pads leading zeros, guaranteeing flawless 16-byte alignment for range partition invariants ($start \le mid < end$).

### Stage 5: Parameter Binding in Range Queries
- **Key Classes / Files**: [RangePreparedStatementSetter.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetter.java), [ColumnForBoundaryQueryPreparedStatementSetter.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java)
- **What Changed**:
  - Added `convertIfUuid(Object val, PartitionColumn pc)` helper to parameter setters.
  - Before binding `start` and `end` boundary values to `PreparedStatement` parameters, if `columnTypeName()` is `"uuid"`, the 16 raw binary bytes are wrapped into a `ByteBuffer` and instantiated as a native `java.util.UUID(long mostSigBits, long leastSigBits)`.
- **Edge Case Covered**: If raw byte arrays were directly passed to `preparedStatement.setObject()`, the PostgreSQL JDBC driver would bind them as `BYTEA` (bytea SQL type). When PostgreSQL evaluates `WHERE uuid_col >= $1::bytea`, query execution halts with `ERROR: operator does not exist: uuid >= bytea`. By converting raw bytes to native `java.util.UUID` instances in Java, the JDBC driver natively binds parameters as PostgreSQL `uuid` types without requiring intrusive SQL query casting.

### Stage 6: In-Memory Range Classification & Merging
- **Key Classes / Files**: [Boundary.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java), [Range.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java)
- **What Changed**:
  - Overrode [Boundary#areValuesEqual](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java#L144-L146), `equals`, and `hashCode` to explicitly perform `java.util.Arrays.equals(b1, b2)` and `java.util.Arrays.hashCode(b)` when boundary values are byte arrays.
  - Refactored [Range#isMergable](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Range.java#L260-L261) and `mergeRange` to evaluate boundary equality via `boundary().areValuesEqual()`.
- **Edge Case Covered**: In Java, array classes (`byte[]`) do not override `Object.equals()`. Standard `Objects.equal(a, b)` evaluates reference identity (`a == b`), not array contents. Two distinct byte arrays in memory containing identical 16 binary bytes previously evaluated as `false`, breaking range merging, splitting, and classification across the pipeline. Our explicit array equality checking ensures robust deep value equality.

---

### Unit Tests
- **Boundary Extraction**: [BoundaryExtractorFactoryTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactoryTest.java#L120-L165)
  - Added `testFromBinaryWithUuid()` to verify flawless 16-byte extraction across text UUID strings, `PGobject` instances, native `java.util.UUID` objects, and null boundary conditions (empty tables).
- **Midpoint Range Splitting**: [BoundarySplitterFactoryTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactoryTest.java#L238-L304)
  - Added `testUuidBoundarySplitter()` verifying exact 16-byte midpoint calculation across the 128-bit number space. Rigorously verified sign-byte extension truncation (values $\ge 0x80...$), zero padding for small numbers, open boundary defaults (`0x00` and `0xFF`), absolute zero padding, and range inversion.
- **Parameter Binding**: [RangePreparedStatementSetterTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetterTest.java#L293-L354) & [ColumnForBoundaryQueryPreparedStatementSetterTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetterTest.java#L584-L640)
  - Added `testSetParameters_withUuidColumn()` verifying that 16 raw binary bytes are successfully converted into native `java.util.UUID` instances when bound to prepared statements.
- **Value Equality and Merging**: [BoundaryTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryTest.java#L365-L410) & [RangeTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangeTest.java#L423-L470)
  - Added `testByteRangeMerge()` and `testByteArrayEquality()` proving that distinct `byte[]` instances with identical contents are correctly evaluated as equal in value and successfully merged into unified ranges.
- **Configuration Inference**: [JdbcIoWrapperTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/iowrapper/JdbcIoWrapperTest.java#L814-L864) & [PostgreSQLDialectAdapterTest.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapterTest.java#L250-L310)
  - Proved successful discovery of UUID indexes as `IndexType.BINARY` with preserved column type names and formatted boundary queries using `CAST(col AS TEXT)`.

### Integration Tests
- **End-to-End Pipeline**: [PostgreSQLWithUniformizationIT.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLWithUniformizationIT.java#L115-L121)
  - Added `t_uuid_pk` table with a UUID primary key and verified end-to-end migration pipeline execution from a live PostgreSQL container to Cloud Spanner.

### Load Tests
Ran all pre-existing load tests in sourcedb-to-spanner and all passed

### Customer workload with UUID PKs
- Ran a [bulk job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-05-19_08_36_33-1761528101412788927;logsSeverity=ERROR;graphView=1?project=span-cloud-ck-testing-external&pageState=%28%22dfTime%22:%28%22l%22:%22dfJobMaxTime%22%29%29&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging) on customer's schema and data for tables with UUID PKs that succeeded (validated row counts and manually checked a few rows) Note that SMT converts UUID to STRING and this is what was used while running this job

- Manually edited one table to have UUID types on Spanner instead of STRING. The [bulk job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-05-19_08_49_48-3855129778049627406;bottomTab=JOB_LOGS;expandBottomPanel=true;logsSeverity=ERROR;graphView=1?project=span-cloud-ck-testing-external&pageState=%28%22dfTime%22:%28%22l%22:%22dfJobMaxTime%22%29%29&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging) failed with error `java.lang.IllegalArgumentException: Unknown spanner type UUID at com.google.cloud.teleport.v2.spanner.ddl.Column.parseSpannerType(Column.java:449)` -  this isnt a partitioning code issue, our template doesnt yet recognise UUID in the parser functions as it's a relatively new datatype


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -129,6 +133,32 @@ public static <T extends Serializable> Builder<T> builder() {
    * delta/min-step based comparison for: - Floating-point types (Float, Double) - Source DB type
    * where there is a defined precision Examples: Float(p, d) TODO: Double(p, d).
    */
+  private static String getCallerInfo() {
```
VardhanThigle: Please avoid instrumentation to extract the caller only for purpose of logging.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -141,6 +171,19 @@ protected boolean areValuesEqual(Object valueA, Object valueB) {
       BigDecimal b2 = new BigDecimal(d2.toString());
       return bigDecimalEqual(b1, b2);
     }
+    if (valueA instanceof byte[] b1 && valueB instanceof byte[] b2) {
+      boolean isEqual = java.util.Arrays.equals(b1, b2);
+      logger.info(
```
VardhanThigle: We might not need this log as it might cause a storm of logs during uniformization phase.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +188,61 @@ private static Boundary<byte[]> fromBinary(
         .build();
   }
 
+  private static String getCallerInfo() {
```
VardhanThigle: Please remove complex instrumentation to extract the caller used only for logging.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +188,61 @@ private static Boundary<byte[]> fromBinary(
...
+      return "null";
+    }
+    StringBuilder sb = new StringBuilder();
+    for (byte b : bytes) {
+      sb.append(String.format("%02X", b));
+    }
+    return sb.toString();
+  }
+
+  private static byte[] extractBinaryOrUuidBytes(ResultSet rs, int colIndex) throws SQLException {
```
VardhanThigle: Please consider to keep it clean by having separate method for UUID extraction and use getBytes for bytes.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +188,61 @@ private static Boundary<byte[]> fromBinary(
...
+    java.sql.ResultSetMetaData metaData = null;
+    try {
+      metaData = rs.getMetaData();
+    } catch (Exception e) {
+      // In mocked unit tests, getMetaData() might not be mocked
```
VardhanThigle: Please consider to fix the mocks in UTs instead of adding a hack in the main code.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +188,61 @@ private static Boundary<byte[]> fromBinary(
...
+    }
+    if (metaData != null && "uuid".equalsIgnoreCase(metaData.getColumnTypeName(colIndex))) {
+      if (rs.getObject(colIndex) == null) {
+        return null;
+      }
+      java.util.UUID uuid = rs.getObject(colIndex, java.util.UUID.class);
+      java.nio.ByteBuffer bb = java.nio.ByteBuffer.wrap(new byte[16]);
+      bb.putLong(uuid.getMostSignificantBits()).putLong(uuid.getLeastSignificantBits());
+      byte[] rawBytes = bb.array();
+      logger.info(
```
VardhanThigle: This log could be avoided.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -234,16 +238,132 @@ private static Date convertLongToSqlDate(long dateLong) {
     return Date.valueOf(LocalDate.ofEpochDay(dateLong));
   }
 
-  private static byte[] splitBytes(byte[] start, byte[] end) {
-    BigInteger startBigInt = (start == null) ? null : new BigInteger(start);
-    BigInteger endBigInt = (end == null) ? null : new BigInteger(end);
+  private static String getCallerInfo() {
```
VardhanThigle: Please consider to avoid getCallerInfor for logs.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java
```diff
@@ -89,6 +94,8 @@ public abstract static class Builder {
 
     public abstract Builder setDatetimePrecision(Integer value);
 
+    public abstract Builder setColumnTypeName(@Nullable String value);
```
VardhanThigle: If you need this field, please consider to not make this nullable and add a type whereever it is initialized. Making it nullable bypasses the strict checks of AutoValue class.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java
```diff
@@ -61,13 +61,18 @@ public abstract class PartitionColumn implements Serializable {
   @Nullable
   public abstract Integer datetimePrecision();
 
+  /** Original database column type name (e.g. 'uuid'). Defaults to null. */
+  @Nullable
```
VardhanThigle: If you need this field, please consider to not make this nullable and add a type whereever it is initialized. Making it nullable bypasses the strict checks of AutoValue class.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +188,61 @@ private static Boundary<byte[]> fromBinary(
...
+    for (byte b : bytes) {
+      sb.append(String.format("%02X", b));
+    }
+    return sb.toString();
+  }
+
+  private static byte[] extractBinaryOrUuidBytes(ResultSet rs, int colIndex) throws SQLException {
+    java.sql.ResultSetMetaData metaData = null;
+    try {
+      metaData = rs.getMetaData();
```
VardhanThigle: Would PartitionColumn's getColumnType (Available with the caller)  help avoid the metadata query alltogether?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
aasthabharill: have added these loggers just for the purposes of debugging and clarity right now, won't keep them in the cleaned PR

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -344,20 +355,31 @@ public ImmutableMap<String, ImmutableList<SourceColumnIndexInfo>> discoverTableI
       try (ResultSet resultSet = statement.executeQuery()) {
         while (resultSet.next()) {
           final String tableName = resultSet.getString("table_name");
+          final String typeCategory = resultSet.getString("type_category");
+          final String typeName = resultSet.getString("type_name");
+          final String columnTypeName = typeName == null ? "" : typeName;
```
VardhanThigle: is there a case where typeName can be Null?

If there isn't it might make sense to add a [checkNotNull](https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkNotNull(T,%20java.lang.Object)) instead of converting null to an empty string.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -344,20 +355,31 @@ public ImmutableMap<String, ImmutableList<SourceColumnIndexInfo>> discoverTableI
...
+          final String columnName = resultSet.getString("column_name");
+          if ("uuid".equalsIgnoreCase(typeName)) {
```
VardhanThigle: If we have this for column discovery, do we need it again for index discovery?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
VardhanThigle, line 568: Would we need to cast UUID as text here as well?
if yes, You can use an additional map clause before the existing one, or just enhance the existing one.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
```diff
@@ -150,6 +153,50 @@ private boolean bigDecimalEqual(BigDecimal b1, BigDecimal b2) {
     return diff.compareTo(decimalStepSize()) < 0;
   }
 
+  @Override
+  public boolean equals(Object obj) {
```
VardhanThigle, line 157: AutoValue should auto generate `equals` and `hashcode` and handle arrays correctly, please refer [use a primitive array for a property value](https://github.com/google/auto/blob/main/value/userguide/howto.md#-use-a-primitive-array-for-a-property-value)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
```diff
@@ -184,6 +190,44 @@ private static Boundary<byte[]> fromBinary(
...
+  /**
+   * Extracts exactly 16 raw binary bytes from a PostgreSQL UUID column.
+   *
+   * <p>{@code rs.getBytes()} returns 36 ASCII string bytes instead of 16 raw binary bytes, which
+   * corrupts range calculations. To support all JDBC driver modes (native {@link java.util.UUID},
+   * {@link String}, or {@code PGobject}), this method safely extracts the text representation via
+   * {@code rs.getString()} before serializing to 16 bytes.
+   */
+  private static byte[] extractUuidBytes(ResultSet rs, int colIndex) throws SQLException {
+    Object obj = rs.getObject(colIndex);
```
VardhanThigle: SInce this function is called for UUID itself, would it make sense to just use
`java.util.UUID id = rs.getObject("my_uuid_column", java.util.UUID.class);` and then convert the UUID to a byte array?

This might avoid all the `instanceOf `checks

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -234,16 +234,76 @@ private static Date convertLongToSqlDate(long dateLong) {
...
+
+    BigInteger startBigInt = (start == null) ? null : new BigInteger(1, start);
+    BigInteger endBigInt = (end == null) ? null : new BigInteger(1, end);
     BigInteger split = splitBigIntegers(startBigInt, endBigInt);
     if (split == null) {
       return null;
     }
+
+    if (start != null && end != null && start.length == end.length) {
+      return toFixedLengthByteArray(split, start.length);
```
VardhanThigle, line 250: I might be missing context so please add more details.

if start.leght == end.legnth, do we need `toFixedLengthByteArray`

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
```diff
@@ -234,16 +234,76 @@ private static Date convertLongToSqlDate(long dateLong) {
...
+    }
+    return padLeadingZeroBytes(array, expectedLength);
+  }
+
+  /**
+   * When BigInteger returns a positive number, it adds an extra leading 0x00 byte for values >=
+   * 0x80... to ensure that the most significant bit is 0 for positive numbers. This method removes
+   * that extra leading 0x00 byte to maintain precise byte alignment.
+   */
+  private static byte[] truncateLeadingSignByte(byte[] array, int expectedLength) {
```
VardhanThigle, line 276: Lets think through to check if this is really needed.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -463,8 +485,14 @@ public String getCountQuery(
   @Override
   public String getBoundaryQuery(
       String tableName, ImmutableList<String> partitionColumns, String colName) {
+    String selectCol = colName;
+    // Cast UUID columns to text so MIN/MAX aggregations return canonical strings consistently
+    // across all JDBC driver versions.
+    if (uuidColumnKeys.contains(new ColumnKey(tableName, colName))) {
+      selectCol = String.format("CAST(%s AS TEXT)", colName);
```
VardhanThigle: since we are passing native UUID type from the RangePreparedStatementSetter, do we need the casting logic?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java
```diff
@@ -61,13 +61,17 @@ public abstract class PartitionColumn implements Serializable {
...
+
   public static Builder builder() {
     return new AutoValue_PartitionColumn.Builder()
         .setStringCollation(null)
         .setStringMaxLength(null)
         .setNumericScale(null)
         .setDecimalStepSize(null)
-        .setDatetimePrecision(null);
+        .setDatetimePrecision(null)
+        .setColumnTypeName("");
```
VardhanThigle, line 74: Cosider removing this and let it assert whereever it's not set.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/schema/SourceColumnIndexInfo.java
```diff
@@ -95,13 +95,16 @@ public abstract class SourceColumnIndexInfo implements Comparable<SourceColumnIn
...
+  public abstract String columnTypeName();
+
   /**
    * Builder for {@link SourceColumnIndexInfo}.
    *
    * @return builder.
    */
   public static Builder builder() {
-    return new AutoValue_SourceColumnIndexInfo.Builder().setCardinality(0L);
+    return new AutoValue_SourceColumnIndexInfo.Builder().setCardinality(0L).setColumnTypeName("");
```
VardhanThigle, line 107: Consider removing the default for TypeName

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
aasthabharill: yes as the BigInteger that was converted a byte array might not match this length

this is needed for both UUID and byte arrays as shown by the edge cases covered in the unit test file (also mentioned in PR description)

testBytesUnsignedBoundarySplitter - fails for byte arrays without this

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
aasthabharill: ran an experiment - deleted these overrides and ran the BoundaryTest file 
testBoundaryAndRangeByteEqualityAndMergability fails without the overrides, but passes with. This proves the requirement for the overrides.

Explanation:
- Because Boundary is a generic class parameterized by T extends Serializable, AutoValue does not know at compile time what type T will be at runtime (Integer, String, Timestamp, or byte[]).
- Due to Java's type erasure, AutoValue compiles start() and end() as generic Object fields and generates standard Object.equals() calls in its generated subclass AutoValue_Boundary.
- When T is instantiated at runtime as a byte[] array, calling standard Object.equals(this.start, that.start) falls back to Java array reference equality (this == that), causing two identical 16-byte arrays in separate memory addresses to evaluate as unequal (false).

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/PartitionColumn.java
aasthabharill: will do as follow up otherwise this will bloat the PR

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
```diff
@@ -551,4 +582,38 @@ private boolean isBlankPaddedType(String typeName, @Nullable Integer typeLength)
...
+  private static final class ColumnKey implements Serializable {
+    private final String tableName;
+    private final String columnName;
+
+    public ColumnKey(String tableName, String columnName) {
+      this.tableName = clean(tableName);
+      this.columnName = clean(columnName);
+    }
+
+    private static String clean(String identifier) {
```
bharadwaj-aditya, line 619: nit: this seems like a util method. Makes sense to move it into utils

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/columnboundary/ColumnForBoundaryQueryPreparedStatementSetter.java
```diff
@@ -52,6 +52,14 @@ public ColumnForBoundaryQueryPreparedStatementSetter(
                     TableSplitSpecification::tableIdentifier, Function.identity()));
   }
 
+  private static Object convertIfUuid(Object val, PartitionColumn pc) {
+    if (val instanceof byte[] bytes && "uuid".equalsIgnoreCase(pc.columnTypeName())) {
```
bharadwaj-aditya: nit: using constants is preferable. In all other places as well. Marking as a nit because this has been done for other types as well and the cleanup can be taken as a follow up

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
bharadwaj-aditya: this is quite a big change. I would be a little worried shipping this without proper testing.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/RangePreparedStatementSetter.java
```diff
@@ -44,6 +45,23 @@ public RangePreparedStatementSetter(
...
+                ImmutableMap.toImmutableMap(
+                    specification -> specification.tableIdentifier(),
+                    specification -> specification.partitionColumns()));
+  }
+
+  /**
+   * Convert raw byte[] to java.util.UUID to prevent PostgreSQL JDBC type mismatch (BYTEA vs UUID).
+   */
+  private static Object convertIfUuid(Object val, PartitionColumn pc) {
+    if (val instanceof byte[] bytes && "uuid".equalsIgnoreCase(pc.columnTypeName())) {
```
bharadwaj-aditya: nit: we should not have DB specific logic here. Some of it exists in other methods as well, but probably needs to be cleaned up.

bharadwaj-aditya (commented):
Seems like a lot of changes for this feature. We should test it carefully.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
aasthabharill: getObject casting does not work for UUID class. 
I made the proposed change to test it, here are the failure results:
```
Error message from worker: java.lang.RuntimeException: java.lang.ClassCastException: Cannot cast java.lang.String to java.util.UUID
	com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms.RangeBoundaryDoFn.processElement(RangeBoundaryDoFn.java:151)
Caused by: java.lang.ClassCastException: Cannot cast java.lang.String to java.util.UUID
	java.base/java.lang.Class.cast(Class.java:3889)
	org.postgresql.jdbc.PgResultSet.getObject(PgResultSet.java:3841)
	org.apache.commons.dbcp2.Jdbc41Bridge.getObject(Jdbc41Bridge.java:169)
	org.apache.commons.dbcp2.DelegatingResultSet.getObject(DelegatingResultSet.java:743)
	org.apache.commons.dbcp2.Jdbc41Bridge.getObject(Jdbc41Bridge.java:169)
	org.apache.commons.dbcp2.DelegatingResultSet.getObject(DelegatingResultSet.java:743)
	org.apache.commons.dbcp2.Jdbc41Bridge.getObject(Jdbc41Bridge.java:169)
	org.apache.commons.dbcp2.DelegatingResultSet.getObject(DelegatingResultSet.java:743)
	org.apache.commons.dbcp2.Jdbc41Bridge.getObject(Jdbc41Bridge.java:169)
	org.apache.commons.dbcp2.DelegatingResultSet.getObject(DelegatingResultSet.java:743)
	com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range.BoundaryExtractorFactory.extractUuidBytes(BoundaryExtractorFactory.java:200)
	com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.range.BoundaryExtractorFactory.fromBinary(BoundaryExtractorFactory.java:181)
	com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms.RangeBoundaryDoFn.processElement(RangeBoundaryDoFn.java:122)
	com.google.cloud.teleport.v2.source.reader.io.jdbc.uniformsplitter.transforms.RangeBoundaryDoFn$DoFnInvoker$e2acb48f.invokeProcessElement(Unknown Source)
	org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.invokeProcessElement(SimpleDoFnRunner.java:216)
	org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.processElement(SimpleDoFnRunner.java:188)
	org.apache.beam.runners.dataflow.worker.SimpleParDoFn.processElement(SimpleParDoFn.java:344)
```
- [Build failure](https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/26019635919/job/76478031222)
- [Dataflow Job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-05-18_00_37_18-2591200974844264205;logsSeverity=ERROR;graphView=0?project=span-cloud-teleport-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging)


#### Explanation:
If this code is executed against a database (like PostgreSQL) where the JDBC connection is configured to return UUIDs as String or PGobject, this specific line will likely throw an SQLException because the driver won't know how to implicitly cast a PGobject to java.util.UUID.class.

Direct Interaction with the Raw ResultSet: If you look at the extraction methods inside BoundaryExtractorFactory.java (like fromIntegers, fromStrings, and the problematic extractUuidBytes), they all require a java.sql.ResultSet as an argument. The ResultSet is the raw, immediate output from the JDBC driver. It has not been sanitized or mapped into a neutral, framework-agnostic Java object yet.

The Point of Failure: Because extractUuidBytes calls rs.getObject(colIndex, java.util.UUID.class) directly on that raw ResultSet, it bears the full brunt of whatever the connection properties dictated. If the connection told PostgreSQL to return a PGobject, that is exactly what the ResultSet is holding at that exact millisecond.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundarySplitterFactory.java
aasthabharill: Edge case: Calculated midpoint has its highest bit set to 1 (e.g., start = 80000000-0000-0000-0000-000000000000, end = ffffffff-ffff-ffff-ffff-ffffffffffff).
BigInteger.toByteArray() will return a 17-byte array with 0x00 at the 0th index. Your logic must drop this sign byte and extract the remaining 16 bytes perfectly to construct the UUID.

Also ran a unit test in the earlier commits called `testBigIntegerToByteArraySignBit` that displays this exact usecase

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/Boundary.java
VardhanThigle: Got the need for overriding equals and hashcode. We can keep the implementation - the implmentation of equals is quite same as the AutoValue implementation.

I will take up wrapping the array in a way that works well with AutoValue as a fast followup later on.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
VardhanThigle: As discussed more over today's call, the casting bit would miss the index and we might need to choose from one of:

1. getting limit 1 in ASC and DESC order for UUID
2. set min as 0 and max as all Fs for UUID specific column (less lucrative from performance POV and will force the need for uniformization always)

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
VardhanThigle: Is this happeening because we are casting UUIDs as text in the SQL queries?

I strongly suspect it might not happen if we avoid the casting.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
VardhanThigle: As discussed casting UUID to string will miss the index and we need alternate strategy here.

aasthabharill: no because addWhereClause only does comparisons, that are supported for UUID. Only MIN/MAX was the issue

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/postgresql/PostgreSQLDialectAdapter.java
aasthabharill: this is actually only used for the specific usecase of ColumnKey - which is only being used in this file. i dont think this is a generic utility being used anywhere else

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/uniformsplitter/range/BoundaryExtractorFactory.java
aasthabharill: yup you were right, tests passing now. retaining the change you suggested

VardhanThigle (approved):

PR #3836, opened 2026-05-20
Pipeline code and pipeline options for cdc data generator

shreyakhajanchi:


v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/CdcDataGenerator.java
```diff
@@ -0,0 +1,220 @@
...
+      if (mySqlConfig.getShards() != null && !mySqlConfig.getShards().isEmpty()) {
+        shardCount = mySqlConfig.getShards().size();
+      }
+      // MySQL is not horizontally scalable like Spanner and performs better with fewer concurrent
+      // write
+      //
+      // operations. We keep key-parallelism low to prevent connection pool exhaustion or lock
+      // contention on the target instance.
+      // 500 DoFn per workers * 10 workers per MySQL shard
+      return 5000 * shardCount;
```
manitgupta, line 223: Still seems too high. I assume this is backward from some sort of testing?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/CdcDataGenerator.java
```diff
@@ -0,0 +1,220 @@
...
+              .setIncludePaneInfo(true)
+              .build());
+    }
+
+    return pipeline.run();
+  }
+
+  static int resolveKeyParallelism(
+      CdcDataGeneratorOptions.SinkType sinkType, SinkConfig sinkConfig) {
+    if (sinkType == CdcDataGeneratorOptions.SinkType.SPANNER) {
```
manitgupta, line 202: Given that there are some best-effort assumptions here, it may make sense to expose this as a parameter as well just for cases where tuning this is a blocker. Otherwise you will have to push a code change.

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/CdcDataGenerator.java
shreyakhajanchi: yes, but this changes with larger machine, I would recommend that since we have an override parameter now added we should be good. I am performing more scale tests to refine the recommendation which will be in readme, can revisit this once the tests are complete. Does that sound okay?

v2/cdc-data-generator/src/main/java/com/google/cloud/teleport/v2/templates/CdcDataGenerator.java
shreyakhajanchi: makes sense , added

manitgupta (approved):

PR #3833, opened 2026-05-19
[DO NOT MERGE] Validate beam 2.74.0 RC1

Amar3tto:

PR #3839, opened 2026-05-20
Resolve race when staging multiple classic templates at the same time

Abacn:
context: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3833#issuecomment-4502649406

There is a racing condition when multiple classic templates get staged, jars are overwrittened by each other. gcsio correctly handles this race, but unfortunately, it's printing an log triggering a JDK bug.


Abacn:
Still stuck on https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/26197041852

PR #3845, opened 2026-05-22
[sourcedb-to-spanner] Fix GCS Avro Export flow

aasthabharill:
https://b.corp.google.com/issues/515646503

## 1. Issues
When migrating data using the `Sourcedb_to_Spanner_Flex` batch Dataflow template with the optional GCS export feature enabled (`gcsOutputDirectory` parameter specified), the Dataflow job fails with two distinct serialization errors depending on the shard configuration:

### Issue 1 (Dynamic Avro Serialization)
For all migrations, successfully migrated records fail to write to GCS with an `UnresolvedUnionException`:
```
org.apache.avro.file.DataFileWriter$AppendWriteException: org.apache.avro.UnresolvedUnionException: Not in union ["null","int"]: table1_transform_fail (field=id)
```

### Issue 2 (Reflect Coder Nullability)
For non-sharded migrations (where `shardId` is null), the job fails with a `NullPointerException` inside the Beam pipeline runner stages:
```
java.lang.NullPointerException: null in string in field shardId
	org.apache.avro.generic.GenericDatumWriter.npe(GenericDatumWriter.java:208)
	org.apache.avro.reflect.ReflectDatumWriter.write(ReflectDatumWriter.java:173)
	org.apache.beam.sdk.extensions.avro.coders.AvroCoder.encode(AvroCoder.java:466)
```

---

### Cause 1 (Index Mismatch in `writeToGCS()`)
In `MigrateTableTransform.java`, the dynamic Avro writer's schema given to the sink was incorrectly configured using:
`record.getPayload().getSchema().toString()` (e.g., database row schema `"table1"` with fields `id` and `firstName`).
However, the elements passed to the writer in the pipeline were the metadata wrapper records:
`record.toGcsRecord()` (schema `SourceRowWithMetadata` with fields `tableName`, `shardId`, `primaryKeys`, and `payload`).

During serialization, Avro's `GenericDatumWriter` evaluated fields by index. For the first field `id` (at index `0` of the writer's schema), it called `datum.get(0)` on the provided `SourceRowWithMetadata` record. Since index `0` of `SourceRowWithMetadata` is the `tableName` field, it returned the String `"table1"`. Attempting to write this String into the integer `id` field threw the `UnresolvedUnionException`.

> [!NOTE]
> The original unit test went undetected because the test table defined only `VARCHAR` (String) columns. By pure coincidence, both the wrapper metadata fields (`tableName`, `shardId`) and the database columns were Strings, causing Avro to silently serialize metadata into database columns.

### Cause 2 (Null `shardId` Serialization in `AvroDestination`)
At the end of the `WriteFiles` stage, Beam serializes a collection of `FileResult<AvroDestination>` metadata records using `AvroCoder.of(AvroDestination.class)`.
In `AvroDestination.java`, the `shardId` field was defined as a raw Java `String` without any nullability annotations. Under Avro reflection, a raw `String` is treated as non-nullable. For standard bulk migrations without sharding, `shardId` is `null`, leading to a `NullPointerException` when the reflection coder attempted to encode it.

---

### Fix 1: Corrected dynamic writer schema
Modified the dynamic Avro destination builder in MigrateTableTransform.java to utilize the wrapper schema:
`record.gcsSchema().toString()))`

This aligns the writer's schema with the actual wrapper schema.

### Fix 2: Annotated `shardId` in `AvroDestination`
Modified [AvroDestination.java](file:///usr/local/google/home/aasthabharill/DataflowTemplates/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java) to annotate the `shardId` field with Avro's `@org.apache.avro.reflect.Nullable` annotation:
`@org.apache.avro.reflect.Nullable public String shardId;`

This instructs Avro's reflection engine to allow `null` values for the field by mapping it to a `["null", "string"]` union type.

---

#### Unit Test
I implemented a comprehensive, unit test `testWriteToGCS()` in MigrateTableTransformTest.java

* **Reproducing production scenarios**: It sets `shardId` to `null` to replicate non-sharded migrations.
* **Type Validation**: It configures a schema with an **`INTEGER id`** field at index `0`, ensuring that any index mismatch instantly fails with `UnresolvedUnionException`.
* **Assertions**: Post-pipeline execution, the test reads the generated `.avro` file back from disk using Avro's `DataFileReader` and asserts that:
  * The schema is correctly structured.
  * The `tableName` matches `"table1"`.
  * The `shardId` is successfully serialized as `null`.
  * The nested payload correctly holds the integer `id` (`123`) and the `firstName` string (`"abc"`).

This test executes locally in sub-second time and compiles and passes successfully.

#### Manual test
Built modified template: gs://ea-functional-tests/templates-aastha-2026-05-22/flex/Sourcedb_to_Spanner_Flex
Job was a success: https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-05-22_03_10_50-3991903184278937362;graphView=1?project=span-cloud-ck-testing-external&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))

### DIFF
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java
@@ -16,6 +16,7 @@
 package com.google.cloud.teleport.v2.templates;
 
 import java.util.Objects;
+import org.apache.avro.reflect.Nullable;
 import org.apache.beam.sdk.coders.DefaultCoder;
 import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
 
@@ -23,7 +24,7 @@ import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
 public class AvroDestination {
   public String name;
   public String jsonSchema;
-  @org.apache.avro.reflect.Nullable public String shardId;
+  @Nullable public String shardId;
 
   // Needed for serialization
   public AvroDestination() {}
--- a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java
+++ b/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java
@@ -23,7 +23,7 @@ import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
 public class AvroDestination {
   public String name;
   public String jsonSchema;
-  public String shardId;
+  @org.apache.avro.reflect.Nullable public String shardId;
 
   // Needed for serialization
   public AvroDestination() {}
--- a/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MigrateTableTransformTest.java
+++ b/v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MigrateTableTransformTest.java
@@ -22,13 +22,21 @@ import static org.mockito.Mockito.when;
 import com.google.cloud.teleport.v2.options.SourceDbToSpannerOptions;
 import com.google.cloud.teleport.v2.source.reader.ReaderImpl;
 import com.google.cloud.teleport.v2.source.reader.io.row.SourceRow;
+import com.google.cloud.teleport.v2.source.reader.io.schema.SchemaTestUtils;
+import com.google.cloud.teleport.v2.source.reader.io.schema.SourceTableSchema;
+import com.google.cloud.teleport.v2.source.reader.io.schema.typemapping.UnifiedTypeMapper;
 import com.google.cloud.teleport.v2.source.reader.io.transform.ReaderTransform;
 import com.google.cloud.teleport.v2.spanner.ddl.Ddl;
 import com.google.cloud.teleport.v2.spanner.migrations.schema.ISchemaMapper;
+import com.google.cloud.teleport.v2.spanner.migrations.schema.SourceColumnType;
+import org.apache.avro.file.DataFileReader;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.io.Compression;
 import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.testing.TestPipeline;
 import org.apache.beam.sdk.transforms.Create;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.windowing.GlobalWindow;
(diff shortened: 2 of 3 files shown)


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/AvroDestination.java
```diff
@@ -23,7 +23,7 @@
 public class AvroDestination {
   public String name;
   public String jsonSchema;
-  public String shardId;
+  @org.apache.avro.reflect.Nullable public String shardId;
```
manitgupta: Add imports at the top

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MigrateTableTransformTest.java
```diff
@@ -132,4 +143,74 @@ public PCollectionTuple expand(PBegin input) {
...
+   * the wrapper schema without throwing UnresolvedUnionException.
+   */
+  @Test
+  public void testWriteToGCS() throws Exception {
+    TestPipeline pipeline = TestPipeline.create().enableAbandonedNodeEnforcement(false);
+
+    final String testTable = "table1";
+    final String shardId = null;
+    final long testReadTime = 1712751118L;
+    var schemaRef = SchemaTestUtils.generateSchemaReference("public", "mydb");
```
manitgupta: Can you UT for both the sharded and the non-sharded cases? If it already exists, please rename appropriately (`testWriteToGCSForNonShardedFlow`) or something similar.

VardhanThigle (dismissed):

manitgupta:
~I just noticed that the IT suite did not run?~

~I see that the workflow seems to be set up correctly - ~

~https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/.github/workflows/spanner-pr.yml#L42~

~Could you please check why it did not trigger?~


EDIT: Nvm it triggered now

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/MigrateTableTransformTest.java
```diff
@@ -132,4 +143,101 @@ public PCollectionTuple expand(PBegin input) {
...
     migrateTableTransform.expand(PBegin.in(p));
   }
+
+  /**
+   * Tests the writeToGCS method end-to-end with a mixture of sharded and non-sharded records,
+   * ensuring that records are dynamically routed to the correct GCS paths based on their tableName
+   * and shardId.
+   */
+  @Test
+  public void testWriteToGCSForShardedAndNonShardedFlows() throws Exception {
```
manitgupta: I think it would be more readable and easier to measure/debug if it were two separate tests? What is the motivation for combining these?

manitgupta:
```
Tests the writeToGCS method end-to-end with a mixture of sharded and non-sharded records,
   * ensuring that records are dynamically routed to the correct GCS paths based on their tableName
   * and shardId.  
  ```

This is not a practical scenario. We will not have a "mixture of sharded and non-sharded records".

manitgupta (approved):

PR #3849, opened 2026-05-22
Verify whether PR was created before labeling it.

tvalentyn:
When there are no doc changes, the suite fails with a confusing error. let's fail with a better error.


tvalentyn:
r: @damccorm

damccorm (approved):
Not sure if we want to fail or succeed, but this is strictly an improvement - thanks

PR #3860, opened 2026-05-26
Add support for Remote UDF

adrw-google:
Support for Spanner Remote UDF 

GSQL dialect has been supported for 3+ mo in prod.  PG dialect is rolling out, but the changes here should be safe to push (existing databases won't have REMOTE UDFs). Added unit tests for PG dialect. Integration tests will be pushed once we deploy support in prod.

Tested using Export/Import integration tests.

Some minor fixes:
* Integration tests require ENTERPRISE_PLUS instance due to full text search (otherwise they fail).
* Simplified prefix in integration tests so that export IT can easily produce input for import IT (with correct checksums and lengths)


it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
```diff
@@ -193,6 +194,7 @@ private synchronized void maybeCreateInstance() {
           InstanceInfo.newBuilder(InstanceId.of(projectId, instanceId))
               .setInstanceConfigId(InstanceConfigId.of(projectId, "regional-" + region))
               .setDisplayName(instanceId)
+              .setEdition(Edition.ENTERPRISE_PLUS) // Needed by Full Text Search.
```
shreyakhajanchi, line 197: same `maybeCreateInstance` method is being used across multiple ITs in both v1 and v2 templates, rather than hardcoding the edition to ENTERPRISE_PLUS in all cases, can we use a setter method?

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Udf.java.orig
```diff
@@ -0,0 +1,318 @@
+/*
```
shreyakhajanchi: I don't think this file was intended to be checked in ?

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
```diff
@@ -687,6 +700,43 @@ public void simpleUdf() throws Exception {
...
+            UdfParameter.builder()
+                .functionSpecificName("s1.remote_udf")
+                .name("y")
+                .type("INT64")
+                .defaultExpression(null)
+                .autoBuild()));
+  }
+
+  @Test
+  public void pgSimpleUdf() throws Exception {
```
shreyakhajanchi, line 732: the test is misleading as it is named `pgSimpleUdf` but has 0 udfs

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/RandomDdlGenerator.java.orig
```diff
@@ -0,0 +1,751 @@
+/*
```
shreyakhajanchi: I think this file needs to be removed. Same comment for other `.orig` files

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
```diff
@@ -635,14 +636,18 @@ public void simpleUdf() throws Exception {
             + "C STRING DEFAULT 'NULL', "
             + "D STRING DEFAULT '') "
             + "RETURNS STRING AS (CONCAT(A, '::', B, '::', C, '::', D))";
+    String udfDef3 =
+        "CREATE FUNCTION s1.remote_udf(x INT64, y INT64) "
+            + "RETURNS INT64 NOT DETERMINISTIC LANGUAGE REMOTE "
+            + "OPTIONS ( endpoint = 'https://us-central1-myproject.cloudfunctions.net/myfunc' )";
```
shreyakhajanchi: nit: can you also add a non string option for improved ut coverage

v1/src/test/resources/ImportPipelineIT/googlesql/UdfSchema-manifest.json
```diff
@@ -0,0 +1,6 @@
+{
```
shreyakhajanchi, line 1: how are we asserting that these udfs were actually created ? Can we please add asserts for them?

cc: @darshan-sj

it/google-cloud-platform/src/main/java/org/apache/beam/it/gcp/spanner/SpannerResourceManager.java
adrw-google: Both need access to full text search capabilities such as text indexes. It's probably easier if all the tests share the same fixture.

v1/src/main/java/com/google/cloud/teleport/spanner/ddl/Udf.java.orig
adrw-google: Done

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
adrw-google: After verifying in devel, I temporarily removed the content of the test. Waiting for PG UDF support to fully roll out to PROD. Added a TODO so that I don't forget to add these tests back.

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/RandomDdlGenerator.java.orig
adrw-google: Done

v1/src/test/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScannerIT.java
adrw-google: Done

v1/src/test/resources/ImportPipelineIT/googlesql/UdfSchema-manifest.json
adrw-google: Done. Added an IS query to verify that. Also I opened b/517150731 to backfill for other types.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
```diff
@@ -109,8 +110,24 @@ private void uploadImportPipelineArtifacts(String subdirectory) throws IOExcepti
         "input/Sequence2-manifest.json",
         Resources.getResource("ImportPipelineIT/" + subdirectory + "/Sequence2-manifest.json")
             .getPath());
+    gcsClient.uploadArtifact(
+        "input/UdfSchema.avro-00000-of-00001",
+        Resources.getResource("ImportPipelineIT/" + subdirectory + "/UdfSchema.avro").getPath());
+    gcsClient.uploadArtifact(
+        "input/UdfSchema-manifest.json",
+        Resources.getResource("ImportPipelineIT/" + subdirectory + "/UdfSchema-manifest.json")
+            .getPath());
```
shreyakhajanchi, line 119: why do we have these files? I don't think we are asserting any non-remote udfs right now in the pipeline.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
```diff
@@ -235,6 +252,15 @@ private void testGoogleSqlImportPipelineBase(
...
     assertThatStructs(float32Records).hasRecordsUnordered(getFloat32TableExpectedRows());
+
+    assertThat(
+            spannerResourceManager
+                .runQuery(
+                    "SELECT CONCAT(ROUTINE_SCHEMA, '.', ROUTINE_NAME) FROM INFORMATION_SCHEMA.ROUTINES")
+                .stream()
+                .map(row -> row.getString(0))
+                .collect(Collectors.toList()))
+        .containsExactly("UdfSchema.Remote");
```
shreyakhajanchi, line 263: Will it be better to assert other fields as well from the schema specially UDF type, options , etc?
```
{
  "type" : "record",
  "name" : "UdfSchema_Remote",
  "namespace" : "spannerexport",
  "fields" : [ ],
  "spannerName" : "UdfSchema.Remote",
  "spannerUdfName" : "UdfSchema.Remote",
  "spannerEntity" : "spannerUdf",
  "spannerOption_0" : "endpoint=\"https://us-central1-myproject.cloudfunctions.net/myfunc\"",
  "spannerOption_1" : "max_batching_rows=10",
  "spannerUdfLanguage" : "REMOTE",
  "googleStorage" : "CloudSpanner",
  "spannerUdfParameter_0" : "`x` INT64",
  "spannerUdfParameter_1" : "`y` INT64",
  "spannerUdfSecurity" : "INVOKER",
  "spannerUdfType" : "INT64",
  "googleFormatVersion" : "1.0.0",
  "spannerUdfDefinition" : ""
}
```

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
```diff
@@ -207,18 +207,19 @@ private void testSpannerToGCSAvroBase(
     // an empty database without any tables.
     spannerResourceManager.executeDdlStatement(setDefaultTimeZoneStatement);
```
shreyakhajanchi, line 208: We should assert the remote udf handling in export pipeline as well

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
adrw-google: I think Jacob has a concurrent PR to add SQL UDFs, which will add these files too. Note that all UDFs (remote or sql) need to be inside a named schema, and these avro files are necessary to create one.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
adrw-google: I think the test should assert that the functions were actually created in the database in the simplest possible way. Using InformationSchemaScanner could introduce false positives here, if there are bugs in InformationSchemaScanner (i.e. it sometimes adds default values). 

We could just take a full dump of INFO_SCHEMA.ROUTINES here, but that might be flaky if we decide to change INFO_SCHEMA implementation. This shouldn't happen frequently though.

We could just re-export the database and compare the avro files to be identical. That could give us a much better coverage.

I opened b/517150731 for the migrations team to ponder and design how these tests should work. I'm adding a minimal viable coverage in this PR.

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
adrw-google: Opened b/517150731 for this. It's up to the migrations teams to design how these tests should work. I'm not familiar with avro enough to pull this off and comparing the binary payload of output to golden doesn't yield diffs that are understandable.

shreyakhajanchi: Will check on this with @bharadwaj-aditya

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
shreyakhajanchi: I didn't mean the exact format , I only meant asserting few other fields like udf type which can be queried using the same format you have added : `"SELECT CONCAT(ROUTINE_SCHEMA, '.', ROUTINE_NAME) FROM INFORMATION_SCHEMA.ROUTINES")` but if you think that is a concern I can check on whether it can be skipped.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
bharadwaj-aditya: If these files are not being used here, please remove them from this PR.

They can be added in the other PR or in a follow up once that is merged.

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
bharadwaj-aditya: If this does not work as expected, the feature is not going to be used. Given the current state of tests, i would suggest running the export/import test manually to verify it works. And add a TODO here to be covered once the schema verification flows are added in the import/export flows.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
bharadwaj-aditya: Let me know if dumping the info_schema routines and checking is a viable option in this PR. That will make sure this functionality works as expected. 
If something as basic as the info_schema implementation changes, it would be fine to have this test fail. We have pre-release tests to catch exactly this sort of issue.

Once we have a more full proof way of testing schema in the templates, we can re-evaluate this check.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
adrw-google: They are used. Again, to create a UDF, we need a named schema. To create that named schema, we need the avro and manifest file that importer will covnert to DDL.

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
adrw-google: No it is not a viable option in this PR. There should be a separate PR to change how integration tests work for all the entity types, not only UDFs. This PR has been waiting for review 2 weeks. Let's submit it and I can follow up.

v1/src/test/java/com/google/cloud/teleport/spanner/ExportPipelineIT.java
adrw-google: I did run these tests manually already.  Added TODO. PTAL

v1/src/test/java/com/google/cloud/teleport/spanner/ImportPipelineIT.java
bharadwaj-aditya: sure. Please add a todo for this so we don't lose track

bharadwaj-aditya (dismissed):
LGTM

adrw-google:
> LGTM

Added the comment to Import test. Needed to rerun the tests. Now turns out that Cassandra tests are flaky :+1:

bharadwaj-aditya (approved):
LGTM

PR #3868, opened 2026-06-02
feat: Add spanner custom release via Github actions

sm745052:
fixes b/514905070

### DIFF
--- a/.github/workflows/spanner-targeted-release.yml
+++ b/.github/workflows/spanner-targeted-release.yml
@@ -33,7 +33,7 @@ on:
       stagePrefix:
         description: 'Prefix for staging the template (e.g., templates-2025-10-31). If blank, defaults to templates-YYYY-MM-DD using current date.'
         type: string
-        required: false
+        required: true
         default: ''
       templateName:
         description: 'Template Name to build and stage'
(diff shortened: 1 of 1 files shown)


.github/workflows/spanner-targeted-release.yml
```diff
@@ -0,0 +1,186 @@
...
+          path: '**/surefire-reports/TEST-*.xml'
+          reporter: java-junit
+          only-summary: 'true'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fail-on-error: 'false'
+          list-suites: 'failed'
+          list-tests: 'failed'
+
+      - name: Build & Stage Template
+        run: |
```
darshan-sj, line 171: This will run irrespective of the results of integration tests?

darshan-sj (dismissed):

.github/workflows/spanner-targeted-release.yml
sm745052: No. Unless a step is configured with a custom if condition (like if: always()): true, it defaults to running only if all previous steps succeeded

darshan-sj (approved):

PR #3875, opened 2026-06-03
[Data Validation] Add support for Custom Transformation

aasthabharill:
b/519845738 

Adds support for Custom Transformation in Data Validation pipeline. 

#### Follow up work:
- Handle exceptions/errors returned by Transformation gracefully (b/519845805)

#### Testing:
Built custom flex template and tested these cases:
1. Happy Path: Transformed Row at spanner was correctly validated with the original row at source
2. Data Mismatch: Corresponding Spanner row didn't have correctly transformed values - was flagged as a mismatch by validation pipeline
3. NULL Handling: NULL values in row handled correctly without throwing error
4. Filtered Rows: Row was dropped from Spanner after transformation, but the original row was NOT flagged with `MISSING_IN_DESTINATION` tag during validation
5. PK Transform: Transformed Row at spanner had a transformed Primary key value but was correctly validated with the original row at source
6. Data type casting: Transformation casted STRING to INT but validation pipeline was able handle this and compare the rows correctly. 

Note: Error scenario was also tested which crashed the pipeline - but since this requires a bigger effort, it's created as a follow up bug as mentioned above.

### DIFF
--- a/v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
+++ b/v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
@@ -169,7 +169,7 @@ public class GCSSpannerDV {
         example = "[{Singers, Vocalists}, {Albums, Records}]",
         helpText =
             "These are the table name overrides from source to spanner. They are written in"
-                + " thefollowing format: [{SourceTableName1, SpannerTableName1}, {SourceTableName2,"
+                + " the following format: [{SourceTableName1, SpannerTableName1}, {SourceTableName2,"
                 + " SpannerTableName2}]This example shows mapping Singers table to Vocalists and"
                 + " Albums table to Records.")
     @Default.String("")
@@ -187,7 +187,7 @@ public class GCSSpannerDV {
             "[{Singers.SingerName, Singers.TalentName}, {Albums.AlbumName, Albums.RecordName}]",
         helpText =
             "These are the column name overrides from source to spanner. They are written in"
-                + " thefollowing format: [{SourceTableName1.SourceColumnName1,"
+                + " the following format: [{SourceTableName1.SourceColumnName1,"
                 + " SourceTableName1.SpannerColumnName1}, {SourceTableName2.SourceColumnName1,"
                 + " SourceTableName2.SpannerColumnName1}]Note that the SourceTableName should"
                 + " remain the same in both the source and spanner pair. To override table names,"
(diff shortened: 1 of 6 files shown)


v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
```diff
@@ -164,9 +168,10 @@ public interface Options extends PipelineOptions {
             "^\\[([[:space:]]*\\{[[:graph:]]+[[:space:]]*,[[:space:]]*[[:graph:]]+[[:space:]]*\\}[[:space:]]*(,[[:space:]]*)*)*\\]$",
         example = "[{Singers, Vocalists}, {Albums, Records}]",
         helpText =
-            "These are the table name overrides from source to spanner. They are written in the"
-                + "following format: [{SourceTableName1, SpannerTableName1}, {SourceTableName2, SpannerTableName2}]"
-                + "This example shows mapping Singers table to Vocalists and Albums table to Records.")
+            "These are the table name overrides from source to spanner. They are written in"
+                + " the following format: [{SourceTableName1, SpannerTableName1}, {SourceTableName2,"
+                + " SpannerTableName2}]This example shows mapping Singers table to Vocalists and"
+                + " Albums table to Records.")
```
manitgupta: Are these formatting changes needed? Please keep that only if there were formatter violations...

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
```diff
@@ -241,12 +286,27 @@ public static PipelineResult run(Options options) {
...
+    if (options.getTransformationJarPath() != null
+        && !options.getTransformationJarPath().isEmpty()
+        && options.getTransformationClassName() != null
+        && !options.getTransformationClassName().isEmpty()) {
+      customTransformation =
+          CustomTransformation.builder(
+                  options.getTransformationJarPath(), options.getTransformationClassName())
+              .setCustomParameters(options.getTransformationCustomParameters())
+              .build();
+    }
```
manitgupta: Very ugly, can we wrap these checks inside a method and optionally see if we can simplify them?

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/templates/GCSSpannerDV.java
aasthabharill: yeah mvn spotless changed them

manitgupta (approved):

PR #3887, opened 2026-06-04
Replace JLine static `Log` reference with SLF4J `Logger` instance

JackPGreen:
The class(es) declare a SLF4J `Logger LOG` field, but sometimes the JLine `Log` static class is referenced instead.

It makes no sense to swap logging implementations mid-class, _my assumption_ is that it's an error by referencing the class instead of the field.


VardhanThigle (dismissed):

JackPGreen:
@VardhanThigle are you able to merge, please?

JackPGreen:
> @VardhanThigle are you able to merge, please?

bump :)

JackPGreen:
Addressed in https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4143

PR #3900, opened 2026-06-09
[DataStreamToSpanner] Deleting GCS failure test for Live template

darshan-sj:
The Live template GCS failure test was doing the following:

* Launch live migration template
* Remove GCS read permission for the worker service account and then write some events on source.
* Add the GCS permissions back and then write few more events on source.
* Finally wait till all the events to be migrated to Spanner and assert all the data exactly matches with source.

But modifying iam permissions in a test is strictly prohibited. We cannot write and run this test case. Hence removing this Failure injection test.


bharadwaj-aditya (approved):
LGTM

PR #3899, opened 2026-06-09
Add project-context.md for sourcedb-to-spanner

aasthabharill:
b/521647942

PR #3903, opened 2026-06-09
feat: Add `project_context_generator` and `project_context_updater` agent skill 

aasthabharill:
b/521646880
b/521646367

- Used this skill to add project_context.md files for `sourcedb-to-spanner`, `spanner-to-sourcedb` and `datastream-to-spanner`
- Good at generating Overview, technical details and architecture diagram
- Needs some manual knowledge transfer and iterations in `AI Agent Tips` section

References:
1. https://docs.google.com/document/d/1wmbXPOXB6sr8xM5VhBXfSos_O0qqZAzkqFltJKiDWys/edit?tab=t.0#heading=h.xrbrshhinr6r
2. https://g3doc.corp.google.com/company/engedu/aisg/vibe/prototyping/lessons/ai-mnesia/index.md?cl=head#anatomy-of-a-context-file
3. Similar skills exist in g3 already, i have modified them from an open-source and github management perspective

### DIFF
--- a/.agents/skills/project_context_updater/OWNERS
+++ b/.agents/skills/project_context_updater/OWNERS
@@ -1,0 +1,1 @@
+aasthabharill
--- a/.agents/skills/project_context_updater/SKILL.md
+++ b/.agents/skills/project_context_updater/SKILL.md
@@ -43,6 +43,7 @@ Copy this checklist and track your progress:
 -   [ ] Step 4: Validate Tech Stack Versions
 -   [ ] Step 5: Update Coding Standards & Testing Frameworks
 -   [ ] Step 6: AI Agent Tips Updates
+-   [ ] Step 7: Prune Outdated Context
 
 ### Step 1: Validate Build Commands
 
@@ -70,13 +71,28 @@ Did you introduce a new design pattern, testing framework (e.g., migrating from 
 ### Step 6: AI Agent Tips Updates
 
 Did you encounter any particular challenges while working on your changes? If
-you found any subtle project quirks, "gotchas", or tricky API interactions,
-document these in the `AI Agent Tips > Areas to be Careful` section.
+you found any subtle project quirks, new architectural decisions, or tricky API interactions,
+document these in the respective `AI Agent Tips` subsections:
+*   **Core Architectural Decisions:** Log finalized "Why we did X instead of Y" decisions.
+*   **Known Issues & Quirks:** Persistent bugs, workarounds, or limitations.
+*   **Lessons Learned & Ah-ha Moments:** Non-obvious solutions or debugging breakthroughs.
 
 *   **Example PRs**: Add a placeholder for your PR (or another representative PR) to the Example
     PRs section, or update it with the actual PR link after creation to help guide future AI agents.
 
-## 3. Creating the PR
+### Step 7: Prune Outdated Context
+
+Review the `AI Agent Tips`. If any entries are no longer relevant due to your recent changes, or if the section has become overly verbose, condense or delete them. Actively manage the file to prevent context window bloat.
+
+## 3. Best Practices for Humans
+
+To keep the `project-context.md` valuable, interact with the agent frequently to explicitly save context:
+
+*   **The Session Starter:** Kick off new chats with: *"Read project-context.md, then let's work on..."*
+*   **The Explicit Save:** If you spend time debugging a complex issue, explicitly prompt the agent: *"Summarize the fix we just found and update the Lessons Learned section in project-context.md so we don't forget it."*
+*   **Keep it Pruned:** If the file is getting large, ask the agent: *"Review project-context.md and condense any outdated or overly verbose points."*
+
+## 4. Creating the PR
 
 Once updates are made, commit the project-context.md changes. If updating an existing PR, push the changes to the remote branch. If creating a new PR, use gh pr create.
 If needed, ask the user to authenticate to git/GitHub (e.g., `gh auth login`).
--- a/.agents/skills/project_context_updater/SKILL.md
+++ b/.agents/skills/project_context_updater/SKILL.md
@@ -73,12 +73,12 @@ Did you encounter any particular challenges while working on your changes? If
 you found any subtle project quirks, "gotchas", or tricky API interactions,
 document these in the `AI Agent Tips > Areas to be Careful` section.
 
-*   **Example PRs**: Add your PR (or another representative PR) to the `Example
-    PRs` section to help guide future AI agents with similar tasks.
+*   **Example PRs**: Add a placeholder for your PR (or another representative PR) to the Example
+    PRs section, or update it with the actual PR link after creation to help guide future AI agents.
 
 ## 3. Creating the PR
 
-Once updates are made, commit the `project-context.md` changes and upload them as part of your existing PR (or create a new one) using `git commit` and `gh pr create`.
+Once updates are made, commit the project-context.md changes. If updating an existing PR, push the changes to the remote branch. If creating a new PR, use gh pr create.
 If needed, ask the user to authenticate to git/GitHub (e.g., `gh auth login`).
 
 ## Reporting Issues
--- a/.agents/skills/project_context_updater/SKILL.md
+++ b/.agents/skills/project_context_updater/SKILL.md
@@ -1,0 +1,86 @@
+---
+name: project-context-updater
+description: >-
+  Reads a project-context.md file to quickly get context on a project's
+  architecture and tools. Update the project-context.md
+  file AFTER completing a code change to ensure the documentation remains
+  accurate and to add any new learnings, troubleshooting tips, or example
+  PRs to the AI Agent Tips section. Use this skill to quickly ramp up on a
+  codebase using project-context.md. Don't use when instructed to create a
+  project-context.md file from scratch (use
+  project-context-generator instead).
+---
+
+# Project Context Updater
+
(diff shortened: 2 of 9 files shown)


.agents/skills/project_context_updater/OWNERS
```diff
@@ -0,0 +1 @@
+aasthabharill
```
manitgupta, line 1: How is `OWNERS` used? I received a PR the other day from @damccorm where he removed an OWNERS file?

.agents/skills/project_context_updater/TEST.md
```diff
@@ -0,0 +1,43 @@
+# Project Context Updater - Agent E2E Test Plan
+
+## Prerequisites
```
manitgupta, line 3: How is `TEST.md` used? Is there some infra we have built around testing the skill?

.agents/skills/project_context_updater/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+name: project-context-updater
+description: >-
+  Reads a project-context.md file to quickly get context on a project's
+  architecture and tools. Update the project-context.md
+  file AFTER completing a code change to ensure the documentation remains
+  accurate and to add any new learnings, troubleshooting tips, or example
+  PRs to the AI Agent Tips section. Use this skill to quickly ramp up on a
+  codebase using project-context.md. Don't use when instructed to create a
+  project-context.md file from scratch (use
+  project-context-generator instead).
```
manitgupta, line 11: Why do we need 2 skills? Can't we have 1 skill that does all CRUD associated with project context? 
Also, how do we ensure that the proposed edits say coherent over days/weeks? 

My initial intuition is that these DataflowTemplates are not very complex for an agent to read and understand, do you have practical examples where having a project context helped? 

My gut tells that this would be more useful across extremely complex codebases, and I want you to weigh the benefits of having this v/s the overhead to maintain a consistent, coherent copy of this file over many weeks/months.

Let me know what you think?

v2/sourcedb-to-spanner/project-context.md
```diff
@@ -54,12 +61,14 @@
...
+*   **Known Issues & Quirks:**
     *   **Foreign Keys:** The pipeline processes parent tables before child tables, but cyclic (self-referencing) foreign keys will cause startup failures and are unsupported.
     *   **Pipeline Logic:** Cross-shard querying logic, causal ordering around the DLQ, and schema mappings parsing are highly complex areas.
+    *   **Integration/Load Tests:** NEVER execute `*IT.java` (Integration) or `*LT.java` (Load) test suites during local coding/machine verification. These require remote environments. Only execute `*Test.java` (Unit) locally.
+*   **Lessons Learned & Ah-ha Moments:**
+    *   **OOM Prevention (MySQL Cursor Fetch):** Always configure `fetchSize` to prevent Out Of Memory errors.
 *   **Example PRs:**
     *   [39a8ae5e0](https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/39a8ae5e0) - Fix GCS Avro Export flow
     *   [90964dca6](https://github.com/GoogleCloudPlatform/DataflowTemplates/commit/90964dca6) - Add Support for UUID-based Partitioning
```
manitgupta, line 74: This is the kind of stuff that will get stale over weeks/months and probably confuse the agent?

.agents/skills_index.md
manitgupta, line 2: How does this help? Who reads this? Are AI agents tuned to automatically look for this file? (Just like they look for a `SKILL.md` file in `.agents/skills/`?

.agents/skills/project_context_updater/OWNERS
aasthabharill: was just following g3 format, will remove it just saw the PR you're referencing to

.agents/skills/project_context_updater/TEST.md
aasthabharill: this file is used in g3 skills - its not an actual test that executes, but its a reference point for a new user trying to use the skill and check if it still works as intended. The user can manually run the test by prompting the agent to perform the tasks though

We need to write EVALs to actually run tests on a skill but think thats a g3 internal framework

.agents/skills/project_context_updater/SKILL.md
aasthabharill: 1. The two skills are serving different purposes and need to be called upon in different scenarios, so don't want to merge them. The generator skill is a one-time call to create a project-context.md file, while the updater skill is to be called upon while making any siginificant change to the repository to ensure project-context.md stays relevant

2. I think the intention behind adding this project-context file is more for these purposes:  
- call out and add developer gotchas and correct things that the AI gets wrong often.
- call out known limitations, and workarounds that we want code to account for

I also think knowing a bunch of our more complex functionality and usecases upfront by reading this doc can help the AI account for edge cases that might not be clearly obvious in code. So its to provide some more direction of thought to the AI. So i dont think the size of the codebase should impact this usecase. 

The current doc might be a bit barebones, the team needs to take up maintaining it by calling the updater tool at the end of each meaningful session. While that might sound like effort, it will help both the AI and other team members to keep gotchas in mind while coding

I do think maybe i can see if some of the sections (like testing practices) are actually required or helpful as that's easily inferable. 

[Here](https://source.corp.google.com/piper///depot/google3/java/com/google/android/glasses/apps/health/MEMORY.md) is an example

v2/sourcedb-to-spanner/project-context.md
aasthabharill: ill add a section in the updater to keep these fresh, had added one earlier but maybe got lost between commits

.agents/skills_index.md
aasthabharill: my bad, it was in the g3 skill directory structure and i assumed this was required to discover skills but that doesnt seem to be the case (for g3 or jetski). i have mainly been following existing skills and docs in g3 to write skills and usually follow the practices i see there. 

I still think its worth keeping this for us to find relevant skills, but definitely not required and can be removed if you think so.

manitgupta (changes requested):
I would encourage you to please take the time to review -

1. The content of the change you're raising
2. The environment you're raising the PR against (DataflowTemplates) and follow its conventions. 

Between this PR and the [previous one](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3899) there are instances of changes being made without self-verification. 

- Commits containing buganizer component paths, email aliases etc.
- Containing redundant files and patterns that will actually not work on this repository.
- Incorrect phrasing (PR v/s CL) which makes it seem like the content was not self-reviewed before raising the PR.

The burden to validate (and catch) what would work and what wouldn't should not be placed on the reviewers. This is a slippery slope that will not scale. 
You are expected to ensure a baseline quality of the change. 

I will review the comments in a bit and respond.

.agents/skills/project_context_updater/TEST.md
manitgupta: I don't think its valuable in its current form. Lets remove it.

v2/sourcedb-to-spanner/project-context.md
manitgupta: How will the updater keep it fresh? How will it find the relevant *new* PRs or commits to link here? Do we think it will be able to do so reliably?

.agents/skills_index.md
manitgupta: I think its not valuable in its current form. Let's remove it.

.agents/skills/project_context_updater/SKILL.md
manitgupta: I have a couple of thoughts here. Let's break it down - 

> The generator skill is a one-time call to create a project-context.md file

I think we own a handful of templates here (4-5). Do we need this one-time skill to be committed when the probable expected usage is a total of 5 times, and then sparingly after that? (whenever we add a net new template or module maybe?)

> updater skill is to be called upon while making any siginificant change to the repository to ensure project-context.md stays relevan

What/Who decides that the changes are significant enough for an update to the project context? As I said earlier, I understand the spirit of what this is trying to do, I'm just not convinced we need a dedicated set of CRUD skills around the context file.


> I think the intention behind adding this project-context file is more for these purposes:
> * call out and add developer gotchas and correct things that the AI gets wrong often.
> * call out known limitations, and workarounds that we want code to account for

This I agree with the most. However, this is an argument supporting the usefulness of having a context file, which was never in question. 

> I also think knowing a bunch of our more complex functionality and usecases upfront by reading this doc can help the AI account for edge cases that might not be clearly obvious in code. 

Again, agree with this. The non-obvious use-cases, gotchas, edge cases, when written in a context, should be useful to AI agents.

---

All of this leads us to the same conclusion - **a curated, accurate `CONTEXT.md` is the real thing of value here.** The *initial* focus of this exercise should be to *write that context accurately first, and not build scaffolding skills around it*.  Once that is done, we can then think about *how* to maintain it.

From the previous PR, I saw a bunch of comments from Vardhan around content in the context file for bulk migration which was not accurate, which got me concerned (hence the pushback). That has a higher disruptive potential since agents may mistake it for truth and get lost. 

#### Suggested Next Steps

1. We should park adding the CRUD skills around the context files for now. 
2. We should focus on writing a good, validated context files for the templates that we own. We can start small (for e.g. the bulk migration context you have already committed). We *should* absolutely use AI to define the format, write up the content, **but** verify that every single fact in that context is human validated against the codebase. If that requires discussions or digging into the code for verification - we should do that.

PR #3942, opened 2026-06-24
Seperated live migration

bharadwaj-aditya:
As a part of the source separation and platformization. Moved all the code specific to each source into a folders specific to each source and sourceConnectors.

The current PR is purely refactoring to isolate the implementation specific to a source. 
Follow up PRs will change the structure of the connectors.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
@@ -101,7 +101,7 @@ public abstract class ChangeEventContext {
   }
 
   // Abstract method to generate shadow table mutation.
-  abstract Mutation generateShadowTableMutation(Ddl ddl, Ddl shadowDdl)
+  protected abstract Mutation generateShadowTableMutation(Ddl ddl, Ddl shadowDdl)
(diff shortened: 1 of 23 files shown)


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
```diff
@@ -101,7 +101,7 @@ protected ChangeEventContext(
   }
 
   // Abstract method to generate shadow table mutation.
-  abstract Mutation generateShadowTableMutation(Ddl ddl, Ddl shadowDdl)
+  protected abstract Mutation generateShadowTableMutation(Ddl ddl, Ddl shadowDdl)
```
pratickchokhani, line 104: Since we are planning to move this to another library, should this be made public?

pratickchokhani (approved):

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/datastream/ChangeEventContext.java
bharadwaj-aditya: we can do that as a follow up when we make the move. Trying to keep the change in logic minimum in this PR

PR #3956, opened 2026-06-30
feat: introduce validateFields method to SourceConnectionConfig and implement validation for all source configurations

pratickchokhani:


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/Shard.java
```diff
@@ -197,4 +197,37 @@ public int hashCode() {
...
+    if (secretManagerUri == null) {
+      throw new IllegalArgumentException("secretManagerUri cannot be null");
+    }
+    if (connectionProperties == null) {
+      throw new IllegalArgumentException("connectionProperties cannot be null");
+    }
+    if (dbNameToLogicalShardIdMap == null) {
+      throw new IllegalArgumentException("dbNameToLogicalShardIdMap cannot be null");
+    }
+  }
```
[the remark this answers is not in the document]
pratickchokhani: We only care of null values. Depending on the templates and source, some fields can be empty.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/AstraConnectionConfig.java
```diff
@@ -54,4 +54,20 @@ public String getAstraDbRegion() {
...
+    if (astraToken == null) {
+      throw new IllegalArgumentException("astraToken cannot be null");
+    }
+    if (keySpace == null) {
+      throw new IllegalArgumentException("keySpace cannot be null");
+    }
+    if (astraDbRegion == null) {
+      throw new IllegalArgumentException("astraDbRegion cannot be null");
+    }
+  }
```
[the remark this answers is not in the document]
pratickchokhani: We only care of null values. Depending on the templates and source, some fields can be empty.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/Shard.java
```diff
@@ -197,4 +197,37 @@ public int hashCode() {
...
+      throw new IllegalArgumentException("password cannot be null");
+    }
+    if (dbName == null) {
+      throw new IllegalArgumentException("dbName cannot be null");
+    }
+    if (namespace == null) {
+      throw new IllegalArgumentException("namespace cannot be null");
+    }
+    if (secretManagerUri == null) {
+      throw new IllegalArgumentException("secretManagerUri cannot be null");
```
bharadwaj-aditya, line 224: secret manager or password can be null. It is possible to pass only one of them.

Can you please verify the logic here ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/JdbcShardConfig.java
```diff
@@ -33,4 +33,14 @@ public List<Shard> getShardConfigs() {
   public void setShardConfigs(List<Shard> shardConfigs) {
     this.shardConfigs = shardConfigs;
   }
+
+  @Override
+  public void validateFields() {
+    if (shardConfigs == null || shardConfigs.isEmpty()) {
```
bharadwaj-aditya, line 39: what happens in the non-sharded use case ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/SourceConnectionConfig.java
```diff
@@ -23,4 +23,11 @@
...
  * <p>This interface will be expanded in Phase 2 during the platformization phase.
  */
-public interface SourceConnectionConfig {}
+public interface SourceConnectionConfig {
+
+  /**
+   * Validates that the source configuration is valid. Throws IllegalArgumentException if the
+   * configuration is invalid.
+   */
+  void validateFields();
```
bharadwaj-aditya, line 32: If the validation is implemented at the source, then it can be private.

This method should not be required.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/Shard.java
pratickchokhani: We have integration tests for both secretManagerUri and password. Unless someone sets the value as null manually in the JSON or HOCON string, these values will not be null.

Hence,  this check is added.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/JdbcShardConfig.java
pratickchokhani: List will have single element.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/SourceConnectionConfig.java
pratickchokhani: This only validates the null checks and basic checks. It is called by the SourceConfigParser before returning the config.

Interface fields cannot be private.

bharadwaj-aditya: If validate is called in each of the flows right after reading the config, then this validation is not really achieving anything.  Especially if the code of what to validate is within the source implementation.

 The value of the validate would be if it lies outside of the source implementation.  Can you think of a scenario where the source specific validation can not be done while reading the source config ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/shard/Shard.java
bharadwaj-aditya: 2 questions
1. Is there a scenario where someone would want to set these fields explicitly as null and not empty strings? Maybe for password ?
2.  Does the integration test handle the scenario where all 4 combinations of these 2 parameters are set ?

PR #3957, opened 2026-06-30
Directrunner fix for live/reverse template

dhwanilpatel:
### Issue

Fix Live/Reverse replication integration tests on DirectRunner

Currently, the Bulk template integration tests (ITs) execute successfully using the DirectRunner, but the Live and Reverse replication ITs fail or time out. This PR resolves these runner-specific issues.

By enabling all template ITs to run on the DirectRunner, this change allows for full local test execution. This significantly reduces the overhead of deploying to the Dataflow runner during test development and decreases local test execution times.

#### 1. Fix TextIO Sharding for Local Testing
                                                                                                                                                                          
  • **Issue:** Writing unbounded data (like dead-letter queues) to files via  TextIO  defaults to runner-determined dynamic sharding.                                         
  • **Why Dataflow Passes:** The fully-managed Dataflow runner inherently supports and dynamically orchestrates unbounded file sharding.                                      
  • **Why DirectRunner Fails:** The  DirectRunner  does not support dynamic sharding for unbounded  PCollections  and immediately throws an  IllegalArgumentException .       
  • **Fix Applied:** Added  .withNumShards(20)  to explicitly define a fixed shard count for direct runner, bypassing dynamic sharding and allowing local  DirectRunner  execution to succeed. 
                                                                                                                                                                          
#### 2. Fix Spanner Change Stream Race Condition
                                                                                                                                                                          
  • **Issue:** Integration tests inject seed data into Spanner, which the pipeline must read via a Change Stream.                                                             
  • **Why Dataflow Passes:** The IT framework uses the Cloud API to actively block test execution until the Dataflow job is  RUNNING  and workers are provisioned, ensuring the Change Stream reader is listening before data is inserted.                                                                                                          
  • **Why DirectRunner Fails:** The IT framework spawns the  DirectRunner  asynchronously in a background Java thread and instantly reports  RUNNING . This causes the test to fire Spanner mutations while the pipeline is still initializing locally, resulting in dropped events.                                                                
  • **Fix Applied:** Injected a  startTimestamp  into the pipeline options set slightly before the test fires its mutations. This forces the Spanner reader to look backward in time, ensuring no events are missed regardless of local initialization delays.

### DIFF
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
@@ -863,18 +863,22 @@ public class DataStreamToSpanner {
     TextIO.Write filterEventsWrite;
     if (options.getRunner() != null && options.getRunner().getSimpleName().equals("DirectRunner")) {
       // DirectRunner does not support dynamic sharding for unbounded PCollections
-      filterEventsWrite = TextIO.write().to(filterEventsDirectory).withSuffix(".json").withWindowedWrites().withNumShards(20);
+      filterEventsWrite =
+          TextIO.write()
+              .to(filterEventsDirectory)
+              .withSuffix(".json")
+              .withWindowedWrites()
+              .withNumShards(20);
     } else {
       // Cloud Dataflow natively supports dynamic sharding
-      filterEventsWrite = TextIO.write().to(filterEventsDirectory).withSuffix(".json").withWindowedWrites();
+      filterEventsWrite =
+          TextIO.write().to(filterEventsDirectory).withSuffix(".json").withWindowedWrites();
     }
 
     transformedRecords
         .get(DatastreamToSpannerConstants.FILTERED_EVENT_TAG)
         .apply(Window.into(FixedWindows.of(org.joda.time.Duration.standardMinutes(1))))
-        .apply(
-            "Write Filtered Events To GCS",
-            filterEventsWrite);
+        .apply("Write Filtered Events To GCS", filterEventsWrite);
 
     spannerConfig =
         SpannerServiceFactoryImpl.createSpannerService(
--- a/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
+++ b/v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
@@ -860,10 +860,15 @@ public class DataStreamToSpanner {
             ? tempLocation + "filteredEvents/"
             : options.getFilteredEventsDirectory();
     LOG.info("Filtered events directory: {}", filterEventsDirectory);
-    TextIO.Write filterEventsWrite = TextIO.write().to(filterEventsDirectory).withSuffix(".json").withWindowedWrites();
+    TextIO.Write filterEventsWrite;
     if (options.getRunner() != null && options.getRunner().getSimpleName().equals("DirectRunner")) {
-      filterEventsWrite = filterEventsWrite.withNumShards(20);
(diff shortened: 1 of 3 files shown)


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
```diff
@@ -274,6 +275,7 @@ public PipelineLauncher.LaunchInfo launchDataflowJob(
             put("numWorkers", "1");
             put("sourceType", sourceType);
             put("workerMachineType", "n2-standard-4");
+            put("startTimestamp", Timestamp.now().toString());
```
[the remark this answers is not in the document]
dhwanilpatel: Incorporated clock skew by providing a timestamp from five minutes ago.

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
```diff
@@ -860,12 +860,16 @@ static Pipeline buildPipeline(Options options) {
             ? tempLocation + "filteredEvents/"
             : options.getFilteredEventsDirectory();
     LOG.info("Filtered events directory: {}", filterEventsDirectory);
+    TextIO.Write filterEventsWrite = TextIO.write().to(filterEventsDirectory).withSuffix(".json").withWindowedWrites();
```
bharadwaj-aditya: suggest to move this to an If-else block or a method for clarity rather than overwriting.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
```diff
@@ -274,6 +275,12 @@ public PipelineLauncher.LaunchInfo launchDataflowJob(
             put("numWorkers", "1");
             put("sourceType", sourceType);
             put("workerMachineType", "n2-standard-4");
+            // Set startTimestamp 5 minutes in the past to account for any clock skew between
+            // the local machine/CI runner and Spanner's server time.
+            put(
```
bharadwaj-aditya: This is not clear. What If the changestream timestamp is in the past, before the spanner db was created ?

v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/templates/DataStreamToSpanner.java
dhwanilpatel: Moved it to if-else.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDbITBase.java
dhwanilpatel: The `startTimestamp` was originally added because the DirectRunner takes a few seconds to spin up its background threads, and it was missing the initial test mutations on Spanner. I initially added a 5-minute padding to handle any clock skew between the CI runner and Spanner. But you're right, since these integration tests spin up the Spanner DB on the fly, looking 5 minutes back can be even before creation of spanner DB.

To fix both the clock skew and the race condition, I updated the code to just run a quick  `SELECT CURRENT_TIMESTAMP()`  directly against Spanner right before launching the job. Passing that server timestamp into the pipeline safely bypasses any local clock skew issues, while still making sure the runner looks backward enough to catch the test mutations."

bharadwaj-aditya (dismissed):
As a follow up. Please also update the readme with how to use the direct runner for these tests.

bharadwaj-aditya (commented):
LGTM

bharadwaj-aditya (approved):
LGTM

PR #3960, opened 2026-06-30
Restore write permission to release.yml

Abacn:


.github/workflows/release.yml
```diff
@@ -300,7 +284,6 @@ jobs:
           ref: ${{ steps.variables.outputs.releaseBranchName }}
           fetch-depth: 0
           token: ${{ secrets.GITHUB_TOKEN }}
-          persist-credentials: false
```
damccorm, line 303: Don't we need this for the next step to work?

Abacn: Sorry wasn't clear. This PR simply revert .github/workflow/release.yml prior to the latest change

damccorm (approved):

PR #3964, opened 2026-07-01
build: bump import-spec

fbiville:
This means JSON Schema library gets updated too, with breaking changes handled in that commit as well.


ali-ince (approved):
LGTM.

Abacn (approved):

PR #3969, opened 2026-07-01
[DO NOT MERGE] Validate beam 2.75.0 RC1

Amar3tto:


Abacn:
```
Caused by: java.lang.NoSuchMethodError: 'com.google.api.services.dataflow.model.DataflowPackage com.google.api.services.dataflow.model.DataflowPackage.setSha256(java.lang.String)'
```

 need to bump dataflow-api version in Templates:

https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/bc35324a051f0f59a52a4ed7d3d9ca6f39a1c35a/it/pom.xml#L34

PR #3970, opened 2026-07-02
Datastream LT fixes

aasthabharill:
b/527103060

### RCA 
In .github/workflows/load-tests.yml file we pass the flag `--it-private-connectivity="datastream-connect-2"` which tells the test to use private connectivity. But in the datastream tests, we pass the host IP as a secret which contains the Public IPs of the CloudSQL instance. This is the reason we were seeing the JDBC Connection error. 

#### Test 1: 
Removing `--it-private-connectivity="datastream-connect-2"` from the spanner-load-tests.yml file worked and the load test succeeded

#### Test 2: 
Keeping the `spanner-load-tests.yml` file same as the original (i.e. still passing private-connectivity flag), i changed the hostIp to use the private IP but this run still failed. 

Here's the current issue: 

VPC Peering in Google Cloud is non-transitive.
- The CloudSQL database we use (blob-table) lives in a Google-managed VPC that is peered to your cloud-teleport-testing VPC.
- Datastream's Private Connectivity (`datastream-connect-2`) lives in a separate Google-managed VPC that is also peered to the cloud-teleport-testing VPC.
- Datastream cannot "hop" through your VPC to reach the CloudSQL VPC because transitive peering is not allowed.

This is exactly why DataStreamToSpanner5kTablesLT works: it doesn't connect to a CloudSQL private IP directly. Instead, it connects to <private IP address>, which is an internal Cloud SQL Auth Proxy running on a VM directly inside the VPC. This proxy bridges the gap.

So to get it working, we will have to:
- Set up a Proxy: You must deploy a proxy (e.g., Cloud SQL Auth Proxy, HAProxy, or Envoy) on a Compute Engine VM or GKE cluster running directly inside the VPC that datastream-connect-2 is peered with.
- Configure Forwarding: Configure this proxy to connect to your CloudSQL instance's Private IP.
- Update Secrets: Update the versions/2 of blob-table-ip (and the IPs for all the other tests) in Secret Manager to point to the Proxy's IP address, not the CloudSQL Private IP directly.

#### Resolution
We are okay with using Public IPs as that is what the test has been using all along anyway. It just broke now probably due to some patch fixes in Datastream, so we just have to skip passing the private-connectivity flag to avoid any confusion/issues.

### DIFF
--- a/.github/workflows/load-tests.yml
+++ b/.github/workflows/load-tests.yml
@@ -35,7 +35,6 @@ jobs:
         -Dproject=cloud-teleport-testing \
         -DartifactBucket=gs://cloud-teleport-testing-it-gitactions \
         -DhostIp=${HOST_IP} \
-        -DprivateConnectivity="datastream-connect-2" \
         -DexportProject=cloud-teleport-testing \
         -DexportDataset=performance_tests \
         -DexportTable=template_performance_metrics -e -fae
--- a/.github/workflows/load-tests.yml
+++ b/.github/workflows/load-tests.yml
@@ -35,6 +35,7 @@ jobs:
         -Dproject=cloud-teleport-testing \
         -DartifactBucket=gs://cloud-teleport-testing-it-gitactions \
         -DhostIp=${HOST_IP} \
+        -DprivateConnectivity="datastream-connect-2" \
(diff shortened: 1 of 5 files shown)


darshan-sj (approved):
LGTM

PR #3990, opened 2026-07-06
Anomaly detection template improvements

claudevdm:
- Drop rows with null measure values to avoid TypeErrors during aggregation
- Handle arithmetic errors during metric evaluation to prevent pipeline stalls
- Nest keys to prevent webhook ID collisions in AsyncWrapper, 
- Reject late elements early in TimestampBufferDoFn to prevent key wedging (this shouldnt technically be possible due to how event timestamp works for bq cdc source.


shunping (approved):
LGTM

PR #3991, opened 2026-07-06
update workflows to be node 24 compatible

derrickaw:
Either upgrading their version or replaced.

### DIFF
--- a/.github/workflows/load-tests.yml
+++ b/.github/workflows/load-tests.yml
@@ -43,13 +43,24 @@ jobs:
         HOST_IP: ${{ steps.variables.outputs.hostIP }}
     - name: Create Github issue on failure
       if: failure()
-      uses: JasonEtco/create-an-issue@1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5 # v2.9.2
+      uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         JOB_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         DATE: ${{ steps.date.outputs.date }}
       with:
-        filename: .github/ISSUE_TEMPLATE/load-test-failure-issue-template.md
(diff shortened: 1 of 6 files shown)


.github/workflows/load-tests.yml
```diff
@@ -43,13 +43,24 @@ jobs:
         HOST_IP: ${{ steps.variables.outputs.hostIP }}
     - name: Create Github issue on failure
       if: failure()
-      uses: JasonEtco/create-an-issue@1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5 # v2.9.2
+      uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
```
Abacn: I remember we saw "Load Test failure" bugs get created in the past, but no one paid attention to it. It seems this auto created bugs were deleted as well:

https://github.com/GoogleCloudPlatform/DataflowTemplates/issues?q=is%3Aissue%20state%3Aclosed%20load%20test

We may just drop this step since it's not practiced.

.github/workflows/release-notes-labels.yml
```diff
@@ -9,17 +9,26 @@ jobs:
       issues: write
       pull-requests: write
     steps:
-      - uses: mheap/github-action-required-labels@0ac283b4e65c1fb28ce6079dea5546ceca98ccbe
```
Abacn, line 12: This action is still actively maintained:

https://github.com/mheap/github-action-required-labels

Can we use the latest version, or open a fix upstream?

Abacn (commented):
Thanks for the fix. LGTM after best avoiding raw scripts parts

.github/workflows/load-tests.yml
derrickaw: Cool, removed this step. Thanks

.github/workflows/release-notes-labels.yml
derrickaw: Will open an upstream PR to support Node 24. Thanks.

https://github.com/mheap/github-action-required-labels/pull/97

Merged already - just need to wait for next release and update the version.

Abacn (approved):

PR #4017, opened 2026-07-12
Refactor spanner-staging-tests 

aasthabharill:
b/534589929

This PR refactors the Spanner staging integration test pipeline to dynamically change the run-name to different testing environments (Production, Staging, Cloud-devel). 

#### 1. Added a dynamic `run-name` expression in `.github/workflows/spanner-staging-tests.yml`
Because this workflow runs against three different Spanner hosts (Prod, Staging, Cloud-devel), distinguishing the runs in the GitHub Actions UI was difficult. Test runs will now clearly identify their target environment (e.g., *"Spanner Cloud-devel Integration Tests"*). 

#### 2. Removed the explicit `spannerHost` input configuration from the `pull_request` block to avoid masking defaults
- `pull_request` triggers don't accept the `inputs` parameter - so the workflow that this PR triggered had no Spanner host set, which is incorrect. ([Workflow link](https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/29193896609/job/86653305439))
- Added fallback to cloud-devel URL in the job `env` block to rectify this issue

#### 3. Removed `@RunWith(JUnit4.class)` from CobyDbIT and CopyDbRandomisedIT test files
- These classes inherit from SpannerTemplateITBase (which defines @RunWith(Parameterized.class) to inject the custom spannerHost), but the JUnit4 annotation overrode the parameterization.
- This caused the spannerHost variable to remain null in those tests. 
- SpannerResourceManager silently bypassed the custom host configuration and defaulted to the standard PROD Spanner endpoint (https://spanner.googleapis.com) even in staging/cloud-devel workflows. 

image

#### 4. Changed `PARTITION_ID` to `default` instead of creating a new one
- When a spanner instance is created, Spanner implicitly provisions a built-in partition named `"default"` that shares the instance's region configuration.
-  Spanner strictly prohibits creating a *new* custom instance partition that shares the exact same region as another partition - so we can't use the spanner instance region due to the default partition that was created
- Therefore, to successfully create a custom partition, the base instance must be created in one configuration (e.g., `nam3`), and the custom partition must be created in a *different* configuration (e.g., `nam6`).
- But across our three environments (Prod, Pre-Prod, and Cloud-devel), there isn't a universally shared set of *two distinct* multi-region configurations. 
- So we will utilize the default partition instead of provisioning a secondary partition.
- This change will ensure the tests work for cloud-devel and production code

### Tests
- Verified that the run name dynamically changes depending on the Spanner Host passed.
- This PR triggers the Spanner Staging Tests. The test correctly used the cloud-devel spanner host URL.

image

https://screenshot.googleplex.com/4Ew96Vt3ckq9mnh.png
- The CopyDbIT tests run on the correct host:
image

https://screenshot.googleplex.com/5YA3kPQctVuYCps.png

### DIFF
--- a/.github/workflows/spanner-staging-tests.yml
+++ b/.github/workflows/spanner-staging-tests.yml
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 name: Spanner Staging integration tests
+run-name: Spanner ${{ inputs.spannerHost == 'https://batch-spanner.googleapis.com' && 'Production' || inputs.spannerHost == 'https://preprod-spanner.sandbox.googleapis.com' && 'Staging' || inputs.spannerHost == 'https://staging-wrenchworks.sandbox.googleapis.com' && 'Cloud-devel' || 'Integration' }} Tests
 
 on:
   pull_request:
--- a/.github/workflows/spanner-staging-tests.yml
+++ b/.github/workflows/spanner-staging-tests.yml
@@ -33,6 +33,11 @@ on:
         type: string
         required: false
         default: ''
+      spannerMultiRegion:
+        description: 'Spanner multi-region to use for tests.'
+        type: string
+        required: false
+        default: 'nam3'
 
 permissions: write-all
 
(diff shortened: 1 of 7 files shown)


darshan-sj (approved):
Looks good

aasthabharill (commented):
Merging despite failures as failure is in InformationSchemaScannerIT that this code does not aim to fix - its just a workflow refactoring PR

PR #4019, opened 2026-07-13
Increasing timeout for SpannerToSpannerIT

shreyakhajanchi:


bharadwaj-aditya (approved):
This looks like we set a timeout of 1 second in case we exceed the 15 minutes. Should be fine.

PR #4031, opened 2026-07-16
[Data Validation] Add Integration Testing Framework and Core Matching IT

aasthabharill:
b/491703995

This PR introduces a robust, centralized integration testing framework for the `gcs-spanner-dv` pipeline, along with the first functional Integration Test (`GCSSpannerDVCoreMatchingIT`). The framework abstracts away boilerplate setup, standardizes Avro schema/data generation, and provides reliable BigQuery assertions to validate pipeline correctness without brittle logic.

#### `GCSSpannerDVITBase.java`
Serves as the foundational base class for all `gcs-spanner-dv` integration tests. It handles provisioning GCP resources (Spanner databases, BigQuery datasets), launching the Dataflow pipeline, and centrally tearing down resources after execution. Includes helper function to upload `.avro` files to GCS to act as the pipeline source.

#### `GCSSpannerDVAvroSetupHelper.java` (and `users.avsc`, `account_roles.avsc`)
A centralized helper class designed to decrease boilerplate code when setting up Avro schemas and mock data for integration tests. It abstracts the process of loading schemas and constructing Avro records.

I introduced a custom `RecordBuilder` within this helper class. 
- Why not `Map.of()`? It has a strict limit of 10 key-value pairs (capping tables at 10 columns) and instantly throws `NullPointerException`s if an explicit `null` value is passed. 
- Falling back to traditional, mutable `HashMap`s requires bulky, multi-line setup logic that bloats the test file. 
- The custom `RecordBuilder`, on the other hand, can scale infinitely, safely handles explicit `null` values, instantly validates column names against the schema, and abstracts the boilerplate nesting of Avro's inner payload records. 

The two schemas defined for this class were carefully constructed keeping the test plan in mind and will be re-used for 80% of the tests that don't require specific source requirements. The test cases will be simulated by modifying the corresponding Spanner schema - which each test will dynamically define and generate.

I considered adding 2 more functions that add a few standard rows in these tables to this file. While that function could have also been reused across test files, i feel it hinders readability of the IT file. In the current framework, the IT file defines the rows inserted in Spanner and Source which paints a clearer picture to the developer as to how the test case is set up. 

#### `GCSSpannerDVTestAsserts.java`
Provides reliable assertion logic for BigQuery output tables. Instead of merely asserting on row counts, it inspects row values for expected column data, confirming that the validation pipeline executed correctly.

#### `GCSSpannerDVCoreMatchingIT.java` (and `Users.sql`)
The first core functional integration test utilizing the new framework. It runs a full end-to-end matching scenario, uploading source data, spinning up the pipeline, and validating the BigQuery output via the assertions class.

- `validationTestWithMatchingAndMismatchedRecords`: This test validates the core functional accuracy of the gcs-spanner-dv batch pipeline by verifying that it can correctly identify and categorize rows into MATCHED, MISSING_IN_DESTINATION, MISSING_IN_SOURCE, and mismatches. 

### Files Modified
Added support for running validation ITs in Spanner PR IT runs


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
```diff
@@ -0,0 +1,147 @@
...
+
+  // --- Private Helper Methods ---
+
+  public GenericRecord createUsersRecord(
+      Long userId,
+      String eventId,
+      String fullName,
+      Integer age,
+      Instant timestamp,
+      String shardId) {
```
manitgupta: This is going to require a change everytime someone wants to edit a field in this record.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
```diff
@@ -0,0 +1,147 @@
...
+            Resources.getResource("GCSSpannerDVAvroSetupHelper/account_roles.avsc").openStream()) {
+      USERS_SCHEMA = new Schema.Parser().parse(usersIs);
+      ACCOUNT_ROLES_SCHEMA = new Schema.Parser().parse(rolesIs);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to load Avro schemas", e);
+    }
+  }
+
+  private final Schema usersSchema;
+  private final Schema accountRolesSchema;
```
manitgupta: These are sample tables for testing, I think they should be encapsulated somewhere.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.generic.GenericRecordBuilder;
+import org.apache.beam.it.gcp.artifacts.ArtifactClient;
+
+/**
+ * A centralized helper class for setting up Avro data in GCS for integration tests. This class
+ * abstracts away the boilerplate of loading Avro schemas, building records, and serializing them
+ * into .avro files uploaded to GCS.
+ */
+public class GCSSpannerDVAvroSetupHelper {
```
manitgupta, line 30: This class is written in a very coupled manner which will keep requiring breaking changes every time something changes. I suggest converting this into a table registry + one schema-driven builder.

Below code is generated, and shows the principle of what I am proposing - 

```
Move the per-table constants into a small registry, and make payload population data-driven from the schema so adding a column never touches the
  helper.

  /** Fixed, per-table metadata. Adding a table = one entry + its .avsc file. */
  public enum TableDef {
    USERS(USERS_SCHEMA, "Users", List.of("user_id", "event_id")),
    ACCOUNT_ROLES(ACCOUNT_ROLES_SCHEMA, "AccountRoles", List.of("role_id"));

    final Schema schema;
    final String tableName;
    final List<String> primaryKeys;

    TableDef(Schema schema, String tableName, List<String> primaryKeys) {
      this.schema = schema;
      this.tableName = tableName;
      this.primaryKeys = primaryKeys;
    }
  }

  /**
   * Builds an envelope record for any table. Payload columns are supplied by name;
   * unknown columns fail fast, and values are coerced to match the schema
   * (e.g. Instant -> timestamp-micros) so callers pass natural Java types.
   * Omit a column to leave it at its schema default (null for nullable fields).
   */
  public GenericRecord createRecord(TableDef table, String shardId, Map<String, Object> columns) {
    GenericRecordBuilder outer = new GenericRecordBuilder(table.schema);
    outer.set("tableName", table.tableName);
    outer.set("shardId", shardId);
    outer.set("primaryKeys", table.primaryKeys);

    Schema payloadSchema = unwrapNullable(table.schema.getField("payload").schema());
    GenericRecordBuilder payload = new GenericRecordBuilder(payloadSchema);
    columns.forEach((name, value) -> {
      Schema.Field field = payloadSchema.getField(name);
      if (field == null) {
            "Unknown column '" + name + "' for table " + table.tableName);
      }
      payload.set(name, coerce(field.schema(), value));
    });

    outer.set("payload", payload.build());
    return outer.build();
  }

  private static Schema unwrapNullable(Schema schema) {
    if (schema.getType() != Schema.Type.UNION) {
      return schema;
    }
    return schema.getTypes().stream()
        .filter(s -> s.getType() != Schema.Type.NULL)
        .findFirst()
        .orElse(schema);
  }

  /** Converts natural Java types to what GenericDatumWriter expects for logical types. */
  private static Object coerce(Schema fieldSchema, Object value) {
    if (value == null) {
      return null;
    }
    LogicalType logicalType = unwrapNullable(fieldSchema).getLogicalType();
    if (value instanceof Instant && logicalType instanceof LogicalTypes.TimestampMicros) {
      Instant i = (Instant) value;
      return i.getEpochSecond() * 1_000_000L + i.getNano() / 1000L;
    }
    return value; // extend here for timestamp-millis, date, etc. as needed
  }

  What the call sites look like

  // Users — note created_at passed as a plain Instant; conversion is automatic
  helper.createRecord(TableDef.USERS, shardId, Map.of(
      "user_id", 1L,
      "event_id", "e1",
      "full_name", "Alice",
      "age", 30,
      "created_at", Instant.now()));

  // AccountRoles
  helper.createRecord(TableDef.ACCOUNT_ROLES, shardId, Map.of(
      "role_id", 5,
      "role_name", "admin"));
```

This approach does lose compile time safety since it uses a generic payload but for test data generation I think its okay.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
```diff
@@ -0,0 +1,147 @@
...
+          dataFileWriter.append(record);
+        }
+      }
+
+      // Upload the temporary file to GCS
+      gcsClient.uploadArtifact(gcsFileName, tempFile.getAbsolutePath());
+    } finally {
+      // Clean up locally
+      tempFile.delete();
+    }
```
manitgupta: Maybe consider splitting it into a different utility that accepts the `List<GenericRecord>` to upload. This utility can restrict itself to generating the records. It will be easier that way I think.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCoreMatchingIT.java
```diff
@@ -0,0 +1,248 @@
...
+  public void setUp() throws IOException {
+    LOG.info("Setting up Spanner and BigQuery resources");
+    spannerResourceManager = setUpSpannerResourceManager();
+    bigQueryResourceManager = setUpBigQueryResourceManager();
+    LOG.info("Creating BigQuery dataset");
+    bigQueryResourceManager.createDataset(REGION);
+    LOG.info("BigQuery dataset created");
+    LOG.info("Creating Spanner DDL");
+    createSpannerDDL(spannerResourceManager, SPANNER_DDL_RESOURCE);
+    LOG.info("Spanner instance created");
```
manitgupta, line 62: Too many info logs seem redundant.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
aasthabharill: shifted this to the BaseIT class. The AvroSetupHelper is now mainly used to just get the GenericRecords

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
aasthabharill: implemented RecordBuilder in place of this - added justification to description

Have currently only added Instant handling in the newly added function convertToAvroFormat (equivalent of coerce from your code snippet) as the current schema only has that - have added comment to update it whenever a new table is introduced

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCoreMatchingIT.java
aasthabharill: removed some

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private static final Schema USERS_SCHEMA;
+  private static final Schema ACCOUNT_ROLES_SCHEMA;
+
+  static {
+    USERS_SCHEMA =
+        GCSSpannerDVITBase.getSchemaFromAvscFile("GCSSpannerDVAvroSetupHelper/users.avsc");
+    ACCOUNT_ROLES_SCHEMA =
+        GCSSpannerDVITBase.getSchemaFromAvscFile("GCSSpannerDVAvroSetupHelper/account_roles.avsc");
+  }
```
manitgupta: Does this still need to be in a static block? Can it not be inline in the enum block? Advantages or disadvantages?

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVTestAsserts.java
```diff
@@ -0,0 +1,101 @@
...
+      String tableName,
+      String status,
+      Long sourceRowCount,
+      Long destinationRowCount,
+      Long matchedRowCount,
+      Long mismatchRowCount) {}
+
+  public record MismatchedRecordDto(
+      String shardId, String schemaName, String tableName, String recordKey, String mismatchType) {}
+}
```
manitgupta, line 101: I think this is a bad idea. Creating these test DTOs here will cause duplication and subtle test errors when the main DTO and this DTO silently diverge over time.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVTestAsserts.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.beam.it.gcp.bigquery.BigQueryResourceManager;
+import org.apache.beam.it.gcp.bigquery.matchers.BigQueryAsserts;
+
+/**
+ * Test helper class for verifying BigQuery output from the gcs-spanner-dv pipeline.
+ *
+ * <p>This class provides strongly-typed Data Transfer Objects (DTOs) and assertion mappers to
+ * safely compare expected validation results against the actual rows written to BigQuery.
+ */
+public final class GCSSpannerDVTestAsserts {
```
manitgupta, line 36: I don't think this is a good idea. Having this sort of class breaks encapsulation of asserts within in a single test. What if two tests using this utility need slightly different behaviours from `assertValidationSummary` in the future? 

I think you should encapsulate the assertion of the values belonging to a test within a single test. This choice also has the ripple effect of creating those test DTO POJOs that I have commented on below.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCoreMatchingIT.java
```diff
@@ -0,0 +1,294 @@
...
+   *
+   * <ul>
+   *   <li><b>ValidationSummary</b>: One global row rolling up metrics correctly across all tables.
+   *   <li><b>TableValidationStats</b>: One row per table with accurate, independent status and
+   *       counts.
+   *   <li><b>MismatchedRecords</b>: Every individual discrepancy is explicitly logged and
+   *       categorized (`MISSING_IN_SOURCE`, `MISSING_IN_DESTINATION`) with the correct primary
+   *       keys.
+   * </ul>
+   */
```
manitgupta: Please validate if these long comments describing the test add value or not. Okay if the test is complicated to understand, not if the test is trivial.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVAvroSetupHelper.java
aasthabharill: changed

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCoreMatchingIT.java
aasthabharill: made the comments shorter

manitgupta (approved):

PR #4045, opened 2026-07-21
feat(spanner-to-sourcedb): Classify NOT_FOUND as retryable to support interleaved table replication

srozsnyai:
This PR classifies `NOT_FOUND` database errors as retryable in the Spanner-to-SourceDb template. 

When replicating interleaved tables, child record mutations can be delivered before their parent records. Currently, this results in a non-retryable `NOT_FOUND` error, causing replication to fail. 

By classifying `NOT_FOUND` as retryable, the pipeline will retry the child write until the parent record is successfully replicated.

**Changes:**
- Removed `ErrorCode.NOT_FOUND` from the non-retryable errors list in `SpannerToSourceDbExceptionClassifier`.
- Added `spannerToSpannerInterleavedOutOfOrderTest` in `SpannerToSpannerIT` to simulate out-of-order delivery (deleting a parent on destination, writing a child update on source, then writing the parent) and verified that replication succeeds after a retry.


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSpannerIT.java
```diff
@@ -396,4 +414,76 @@ private void assertReplication(
...
+        Mutation.newInsertOrUpdateBuilder("Child")
+            .set("id")
+            .to(1L)
+            .set("child_id")
+            .to(1L)
+            .set("description")
+            .to("Updated Child")
+            .build());
+
+    Thread.sleep(5000);
```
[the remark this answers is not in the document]
srozsnyai: I'd like to keep the Thread.sleep because I'd expect polling the DLQ would drastically slow down the test. In a streaming pipeline, writing to a GCS DLQ requires windowing, meaning we would have to wait a minute or more just for the file to be emitted. 

Sleep seems to be used across of other our integration tests and 5secs provides enough time to propagte recordsin an isolated environment without unnecessarily inflating the test duration.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/utils/SpannerToSourceDbExceptionClassifier.java
```diff
@@ -35,7 +35,6 @@ public class SpannerToSourceDbExceptionClassifier {
           ErrorCode.ALREADY_EXISTS,
           ErrorCode.OUT_OF_RANGE,
           ErrorCode.INVALID_ARGUMENT,
-          ErrorCode.NOT_FOUND,
           ErrorCode.FAILED_PRECONDITION,
```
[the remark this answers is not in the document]
srozsnyai: I'd like to push back for the following reasons:
1. Traversing the chain: Writes occur within Spanner's runInTransaction callback, which wraps the immediate exception in a SpannerException. Checking the immediate getCause() should be sufficient; recursion is unnecessary complexity.

2. Parsing error messages: Distinguishing schema vs. row-level NOT_FOUND errors requires string-matching error messages. This is a brittle as message strings are not a stable API contract.
   
3. Operational resilience: In CDC, treating a missing table as "permanent" forces data into a DLQ. Keeping it retryable should give the pipeline the opportunity to "self-heal" once an operator applies the missing DDL.

v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbIT/spanner-schema.sql
```diff
@@ -58,6 +58,18 @@ CREATE TABLE IF NOT EXISTS `generated_to_non_generated_column_table` (
...
+    id INT64 NOT NULL,
+    description STRING(25)
+) PRIMARY KEY(id);
+
+CREATE TABLE IF NOT EXISTS Child (
+    id INT64 NOT NULL,
+    child_id INT64 NOT NULL,
+    description STRING(25)
+) PRIMARY KEY(id, child_id),
+INTERLEAVE IN PARENT Parent ON DELETE CASCADE;
```
bharadwaj-aditya: This should be on delete no action.

When you on delete cascade, the child will get removed as well.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSpannerIT.java
```diff
@@ -396,4 +414,76 @@ private void assertReplication(
...
+    // 4. The pipeline will retry the Child write. We then restore the Parent,
+    //    allowing the retrying Child to finally succeed.
+    spannerResourceManager.write(
+        Mutation.newInsertOrUpdateBuilder("Parent")
+            .set("id")
+            .to(1L)
+            .set("description")
+            .to("Parent row")
+            .build());
+    spannerResourceManager.write(
```
bharadwaj-aditya: The first part of the test does not accomplish much. We can just do the 2nd part, which is create the child first and parent 2nd, and that will assert the DLQ flow

v2/spanner-to-sourcedb/src/test/resources/SpannerToSourceDbIT/spanner-schema.sql
srozsnyai: fixed - reworked schema and the actual test

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSpannerIT.java
srozsnyai: fixed - reworked schema and the actual test

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya (approved):
LGTM

PR #4054, opened 2026-07-23
[Data Validation] Add GCSSpannerDvSchemaIT

aasthabharill:
**Call out:**
- This test file originally had another test case: "Handle source Avro tables completely missing from Spanner DDL" but was removed due to the bug where if source table is not found in Spanner DDL, the validation pipeline crashes.
- Once that is fixed or we introduce Table Configuration parameter, ill add that test


v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/dofn/CreateSpannerReadOpsFn.java
```diff
@@ -35,10 +35,13 @@ public void processElement(ProcessContext c) {
...
         tableName -> {
+          String quote = ddl.dialect() == com.google.cloud.spanner.Dialect.POSTGRESQL ? "\"" : "`";
           // We encode the tableName in the query itself to push table information dynamically
           // and avoid table level stages.
           String query =
-              String.format("SELECT *, '%s' as __tableName__ FROM %s", tableName, tableName);
+              String.format(
+                  "SELECT *, '%s' as __tableName__ FROM %s%s%s",
+                  tableName, quote, tableName, quote);
           c.output(ReadOperation.create().withQuery(query));
```
manitgupta, line 46: PG dialect support verification is pending right? 

Add a TODO here to come back and see if there's a better way to generalize dialect specific changes that we need to introduce.

manitgupta (dismissed):
One minor comment, LGTM

manitgupta (approved):

PR #4055, opened 2026-07-23
[Data Validation] Add GCSSpannerDvRenamedColumnsIT and GCSSpannerDvRenamedTableIT

aasthabharill:
**Call out:**
`testWithoutSchemaMappingFile` testcase in `GCSSpannerDvRenamedTableIT` currently has the `@Ignore` tag due to the bug where if source table is not found in Spanner DDL, the validation pipeline crashes.

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVRenamedColumnsIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVRenamedColumnsIT.java
@@ -39,9 +39,8 @@ import org.slf4j.LoggerFactory;
  * Integration test for GCSSpannerDV pipeline covering column renaming scenarios.
  *
  * <p>This test validates that the pipeline correctly maps source columns to Spanner columns when
- * they have been renamed during the migration, given the proper configuration files (Overrides or
- * Session files). It also validates that mismatches are correctly identified if the configuration
- * file is omitted.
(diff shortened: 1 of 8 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVRenamedColumnsIT.java
```diff
@@ -0,0 +1,260 @@
...
+
+  @Before
+  public void setUp() throws IOException {
+    LOG.info("Setting up Spanner and BigQuery resources");
+    spannerResourceManager = setUpSpannerResourceManager();
+    bigQueryResourceManager = setUpBigQueryResourceManager();
+    bigQueryResourceManager.createDataset(REGION);
+    LOG.info("BigQuery dataset created");
+    createSpannerDDL(spannerResourceManager, SPANNER_DDL_RESOURCE);
+    LOG.info("Spanner instance created");
```
manitgupta: When all of these tests run will be a lot of similar logs from each test? Either add the test name or remove them? Don't think they are very useful. Validate other logs as well

aasthabharill: I saw the github run and the resourceManagers do log everything quite comprehensively already, so the logs i added were redundant - have removed all

The test name is mentioned before each log automatically as its source btw

manitgupta (approved):

PR #4057, opened 2026-07-24
Run Validation ITs using DirectRunner

aasthabharill:
b/539158138

The tests to be run with DirectRunner will run as a part of Integration Smoke tests. Dataflow runner tests will run as part of the Integration Test. 

- Resolves a bug where class annotations weren't being correctly read and consumed
- Resolves a bug where test method names aren't correctly read


manitgupta (commented):
Please also ensure that these smoke tests run on the Spanner specfic github action that we have and don't trigger the Java PR?

manitgupta (approved):
Let's gooo!!! 🚀🚀🚀🚀

PR #4062, opened 2026-07-27
[Data Validation] Add Custom Transformation ITs

aasthabharill:
**Call out:**
Codecov will constantly fail on this PR because of the Custom Transformation file (`CustomTransformationForDVIT`)- but there is no need to add a test file for this so please ignore it.

Using DataflowRunner for now as the Custom transformation gets cached in the DirectRunner and the subsequent tests start failing 

- **The Issue:** `GCSSpannerDVShardedIT` integration tests unexpectedly began failing with `NullPointerException`s in the CI/CD pipeline immediately after `GCSSpannerDVCustomTransformationIT` was introduced.
```
Caused by: java.lang.RuntimeException: Error mapping GenericRecord to ComparisonRecord: Cannot invoke "java.util.Map.entrySet()" because "transformedCols" is null
```
The NullPointerException is a seperate bug that will be fixed in the next PR, but for the test `GCSSpannerDVShardedIT`, it should not even have reached this part of the code as `customTransformer = null`. This points to a resource leak.
-   **The Root Cause (Test Isolation Leak):** `CustomTransformationImplFetcher` utilized a single `static` variable to cache the loaded transformer. Because Maven runs integration tests sequentially in the same JVM, the transformer loaded by `GCSSpannerDVCustomTransformationIT` persisted in memory and was erroneously injected into `GCSSpannerDVShardedIT`, even though the latter had an empty transformation configuration.

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCustomTransformationIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCustomTransformationIT.java
@@ -1,6 +1,22 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package com.google.cloud.teleport.v2.templates;
 
 import com.google.cloud.ByteArray;
+import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.teleport.metadata.DirectRunnerTest;
 import com.google.cloud.teleport.metadata.TemplateIntegrationTest;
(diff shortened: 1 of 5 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVITBase.java
```diff
@@ -200,6 +201,25 @@ protected LaunchInfo launchDataflowJob(
     return jobInfo;
   }
 
+  public void createAndUploadJarToGcs(String gcsPathPrefix)
+      throws IOException, InterruptedException {
+    ProcessBuilder processBuilder =
+        new ProcessBuilder("/bin/bash", "-c", "mvn clean package -DskipTests")
+            .directory(new File("../spanner-custom-shard"));
```
manitgupta: Method name seems generic ("upload JAR") but the file/dir is hardcoded?

manitgupta: This ProcessBuilder thing, can it fail on different versions of Linux since we are directly coding the command, is there a better way to join the args?

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCustomTransformationIT.java
```diff
@@ -0,0 +1,182 @@
...
+
+    LaunchInfo jobInfo =
+        launchDataflowJob(
+            options,
+            testName,
+            PROJECT,
+            spannerResourceManager,
+            bigQueryResourceManager.getDatasetId(),
+            gcsInputDirectory,
+            "GCSSpannerDVCustomTransformationIT/session.json",
```
manitgupta, line 148: Can you add a comment on why session file was needed?

manitgupta (commented):
Can you elaborate on the caching of the JAR on the direct runner? It wasn't clear what was happening.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVITBase.java
aasthabharill: Yes ProcessBuilder will work with all environments. And since we're only executing a mvn command and our tests run using that we can be sure this will work on the machines running the test.

I have removed `/bin/bash` though - this was a leftover shell wrapper in the original uploadJar method in SourcedbToSpannerBaseIT which we dont need anymore. 

I have modified the function name to be more specific, thanks for the callout

manitgupta (approved):
Can you please flag in a team meeting about this open work from modularization around support for shardIdColumn that was never implemented?

We need to close the gap there and remove dependency on session file for this use-case.

PR #4067, opened 2026-07-28
[Data Validation] Add Sharded Setup IT

aasthabharill:
Call outs:
- Modified `ValidationSummary` table `setTablesWithMismatches` to sort table names before writing as currently it's completely random and non-deterministic. This makes user readability difficult (and also test assertions). Spanner supports a max of 5000 tables, so a one-time sort should not be a problem.
- The second test checks duplicate scenario which is non-ideal IMO so have added a TODO item on me to investigate further but keeping the test in to check current behaviour anyway
- validationTestWithEmptyTables was flaky due to lack of wait time so added

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaIT.java
@@ -69,6 +69,9 @@ public class GCSSpannerDVSchemaIT extends GCSSpannerDVITBase {
         GCSSpannerDVAvroSetupHelper.TableDef.ACCOUNT_ROLES.schema,
         Collections.emptyList());
 
+    // Wait for the 15-second Spanner staleness bound so that tables created in setUp are visible
+    Thread.sleep(20000);
+
     // Launch Pipeline
     LaunchConfig.Builder options = LaunchConfig.builder(testName, specPath);
     LaunchInfo jobInfo =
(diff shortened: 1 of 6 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVShardedIT.java
```diff
@@ -0,0 +1,452 @@
...
+                .to(Timestamp.parseTimestamp(t2.toString()))
+                .build(),
+            Mutation.newInsertOrUpdateBuilder("AccountRoles")
+                .set("role_id")
+                .to(1L)
+                .set("role_name")
+                .to("ADMIN")
+                .build()));
+
+    Thread.sleep(20000);
```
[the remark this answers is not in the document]
aasthabharill: this is required for 15 second stale read - otherwise test will be flaky

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaIT.java
```diff
@@ -69,6 +69,9 @@ public void validationTestWithEmptyTables() throws Exception {
         GCSSpannerDVAvroSetupHelper.TableDef.ACCOUNT_ROLES.schema,
         Collections.emptyList());
 
+    // Wait for the 15-second Spanner staleness bound so that tables created in setUp are visible
+    Thread.sleep(20000);
```
manitgupta: Didn't get this? How is staleness bound related to DDL execution?

manitgupta (dismissed):

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaIT.java
aasthabharill: So the test fails on some runs with the error: `INVALID_ARGUMENT: Table not found: ORDER`

The validation job tries to query the spanner tables it found but fails on the SELECT query due to 15 second staleness - the table wasn't created back then

will make the comment clearer

manitgupta (approved):

PR #4072, opened 2026-07-29
Fix bug where Spanner STORING arrays were incorrectly parsed as VECTOR index keys

sm745052:
**Bug**
Previous code assumed any `ARRAY` column with a `NULL` sort order was a Vector search key. This broke standard indexes that used arrays in their `STORING` clause. Instead of treating the array as stored payload data, the export template parsed it as an index key and generated corrupted DDL.

```sql
CREATE INDEX MyIndex ON MyTable(StandardKey) STORING (MyStoredArray)
```

❌ **Bad DDL generated before this fix:**
```sql
-- The array was wrongly forced into the key list
CREATE INDEX MyIndex ON MyTable(StandardKey, MyStoredArray)
```
✅ Correct DDL generated after this fix:

```sql
CREATE INDEX MyIndex ON MyTable(StandardKey) STORING (MyStoredArray)
```

### DIFF
--- a/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
+++ b/v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
@@ -602,13 +602,10 @@ public class InformationSchemaScanner {
         boolean isStoring = resultSet.isNull(7);
         if (isStoring) {
           indexColumnsBuilder.storing();
-        } else if (spannerType != null
-            && (spannerType.equals(tokenlistType)
-                || spannerType.startsWith("ARRAY")
-                || spannerType.contains("vector length"))) {
-          // Tokenlist columns do not have ordering.
+        } else if (ordering == null) {
+          // Unordered keys (like Vector ARRAYs and Search TOKENLISTs) have no column ordering
           indexColumnsBuilder.none();
-        } else if (ordering != null) {
+        } else {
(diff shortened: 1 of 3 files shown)


v1/src/main/java/com/google/cloud/teleport/spanner/ddl/InformationSchemaScanner.java
```diff
@@ -599,14 +599,12 @@ private void listIndexColumns(
...
-            && (spannerType.equals(tokenlistType)
-                || spannerType.startsWith("ARRAY")
-                || spannerType.contains("vector length"))) {
-          indexColumnsBuilder.none();
-        } else if (ordering == null) {
+        boolean isStoring = resultSet.isNull(7);
+        if (isStoring) {
           indexColumnsBuilder.storing();
+        } else if (ordering == null) {
+          // Unordered keys (like Vector ARRAYs and Search TOKENLISTs) have no column ordering
```
shreyakhajanchi, line 606: can we add a test for this case as well?

sm745052: There actually already are dedicated integration tests for this case! 
`vectorIndexes()`, `pgVectorIndexes()`
they explicitly create `CREATE VECTOR INDEX` statements.

what we did not have coverage for was `STORED ARRAYS`

shreyakhajanchi (approved):

PR #4073, opened 2026-07-29
test(spanner): add tests for creating tables w/o primary keys

hengfengli:


bharadwaj-aditya (commented):
These changes look fine. But we also need to add a test in copydb test.

hengfengli:
> These changes look fine. But we also need to add a test in copydb test.

Good suggestion. Just added.

v1/src/test/java/com/google/cloud/teleport/spanner/AvroSchemaToDdlConverterTest.java
```diff
@@ -326,6 +326,43 @@ public void simple() {
...
+    AvroSchemaToDdlConverter converter = new AvroSchemaToDdlConverter();
+    Ddl ddl = converter.toDdl(Collections.singleton(schema));
+    assertThat(ddl.allTables(), hasSize(1));
+    assertThat(
+        ddl.prettyPrint(),
+        equalToCompressingWhiteSpace(
+            "CREATE TABLE `Users` ("
+                + " `rowid` INT64 NOT NULL GENERATED BY DEFAULT AS IDENTITY (BIT_REVERSED_POSITIVE) HIDDEN,"
+                + " `name` STRING(MAX),"
+                + " ) PRIMARY KEY (`rowid` ASC)"));
```
darshan-sj, line 363: This is equivalent to 

CREATE TABLE `Users` ( `name` STRING(MAX) );  ?

v1/src/test/java/com/google/cloud/teleport/spanner/AvroSchemaToDdlConverterTest.java
```diff
@@ -326,6 +326,43 @@ public void simple() {
...
+            + "    \"hidden\" : \"true\""
+            + "  }, {"
+            + "    \"name\" : \"name\","
+            + "    \"type\" : [ \"null\", \"string\" ],"
+            + "    \"sqlType\" : \"STRING(MAX)\""
+            + "  }],"
+            + "  \"googleStorage\" : \"CloudSpanner\","
+            + "  \"spannerPrimaryKey_0\" : \"`rowid` ASC\","
+            + "  \"googleFormatVersion\" : \"booleans\""
+            + "}";
```
darshan-sj, line 351: How did you form this json? Does the Avro export template create this after exporting from a table without PK ?

v1/src/test/resources/CopyDbIT-AllSchemaAndData-gsql.sql
```diff
@@ -203,4 +203,8 @@ REMOTE OPTIONS (endpoint='//aiplatform.googleapis.com/projects/%PROJECT_ID%/loca
 -- Functions (UDFs)
 CREATE FUNCTION `my_schema`.`MyCustomFunction`(`input_val` INT64) RETURNS INT64 AS (`input_val` * 2);
 
+-- Tables w/o PKs
+CREATE TABLE `MyTable_NoPks` (
+	`val` INT64
+);
```
darshan-sj, line 209: Any other edge cases we need to test that you can think of?

darshan-sj (approved):
There are only test changes in this PR. The Import Export templates handle the tables without PK already is it?

v1/src/test/java/com/google/cloud/teleport/spanner/AvroSchemaToDdlConverterTest.java
hengfengli: Yes, that's correct.

v1/src/test/java/com/google/cloud/teleport/spanner/AvroSchemaToDdlConverterTest.java
hengfengli: Yes, this is right. `CREATE TABLE Users ( name STRING(MAX) );` is supported now in Spanner. We automatically add a hidden identity column `rowid` as the primary key.

v1/src/test/resources/CopyDbIT-AllSchemaAndData-gsql.sql
hengfengli: This is the main case we would like to test here. This is basically a syntax sugar. It doesn't really add new features.

hengfengli:
> There are only test changes in this PR. The Import Export templates handle the tables without PK already is it?

Yes, the import/export templates already handle it since this is like a syntax sugar. It doesn't add new underlying features.

PR #4077, opened 2026-07-29
[Data Validation] Add Primary Key IT

aasthabharill:
**Call out:**
Three test cases:
1. Transformed primary keys (value change using custom transformation)
2. Modified primary key columns (where PK columns are different in SQL and Spanner)
3. Complex Avro datatype as PK column

The first one uses a transformation, and as seen in [this](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4062) PR, this needs to be run on **DataflowRunner** to avoid resource leaks. The rest of the test methods are functional and so have been configured to run on **DirectRunner**

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVPrimaryKeyIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVPrimaryKeyIT.java
@@ -1,0 +1,318 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates;
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.teleport.metadata.DirectRunnerTest;
+import com.google.cloud.teleport.metadata.TemplateIntegrationTest;
+import com.google.cloud.teleport.v2.spanner.migrations.transformation.CustomTransformation;
+import com.google.cloud.teleport.v2.templates.GCSSpannerDVAvroSetupHelper.RecordBuilder;
+import com.google.cloud.teleport.v2.templates.GCSSpannerDVAvroSetupHelper.TableDef;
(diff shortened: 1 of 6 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVPrimaryKeyIT.java
```diff
@@ -0,0 +1,317 @@
...
+    // full_name is changed to hex string by the custom transformation
+    spannerResourceManager.write(
+        Arrays.asList(
+            Mutation.newInsertOrUpdateBuilder("Users")
+                .set("user_id")
+                .to(1L)
+                .set("event_id")
+                .to("E1")
+                .set("full_name")
+                .to("416c696365") // Hex for "Alice"
```
manitgupta: Nit: Is a hex transform is happening lets just call the column name appropriately, why `full_name`? Maybe `user_id_in_hex` or something?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/avro/GenericRecordTypeConvertor.java
```diff
@@ -261,6 +261,9 @@ private Map<String, Value> populateCustomTransformations(
       return null;
     }
     Map<String, Object> transformedCols = migrationTransformationResponse.getResponseRow();
+    if (transformedCols == null) {
+      return result;
+    }
```
manitgupta, line 266: Why was this needed?

aasthabharill: The next line includes `transformedCols.entrySet()` which throws `NullPointerException` if transformedCols=null

Custom Transformation can return null in a lot of cases where a particular table is not transforming any rows and the code then fails with NullPointerException

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVPrimaryKeyIT.java
```diff
@@ -0,0 +1,317 @@
...
+                .set("age")
+                .to(30L)
+                .set("created_at")
+                .to(
+                    Timestamp.parseTimestamp(
+                        "2024-01-01T11:00:00Z")) // CustomTransformation adds 1 hour
+                .build()));
```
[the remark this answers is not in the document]
aasthabharill: we don't need to add this for tests that run using DataflowRunner as the test setup itself takes this long

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVPrimaryKeyIT.java
aasthabharill: because i was re-using the pre-defined USERS schema it had to be `full_name` as i didn't want to do column renaming in this test. 

But in that spirit, i have changed the logic to use a different table_name so there is a dedicated transformation for this test

manitgupta (approved):

PR #4079, opened 2026-07-29
 feat(mongodb-to-mongodb]: Read/Write Performance and Reliability Improvements

michaeltle-goog:
## Overview

  This PR refactors the mongodb-to-mongodb Dataflow template to improve read-phase scalability, write concurrency, and target cluster protection. It replaces legacy splitting operations with a dynamic, type-aware bounds generator and decouples read/write execution in high-throughput pipelines.

  Performance: Migrated 2.5M documents in ~12 minutes.

  ## Key Changes

  ### 1. Type-Aware Data-Driven Read Splitting (ReadSplitGenerator)

  Replaced splitVector and autoBucket with a deterministic, data-driven splitting algorithm that safely handles single-type and mixed-type _id collections:

- Active Type Discovery: Scans the source collection (find().limit(1)) to identify which BSON _id types are present (NUMBER, STRING, OBJECT_ID, OTHER).
- Per-Type Sampling ($sample): Executes an aggregation pipeline independently for each active BSON type:
```
[
      { "$match": { "_id": { "$type": "<type>" } } },
      { "$sample": { "size": sampleSize } },
      { "$project": { "_id": 1 } },
      { "$sort": { "_id": 1 } }
]
```
 (Sample size scales dynamically as max(1000, numSplits * 64)).
 
- Quantile Boundary Extraction: Divides the sorted sample keys into numSplits equal buckets to extract quantile boundaries (25th, 50th, 75th percentiles).
- Type-Scoped Range Filtering: Wraps every generated range query in an explicit BSON $type guard (e.g., _id >= min AND _id < max AND $type: "string"). This prevents MongoDB from silently dropping documents during cross-type range evaluations.
- Multi-Type Merging: Combines the type branches for each split index using an $or wrapper. If only one BSON type exists in the collection, the $or wrapper is automatically omitted.
- Deterministic Fallback: If a collection has too few documents for sampling or duplicate quantile boundaries are detected, it falls back to uniform partition splitting per type ($mod for numbers, hex prefixes for strings, timestamp prefixes for ObjectIDs).

### 2. Non-Blocking Asynchronous Writes (MongoDbTransforms)

- Decoupled Read and Write Loops: Refactored write execution so that MongoDB batch futures (CompletableFuture.runAsync) no longer block upstream reader threads.
- Bounded Concurrency: Capped parallel write tasks per worker (maxConcurrentAsyncWrites) to pipeline document reads without causing memory exhaustion or GC pressure.

### 3. Write-Path Rate Limiting

- Linear Ramp-Up Throttling: Added a configurable rate limiter (updateRateLimiterIfNeeded()) that linearly increases write throughput per worker from initialWriteRatePerWorker to maxWriteRatePerWorker over writeRateRampUpMinutes.

### 4. Logging

- Added more info logging to help customers understand/parse the Dataflow pipeline


v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -190,8 +180,67 @@ public interface Options extends JavascriptTextTransformer.JavascriptTextTransfo
...
     @TemplateParameter.Integer(
+        order = 11,
+        groupName = "Target",
+        optional = true,
+        description = "Initial Write Rate Per Worker",
+        helpText =
+            "Initial maximum documents/second written per worker thread during linear write rate"
+                + " ramp-up. Set to <= 0 to disable throttling.")
+    @Default.Integer(100)
+    Integer getInitialWriteRatePerWorker();
```
namita-l, line 191: we can make the initial ramp and the max per worker much higher: see https://screenshot-v2.corp.google.com/7aftn23jtkgeo

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -190,8 +180,67 @@ public interface Options extends JavascriptTextTransformer.JavascriptTextTransfo
...
+    String getDlqDirectory();
+
+    void setDlqDirectory(String value);
+
+    @TemplateParameter.Integer(
+        order = 16,
+        optional = true,
         description = "DLQ Max Retries",
         helpText = "Maximum number of times to retry events from DLQ.")
     @Default.Integer(3)
```
namita-l, line 246: is there a configuration for how many time we retry in memory before sending to the retryable dlq?

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -190,8 +180,67 @@ public interface Options extends JavascriptTextTransformer.JavascriptTextTransfo
...
+        helpText = "Number of discrete linear step increases over the ramp-up period.")
+    @Default.Integer(5)
+    Integer getWriteRateRampUpSteps();
+
+    void setWriteRateRampUpSteps(Integer value);
+
+    @TemplateParameter.Text(
         order = 15,
         optional = true,
+        description = "DLQ Directory",
```
namita-l, line 233: just want to confirm that we store the original doc / event in the dlq and not the post UDF transformed doc.

namita-l (dismissed):

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Yes we have  a `--maxWriteRetries` flag that retries with exponential backoff.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Yup we store it under "_original_document" in the DLQ json.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Done raised the write rate to be 5k -> 25k

namita-l (approved):

PR #4080, opened 2026-07-30
[Data Validation] Add GCSSpannerDVSchemaTransformationsIT

aasthabharill:
1. Source data type differs from the Spanner schema but can be handled automatically without requiring custom transformation
2. Columns present in the source are dropped in Spanner
3. New columns are added in Spanner

The third one uses a transformation, and as seen in [this](https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4062) PR, this needs to be run on **DataflowRunner** to avoid resource leaks. The rest of the test methods are functional and so have been configured to run on **DirectRunner**

Also fixes this issue: b/541972060
The other ITs using custom transformations were relying on the mvn build during IT execution. The mvn command i had introduced in the test was rebuilding the JAR and so any parallelly executing IT that relied on the JAR would fail due to it being deleted during this process

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
@@ -1,0 +1,419 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.templates;
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.teleport.metadata.DirectRunnerTest;
+import com.google.cloud.teleport.metadata.SkipDirectRunnerTest;
+import com.google.cloud.teleport.metadata.TemplateIntegrationTest;
+import com.google.cloud.teleport.v2.spanner.migrations.transformation.CustomTransformation;
+import com.google.cloud.teleport.v2.templates.GCSSpannerDVTestAsserts.MismatchedRecordDto;
+import com.google.cloud.teleport.v2.templates.GCSSpannerDVTestAsserts.TableValidationStatsDto;
+import com.google.cloud.teleport.v2.templates.GCSSpannerDVTestAsserts.ValidationSummaryDto;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.beam.it.common.PipelineLauncher.LaunchConfig;
+import org.apache.beam.it.common.PipelineLauncher.LaunchInfo;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Integration tests covering edge cases where Source and Spanner schemas have differences. Tests:
+ *
+ * <p>1. Data type differences (that don't require custom transformation)
+ *
+ * <p>2. Dropped columns in Spanner
+ *
+ * <p>3. Newly added columns in Spanner (populated via a custom transformation).
+ */
+@Category(TemplateIntegrationTest.class)
+@RunWith(JUnit4.class)
+@TemplateIntegrationTest(GCSSpannerDV.class)
+public class GCSSpannerDVSchemaTransformationsIT extends GCSSpannerDVITBase {
+
+  private static final String SPANNER_DDL_RESOURCE =
+      "GCSSpannerDVSchemaTransformationsIT/spanner-schema.sql";
+
+  @Before
+  public void setUp() throws IOException {
+    spannerResourceManager = setUpSpannerResourceManager();
+    bigQueryResourceManager = setUpBigQueryResourceManager();
+    bigQueryResourceManager.createDataset(REGION);
+    createSpannerDDL(spannerResourceManager, SPANNER_DDL_RESOURCE);
+  }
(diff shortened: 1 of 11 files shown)


v2/spanner-custom-shard/src/main/java/com/custom/CustomTransformationForDVIT.java
```diff
@@ -46,7 +46,7 @@ public MigrationTransformationResponse toSpannerRow(MigrationTransformationReque
       throws InvalidTransformationException {
 
     Map<String, Object> row = request.getRequestRow();
-    Map<String, Object> responseRow = new HashMap<>(row);
+    Map<String, Object> responseRow = new HashMap<>();
```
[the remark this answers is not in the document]
aasthabharill: already ran tests and this is not an issue

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
```diff
@@ -0,0 +1,409 @@
...
+                .set("role_id", 1)
+                .set("role_name", "1234")
+                .build(), // Match scenario
+            new GCSSpannerDVAvroSetupHelper.RecordBuilder(tableDef, null)
+                .set("role_id", 2)
+                .set("role_name", "5678")
+                .build() // Mismatch scenario
+            );
+
+    String gcsInputDirectory = getGcsPath("input");
```
manitgupta, line 83: I didn't notice this earlier but defnitely weird - its not clear what `getGcsPath` does and what the argument is - it takes a string and returns a string..consider renaming it what it does

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
```diff
@@ -0,0 +1,409 @@
...
+                /* schemaName= */ null,
+                /* tableName= */ "AccountRoles",
+                /* status= */ "MISMATCH",
+                /* sourceRowCount= */ 2L,
+                /* destinationRowCount= */ 2L,
+                /* matchedRowCount= */ 1L,
+                /* mismatchRowCount= */ 2L)));
+
+    // Verify role_name column mismatch (Source: 5678, Spanner: 9999)
+    // Note: gcs-spanner-dv currently lacks a MISMATCHED_VALUE category. Differing row values are
```
manitgupta: There is no way to achieve that (in the current implementation) since the hash of two mismatching records will be different and will materialize as two rows :)
Maybe consider removing can be confusing in the future.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
```diff
@@ -0,0 +1,409 @@
...
+    createSpannerDDL(spannerResourceManager, SPANNER_DDL_RESOURCE);
+  }
+
+  /**
+   * Tests data type change (STRING to INT64) of a non-PK column that doesn't require explicit
+   * handling via custom transformations and are automatically handled by the pipeline.
+   */
+  @Test
+  @Category({TemplateIntegrationTest.class, DirectRunnerTest.class})
+  public void testDataTypeDifference() throws Exception {
```
manitgupta, line 67: Maybe we've added tests in the similar category already, but worth calling out and discussing.

Does this **really** need to be an integration test? What are we hoping to catch here that can't be asserted in the PTransform or the hasher UT?

If two records don't match there hash will be different. I think because this runs on DirectRunner its cheap so it might be okay, but its worth thinking if this is just a duplicate.

Same goes for the tests added till now - just assess once. Deleting tests that are duplicate of testing done elsewhere is  a *good* thing.

aasthabharill: The intention behind this IT is to test the two steps that are most important for automatic data type casting:
1. Cast Avro datatype to Spanner Datatype
2. Hashing that includes datatype specific handling (specifically one that will differentiate a STRING "32" and INT 32)

While we can cover their specific behaviour in unit testing, the IT is to tie them together - if someone changes the logic in the first step, the second should break because of type-aware hashing. The major test case in this method is the MATCH scenario to ensure the casting is handled correctly leading to the same hash, the MISMATCH scenario is added just for sanity. 

One call out though is that the custom transformation IT does test data type casting when custom transformation is used for that column (STRING to BYTES). The functions used are _similar_ for type casting in custom transformation scenario vs no custom transformation scenario, but the code flow is different.

Since its a DirectRunner test and takes ~2 minutes i personally would incline on keeping this test. But i'm not super opinionated on this and if you think its overkill I can remove testDataTypeDifference()

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
aasthabharill: I wanted to add this comment to clarify that in case of data mismatch, getting two separate rows is the expected behaviour as it can be slightly un-intuitive.

Maybe I'll re-phrase this comment?

manitgupta: Yes sounds good, please rephrase

manitgupta (commented):
I see that Spanner PR has failed, please investigate and fix.

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
aasthabharill: getGcsPath is defined in `TemplateBaseClass` and is a commonly used function across a lot of DataflowTemplates so i dont think i should change it

The parameter "input" is just the name of a sub-directory where the test will place its input files within a temporary, isolated GCS bucket created for this specific test run. So in our case, it generates a GCS URI like `gs://my-test-bucket/test-run-12345/input`

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
aasthabharill: changed it for this test + others with a similar comment

manitgupta (dismissed):

darshan-sj (approved):
Looks good

PR #4084, opened 2026-07-30
build: update to latest import-spec version

fbiville:
This includes a serializability fix for key mappings.

This also includes a downgrade of the JSON schema library, which silently broke the Java 11 compatibility promise of import-spec (Dataflow works with Java 17+ so it's not an issue here, but it's an issue in other import-spec user projects).
This is basically a partial revert of f22c284695e6324bd6bb0ddea53b55f7b9eb5972


fbiville:
Failure seems to come from `PubSubToAvroIT`

nord-erik (approved):

damccorm (approved):
Thanks! Rerunning checks to try to get green before merging

PR #4090, opened 2026-07-31
[Data Validation] Add Wide Row IT for Max PK Columns

aasthabharill:
https://docs.cloud.google.com/spanner/quotas#tables

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVWideRowMax16KeyTableIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVWideRowMax16KeyTableIT.java
@@ -125,7 +125,9 @@ public class GCSSpannerDVWideRowMax16KeyTableIT extends GCSSpannerDVITBase {
     StringBuilder mismatchedPks = new StringBuilder("[");
     for (int i = 1; i <= 16; i++) {
       mismatchedPks.append("col_").append(i).append(":2");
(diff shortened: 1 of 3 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVWideRowMax16KeyTableIT.java
```diff
@@ -0,0 +1,151 @@
...
+    StringBuilder mismatchedPks = new StringBuilder("[");
+    for (int i = 1; i <= 16; i++) {
+      mismatchedPks.append("col_").append(i).append(":2");
+      if (i < 16) {
+        mismatchedPks.append(", ");
+      }
+    }
+    mismatchedPks.append("]");
+
+    GCSSpannerDVTestAsserts.assertMismatchedRecords(
```
darshan-sj, line 135: We don't need to assert Matched records?

darshan-sj (approved):
LGTM

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVWideRowMax16KeyTableIT.java
aasthabharill: we don't write those to the BigQuery tables - only MISMATCHED records are written; so there's no way to assert matched records except for deriving it from this table - its assumed if a record is not present in this table it was MATCHED

PR #4093, opened 2026-08-01
[Data Validation] Add wide row tests for max column and table name lengths

aasthabharill:


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVWideRowMaxColumnNameIT.java
```diff
@@ -0,0 +1,149 @@
+/*
```
shreyakhajanchi, line 1: Can we merge both the ITs into one schema with max table names and max column names?

shreyakhajanchi (approved):

PR #4094, opened 2026-08-01
[Data Validation] Add wide row tests for max string length

aasthabharill:
https://docs.cloud.google.com/spanner/quotas#tables


darshan-sj (approved):
Looks good

PR #4098, opened 2026-08-03
update gsutils ref

derrickaw:
1. Checked gsutils usage in repo and this was the only reference to it left.
2. gsutils is being deprecated and replaced with gcloud storage command.


derrickaw:
R: @Abacn

Abacn (approved):

PR #4116, opened 2026-08-06
[Data Validation] Duplicate Rows + Generated Spanner col IT

aasthabharill:
Created bug: b/543222130 for duplicate row

### DIFF
--- a/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
+++ b/v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
@@ -410,7 +410,12 @@ public class GCSSpannerDVSchemaTransformationsIT extends GCSSpannerDVITBase {
                 /* mismatchType= */ "MISSING_IN_SOURCE")));
   }
 
-  /** Tests pipeline resilience when columns present in the source are generated in Spanner. */
+  /**
+   * Tests pipeline resilience when columns present in the source are generated in Spanner. The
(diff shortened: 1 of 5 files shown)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
```diff
@@ -409,4 +409,112 @@ public void testAddedColumnInSpanner() throws Exception {
...
+            Arrays.asList("user_id", "event_id"));
+
+    Instant t1 = Instant.parse("2024-01-01T10:00:00Z");
+
+    List<GenericRecord> sourceRecords =
+        Arrays.asList(
+            new GCSSpannerDVAvroSetupHelper.RecordBuilder(tableDef, null)
+                .set("user_id", 1L)
+                .set("event_id", "E1")
+                .set("full_name", "1E1") // 1E1 matches Spanner generated value
```
shreyakhajanchi, line 435: what will happen if at source the column name is different?

shreyakhajanchi (dismissed):
Just one question, rest LGTM

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSchemaTransformationsIT.java
aasthabharill: thanks for the call out - the code does work as the generated column is filtered out from both source and spanner side (it was already handled it source, but not from spanner - which i fixed)

but since this is a good edge case, i have modified this test to check this on a renamed column

shreyakhajanchi (approved):

PR #4124, opened 2026-08-10
[Data Validation] Add Spanner failure injection test + Base class

aasthabharill:
Successful run on latest code: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/32717186875/job/97400820802

Dataflow job: https://pantheon.corp.google.com/dataflow/jobs/us-central1/2026-08-24_03_40_35-17616644887397284197;logsSeverity=ERROR;graphView=0?project=span-cloud-teleport-testing&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&e=PangolinKitchenLaunch::PangolinKitchenEnabled&mods=logs_tg_staging

Errors were correctly simulated but LESS than 4 times so pipeline can recover on retry


v2/failure-injection-policies/src/main/java/com/google/cloud/teleport/v2/failureinjection/InitialLimitedDurationErrorInjectionPolicy.java
```diff
@@ -37,12 +38,12 @@ public class InitialLimitedDurationErrorInjectionPolicy
       LoggerFactory.getLogger(InitialLimitedDurationErrorInjectionPolicy.class);
   private static final long serialVersionUID = 1L;
 
-  private Instant startTime;
+  private static volatile Instant startTime = null;
+  private static final AtomicLong callCount = new AtomicLong(0);
```
manitgupta, line 42: Why was `AtomicLong` needed?

v2/failure-injection-policies/src/main/java/com/google/cloud/teleport/v2/failureinjection/InitialLimitedDurationErrorInjectionPolicy.java
```diff
@@ -124,22 +125,20 @@ public InitialLimitedDurationErrorInjectionPolicy(JsonNode inputParameter, Clock
    */
   @Override
   public boolean shouldInjectionError() {
-    if (this.startTime == null) {
-      synchronized (this) {
-        if (this.startTime == null) {
-          this.startTime = Instant.now(clock);
+    if (startTime == null) {
```
manitgupta: Why is `this.` removed? Let's try to minimise unnecessary diffs.

v2/failure-injection-policies/src/main/java/com/google/cloud/teleport/v2/failureinjection/InitialLimitedDurationErrorInjectionPolicy.java
```diff
@@ -124,22 +125,20 @@ public InitialLimitedDurationErrorInjectionPolicy(JsonNode inputParameter, Clock
...
+      synchronized (InitialLimitedDurationErrorInjectionPolicy.class) {
+        if (startTime == null) {
+          startTime = Instant.now(clock);
```
manitgupta, line 131: We are synchronising on the class now instead of the object of the class? Do I understand that right? What was the issue happening earlier?

v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVSpannerReadFT.java
```diff
@@ -0,0 +1,163 @@
...
+
+  // Tests transient UNAVAILABLE errors on SpannerIO.readAll() to ensure native exponential backoff
+  // succeeds.
+  @Test
+  public void testTransientReadFailure() throws IOException, InterruptedException {
+    List<GenericRecord> records = new ArrayList<>();
+    List<Mutation> mutations = new ArrayList<>();
+    Instant now = Instant.now().truncatedTo(java.time.temporal.ChronoUnit.MILLIS);
+
+    // We insert 500 rows to ensure Dataflow has enough data to potentially split bundles
```
manitgupta, line 75: nit: replace 500 with NUM_RECORDS

v2/failure-injection-policies/src/main/java/com/google/cloud/teleport/v2/failureinjection/InitialLimitedDurationErrorInjectionPolicy.java
aasthabharill: Since we made the state static, all the threads on a Dataflow worker are now hitting the exact same counter. So if we were to use synchronized block, every thread would have to acquire the lock and wait. AtomicLong is better as `incrementAndGet()` allows threads to update the counter without blocking each other.

v2/failure-injection-policies/src/main/java/com/google/cloud/teleport/v2/failureinjection/InitialLimitedDurationErrorInjectionPolicy.java
aasthabharill: What was happening earlier: 
The original issue was that Beam actually deserializes multiple instances of this policy per worker (usually one per thread or bundle). When the variables weren't static, each instance got its own isolated timer. This meant the duration kept resetting per-thread instead of applying globally - so i kept getting the same fake exception for each retry ultimately failing the job.

I fixed that by making the state static so it's shared across the worker. But once the state is static, locking on this (the instance) is unsafe because threads using different instances would acquire different locks, leading to race conditions. Synchronizing on the Class object ensures all instances share the exact same lock to initialize the global timer safely.

aasthabharill (commented):
converting to draft as i triggered the failure injection tests using `InitialLimitedDurationErrorInjectionPolicy` and the bulk and reverse FTs are failing - will debug + add change to trigger FT github workflow on such PRs

PR #4127, opened 2026-08-10
Spanner to spanner transactional handling fixes

bharadwaj-aditya:
1. Corrected 2PC transaction handling in Spanner to Spanner replication.
2. Support cross project Spanner to Spanner replication
3. The special handling for Spanner to Spanner in a single transaction has been removed from the TransactionWriter. This will need to be re-added within the TargetSpanner DAO as a follow up

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
@@ -54,8 +54,7 @@ public class SpannerInformationSchemaScanner implements SourceSchemaScanner {
     this.spannerConfig = spannerConfig;
   }
 
-  @Override
-  public SourceSchema scan() {
+  public Ddl scanDdl() {
     SpannerAccessor accessor = SpannerAccessor.getOrCreate(spannerConfig);
     try {
       BatchClient batchClient = accessor.getBatchClient();
@@ -63,13 +62,19 @@ public class SpannerInformationSchemaScanner implements SourceSchemaScanner {
       InformationSchemaScanner scanner = new InformationSchemaScanner(txn);
       Ddl ddl = scanner.scan();
       LOG.info("Scanned Spanner schema for database '{}'", spannerConfig.getDatabaseId().get());
-      return convertDdlToSourceSchema(ddl);
+      return ddl;
     } finally {
       accessor.close();
     }
   }
 
-  SourceSchema convertDdlToSourceSchema(Ddl ddl) {
+  @Override
+  public SourceSchema scan() {
+    Ddl ddl = scanDdl();
+    return convertDdlToSourceSchema(ddl);
+  }
+
+  public SourceSchema convertDdlToSourceSchema(Ddl ddl) {
     Map<String, SourceTable> tables = new HashMap<>();
     for (Table spannerTable : ddl.allTables()) {
       SourceTable sourceTable = convertTable(spannerTable);
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
@@ -83,6 +83,7 @@ public class SpannerInformationSchemaScanner implements SourceSchemaScanner {
     return SourceSchema.builder(sourceType)
         .databaseName(spannerConfig.getDatabaseId().get())
         .tables(ImmutableMap.copyOf(tables))
+        .rawDdl(ddl)
         .build();
   }
 
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
@@ -27,10 +27,7 @@ import com.google.cloud.teleport.v2.spanner.ddl.Table;
 import com.google.cloud.teleport.v2.spanner.migrations.schema.ISchemaMapper;
 import com.google.cloud.teleport.v2.spanner.sourceddl.SourceTable;
 import com.google.cloud.teleport.v2.spanner.type.Type;
-import com.google.cloud.teleport.v2.templates.constants.Constants;
 import com.google.cloud.teleport.v2.templates.dbutils.dml.IDMLGenerator;
-import com.google.cloud.teleport.v2.templates.dbutils.processor.ISpToSrcSourceConnector;
-import com.google.cloud.teleport.v2.templates.dbutils.processor.SourceProcessorFactory;
 import com.google.cloud.teleport.v2.templates.exceptions.InvalidDMLGenerationException;
 import com.google.cloud.teleport.v2.templates.models.DMLGeneratorRequest;
 import com.google.cloud.teleport.v2.templates.models.DMLGeneratorResponse;
@@ -75,7 +72,13 @@ public class SpannerDMLGenerator implements IDMLGenerator {
       throw new InvalidDMLGenerationException("Spanner DDL must not be null.");
     }
     if (targetSpannerSchema == null) {
-      throw new InvalidDMLGenerationException("SourceSchema must not be null.");
+      throw new InvalidDMLGenerationException("target spanner schema could not be fetched.");
+    }
+
+    Ddl targetDdl =
+        (targetSpannerSchema.rawDdl() instanceof Ddl) ? (Ddl) targetSpannerSchema.rawDdl() : null;
+    if (targetDdl == null) {
+      throw new InvalidDMLGenerationException("target spanner ddl could not be fetched.");
     }
 
     Table origSpannerTable = originalSpannerDdl.table(origSpannerTableName);
@@ -109,10 +112,10 @@ public class SpannerDMLGenerator implements IDMLGenerator {
     String modType = request.getModType();
     if ("INSERT".equals(modType) || "UPDATE".equals(modType)) {
       return buildUpsertMutation(
-          origSpannerTable, targetSpannerTable, schemaMapper, request, targetTableName);
+          origSpannerTable, targetSpannerTable, schemaMapper, request, targetTableName, targetDdl);
     } else if ("DELETE".equals(modType)) {
       return buildDeleteMutation(
-          origSpannerTable, targetSpannerTable, schemaMapper, request, targetTableName);
+          origSpannerTable, targetSpannerTable, schemaMapper, request, targetTableName, targetDdl);
     } else {
       throw new InvalidDMLGenerationException(
           "Unsupported modType '" + modType + "' for table " + origSpannerTableName);
@@ -124,7 +127,8 @@ public class SpannerDMLGenerator implements IDMLGenerator {
       SourceTable targetSpannerTable,
       ISchemaMapper schemaMapper,
       DMLGeneratorRequest request,
-      String targetTableName) {
+      String targetTableName,
+      Ddl targetDdl) {
 
     Mutation.WriteBuilder builder = Mutation.newInsertOrUpdateBuilder(targetTableName);
     JSONObject newValuesJson = request.getNewValuesJson();
@@ -178,6 +182,7 @@ public class SpannerDMLGenerator implements IDMLGenerator {
     Key primaryKey =
         buildTargetPrimaryKey(
             origSpannerTable,
+            targetDdl,
             targetSpannerTable,
             schemaMapper,
             newValuesJson,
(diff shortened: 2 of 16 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
```diff
@@ -169,7 +169,24 @@ private static DMLGeneratorResponse buildUpsertMutation(
...
+          && request.getCustomTransformationResponse().containsKey(sourceColName)) {
+        customVal = request.getCustomTransformationResponse().get(sourceColName);
+      }
+      JSONObject valuesJson = keyValuesJson.has(sourceColName) ? keyValuesJson : newValuesJson;
+      if (customVal != null) {
+        appendCustomKeyComponent(keyBuilder, sourceCol, customVal);
+      } else if (valuesJson.has(sourceColName)) {
+        appendKeyComponent(keyBuilder, sourceCol, valuesJson, sourceColName);
+      }
+    }
```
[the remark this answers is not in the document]
bharadwaj-aditya: the keyValuesJson will always have the primary key.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
```diff
@@ -32,6 +33,7 @@
 public class SpannerSpToSrcSourceConnector implements ISpToSrcSourceConnector {
 
   private final IConnectionHelper connectionHelper;
+  private Ddl targetDdl;
```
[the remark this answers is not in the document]
bharadwaj-aditya: The migration expects the ddl to be the same so this is not valid.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
```diff
@@ -64,9 +74,24 @@ String getConnectionUrl(Shard shard) {
...
+    if (targetDdl == null) {
+      SpannerConfig targetSpannerConfig =
+          SpannerConfig.create()
+              .withProjectId(spannerShard.getProjectId())
+              .withInstanceId(spannerShard.getInstanceId())
+              .withDatabaseId(spannerShard.getDatabaseId());
+      targetDdl = new SpannerInformationSchemaScanner(targetSpannerConfig).scanDdl();
+    }
+    return targetDdl;
   }
```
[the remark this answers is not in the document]
bharadwaj-aditya: the targetDdl needs to be the same.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
```diff
@@ -120,7 +129,10 @@ public SourceSchema getInformationSchema(List<Shard> shards) throws Exception {
             .withProjectId(spannerShard.getProjectId())
             .withInstanceId(spannerShard.getInstanceId())
             .withDatabaseId(spannerShard.getDatabaseId());
-    return new SpannerInformationSchemaScanner(targetSpannerConfig).scan();
+    SpannerInformationSchemaScanner scanner =
+        new SpannerInformationSchemaScanner(targetSpannerConfig);
+    targetDdl = scanner.scanDdl();
+    return scanner.convertDdlToSourceSchema(targetDdl);
```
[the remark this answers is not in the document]
bharadwaj-aditya: closing same as above.

shreyakhajanchi (commented):
Can you please list the fixes made? It will help review the PR better.

bharadwaj-aditya:
> Can you please list the fixes made? It will help review the PR better.

Updated the PR description

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
```diff
@@ -64,9 +74,24 @@ String getConnectionUrl(Shard shard) {
...
+  private synchronized Ddl checkAndInitTargetDdl(SpannerShard spannerShard) {
+    if (targetDdl == null) {
+      SpannerConfig targetSpannerConfig =
+          SpannerConfig.create()
+              .withProjectId(spannerShard.getProjectId())
+              .withInstanceId(spannerShard.getInstanceId())
+              .withDatabaseId(spannerShard.getDatabaseId());
+      targetDdl = new SpannerInformationSchemaScanner(targetSpannerConfig).scanDdl();
+    }
+    return targetDdl;
```
shreyakhajanchi: We are not using the return value, maybe make the method void ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
```diff
@@ -94,22 +119,6 @@ public void validate(List<Shard> shards, PipelineOptions options) throws Excepti
       throw new IllegalArgumentException(
           "Expected SpannerShard but got: " + shards.get(0).getClass());
     }
```
shreyakhajanchi, line 127: We should add an IT for the case where metadata db is different from the target db

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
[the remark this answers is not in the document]
shreyakhajanchi: what about the case where PK column in target spanner instance is a new column all together and is not in custom transformation as well, shouldn't we throw exception in that case ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
```diff
@@ -169,7 +169,24 @@ private static DMLGeneratorResponse buildUpsertMutation(
       }
     }
 
-    return new SpannerMutationResponse(builder.build());
+    Key.Builder keyBuilder = Key.newBuilder();
```
shreyakhajanchi, line 199: We should cover this assertion in UT

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerTargetDao.java
```diff
@@ -67,35 +83,47 @@ public void write(
...
+        .allowNestedTransaction()
+        .run(
+            (TransactionRunner.TransactionCallable<Void>)
+                mainTxn -> {
+                  if (targetDdl != null && primaryKey != null) {
+                    readDataTableRowWithExclusiveLock(
+                        mainTxn, mutation.getTable(), primaryKey, targetDdl);
+                  }
+                  if (transactionalCheck != null) {
+                    transactionalCheck.check();
```
shreyakhajanchi: there is an case here with Spanner's TransactionRunner that will cause Dataflow workers to stall in an infinite loop if the outer shadowTransaction aborts.

transactionalCheck.check() executes a read using the outer shadowTransaction. If the Spanner backend aborts shadowTransaction (e.g., due to lock contention or timeout), this check will throw a SpannerException with ErrorCode.ABORTED.

Because this exception is thrown inside mainTxn's closure, mainTxn's TransactionRunner intercepts it, assumes that mainTxn was aborted, and automatically retries mainTxn. On the retry, it calls check() again, hits the aborted shadowTransaction, throws the same exception, and loops infinitely.

To fix this, we need to wrap the exception so it isn't swallowed by the inner TransactionRunner. This forces mainTxn to fail immediately and allows the error to bubble up to the outer shadowTransaction runner where it can be properly retried:

```
if (transactionalCheck != null) {
  try {
    transactionalCheck.check();
  } catch (SpannerException e) {
    if (e.getErrorCode() == com.google.cloud.spanner.ErrorCode.ABORTED) {
      // Wrap the exception to prevent mainTxn from infinitely retrying an aborted outer transaction.
      throw new IllegalStateException("Shadow transaction aborted during check", e);
    }
    throw e;
  }
}
```

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
bharadwaj-aditya: done

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerTargetDao.java
bharadwaj-aditya: this has been fixed.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerSpToSrcSourceConnector.java
bharadwaj-aditya: added

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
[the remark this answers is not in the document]
bharadwaj-aditya: fixed

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
bharadwaj-aditya: added

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerTargetDao.java
```diff
@@ -67,35 +84,57 @@ public void write(
...
-    Mutation mutation = ((SpannerMutationResponse) dmlGeneratorResponse).getMutation();
-    client.writeAtLeastOnce(ImmutableList.of(mutation));
-    LOG.debug("Successfully wrote mutation via SpannerTargetDao for key: {}", connectionKey);
+    Mutation mutation = spannerMutationResponse.getMutation();
+    Key primaryKey = spannerMutationResponse.getPrimaryKey();
+
+    // TODO - optimize this flow to avoid the nested transaction if the shadow transaction
+    // (in the transaction check) and the main transaction are on the same database
+    client
+        .readWriteTransaction(Options.priority(RpcPriority.HIGH))
```
manitgupta, line 93: This is supposed to be hardcoded to always be high?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
```diff
@@ -133,95 +138,139 @@ private static DMLGeneratorResponse buildUpsertMutation(
...
+    Ddl targetDdl = null;
+    try {
+      ISpToSrcSourceConnector sourceConnector =
+          SourceProcessorFactory.getSource(Constants.SOURCE_SPANNER);
+      if (sourceConnector instanceof SpannerSpToSrcSourceConnector) {
+        targetDdl = ((SpannerSpToSrcSourceConnector) sourceConnector).getTargetDdl();
+      }
+    } catch (Exception e) {
+      LOG.warn("could not fetch spanner source", e);
+    }
```
manitgupta: If the try block fails in fetching the targetDdl this won't fail the pipeline. Given that this being called per record, is that expected? Should we be raising an exception here?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
```diff
@@ -54,22 +54,27 @@ public SpannerInformationSchemaScanner(SpannerConfig spannerConfig) {
...
       accessor.close();
     }
   }
 
-  SourceSchema convertDdlToSourceSchema(Ddl ddl) {
+  @Override
+  public SourceSchema scan() {
+    Ddl ddl = scanDdl();
+    return convertDdlToSourceSchema(ddl);
+  }
```
manitgupta, line 75: I think we need the raw `Ddl` so we had to make `scanDdl` a public method. I think we can consider making something like `rawDdl` a member field of SourceSchema to expose it..that will allow us to keep the existing `scan()` based pattern..

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
```diff
@@ -133,95 +138,139 @@ private static DMLGeneratorResponse buildUpsertMutation(
...
       if (customVal != null) {
-        appendCustomKeyComponent(keyBuilder, sourceCol, customVal);
+        appendCustomKeyComponent(keyBuilder, targetCol, customVal);
+      } else if (valuesJson.has(targetColName)) {
+        appendCustomKeyComponent(keyBuilder, targetCol, valuesJson.get(targetColName));
       } else {
-        appendKeyComponent(keyBuilder, sourceCol, valuesJson, sourceColName);
+        LOG.warn("Primary key column '{}' could not be resolved.", targetColName);
+        throw new InvalidDMLGenerationException(
+            "Primary key column '" + targetColName + "' could not be resolved.");
```
manitgupta: I think this may fail when the target PK is renamed. `keyValuesJson` would be keyed by the original PK names right? We might need to verify the same for custom transformation as well..

I would recommend adding a UT where the PK is being renamed and checking if this works as expected..

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerTargetDao.java
bharadwaj-aditya: Ideally this should come from the pipeline config. Changing this is marked as a TODO so ignoring this comment for now.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/SpannerInformationSchemaScanner.java
bharadwaj-aditya: Agreed. This is being created with Autovalue, so it is not done as of now. Will take this up as a part of passing the targetDdl flow in a follow up.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
bharadwaj-aditya: this flow will never be reached. But added the check for hygiene.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/spanner/SpannerDMLGenerator.java
bharadwaj-aditya: added

manitgupta (dismissed):

manitgupta (approved):

PR #4133, opened 2026-08-12
Add Bulk support for ORACLE

dhwanilpatel:
Add support of ORACLE source in BULK Template using source Skill.

Implemented all ITs similar to MySQL using test skill and all ITs are passing in local. Will raise separate PR of it.


v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+  public String getBoundaryQuery(
+      String tableName, ImmutableList<String> partitionColumns, String colName) {
+    return addWhereClause(
+        "SELECT MIN(" + quote(colName) + "), MAX(" + quote(colName) + ") FROM " + quote(tableName),
+        partitionColumns);
+  }
+
+  @Override
+  public boolean checkForTimeout(SQLException exception) {
+    return false; // Simplified
```
bharadwaj-aditya: We should check for timeout exceptions here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  @Override
+  public String getCollationsOrderQuery(String dbCharset, String dbCollation, boolean padSpace) {
+    return "SELECT 1 FROM DUAL WHERE 1=0";
```
bharadwaj-aditya, line 115: can you double check this query ? The where clause looks odd.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+      try (ResultSet rs = metaData.getTables(null, schemaPattern, null, new String[] {"TABLE"})) {
+        while (rs.next()) {
+          tablesBuilder.add(rs.getString("TABLE_NAME"));
+        }
+      }
+    } catch (SQLException e) {
+      throw new SchemaDiscoveryException(e);
+    }
+    ImmutableList<String> tables = tablesBuilder.build();
+    org.slf4j.LoggerFactory.getLogger(OracleDialectAdapter.class)
```
bharadwaj-aditya: Declare the logger as a constant at the start of the class.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+    return tables;
+  }
+
+  @Override
+  public ImmutableMap<String, ImmutableMap<String, SourceColumnType>> discoverTableSchema(
+      javax.sql.DataSource dataSource,
+      JdbcSchemaReference schemaReference,
+      ImmutableList<String> tables)
+      throws SchemaDiscoveryException, RetriableSchemaDiscoveryException {
+    Map<String, ImmutableMap.Builder<String, SourceColumnType>> builders = new HashMap<>();
```
bharadwaj-aditya: nit: this can just be built in one lambda, rather than creating and then running a forEach

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+      for (String table : tables) {
+        try (ResultSet rs = metaData.getColumns(null, schemaPattern, table, null)) {
+          while (rs.next()) {
+            String colName = rs.getString("COLUMN_NAME");
+            String typeName = rs.getString("TYPE_NAME");
+            if (typeName != null) {
+              typeName = typeName.replaceAll("\\([0-9]+\\)", "");
+            }
+            long colSize = rs.getLong("COLUMN_SIZE");
+            long descinalDigits = rs.getLong("DECIMAL_DIGITS"); // May be 0 if null
```
bharadwaj-aditya: nit: naming - decimalDigits

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,237 @@
...
+                    .setColumnName(colName)
+                    .setIsPrimary(true)
+                    .setIsUnique(true)
+                    .setOrdinalPosition(seq)
+                    .setIndexName(pkName != null ? pkName : "PRIMARY")
+                    .setIndexType(type)
+                    .setColumnTypeName("");
+
+            if (type == SourceColumnIndexInfo.IndexType.STRING) {
+              com.google.cloud.teleport.v2.reader.io.jdbc.uniformsplitter.stringmapper
```
bharadwaj-aditya: can you declare this in an import and just use the CollationReference here ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,198 @@
...
+                infoBuilder.setCollationReference(emptyCollation);
+                infoBuilder.setStringMaxLength(200);
+            }
+
+            builders.get(table).add(infoBuilder.build());
+          }
+        }
+      }
+    } catch (SQLException e) {
+    }
```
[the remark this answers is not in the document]
bharadwaj-aditya: Please check this.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/iowrapper/config/defaults/OracleConfigDefaults.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.cloud.teleport.v2.source.oracle.reader.io.jdbc.rowmapper.provider.OracleJdbcValueMappings;
+import com.google.common.collect.ImmutableList;
+import org.apache.beam.sdk.util.FluentBackoff;
+import org.joda.time.Duration;
+
+public class OracleConfigDefaults {
+  public static final DialectAdapter DEFAULT_ORACLE_DIALECT_ADAPTER = new OracleDialectAdapter();
+  public static final JdbcValueMappingsProvider DEFAULT_ORACLE_VALUE_MAPPING_PROVIDER =
+      new OracleJdbcValueMappings();
+  public static final Long DEFAULT_ORACLE_MAX_CONNECTIONS = 160L;
```
bharadwaj-aditya, line 30: Any reasoning for this default ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
```diff
@@ -0,0 +1,132 @@
...
+      (value, schema) -> {
+        if (value == null) {
+          return null;
+        }
+        return ByteBuffer.wrap(value);
+      };
+
+  private static final JdbcMappings JDBC_MAPPINGS =
+      JdbcMappings.builder()
+          .put("VARCHAR", ResultSet::getString, valuePassThrough, 1024)
```
bharadwaj-aditya: please cross verify these sizes

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/OracleSrcToSpSourceConnector.java
```diff
@@ -0,0 +1,137 @@
...
+      int port,
+      String dbName,
+      String connectionProperties,
+      String namespace,
+      Integer fetchSize) {
+    String url = "jdbc:oracle:thin:@" + host + ":" + port + "/" + dbName;
+    if (StringUtils.isNotBlank(connectionProperties)) {
+      url = url + "?" + connectionProperties;
+    }
+    // Set explicit fetchSize via options builder mapping...
```
bharadwaj-aditya: check if fetchSize is relevant here

v2/sourcedb-to-spanner/pom.xml
```diff
@@ -247,6 +247,12 @@
             <scope>test</scope>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
bharadwaj-aditya, line 250: please declare this as a compile time requirement, similar to mysql and pg above

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
```diff
@@ -112,7 +113,7 @@ public JsonNode transformChangeEventViaSessionFile(JsonNode changeEvent) {
   }
 
   public String getShardId(JsonNode changeEvent) {
-    if (!MYSQL_SOURCE_TYPE.equals(this.sourceType)
+    if ((!MYSQL_SOURCE_TYPE.equals(this.sourceType) && !ORACLE_SOURCE_TYPE.equals(this.sourceType))
```
bharadwaj-aditya, line 116: why is PG not a part of this check. Please cross verify if this is specific to mysql

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/SourceType.java
```diff
@@ -25,6 +25,7 @@ public enum SourceType {
 
   /** MySQL source database type. */
   MYSQL(new String[] {"mysql"}),
+  ORACLE(new String[] {"oracle"}),
```
bharadwaj-aditya: nit: Please place under PG like all the other declarations for uniformity

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertorTest.java
```diff
@@ -433,7 +433,7 @@ public void getShardIdInvalidSource() {
     ShardingContext shardingContext = getShardingContext();
     ChangeEventSessionConvertor changeEventSessionConvertor =
         new ChangeEventSessionConvertor(
-            schema, null, new TransformationContext(), shardingContext, "oracle", false);
+            schema, null, new TransformationContext(), shardingContext, "sqlserver", false);
```
bharadwaj-aditya: If this test is for invalid source, please use some other name that will not fail when we add sqlserver

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/OracleSrcToSpSourceConnector.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Override
+  public String getJdbcUrl(String host, int port, String dbName, String connectionProperties, String namespace, Integer fetchSize) {
+    String url = "jdbc:oracle:thin:@" + host + ":" + port + "/" + dbName;
+    if (StringUtils.isNotBlank(connectionProperties)) {
+      url = url + "?" + connectionProperties;
+    }
+    // Set explicit fetchSize via options builder mapping...
+    return url;
+  }
```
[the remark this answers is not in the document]
dhwanilpatel: Good Catch. I will update the JDBC URL to include "//".

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,198 @@
...
+  public ImmutableList<String> discoverTables(
+      javax.sql.DataSource dataSource, JdbcSchemaReference sourceSchemaReference)
+      throws SchemaDiscoveryException, RetriableSchemaDiscoveryException {
+    ImmutableList.Builder<String> tablesBuilder = ImmutableList.builder();
+    try (Connection conn = dataSource.getConnection()) {
+      DatabaseMetaData metaData = conn.getMetaData();
+      String schemaPattern = sourceSchemaReference.namespace() != null ? sourceSchemaReference.namespace() : metaData.getUserName();
+      if (schemaPattern != null) {
+          schemaPattern = null;
+      }
```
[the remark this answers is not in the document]
dhwanilpatel: If block indeed sound like bug, but I think we shouldn't move it to uppercase, as it might break with case sensitive table/index names. 

I will remove if block and pass the schemaPattern as it is. I.e. if user provided sourceSchema then will pass it, if not provided will pass metadata.getUserName so it will scan for the tables for provided user only.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,198 @@
...
+      throws SchemaDiscoveryException, RetriableSchemaDiscoveryException {
+    Map<String, ImmutableMap.Builder<String, SourceColumnType>> builders = new HashMap<>();
+    tables.forEach(table -> builders.put(table, ImmutableMap.builder()));
+    
+    try (Connection conn = dataSource.getConnection()) {
+      DatabaseMetaData metaData = conn.getMetaData();
+      String schemaPattern = schemaReference.namespace() != null ? schemaReference.namespace() : metaData.getUserName();
+      if (schemaPattern != null) {
+          schemaPattern = null;
+      }
```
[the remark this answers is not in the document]
dhwanilpatel: Similar change as https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4133#discussion_r3772434622

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,198 @@
...
+        try (ResultSet rs = metaData.getPrimaryKeys(null, null, table)) {
+          while (rs.next()) {
+            String colName = rs.getString("COLUMN_NAME");
+            String pkName = rs.getString("PK_NAME");
+            long seq = rs.getShort("KEY_SEQ");
+            
+            SourceColumnIndexInfo.IndexType type = SourceColumnIndexInfo.IndexType.OTHER;
+            try (ResultSet crs = metaData.getColumns(null, null, table, colName)) {
+                if (crs.next()) {
+                    String typeName = crs.getString("TYPE_NAME");
```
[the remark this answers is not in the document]
dhwanilpatel: similar to https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4133#discussion_r3772434622

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
```diff
@@ -0,0 +1,80 @@
...
+  public SourceSchema scan() {
+    Map<String, SourceTable> tablesMap = new HashMap<>();
+    SourceSchema.Builder builder = SourceSchema.builder(SourceDatabaseType.ORACLE).databaseName("ORACLE");
+    try {
+      DatabaseMetaData metaData = connection.getMetaData();
+      String schemaPattern = null;
+      try {
+        schemaPattern = connection.getSchema();
+      } catch (Exception e) {
+      }
```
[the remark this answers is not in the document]
dhwanilpatel: similar to https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4133#discussion_r3772434622

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
```diff
@@ -0,0 +1,100 @@
...
+      
+  private static final ResultSetValueMapper<java.sql.Timestamp> timestampTzToAvroRecord =
+      (value, schema) -> {
+          long micros = TimeUnit.SECONDS.toMicros(value.getTime() / 1000)
+              + TimeUnit.NANOSECONDS.toMicros(value.getNanos());
+          return new GenericRecordBuilder(TimeStampTz.SCHEMA)
+              .set(TimeStampTz.TIMESTAMP_FIELD_NAME, micros)
+              .set(TimeStampTz.OFFSET_FIELD_NAME, 0)
+              .build();
+      };private static final ResultSetValueMapper<java.math.BigDecimal> bigDecimalToByteArray =
```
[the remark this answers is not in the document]
dhwanilpatel: Fixed.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Updated

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Fixed.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/iowrapper/config/defaults/OracleConfigDefaults.java
dhwanilpatel: It seems like Skill got this default [160] from MySQL/PostgreSQL Config as both has the default of 160 max connection limit.

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertorTest.java
dhwanilpatel: Changed to "mock_db"

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/source/config/SourceType.java
dhwanilpatel: Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/OracleSrcToSpSourceConnector.java
dhwanilpatel: Yes it should be added but looks like skill missed this. 

Added fetchsize parameter in JDBC url.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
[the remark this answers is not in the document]
dhwanilpatel: Added logger.

v2/sourcedb-to-spanner/pom.xml
dhwanilpatel: removed runtime scope.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Added check for timeout execption here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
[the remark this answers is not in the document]
dhwanilpatel: @bharadwaj-aditya is it okay to pass userName if provided sourceSchema is null?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/iowrapper/config/defaults/OracleConfigDefaults.java
dhwanilpatel: Default connection limit for Mysql [151] and PG [100] are less than default connection limit for Oracle [300], So this default connection limit of 160 should be fine for Oracle.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: The WHERE `1=0` clause intentionally returns an empty result set. Because Oracle lacks a simple, unified collation logic like MySQL, returning an empty result safely forces Dataflow to fall back to its default partitioner logic.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
dhwanilpatel: Datastream uses the same JSON structure for both MySQL and Oracle, so this Shard ID parsing logic needs to be implemented for both. For PG it uses a different structure, so that's why it would have intentionally excluded from here to avoid parsing errors.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
dhwanilpatel: Cross verified size from documentation and updated numbers. (https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html)

Few points to highlight
* VARCHAR/NCHAR and other char type has two config : STANDARD (4000) and EXTENDED(32767) : we have used standard number only.
* FOR BLOB/CLOB and other blob type: max is 4 gb: We have mapped it to INTEGER.MAX_VALUE similar to LONGTEXT on MYSQL.

@bharadwaj-aditya please confirm this behaviour is okay.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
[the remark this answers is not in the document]
bharadwaj-aditya: I don't understand the use of this block to find a value and set it to null.

If oracle always sets it to Uppercase, then we should do this. If there is a scenario where it can support case sensitivity, then lets not make this change.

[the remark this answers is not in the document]
dhwanilpatel: > I don't understand the use of this block to find a value and set it to null.

That if block indeed was a bug, which I have removed now.

 > If oracle always sets it to Uppercase, then we should do this. If there is a scenario where it can support case sensitivity, then lets not make this change.
 
By default, Oracle converts resource names to Uppercase(similar to PG which converts to lowercase). However Case sensitivity can be retained in Oracle if resources are created with double quotes (e.g. CREATE TABLE "SampleTable"). Since oracle supports preserving case-sensitivity, we should not convert it to uppercase as Gemini suggested.
 
 I wanted your opinion on this block 
 
 ```
 String schemaPattern = sourceSchemaReference.namespace() != null ? sourceSchemaReference.namespace() : metaData.getUserName();
 ```
 
If user doesn't provided sourceSchemaReference, should we leave it as null? which will scan all table schemas across all users, which can be extremely slow. Should we continue passing `username`, so it will scan the table schemas of this user only.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
```diff
@@ -0,0 +1,303 @@
...
+            .set(TimeStampTz.OFFSET_FIELD_NAME, 0)
+            .build();
+      };
+
+  private static final ResultSetValueMapper<java.math.BigDecimal> bigDecimalToByteArray =
+      (value, schema) -> {
+        if (value == null) {
+          return null;
+        }
+        return ByteBuffer.wrap(value.unscaledValue().toByteArray());
```
bharadwaj-aditya: Can you check the working of this with tests involving decimals of different scales ?

Gemini threw up this error:
Issue: Avro logical type decimal(precision, scale) requires the byte array representation to match the fixed scale defined in the schema. If value.unscaledValue().toByteArray() is called directly without adjusting the scale (e.g., value 123.4 with scale 1 against an Avro
  field with scale 2), the unscaled integer 1234 is serialized instead of 12340, causing decimal value corruption upon deserialization (e.g. read as 12.34).
  • Fix: Set the scale to match the Avro schema property as done in MysqlJdbcValueMappings:

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
```diff
@@ -0,0 +1,303 @@
...
+      };
+
+  private static final ResultSetValueMapper<java.sql.Timestamp> timestampTzToAvroRecord =
+      (value, schema) -> {
+        long micros =
+            TimeUnit.SECONDS.toMicros(value.getTime() / 1000)
+                + TimeUnit.NANOSECONDS.toMicros(value.getNanos());
+        return new GenericRecordBuilder(TimeStampTz.SCHEMA)
+            .set(TimeStampTz.TIMESTAMP_FIELD_NAME, micros)
+            .set(TimeStampTz.OFFSET_FIELD_NAME, 0)
```
bharadwaj-aditya: The timestamp offset gets stripped here and it will get read as the JDBC offset. Can you check with different timezones ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,245 @@
...
+        try (ResultSet rs = metaData.getColumns(null, schemaPattern, table, null)) {
+          while (rs.next()) {
+            String colName = rs.getString("COLUMN_NAME");
+            String typeName = rs.getString("TYPE_NAME");
+            if (typeName != null) {
+              typeName = typeName.replaceAll("\\([0-9]+\\)", "");
+            }
+            long colSize = rs.getLong("COLUMN_SIZE");
+            long decimalDigits = rs.getLong("DECIMAL_DIGITS"); // May be 0 if null
+            SourceColumnType record = new SourceColumnType(typeName, new Long[] {colSize}, null);
```
bharadwaj-aditya: seems like decimal digits is not used here. This should get set correctly.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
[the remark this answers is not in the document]
bharadwaj-aditya: This should ideally get thrown, because if the table does not have a primary key we can not run a bulk migration.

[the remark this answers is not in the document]
bharadwaj-aditya: please cross verify against other sources as to what the handling is for tables without pks

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,245 @@
...
+
+            SourceColumnIndexInfo.Builder infoBuilder =
+                SourceColumnIndexInfo.builder()
+                    .setColumnName(colName)
+                    .setIsPrimary(true)
+                    .setIsUnique(true)
+                    .setOrdinalPosition(seq)
+                    .setIndexName(pkName != null ? pkName : "PRIMARY")
+                    .setIndexType(type)
+                    .setColumnTypeName("");
```
bharadwaj-aditya: why is the column type empty here ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,245 @@
...
+
+            if (type == SourceColumnIndexInfo.IndexType.STRING) {
+              CollationReference emptyCollation =
+                  CollationReference.builder()
+                      .setDbCharacterSet("UTF8")
+                      .setDbCollation("UTF8_BIN")
+                      .setPadSpace(false)
+                      .build();
+              infoBuilder.setCollationReference(emptyCollation);
+              infoBuilder.setStringMaxLength(200);
```
bharadwaj-aditya: What is the significance of 200 here ?
Shouldn't this be based on the schema definition ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/OracleSrcToSpSourceConnector.java
```diff
@@ -0,0 +1,142 @@
...
+          .put("FLOAT", UnifiedMappingProvider.Type.FLOAT)
+          .put("DOUBLE PRECISION", UnifiedMappingProvider.Type.DOUBLE)
+          .put("BINARY_FLOAT", UnifiedMappingProvider.Type.FLOAT)
+          .put("BINARY_DOUBLE", UnifiedMappingProvider.Type.DOUBLE)
+          .put("INTEGER", UnifiedMappingProvider.Type.LONG)
+          .put("INT", UnifiedMappingProvider.Type.LONG)
+          .put("SMALLINT", UnifiedMappingProvider.Type.LONG)
+          .put("DATE", UnifiedMappingProvider.Type.TIMESTAMP)
+          .put("TIMESTAMP", UnifiedMappingProvider.Type.TIMESTAMP)
+          .put("TIMESTAMP WITH TIME ZONE", UnifiedMappingProvider.Type.TIMESTAMP_WITH_TIME_ZONE)
```
bharadwaj-aditya, line 56: seems like the interval types are missing here. What happens if we encounter those ?

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,245 @@
...
+      String schemaPattern =
+          sourceSchemaReference.namespace() != null
+              ? sourceSchemaReference.namespace()
+              : metaData.getUserName();
+      try (ResultSet rs = metaData.getTables(null, schemaPattern, null, new String[] {"TABLE"})) {
+        while (rs.next()) {
+          tablesBuilder.add(rs.getString("TABLE_NAME"));
+        }
+      }
+    } catch (SQLException e) {
```
bharadwaj-aditya, line 153: we should be emitting metrics here and handle transient errors. Can you check the Mysql Dialect Adaptor for the implmentation.
Check for MetricCounters.READER_SCHEMA_DISCOVERY_ERRORS

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
```diff
@@ -33,7 +34,8 @@ public interface SourceDbToSpannerOptions extends CommonTemplateOptions {
         @TemplateParameter.TemplateEnumOption(ASTRA_DB_SOURCE_DIALECT),
         @TemplateParameter.TemplateEnumOption(CASSANDRA_SOURCE_DIALECT),
         @TemplateParameter.TemplateEnumOption(MYSQL_SOURCE_DIALECT),
-        @TemplateParameter.TemplateEnumOption(PG_SOURCE_DIALECT)
+        @TemplateParameter.TemplateEnumOption(PG_SOURCE_DIALECT),
+        @TemplateParameter.TemplateEnumOption(ORACLE_SOURCE_DIALECT)
       },
       description = "Dialect of the source database",
       helpText = "Possible values are `CASSANDRA`, `MYSQL` and `POSTGRESQL`.")
```
bharadwaj-aditya: update the help text with the supported sources

bharadwaj-aditya (commented):
Additional Gemini comments

1.  extractDbFromURL uses new URI(sourceDbUrl.substring(5)).getPath(). For Oracle thin URLs like jdbc:oracle:thin:@//host:port/service_name, standard URI parsing evaluates to an opaque URI and .getPath() returns null, resulting in a NullPointerException. Ensure extractDbFromURL
  properly parses Oracle JDBC URLs.

2. Can you add the unit tests for the methods ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
bharadwaj-aditya: looks odd. Can you add a TODO to check this for pg.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Looks like miss. Set the columnTypeName here.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Agree, it should be based on col size. Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/SourceDbToSpannerOptions.java
dhwanilpatel: Updated.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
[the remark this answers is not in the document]
dhwanilpatel: On MYSQL we throw the exception, updated to throw the exception in all three method of discovering table/schema/indices.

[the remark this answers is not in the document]
dhwanilpatel: Updated to throw the exception: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4133#discussion_r3781241138

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
dhwanilpatel: Used similar approach as MySQLJdbcValueMappings

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
dhwanilpatel: Updated and changed Oracle Timezone IT to test with different timezone as well.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/OracleSrcToSpSourceConnector.java
dhwanilpatel: Added and verified and added few missing types like JSON/LONG/LONG RAW/etc.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/convertors/ChangeEventSessionConvertor.java
dhwanilpatel: Added TODO

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Added metrics

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/reader/io/jdbc/uniformsplitter/transforms/JdbcUrl.java
```diff
@@ -103,12 +104,17 @@ abstract class JdbcUrl {
         if (startHost == -1) {
           return null;
         }
-        List<String> components = Splitter.on(':').splitToList(cleanUri.substring(startHost + 1));
-        if (components.size() < 3) {
-          return null;
+        String hostPart = cleanUri.substring(startHost + 1);
+        if (hostPart.contains("/")) {
+          cleanUri = "oracle://" + hostPart;
```
bharadwaj-aditya: oracle specific logic should move to the connector

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,351 @@
...
+  @Override
+  public ImmutableList<String> discoverTables(
+      javax.sql.DataSource dataSource, JdbcSchemaReference sourceSchemaReference)
+      throws SchemaDiscoveryException, RetriableSchemaDiscoveryException {
+    ImmutableList.Builder<String> tablesBuilder = ImmutableList.builder();
+    try (Connection conn = dataSource.getConnection()) {
+      DatabaseMetaData metaData = conn.getMetaData();
+      String schemaPattern =
+          sourceSchemaReference.namespace() != null
+              ? sourceSchemaReference.namespace()
```
bharadwaj-aditya, line 127: can users have access to limited tables within a namespace - to check

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
```diff
@@ -0,0 +1,351 @@
...
+        tables.stream().collect(Collectors.toMap(t -> t, t -> ImmutableList.builder()));
+
+    try (Connection conn = dataSource.getConnection()) {
+      DatabaseMetaData metaData = conn.getMetaData();
+      String schemaPattern =
+          sourceSchemaReference.namespace() != null
+              ? sourceSchemaReference.namespace()
+              : metaData.getUserName();
+      for (String table : tables) {
+        try (ResultSet rs = metaData.getPrimaryKeys(null, schemaPattern, table)) {
```
bharadwaj-aditya, line 269: verify if other sources fetch non primary keys as well

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
```diff
@@ -0,0 +1,336 @@
...
+   * Data Types</a> If type is unknown, it defaults safely to 100 bytes to tightly constrain bundle
+   * metrics.
+   */
+  @Override
+  public int estimateColumnSize(SourceColumnType sourceColumnType) {
+    String typeName = sourceColumnType.getName().toUpperCase();
+    if (JDBC_MAPPINGS.sizeEstimators().containsKey(typeName)) {
+      return JDBC_MAPPINGS.sizeEstimators().get(typeName).apply(sourceColumnType);
+    }
+    return 100;
```
bharadwaj-aditya: add warn logger here. This should never happen

bharadwaj-aditya (commented):
Minor comments

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: In Oracle, users and namespace has one to one relationship.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/rowmapper/provider/OracleJdbcValueMappings.java
dhwanilpatel: Done.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/reader/io/jdbc/uniformsplitter/transforms/JdbcUrl.java
dhwanilpatel: This change was added for handling URLs without `//` as well, but we specifically add the jdbcurl with `//` on `OracleSrcToSpConnector.getJdbcURL()`, it should work as it is. 

Will revert this change.

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: Oracle fetches the primary key index only. (`metaData.getPrimaryKeys`). I checked Mysql/PG, they fetched all primary/secondary index. 

Do we want to keep behaviour same for oracle as well i.e. fetch all indices? or keep primary index only?

bharadwaj-aditya (approved):
LGTM

v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/oracle/reader/io/jdbc/dialectadapter/oracle/OracleDialectAdapter.java
dhwanilpatel: We are keeping to primary only for now.

PR #4137, opened 2026-08-13
Add Reverse support for ORACLE

dhwanilpatel:
Add support of ORACLE source in REVERSE Template using source Skill.

Implemented all ITs similar to MySQL using test skill and all ITs are passing in local. Will raise separate PR of it.

### DIFF
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.teleport.v2.spanner.sourceddl;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class OracleInformationSchemaScanner implements SourceSchemaScanner {
+  private final Connection connection;
+
+  public OracleInformationSchemaScanner(Connection connection) {
+    this.connection = connection;
+  }
+
+  @Override
+  public SourceSchema scan() {
+    Map<String, SourceTable> tablesMap = new HashMap<>();
+    SourceSchema.Builder builder =
+        SourceSchema.builder(SourceDatabaseType.ORACLE).databaseName("ORACLE");
+    try {
+      DatabaseMetaData metaData = connection.getMetaData();
+      String schemaPattern = null;
+      try {
+        schemaPattern = connection.getSchema();
+      } catch (Exception e) {
+      }
+      if (schemaPattern == null) {
+        try {
+          schemaPattern = metaData.getUserName();
+        } catch (Exception e) {
+        }
+      }
+
+      try (ResultSet rs = metaData.getTables(null, schemaPattern, "%", new String[] {"TABLE"})) {
+        while (rs.next()) {
+          String tableName = rs.getString("TABLE_NAME");
+          SourceTable table = scanTable(metaData, schemaPattern, tableName);
+          tablesMap.put(tableName, table);
+        }
+      }
+    } catch (SQLException e) {
+      throw new RuntimeException("Failed to scan Oracle information schema", e);
+    }
+    return builder.tables(ImmutableMap.copyOf(tablesMap)).build();
+  }
+
+  private SourceTable scanTable(DatabaseMetaData metaData, String schemaPattern, String tableName)
+      throws SQLException {
+    SourceTable.Builder tableBuilder =
+        SourceTable.builder(SourceDatabaseType.ORACLE).name(tableName).schema(schemaPattern);
+    List<SourceColumn> columns = new ArrayList<>();
+
+    try (ResultSet colsRs = metaData.getColumns(null, schemaPattern, tableName, "%")) {
+      while (colsRs.next()) {
+        String columnName = colsRs.getString("COLUMN_NAME");
+        String dataType = colsRs.getString("TYPE_NAME");
+
+        SourceColumn.Builder colBuilder =
+            SourceColumn.builder(SourceDatabaseType.ORACLE)
+                .name(columnName)
+                .type(dataType)
+                .isNullable("YES".equalsIgnoreCase(colsRs.getString("IS_NULLABLE")));
+
+        String isGeneratedStr = "";
+        try {
+          isGeneratedStr = colsRs.getString("IS_GENERATEDCOLUMN");
+        } catch (Exception e) {
+        }
+        colBuilder.isGenerated("YES".equalsIgnoreCase(isGeneratedStr));
+        columns.add(colBuilder.build());
+      }
+    }
+
+    List<String> pks = new ArrayList<>();
+    try (ResultSet pkRs = metaData.getPrimaryKeys(null, schemaPattern, tableName)) {
+      while (pkRs.next()) {
+        pks.add(pkRs.getString("COLUMN_NAME"));
+      }
+    }
+
+    tableBuilder.columns(ImmutableList.copyOf(columns));
+    tableBuilder.primaryKeyColumns(ImmutableList.copyOf(pks));
+    return tableBuilder.build();
+  }
+}
--- a/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
+++ b/v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
@@ -55,6 +55,9 @@ public class OracleInformationSchemaScanner implements SourceSchemaScanner {
       try (ResultSet rs = metaData.getTables(null, schemaPattern, "%", new String[] {"TABLE"})) {
         while (rs.next()) {
           String tableName = rs.getString("TABLE_NAME");
+          if (tableName.startsWith("BIN$")) {
+            continue;
+          }
           SourceTable table = scanTable(metaData, schemaPattern, tableName);
           tablesMap.put(tableName, table);
         }
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
@@ -52,8 +52,24 @@ public class JdbcDao implements IDao {
         throw new ConnectionException("Connection is null");
       }
       connObj.setAutoCommit(false);
-      statement = connObj.createStatement();
-      statement.executeUpdate(dmlStatement);
+      java.util.List<Object> params = dmlGeneratorResponse.getPreparedStatementParameters();
+      if (params != null && !params.isEmpty()) {
+        java.sql.PreparedStatement pstmt = connObj.prepareStatement(dmlStatement);
+        statement = pstmt;
+        int paramIdx = 1;
+        for (Object param : params) {
+          if (param instanceof byte[]) {
+            byte[] bytes = (byte[]) param;
+            pstmt.setBlob(paramIdx++, new java.io.ByteArrayInputStream(bytes), bytes.length);
+          } else {
+            pstmt.setObject(paramIdx++, param);
+          }
+        }
+        pstmt.executeUpdate();
+      } else {
+        statement = connObj.createStatement();
+        statement.executeUpdate(dmlStatement);
+      }
 
       if (transactionalCheck != null) {
         transactionalCheck.check();
--- a/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/DMLGeneratorUtils.java
+++ b/v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/DMLGeneratorUtils.java
@@ -40,7 +40,8 @@ public class DMLGeneratorUtils {
         Column spannerColDef,
         SourceColumn sourceColDef,
         JSONObject valuesJson,
-        String sourceDbTimezoneOffset);
+        String sourceDbTimezoneOffset,
+        List<Object> preparedStatementParameters);
   }
 
   public static String convertBase64ToRawHex(String base64EncodedString) {
@@ -71,7 +72,8 @@ public class DMLGeneratorUtils {
       JSONObject keyValuesJson,
       String sourceDbTimezoneOffset,
       Map<String, Object> customTransformationResponse,
-      ColumnValueMapper columnValueMapper) {
+      ColumnValueMapper columnValueMapper,
+      List<Object> preparedStatementParameters) {
     Map<String, String> response = new java.util.LinkedHashMap<>();
 
     List<String> sourcePKs = sourceTable.primaryKeyColumns();
@@ -110,7 +112,7 @@ public class DMLGeneratorUtils {
         }
         columnValue =
             columnValueMapper.getMappedColumnValue(
-                spannerColDef, sourceColDef, keyValuesJson, sourceDbTimezoneOffset);
+                spannerColDef, sourceColDef, keyValuesJson, sourceDbTimezoneOffset, preparedStatementParameters);
       } else if (newValuesJson.has(actualColName)) {
         if (newValuesJson.isNull(actualColName)) {
           response.put(colName, null);
@@ -118,7 +120,7 @@ public class DMLGeneratorUtils {
         }
         columnValue =
             columnValueMapper.getMappedColumnValue(
-                spannerColDef, sourceColDef, newValuesJson, sourceDbTimezoneOffset);
+                spannerColDef, sourceColDef, newValuesJson, sourceDbTimezoneOffset, preparedStatementParameters);
       } else {
         continue;
       }
@@ -137,7 +139,8 @@ public class DMLGeneratorUtils {
       JSONObject keyValuesJson,
       String sourceDbTimezoneOffset,
       Map<String, Object> customTransformationResponse,
-      ColumnValueMapper columnValueMapper) {
+      ColumnValueMapper columnValueMapper,
+      List<Object> preparedStatementParameters) {
     Map<String, String> response = new java.util.LinkedHashMap<>();
 
     List<String> sourcePKs = sourceTable.primaryKeyColumns();
@@ -191,7 +194,7 @@ public class DMLGeneratorUtils {
         }
         columnValue =
             columnValueMapper.getMappedColumnValue(
-                spannerColDef, sourceColDef, keyValuesJson, sourceDbTimezoneOffset);
+                spannerColDef, sourceColDef, keyValuesJson, sourceDbTimezoneOffset, preparedStatementParameters);
       } else if (newValuesJson.has(actualColName)) {
         if (newValuesJson.isNull(actualColName)) {
           response.put(sourceColName, null);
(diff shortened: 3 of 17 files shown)


v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+                  + ", 'YYYY-MM-DD\"T\"HH24:MI:SS.FF\"Z\"') AS TIMESTAMP), 'UTC') AT TIME ZONE '"
+                  + sourceDbTimezoneOffset
+                  + "' AS TIMESTAMP)";
+        } else {
+          response =
+              "TO_TIMESTAMP("
+                  + getQuotedEscapedString(colValue, spannerColType)
+                  + ", 'YYYY-MM-DD\"T\"HH24:MI:SS.FF\"Z\"')";
+        }
+        break;
```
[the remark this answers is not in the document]
bharadwaj-aditya: please check this

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+
+      try (ResultSet rs = metaData.getTables(null, schemaPattern, "%", new String[] {"TABLE"})) {
+        while (rs.next()) {
+          String tableName = rs.getString("TABLE_NAME");
+          SourceTable table = scanTable(metaData, schemaPattern, tableName);
+          tablesMap.put(tableName, table);
+        }
+      }
```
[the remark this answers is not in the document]
bharadwaj-aditya: I'm guessing this would be there in the bulk migration skill as well. so there will be overlap of code.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
```diff
@@ -52,8 +52,24 @@ public void write(
         throw new ConnectionException("Connection is null");
       }
       connObj.setAutoCommit(false);
-      statement = connObj.createStatement();
-      statement.executeUpdate(dmlStatement);
+      java.util.List<Object> params = dmlGeneratorResponse.getPreparedStatementParameters();
+      if (params != null && !params.isEmpty()) {
+        java.sql.PreparedStatement pstmt = connObj.prepareStatement(dmlStatement);
```
bharadwaj-aditya, line 57: creating a prepared statement for every query is potentially more expensive than a direct update. Unless we're caching the prepared statements, we should not be doing this.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/DMLGeneratorUtils.java
```diff
@@ -73,7 +72,7 @@ public static Map<String, String> getColumnValues(
       String sourceDbTimezoneOffset,
       Map<String, Object> customTransformationResponse,
       ColumnValueMapper columnValueMapper) {
-    Map<String, String> response = new HashMap<>();
+    Map<String, String> response = new java.util.LinkedHashMap<>();
```
bharadwaj-aditya, line 77: There's no harm doing this, but does this need to be a linked hashmap ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.json.JSONObject;
+
+/** Creates DML statements for PostgreSQL. */
+public class OracleDMLGenerator implements IDMLGenerator {
```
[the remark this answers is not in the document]
bharadwaj-aditya: please change this

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+
+  private static final ThreadLocal<java.util.List<Object>> threadLocalParameters =
```
bharadwaj-aditya: why are we adding thread locals here ? It seems the same can be achieved with just method parameters.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+            String.format(
+                "Unsupported modType: %s for table %s",
+                dmlGeneratorRequest.getModType(), spannerTableName));
+      }
+    } finally {
+      threadLocalParameters.remove();
+    }
+  }
+
+  private static DMLGeneratorResponse getUpsertStatement(
```
bharadwaj-aditya, line 136: couple of oracle comments from gemini - 
Issue 1 (? in USING subquery): In Oracle SQL, untyped bind parameters in SELECT ? AS "col" FROM DUAL within a USING (...) subquery fail with ORA-00932: inconsistent datatypes because Oracle cannot infer the data type of ? in the projection without an explicit cast (e.g.
  CAST(? AS VARCHAR2(...))).
  • Issue 2 (Untyped NULL): SELECT NULL AS "col" FROM DUAL produces an untyped null column which fails when referenced in ON (t."id" = s."id") or UPDATE SET.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+  private static String getColumnValueByType(
+      String columnType, String colValue, String sourceDbTimezoneOffset, String spannerColType) {
+    String response = "";
+    // TODO: Add support for array types (e.g., varchar[], integer[]) to generate valid PostgreSQL
+    // array literals.
+    if (columnType != null && columnType.contains("(")) {
+      columnType = columnType.substring(0, columnType.indexOf("(")).trim();
+    }
+    switch (columnType) {
+      case "nvarchar2":
```
bharadwaj-aditya, line 386: boolean seems to be missing here.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleSpToSrcSourceConnector.java
```diff
@@ -0,0 +1,128 @@
...
+  public IDMLGenerator getDmlGenerator() {
+    return new OracleDMLGenerator();
+  }
+
+  @Override
+  public IConnectionHelper getConnectionHelper() {
+    return this.connectionHelper;
+  }
+
+  String getConnectionUrl(Shard shard) {
```
bharadwaj-aditya, line 55: Gemini comment - 

Issue: Oracle Easy Connect syntax requires // after @: "jdbc:oracle:thin:@//" + host + ":" + port + "/" + dbName. Missing // can fail connection resolution depending on the Oracle client environment. Note that PR #4133 used @//

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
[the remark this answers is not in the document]
dhwanilpatel: Updated it.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/sourceddl/OracleInformationSchemaScanner.java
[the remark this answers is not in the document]
dhwanilpatel: Sure will check if it is the case for bulk as well or not. 

Fixed it for reverse.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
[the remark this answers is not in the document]
dhwanilpatel: Updated

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
dhwanilpatel: I looked into the need for prepared statements here. Oracle enforces a 4,000-byte limit on SQL literals, which breaks for datatypes like BYTES/BLOB and triggers `ORA-01704: string literal too long`. Because of this, we must rely on prepared statements for these large data types.

This is exactly why the `ThreadLocal` structure you pointed out in `OracleDMLGenerator` was originally introduced. it acts as a workaround to track which queries contain bytes/blob and explicitly require a prepared statement.

I agree with your suggestion to cache the prepared statements, as it is expensive for the database to recreate them for every row.

However, because our DML Generator currently only parameterizes BYTES/BLOB columns while injecting concrete literal values for everything else (e.g., `UPDATE table SET name='John', blob_col=? WHERE id=123`), caching them in this state would lead to a 100% cache miss rate since the physical SQL string text changes on every row.

To secure those cache hits, we need to use fully generic queries (e.g., `UPDATE table SET name=?, blob_col=? WHERE id=?`). We can keep things simple and structurally enforce generic parameterized statements for all data types.

**Final Proposal:**

We'll change the `OracleDMLGenerator` to natively use fully generic, parameterized queries (e.g., `UPDATE table SET name=?, blob_col=? WHERE id=?`) for all data types and cache them dynamically. This will provide a high cache hit rate, improve execution efficiency, and cleanly remove the need for the ThreadLocal checks entirely.

@bharadwaj-aditya  What are your thoughts on this proposal? I might be missing something over here. If it looks good, I will make the changes.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Will make changes as per finalised proposal in thread: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4137#discussion_r3781870963

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Updated.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Updated ? and NULL both to cast with type.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -0,0 +1,469 @@
...
+
+  private static String escapeString(String input) {
+    String cleanedNullBytes = StringUtils.replace(input, "\u0000", "");
+    cleanedNullBytes = StringUtils.replace(cleanedNullBytes, "'", "''");
+    // PostgreSQL defaults to standard conforming strings, so backslash is just a
+    // backslash.
+    // For standard string literals '', we just need to escape the single quote as
+    // ''
+    return cleanedNullBytes;
+  }
```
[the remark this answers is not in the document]
dhwanilpatel: Fixed doc.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleSpToSrcSourceConnector.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @Override
+  public org.apache.beam.sdk.values.TupleTag<String> classifyException(Throwable cause) {
+    if (cause instanceof java.sql.SQLSyntaxErrorException
+        || cause instanceof java.sql.SQLDataException) {
+      return com.google.cloud.teleport.v2.templates.constants.Constants.PERMANENT_ERROR_TAG;
+    }
+    return null;
+  }
```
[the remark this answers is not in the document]
dhwanilpatel: Updated this to recursively check cause.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleSpToSrcSourceConnector.java
```diff
@@ -0,0 +1,128 @@
...
+
+  Connection createConnection(Shard shard) throws Exception {
+    HikariConfig config = new HikariConfig();
+    config.setJdbcUrl(getConnectionUrl(shard));
+    config.setUsername(shard.getUserName());
+    config.setPassword(shard.getPassword());
+    config.setDriverClassName("oracle.jdbc.OracleDriver");
+    HikariDataSource ds = new HikariDataSource(config);
+    return ds.getConnection();
+  }
```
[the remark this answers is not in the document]
dhwanilpatel: I have fixed this for oracle, but there has been TODO for this for MySQL/PG as well.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleSpToSrcSourceConnector.java
dhwanilpatel: Updated

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dml/DMLGeneratorUtils.java
dhwanilpatel: This was needed for prepared statement, where we needed to maintain order for columns.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
bharadwaj-aditya: lets use prepared statement for all flows and cache them. Note that this needs to be per shard.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Removed the thread local use.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/dbutils/dao/source/JdbcDao.java
dhwanilpatel: [Update]: We will use the logic, where we will use the prepare statement for the type values having size > 3500 [with some buffer against 4000] and for other will use normal statement.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -263,20 +278,23 @@ private static DMLGeneratorResponse generateUpsertStatement(
                     spannerColDef,
                     col,
                     dmlGeneratorRequest.getNewValuesJson(),
-                    dmlGeneratorRequest.getSourceDbTimezoneOffset()));
+                    dmlGeneratorRequest.getSourceDbTimezoneOffset(),
+                    currentParams));
           }
-        } catch (Exception e) {
+        } catch (NoSuchElementException e) {
+          LOG.warn("Spanner column not found for generated Oracle column: {}", col.name());
```
bharadwaj-aditya: this will log for every row of this table, leading to log throttling. Please remove this logger.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -386,9 +418,9 @@ private static String getColumnValueByType(
               "TO_DATE(" + getQuotedEscapedString(colValue, spannerColType) + ", 'YYYY-MM-DD')";
         } else {
           response =
-              "TO_TIMESTAMP_TZ("
+              "FROM_TZ(TO_TIMESTAMP("
                   + getQuotedEscapedString(colValue, spannerColType)
-                  + ", 'YYYY-MM-DD\"T\"HH24:MI:SS.FF\"Z\"')";
+                  + ", 'YYYY-MM-DD\"T\"HH24:MI:SS.FF\"Z\"'), 'UTC')";
```
bharadwaj-aditya, line 422: it seems odd to default this to UTC. Can you confirm the method behaviour ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
```diff
@@ -285,11 +303,35 @@ static String getMappedColumnValue(
...
+     * If Spanner sends a massive payload (e.g. a large text CLOB or a large binary BLOB) and we
+     * format it as an inline literal, Oracle will violently crash.
+     * 
+     * To prevent this, if the exact payload size breaches ~3,500 characters, we dynamically intercept
+     * the raw pristine object before it is wrapped in SQL syntax, inject it into the PreparedStatement 
+     * Java array, and natively return a '?' placeholder. 
+     * Smaller payloads safely bypass this to retain blistering fast simple Statement execution.
+     */
+    String rawJsonStr = valuesJson.get(colName).toString();
+    if (preparedStatementParameters != null && rawJsonStr != null && rawJsonStr.length() >= 3500) {
```
bharadwaj-aditya, line 323: it seems like prepared statement parameters are always passed. Does that mean that every write happens via prepared statements ?

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Ack, will remove it.

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: We add the parameters if the length of the value is > 3500. 
```
if (preparedStatementParameters != null && rawJsonStr != null && rawJsonStr.length() >= 3500) {
```

Parameters are empty for the normal cases. JdbcDao creates the prepared statement if provided parameter list is non-empty. 

```
      if (params != null && !params.isEmpty()) {
        java.sql.PreparedStatement pstmt = connObj.prepareStatement(dmlStatement);
```

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
dhwanilpatel: Spanner exports timestamps in UTC (ending with a 'Z'), but Oracle's TO_TIMESTAMP ignores the 'Z' and treats it as a timezone-less value. We have to explicitly use FROM_TZ(..., 'UTC') so Oracle doesn't mistakenly apply its local server timezone to the data.

Ref: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4137#discussion_r3775059689

bharadwaj-aditya: ok

v2/spanner-to-sourcedb/src/main/java/com/google/cloud/teleport/v2/templates/source/oracle/OracleDMLGenerator.java
bharadwaj-aditya: ok.

bharadwaj-aditya (approved):
LGTM

PR #4138, opened 2026-08-13
[Data Validation] Postgres AllDatatypes E2E test

aasthabharill:
- Fixed UUID handling during Hashing to add UUID support
- go/pg-bulk-migration-support-dd was followed as reference for mappings:

### DIFF
--- a/v2/gcs-spanner-dv/src/test/resources/BulkMigrationAndValidationPostgreSQLAllDataTypesE2EIT/postgresql-data-types.sql
+++ b/v2/gcs-spanner-dv/src/test/resources/BulkMigrationAndValidationPostgreSQLAllDataTypesE2EIT/postgresql-data-types.sql
@@ -22,7 +22,7 @@ CREATE TABLE bigint_to_int64 (
     col bigint
 );
 INSERT INTO bigint_to_int64 (col) VALUES
-    ('-9223372036854775808'), ('9223372036854775807'), ('0'), (NULL);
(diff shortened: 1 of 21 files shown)


v2/gcs-spanner-dv/src/test/resources/BulkMigrationAndValidationPostgreSQLAllDataTypesE2EIT/postgresql-data-types.sql
```diff
@@ -0,0 +1,1042 @@
+/*
+  NOTE: Boundary values injected in this schema strictly respect the target Spanner datatype limits.
```
darshan-sj, line 2: Does this test run successfully? I think these comments would give errors when we try to apply to the database.

darshan-sj (approved):
Looks good

v2/gcs-spanner-dv/src/test/resources/BulkMigrationAndValidationPostgreSQLAllDataTypesE2EIT/postgresql-data-types.sql
aasthabharill: it does run successfully, because `/* .. */` is also a standard comment syntax in postgresql, its applied by the db engine and is ignored

aasthabharill: test run passed as well

PR #4142, opened 2026-08-14
Bump netty version to 4.2.17.Final

kennknowles:


damccorm (approved):

chamikaramj (approved):

kennknowles:
Spanner PR job has been timing out at 3 hours since August 14. Filed #4149. I think there's enough green here to justify merging before resolving that bug. WDYT?

damccorm:
> Spanner PR job has been timing out at 3 hours since August 14. Filed #4149. I think there's enough green here to justify merging before resolving that bug. WDYT?

I agree

PR #4148, opened 2026-08-17
Exclude jline from transitive dependencies of hadoop

kennknowles:
jline was being pulled in transitively via hadoop-yarn-client, which is used by several hadoop dependencies (hadoop-common, hadoop-client, hadoop-mapreduce-client-core).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,6 +179,17 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-yarn-client</artifactId>
+        <version>${hadoop.version}</version>
+        <exclusions>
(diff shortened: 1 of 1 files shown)


kennknowles:
This won't pass until rebased on #4143 , which is failing to get green (unknown cause currently).

kennknowles:
The failing test is not healthy: 
image

However, I want to kick it until it goes green if I can.

Abacn:
Yeah DataStream test is not relevant to this change, trying to fix in #4159

kennknowles:
OK - if you think it is safe to merge, then I will do so after approval. My understanding is that jline is only used in a hadoop CLI that someone made, so it is not needed for use of the library. I think the transitive dep on hadoop is from kafka and those tests are green.

Abacn (approved):

PR #4152, opened 2026-08-18
Add Foreign Key ON DELETE SET NULL support for Cloud Spanner Avro tem…

sachinagarwal1111:
…plate


sachinagarwal1111:
Fixed test failure.

pratickchokhani:
I didn't see any Integration tests? I believe it's because the feature is not launched in Spanner. 

How about create a tracking bug and linking it as TODO?

sachinagarwal1111:
> I didn't see any Integration tests? I believe it's because the feature is not launched in Spanner.
> 
> How about create a tracking bug and linking it as TODO?

Yes, that's correct. Since the feature is not launched yet, I didn't add integration test. Added a TODO with tracking bug id.

pratickchokhani (approved):

PR #4183, opened 2026-08-26
Terraform module for gcs-spanner-dv

aasthabharill:
b/514941482

Since validation job is source agnostic, we only need this sample folder. For sharded flow or postgres flow, nothing will really change in the setup, so there's no point creating new samples


v2/gcs-spanner-dv/terraform/samples/simple-validation-job/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+Once the jobs have finished running, you can cleanup by running -
+
+```shell
+terraform destroy
+```
+
+## Observability
+
+To monitor the data validation job, you can view the Dataflow job in the Google Cloud Console. The resulting metrics from the validation job such as matched records, mismatched records, and missing records are available in the BigQuery dataset specified in `var.bigquery_dataset`.
```
manitgupta, line 146: Add a couple of examples of queries the user can run to inspect the validation results.

v2/gcs-spanner-dv/terraform/samples/simple-validation-job/main.tf
```diff
@@ -0,0 +1,84 @@
...
+      sessionFilePath                = var.dataflow_params.template_params.session_file_path
+      schemaOverridesFilePath        = var.dataflow_params.template_params.schema_overrides_file_path
+      tableOverrides                 = var.dataflow_params.template_params.table_overrides
+      columnOverrides                = var.dataflow_params.template_params.column_overrides
+      runId                          = var.dataflow_params.template_params.run_id
+      transformationJarPath          = var.dataflow_params.template_params.transformation_jar_path
+      transformationClassName        = var.dataflow_params.template_params.transformation_class_name
+      transformationCustomParameters = var.dataflow_params.template_params.transformation_custom_parameters
+    } : k => v if v != null && v != ""
+  }
```
manitgupta: This is to simplify the null check right? Seems over-optimized..I would prefer readability over this `k, v` loop. Also is everything in this optional? Like stuff like `databaseId` and `instanceId`?

v2/gcs-spanner-dv/terraform/samples/simple-validation-job/README.md
```diff
@@ -0,0 +1,329 @@
...
+If the worker counts are changed in `tfvars` and a Terraform apply is run, Terraform will attempt to cancel the existing Dataflow job and replace it with a new one.
+**This is not recommended**. Instead, use the `gcloud` CLI to update the worker counts of a launched Dataflow job:
+
+```shell
+gcloud dataflow jobs update-options \
+        --region=us-central1 \
+        --min-num-workers=5 \
+        --max-num-workers=20 \
+      <YOUR-JOB-ID>
+```
```
manitgupta: This is also valid for a batch job?

aasthabharill: its not, thanks for calling out - have removed this block from here and the Bulk terraform READMEs

v2/gcs-spanner-dv/terraform/samples/simple-validation-job/main.tf
aasthabharill: passing `null` without this check is actually safe, this is more of a defensive check in case user passes empty strings `""` in any of these parameters as the job would fail

but have removed it anyway, this is not something we check in other terraforms either but i thought it could be useful in failing fast in case of misconfig

no, everything is not optional - these variables were defined as non-optional in the variables.tf file

manitgupta (approved):

PR #4188, opened 2026-08-28
[Data Validation] Add table configuration

aasthabharill:
b/514940857

Added two new pipeline parameters:
1. `tables`: comma-separated list of source tables for small configurations
2. `tableConfigurationFilePath`: GCS File path containing list of source tables in a text file for large workloads

### Why two different template params?
`tables` will be extremely inconvenient to configure in case user needs to configure large number of tables

### `tableConfigurationFilePath` is JSON file
Extensible for future additions 
Example:
```
{
  "tableNames": ["table1", "table2"],
  "tableConfigurations": {
    "table2": { "columnsToValidate": [...] }
  }
}
```

### Table names are to be specified as per their source names
More intuitive for users as they are more familiar with their source schemas. In case of Spanner table renames, the user specifies an overrides file/session file - which will be used to map their configured tables to their corresponding Spanner names

### Fail fast if both template params are provided
There should be only one source of truth for the configured tables and since the purpose for both template params is the same, it doesn't make sense to pass both. 

### No mechanism to fail fast if configured tables do not exist in source
Since our source is the GCS Avro directory, if a source table exists but has zero rows, it will not have any entry in GCS. 
Therefore, GCS records for a table not existing could mean either of two things: 
1. The table doesn't exist on source and was misconfigured by the user
2. The table does exist on source but has 0 rows.
Since there's no way to differentiate between these two cases, we cannot introduce a fail-fast loop to identify a misconfigured table list.

## Testing
- Added comprehensive unit tests for all changes
- Added an integration test for the scenario where `tables` mentions a subset of tables which were renamed in spanner (so tests table config in combination with overrides)


v2/gcs-spanner-dv/src/test/java/com/google/cloud/teleport/v2/templates/GCSSpannerDVCoreMatchingIT.java
```diff
@@ -383,4 +383,113 @@ public void validationTestWithDuplicateAvroRecords() throws Exception {
...
+            // AccountRoles: Mismatched record (role_name = ADMINSTRATOR instead of ADMIN)
+            Mutation.newInsertOrUpdateBuilder("AccountRoles")
+                .set("role_id")
+                .to(1L)
+                .set("role_name")
+                .to("ADMINSTRATOR")
+                .build()));
+
+    // Wait for Spanner's 20-second exact staleness read bound in SpannerReaderTransform
+    Thread.sleep(20000);
```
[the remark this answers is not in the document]
aasthabharill: this is a DirectRunnerTest

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
```diff
@@ -0,0 +1,143 @@
...
+   * @param schemaMapper The schema mapper to translate the table name.
+   * @return true if allowed or no filters are configured, false otherwise.
+   */
+  public boolean isSpannerTableAllowed(String spannerTableName, ISchemaMapper schemaMapper) {
+    if (!hasFilters()) {
+      return true;
+    }
+    try {
+      String sourceTable = schemaMapper.getSourceTableName("", spannerTableName);
+      return configuredSourceTables.contains(sourceTable);
```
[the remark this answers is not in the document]
aasthabharill: schemaMapper will never be null. `SchemaMapperProviderFn.java` always returns an instance: it falls back to an IdentityMapper if no session file or overrides are provided.

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/transforms/SourceReaderTransform.java
```diff
@@ -38,36 +42,49 @@ public class SourceReaderTransform
...
 
-  private static String createAvroFilePattern(String inputPath) {
+  static List<String> getFilePatterns(String gcsInputDirectory, TableSelectionConfig tableConfig) {
+    List<String> filePatterns = new ArrayList<>();
     String cleanPath =
-        inputPath.endsWith("/") ? inputPath.substring(0, inputPath.length() - 1) : inputPath;
-    return cleanPath + "/**.avro";
+        gcsInputDirectory.endsWith("/")
+            ? gcsInputDirectory.substring(0, gcsInputDirectory.length() - 1)
+            : gcsInputDirectory;
```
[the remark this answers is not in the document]
aasthabharill: made it non-optional

manitgupta:
```
tableListFilePath is flat text file
Very easy to make, maintain and edit for customers as it's just a list of tables. JSON for this would be overkill and would require formatters for no real advantage.
```

I don't think it should be flat text file, I think it should be a `JSON`. Reason - extensibility. 
A TableConfiguration can have more aspects about a table such as which columns to include/exclude etc. and keeping a JSON (backed by a POJO in the code) will keep this configuration flexible and extendible.

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableSelectionConfig.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.beam.sdk.io.FileSystems;
+import org.apache.beam.sdk.io.fs.ResourceId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Configuration class for table-based filtering in Data Validation pipeline. Encapsulates parsing,
+ * matching, and validation of source and Spanner tables.
+ */
+public class TableSelectionConfig implements Serializable {
```
manitgupta: I think this should be a named `TableConfiguration` (mentioned the reason in a different comment - extensibility). 

The `TableConfiguration` should have a field `tableName` which is used for filtering.

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    TableSelectionConfig config = new TableSelectionConfig(configuredTables);
+
+    return config;
+  }
+
+  public boolean hasFilters() {
+    return configuredSourceTables != null && !configuredSourceTables.isEmpty();
+  }
```
manitgupta, line 110: What does this do?

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Checks if a Spanner table is allowed by the configuration. Translates the Spanner table name to
+   * its source table counterpart using the schema mapper.
+   *
+   * @param spannerTableName The Spanner table name.
+   * @param schemaMapper The schema mapper to translate the table name.
+   * @return true if allowed or no filters are configured, false otherwise.
+   */
+  public boolean isSpannerTableAllowed(String spannerTableName, ISchemaMapper schemaMapper) {
```
manitgupta, line 137: Didn't the PR description say we are supporting configuration using source table names?

manitgupta (changes requested):
I haven't looked at the full PR yet, lets align on the existing comments and then will look at the rest

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
aasthabharill: yes so the customer is expected to use source table names.
But this function goes backward as in it checks Spanner table name -> mapped to source because of the way the code is written in `CreateSpannerReadOpsFn`. It calls this `List<String> tableNames = ddl.getTablesOrderedByReference();` - which returns the tables existing in the Spanner DDL - so we check which ones out of these to filter out by reverse mapping it

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
aasthabharill: just an helper function to identify if we need to apply the table filtering logic or not

was used in `isSpannerTableAllowed` and `isSourceTableAllowed` to return `true` by default if no filtering and in `SourceReaderTransform` to know what filePatterns to look for

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableSelectionConfig.java
aasthabharill: will make these changes + JSON and get back

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/config/TableConfiguration.java
manitgupta: I see, ideally shouldn't we be passing a filtered list into that DoFn?

aasthabharill: So I had actually considered this, but the issue is we need the Spanner DDL object to materialize the schemaMapper (`ISchemaMapper schemaMapper = schemaMapperProvider.apply(ddl);`). So to actually get the list of Spanner tables I need the Spanner DDL, which we get through the get Spanner infoschema step, meaning it's not initialized at pipeline startup. The first place this DDL becomes accessible as a side-input in our read branch is in CreateSpannerReadOpsFn itself

aasthabharill: Also, in case we want to add this to bulk, etc. later i think keeping it like this would be better. This is what we do in validations code:
```
List<String> tableNames = ddl.getTablesOrderedByReference();
for (String tableName : tableNames) { <filtering logic> )
```
Because we filter on this list using our function, the filtered list is also ordered. I dont think the order would matter in validations as its just read ops, but im not sure if it would actually matter in bulk. (please correct me if im wrong here)

manitgupta:
Can you put a sample `tableConfiguration` JSON in the description? I would like to have a look once

v2/gcs-spanner-dv/src/main/java/com/google/cloud/teleport/v2/transforms/SourceReaderTransform.java
```diff
@@ -38,36 +42,49 @@ public class SourceReaderTransform
...
     return input
+        .apply("CreateFilePatterns", Create.of(getFilePatterns(gcsInputDirectory, tableConfig)))
         .apply(
             "ReadSourceAvroRecords",
-            AvroIO.parseGenericRecords(new IdentityGenericRecordFn())
-                .from(createAvroFilePattern(gcsInputDirectory))
-                .withCoder(GenericRecordCoder.of())
-                .withHintMatchesManyFiles())
+            AvroIO.parseAllGenericRecords(new IdentityGenericRecordFn())
+                .withCoder(GenericRecordCoder.of()))
```
manitgupta: Why is the `withHintMatchesManyFiles()` removed!!?

manitgupta (changes requested):

manitgupta (approved):

PR #4206, opened 2026-09-01
Add automated PR watchdog for codelab parameters

aasthabharill:
- Comment gets added on changes to the Options files ONLY. 
- The workflow adds one comment and edits the same one in case of repeated runs
- Follow up work: write up a small guide to updating codelabs to go/harbourbridge so we dont expose these links in the comments

Test how this works by playing around with these PRs:
- https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4200
- external repo: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/4207 (we dont get github write permissions on forked repos so there will be no comment, but the workflow fails with a clear error message)


.github/workflows/codelab-watchdog.yml
```diff
@@ -0,0 +1,67 @@
+name: Codelab Parameter Watchdog
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+    paths:
+      - '**/SourceDbToSpannerOptions.java'
+      - '**/DataStreamToSpannerOptions.java'
+      - '**/SpannerToSourceDbOptions.java'
```
manitgupta, line 9: Validations?

manitgupta:
CLA seems to be failing?

.github/workflows/codelab-watchdog.yml
aasthabharill: the codelab doesnt have validations currently so havent added it here either

manitgupta (approved):

.github/workflows/codelab-watchdog.yml
manitgupta: Gotcha, thanks

PR #4209, opened 2026-09-01
Refactor pipeline parameter options to seperate Options.java file

aasthabharill:


v2/datastream-to-spanner/src/main/java/com/google/cloud/teleport/v2/options/DataStreamToSpannerOptions.java
```diff
@@ -0,0 +1,459 @@
+/*
```
shreyakhajanchi, line 1: nit: the formatting of this license header seems different from existing files , can you PTAL

shreyakhajanchi (approved):

PR #4216, opened 2026-09-03
fix(spanner): skip locality group option in information schema until full support

finnzzf:
Update information schema IT to accept default locality group in column options.

The test is updated to accept both old and new formats until the change is fully rolled out in spanner release.


finnzzf:
since the locality_group information_schema enhancement is not yet fully rolled out, and the locality group option support is not in the dataflow repo, the proposed workaround for now is preventing locality group to show up in the DDL until the support is in place.

shreyakhajanchi (approved):
LGTM

PR #4224, opened 2026-09-06
refactor: add Jdbc url to shard pojo. This will be used by jdbc connection helper. It will allow the URL to be created by the respective source connector. 

pratickchokhani:


bharadwaj-aditya (commented):
Moving the connection UTL to the helper makes sense. But do we need to expose this map to external entities to read against the URL ? Seems like implementation is leaking out of the helper.

The map should ideally be an internal detail. And this still emits shards and consumers use it however they see fit.

pratickchokhani:
> Moving the connection UTL to the helper makes sense. But do we need to expose this map to external entities to read against the URL ? Seems like implementation is leaking out of the helper.
> 
> The map should ideally be an internal detail. And this still emits shards and consumers use it however they see fit.

The goal of the refactoring is to have a default JDBC URL pattern. For SQL server, the other constructor will be used to pass the entire map.

The exposed map is required [JdbcConnectionHelper.java] for getting URL and Shard pair.

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/migrations/connection/ConnectionHelperRequest.java
```diff
@@ -25,23 +29,29 @@
...
- *   <li>A list of {@link Shard} objects representing the database shards.
+ *   <li>A map of connection URLs to {@link Shard} objects representing the database shards.
  *   <li>Optional connection properties as a {@link String}.
  *   <li>The maximum number of connections allowed.
  *   <li>The name of the driver to connect to source.
  *   <li>Optional connection initialisation queries as a {@link String}.
  * </ul>
  */
 public class ConnectionHelperRequest {
-  private List<Shard> shards;
```
dhwanilpatel, line 36: Should we keep this connectionUrl as shard property itself?

pratickchokhani: Shard pojo is used for parsing config files passed by customer. I didn't wanted to bloat it.

bharadwaj-aditya (approved):
LGTM

PR #4235, opened 2026-09-08
feat(mongodb-to-mongodb): High-Throughput Backfill Migration (with Experimental CDC)

michaeltle-goog:
## Overview

This pull request enhances the `mongodb-to-mongodb` Dataflow template to deliver high-throughput, horizontally scalable historical backfills and introduces experimental support for Change Data Capture (CDC).

The pipeline introduces three standardized migration modes via the `migrationMode` parameter:

1. **`BACKFILL` (Historical Backfill Only)**:
   - Executes parallel, adaptive volume-based reads using Splittable DoFns with keyset pagination and concurrency slot bounding.
   - **Bypasses the Stateful Deduplication stage entirely**: Because historical backfill partitions are mutually disjoint and target writes are idempotent (`ReplaceOneModel` with `upsert`), deduplication is unnecessary. Bypassing this stage eliminates the `KeyByDedupKey` shuffle and Windmill state barrier, enabling maximum ingestion throughput.
2. **`BACKFILL_AND_STREAMING` (Dual Backfill + Live CDC)**:
   - Runs historical backfill concurrently with live MongoDB change stream CDC, capturing initial cluster time ($T_0$) before backfill execution begins.
   - **Includes the Stateful Deduplication stage**: This is the **only mode** where stateful deduplication (`StatefulDeduplication.of()`) is present. It reconciles historical backfill documents against concurrent live mutations using a 3-tier monotonic sort key (`TimestampSortKey`) per document key, guaranteeing that older backfill elements never overwrite newer CDC mutations.
3. **`STREAMING_CDC` (Live CDC Only)**:
   - Continuously replicates real-time changes via MongoDB change streams (`db.watch()` or `collection.watch()`) using server-side `$toHashedIndexKey` modulo partitioning and checkpointed resume tokens.
   - Streams live insert, update, replace, and delete operations directly to target collections without backfill or deduplication overhead.

> [!NOTE]
> **Change Data Capture (CDC)** streaming via MongoDB change streams is currently in **experimental mode**.

---

### 1. Adaptive Splitting & Virtual Concurrency Slots (`MongoDbBackfillReader`)
- **Adaptive Volume-Based Splitting**: Computes partition counts adaptively based on collection volume using `targetBackfillChunkSize` (e.g., 25k docs) and `maxBackfillSplits` (up to 10,000), ensuring balanced partition sizing across heterogeneous collection sizes.
- **Bounded Cursor Concurrency (`maxConcurrentBackfillReads`)**: Routes backfill partitions into bounded virtual concurrency slots (`BackfillSlotTask`, `ProcessBackfillSlotFn`), guaranteeing that active cursors and sockets on MongoDB routers (`mongos`) remain strictly bounded regardless of split count.
- **Keyset-Paged Resumption**: Slices are streamed in micro-batches (`MAX_DOCS_PER_SLICE = 2,000`) using indexed keyset pagination (`_id > lastSeenId`), allowing smooth Windmill checkpointing, worker preemption recovery, and dynamic rebalancing without duplicate reads.

### 2. Fast Type-Isolated Split Generation (`ReadSplitGenerator`)
- **Type-Aware Boundary Sampling**: Discovers active BSON `_id` types (ObjectId, String, Number, Binary, Date, Document) and computes quantile boundaries independently per type, preventing cross-type query skew.
- **Parallel Type Probing**: Probes active type buckets concurrently using a dedicated `ExecutorService`, significantly reducing pipeline initialization latency on databases with many collections.
- **Canonical Extended JSON Serialization**: Serializes boundaries via `JsonMode.EXTENDED` to preserve exact BSON types across worker nodes.
- **Algorithmic Fallback**: Gracefully falls back to uniform partition bounds ($mod for numbers, hex intervals for strings/ObjectIDs) if sampling is unavailable.

### 3. High-Throughput Asynchronous Micro-Batch Writer (`MongoDbTransforms`)
- **Configurable Write Sharding**: Downstream writes are keyed by collection and shard ID (`numWriteShards`), distributing Windmill state, timers, and bundle commitments evenly across workers.
- **Asynchronous Bulk Writing**: Non-blocking bulk writes (`maxConcurrentAsyncWrites`) saturate target ingestion capacity (including Cloud Firestore MongoDB endpoints).
- **Optimized Connection Pooling**: Configured connection pools (`maxSize(256)`, `minSize(0)`) to dynamically scale under worker load while preventing connection churn from short-lived driver queries.

---

## Experimental CDC Overview

- Database-level (`db.watch()`) and collection-level change stream support.
- Partitioned parallel streams using server-side `$toHashedIndexKey` modulo filtering on document keys.
- Continuous idle watermark advancement and optimized resume token stride serialization.
- *Note*: CDC remains experimental while optimizations for large-scale sharded oplog throughput and update-lookup decoupling continue.

---

## Verification & Testing
- **Unit Tests**: All 132 unit tests passing (`mvn test -pl v2/mongodb-to-mongodb`).
- **Checkstyle & Formatting**: 0 Checkstyle violations and 100% Google Java Format compliance verified via Spotless.
- **End-to-End Scale Validation**: Verified full migration of 95.25M documents from a MongoDB 7.0 sharded cluster to Cloud Firestore with DLQ replay and 100.0% deep field-by-field sample parity.

### DIFF
--- a/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
+++ b/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
@@ -143,20 +143,6 @@ public class MongoDbToMongoDb {
     void setTargetCollection(String value);
 
     @TemplateParameter.Integer(
-        order = 7,
-        groupName = "Source",
-        optional = true,
-        description = "Number of Backfill Read Splits",
-        helpText =
-            "Number of parallel coarse read splits generated per collection during historical"
-                + " backfill (e.g. 1, 4, 8). Each split opens a sequential index-range cursor."
-                + " Default is 1.")
-    @Default.Integer(1)
-    Integer getNumBackfillSplits();
-
-    void setNumBackfillSplits(Integer value);
-
-    @TemplateParameter.Integer(
         order = 8,
         groupName = "Target",
         optional = true,
@@ -372,6 +358,46 @@ public class MongoDbToMongoDb {
     Integer getMaxBufferingDurationMs();
 
     void setMaxBufferingDurationMs(Integer value);
+
+    @TemplateParameter.Integer(
+        order = 25,
+        groupName = "Source",
+        optional = true,
+        description = "Target Backfill Chunk Size",
+        helpText =
+            "Target number of documents per backfill split. Adaptive volume-based splitting uses this"
+                + " to calculate the number of splits per collection based on estimated count. Default is 200000.")
+    @Default.Integer(200000)
+    Integer getTargetBackfillChunkSize();
+
+    void setTargetBackfillChunkSize(Integer value);
+
+    @TemplateParameter.Integer(
+        order = 26,
+        groupName = "Source",
+        optional = true,
+        description = "Max Backfill Splits",
+        helpText =
+            "Maximum number of backfill read splits allowed per collection during adaptive splitting."
+                + " Default is 256.")
+    @Default.Integer(256)
+    Integer getMaxBackfillSplits();
+
+    void setMaxBackfillSplits(Integer value);
+
+    @TemplateParameter.Integer(
+        order = 27,
+        groupName = "Source",
+        optional = true,
+        description = "Max Concurrent Backfill Reads",
+        helpText =
+            "Maximum number of concurrent in-flight backfill cursors allowed across the cluster."
+                + " Partitions are distributed across virtual concurrency slots to strictly bound"
+                + " source database connections and cursor memory. Default is 128.")
+    @Default.Integer(128)
+    Integer getMaxConcurrentBackfillReads();
+
+    void setMaxConcurrentBackfillReads(Integer value);
   }
 
   public static void main(String[] args) {
@@ -453,9 +479,6 @@ public class MongoDbToMongoDb {
     LOG.info("  Target Database:         {}", options.getTargetDatabase());
     LOG.info("  Source Collections:      {}", sourceCollections);
     LOG.info(
-        "  Backfill Configuration:  coarse-split sequential cursor streaming (numBackfillSplits={})",
-        options.getNumBackfillSplits() != null ? options.getNumBackfillSplits() : 1);
-    LOG.info(
         "  Write Configuration:     batchSize={}, maxConcurrentAsyncWrites={}, maxWriteRetries={},"
             + " dlqMaxRetries={}",
         options.getBatchSize(),
@@ -472,6 +495,9 @@ public class MongoDbToMongoDb {
     LOG.info("  Migration Mode:          {}", options.getMigrationMode());
     LOG.info("  Change Stream Splits:    {}", options.getNumChangeStreamSplits());
     LOG.info("  Change Stream Strategy:  {}", options.getChangeStreamFullDocument());
+    LOG.info("  Backfill Chunk Size:     {}", options.getTargetBackfillChunkSize());
+    LOG.info("  Max Backfill Splits:     {}", options.getMaxBackfillSplits());
+    LOG.info("  Max Concurrent Reads:    {}", options.getMaxConcurrentBackfillReads());
     LOG.info("  DLQ Base Directory:      {}", baseDlqPath + timestampPath);
     LOG.info("  DLQ Retryable Directory: {}", retryableDlqPath);
     LOG.info("  DLQ Permanent Directory: {}", permanentDlqPath);
@@ -550,8 +576,12 @@ public class MongoDbToMongoDb {
                   : targetCollectionRaw;
 
           if (includeBackfill) {
-            int numBackfillSplits =
-                options.getNumBackfillSplits() != null ? options.getNumBackfillSplits() : 1;
+            int targetChunkSize =
+                options.getTargetBackfillChunkSize() != null
+                    ? options.getTargetBackfillChunkSize()
+                    : 200000;
+            int maxSplits =
+                options.getMaxBackfillSplits() != null ? options.getMaxBackfillSplits() : 256;
             try {
               backfillPartitions.addAll(
                   MongoDbBackfillReader.generatePartitions(
@@ -560,15 +590,15 @@ public class MongoDbToMongoDb {
                       options.getSourceDatabase(),
                       inputCollection,
                       targetCollection,
-                      numBackfillSplits,
+                      targetChunkSize,
+                      maxSplits,
                       t0));
             } catch (Exception e) {
               LOG.warn(
                   "Could not generate partitioned splits for collection '{}' ({}). Falling back to"
-                      + " algorithmic splits (splits={}).",
+                      + " algorithmic splits.",
                   inputCollection,
-                  e.getMessage(),
-                  numBackfillSplits);
+                  e.getMessage());
               backfillPartitions.addAll(
                   MongoDbBackfillReader.generatePartitions(
                       null,
@@ -576,7 +606,8 @@ public class MongoDbToMongoDb {
                       options.getSourceDatabase(),
                       inputCollection,
                       targetCollection,
-                      numBackfillSplits,
+                      targetChunkSize,
+                      maxSplits,
                       t0));
             }
           }
@@ -658,11 +689,16 @@ public class MongoDbToMongoDb {
       }
 
       if (includeBackfill && !backfillPartitions.isEmpty()) {
+        int maxConcurrentReads =
+            options.getMaxConcurrentBackfillReads() != null
+                ? options.getMaxConcurrentBackfillReads()
+                : MongoDbBackfillReader.ReadPartitions.DEFAULT_MAX_CONCURRENT_READS;
         PCollection<DocumentWithMetadata> backfillDocs =
             pipeline
                 .apply(
                     "ReadBackfill",
-                    new MongoDbBackfillReader.ReadPartitions(backfillPartitions))
+                    new MongoDbBackfillReader.ReadPartitions(
+                        backfillPartitions, maxConcurrentReads))
                 .setCoder(DocumentWithMetadataCoder.of());
         allStreams.add(backfillDocs);
       }
--- a/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
+++ b/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
@@ -392,6 +392,13 @@ public class MongoDbToMongoDb {
     String sourceUri = options.getSourceUri();
     String sourceDatabase = options.getSourceDatabase();
     String sourceCollection = options.getSourceCollection();
+    String targetCollectionRaw = options.getTargetCollection();
+
+    if ((sourceCollection == null || sourceCollection.isEmpty())
+        && (targetCollectionRaw != null && !targetCollectionRaw.isEmpty())) {
+      throw new IllegalArgumentException(
+          "targetCollection cannot be specified when migrating an entire database without specifying sourceCollection.");
+    }
 
     List<String> sourceCollections = new ArrayList<>();
     if (sourceCollection != null && !sourceCollection.isEmpty()) {
@@ -438,7 +445,7 @@ public class MongoDbToMongoDb {
     LOG.info("  Source Collections:      {}", sourceCollections);
     LOG.info(
         "  Backfill Configuration:  coarse-split sequential cursor streaming (numBackfillSplits={})",
-        options.getNumBackfillSplits() != null ? options.getNumBackfillSplits() : 4);
+        options.getNumBackfillSplits() != null ? options.getNumBackfillSplits() : 1);
     LOG.info(
         "  Write Configuration:     batchSize={}, maxConcurrentAsyncWrites={}, maxWriteRetries={},"
             + " dlqMaxRetries={}",
@@ -528,7 +535,6 @@ public class MongoDbToMongoDb {
 
       try {
         for (String inputCollection : sourceCollections) {
-          String targetCollectionRaw = options.getTargetCollection();
           final String targetCollection =
               (targetCollectionRaw == null || targetCollectionRaw.isEmpty())
                   ? inputCollection
@@ -565,7 +571,7 @@ public class MongoDbToMongoDb {
             }
           }
 
-          if (includeCdc) {
+          if (includeCdc && (sourceCollection != null && !sourceCollection.isEmpty())) {
             int numCdcSplits =
                 options.getNumChangeStreamSplits() != null
                     ? options.getNumChangeStreamSplits()
@@ -598,6 +604,38 @@ public class MongoDbToMongoDb {
                       t0,
                       options.getChangeStreamFullDocument()));
             }
+          }
+        }
+
+        if (includeCdc && (sourceCollection == null || sourceCollection.isEmpty())) {
+          int numCdcSplits =
+              options.getNumChangeStreamSplits() != null ? options.getNumChangeStreamSplits() : 1;
+          LOG.info(
+              "Configuring database-level change stream on database '{}' with {} parallel splits",
+              options.getSourceDatabase(),
+              numCdcSplits);
+          try {
+            cdcPartitions.addAll(
+                MongoDbChangeStreamReader.generateDatabasePartitions(
+                    setupClient,
+                    options.getSourceUri(),
+                    options.getSourceDatabase(),
+                    numCdcSplits,
+                    t0,
+                    options.getChangeStreamFullDocument()));
+          } catch (Exception e) {
+            LOG.warn(
+                "Could not generate database-level change stream splits ({}). Falling back to single"
+                    + " split.",
+                e.getMessage());
+            cdcPartitions.addAll(
+                MongoDbChangeStreamReader.generateDatabasePartitions(
+                    null,
+                    options.getSourceUri(),
+                    options.getSourceDatabase(),
+                    1,
+                    t0,
+                    options.getChangeStreamFullDocument()));
           }
         }
       } finally {
--- a/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
+++ b/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
@@ -525,14 +525,10 @@ public class MongoDbToMongoDb {
               "Could not parse startAtOperationTime '{}' as epoch seconds. Capturing current"
                   + " cluster time.",
               options.getStartAtOperationTime());
-          t0 =
-              MongoDbChangeStreamReader.captureCurrentClusterTime(
-                  sourceUri, sourceDatabase);
+          t0 = MongoDbChangeStreamReader.captureCurrentClusterTime(sourceUri, sourceDatabase);
         }
       } else {
-        t0 =
-            MongoDbChangeStreamReader.captureCurrentClusterTime(
-                sourceUri, sourceDatabase);
+        t0 = MongoDbChangeStreamReader.captureCurrentClusterTime(sourceUri, sourceDatabase);
       }
       LOG.info(
           "Captured initial cluster time T0: {} (seconds={}, inc={})",
@@ -622,9 +618,7 @@ public class MongoDbToMongoDb {
 
           if (includeCdc && (sourceCollection != null && !sourceCollection.isEmpty())) {
             int numCdcSplits =
-                options.getNumChangeStreamSplits() != null
-                    ? options.getNumChangeStreamSplits()
-                    : 1;
+                options.getNumChangeStreamSplits() != null ? options.getNumChangeStreamSplits() : 1;
             try {
               cdcPartitions.addAll(
                   MongoDbChangeStreamReader.generatePartitions(
@@ -714,9 +708,7 @@ public class MongoDbToMongoDb {
       if (includeCdc && !cdcPartitions.isEmpty()) {
         PCollection<DocumentWithMetadata> cdcDocs =
             pipeline
-                .apply(
-                    "ReadCDC",
-                    new MongoDbChangeStreamReader.ReadPartitions(cdcPartitions))
+                .apply("ReadCDC", new MongoDbChangeStreamReader.ReadPartitions(cdcPartitions))
                 .setCoder(DocumentWithMetadataCoder.of());
         allStreams.add(cdcDocs);
       }
@@ -725,9 +717,7 @@ public class MongoDbToMongoDb {
       if (allStreams.size() == 1) {
         documents = allStreams.get(0);
       } else if (allStreams.size() > 1) {
-        documents =
-            PCollectionList.of(allStreams)
-                .apply("MergeStreams", Flatten.pCollections());
+        documents = PCollectionList.of(allStreams).apply("MergeStreams", Flatten.pCollections());
       } else {
         throw new IllegalStateException("No streams configured to read.");
       }
@@ -751,8 +741,8 @@ public class MongoDbToMongoDb {
   }
 
   /**
-   * PTransform that processes documents: stateful deduplication, metric counting,
-   * UDF transformation, and validation, routing process failures to DLQ.
+   * PTransform that processes documents: stateful deduplication, metric counting, UDF
+   * transformation, and validation, routing process failures to DLQ.
    */
   public static class ProcessDocuments
       extends PTransform<PCollection<DocumentWithMetadata>, PCollection<DocumentWithMetadata>> {
@@ -811,13 +801,9 @@ public class MongoDbToMongoDb {
         udfProcessed
             .get(udfFailureTag)
             .apply(
-                "WriteToDlq_UDF",
-                new WriteToDlq(retryableDlqPath, permanentDlqPath, tmpDirectory));
+                "WriteToDlq_UDF", new WriteToDlq(retryableDlqPath, permanentDlqPath, tmpDirectory));
 
-        documents =
-            udfProcessed
-                .get(udfSuccessTag)
-                .setCoder(DocumentWithMetadataCoder.of());
+        documents = udfProcessed.get(udfSuccessTag).setCoder(DocumentWithMetadataCoder.of());
       }
 
       // Validation Stage with DLQ
@@ -842,10 +828,10 @@ public class MongoDbToMongoDb {
   }
 
   /**
-   * PTransform that writes valid documents in bulk to target MongoDB, routing write failures to DLQ.
+   * PTransform that writes valid documents in bulk to target MongoDB, routing write failures to
+   * DLQ.
    */
-  public static class WriteDocuments
-      extends PTransform<PCollection<DocumentWithMetadata>, PDone> {
+  public static class WriteDocuments extends PTransform<PCollection<DocumentWithMetadata>, PDone> {
     private final transient Options options;
     private final String retryableDlqPath;
     private final String permanentDlqPath;
@@ -891,8 +877,7 @@ public class MongoDbToMongoDb {
                           : DEFAULT_WRITE_RATE_RAMP_UP_STEPS));
 
       writeFailures.apply(
-          "WriteToDlq_Write",
-          new WriteToDlq(retryableDlqPath, permanentDlqPath, tmpDirectory));
+          "WriteToDlq_Write", new WriteToDlq(retryableDlqPath, permanentDlqPath, tmpDirectory));
 
       return PDone.in(validDocs.getPipeline());
     }
@@ -933,9 +918,7 @@ public class MongoDbToMongoDb {
             failureTag,
             item != null
                 ? item.withFailure(
-                    "Null document payload",
-                    ErrorType.PERMANENT,
-                    FailureStage.VALIDATE)
+                    "Null document payload", ErrorType.PERMANENT, FailureStage.VALIDATE)
                 : DocumentWithMetadata.of(
                     null,
                     null,
--- a/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
+++ b/v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
@@ -19,9 +19,14 @@ import com.google.cloud.teleport.metadata.Template;
 import com.google.cloud.teleport.metadata.TemplateCategory;
 import com.google.cloud.teleport.metadata.TemplateParameter;
 import com.google.cloud.teleport.v2.transforms.DocumentWithMetadata;
+import com.google.cloud.teleport.v2.transforms.DocumentWithMetadataCoder;
 import com.google.cloud.teleport.v2.transforms.JavascriptTextTransformer;
+import com.google.cloud.teleport.v2.transforms.MongoDbBackfillReader;
+import com.google.cloud.teleport.v2.transforms.MongoDbBackfillReader.BackfillPartition;
+import com.google.cloud.teleport.v2.transforms.MongoDbChangeStreamReader;
+import com.google.cloud.teleport.v2.transforms.MongoDbChangeStreamReader.ChangeStreamPartition;
 import com.google.cloud.teleport.v2.transforms.MongoDbTransforms;
-import com.google.cloud.teleport.v2.transforms.ReadSplitGenerator;
+import com.google.cloud.teleport.v2.transforms.StatefulDeduplication;
 import com.google.cloud.teleport.v2.transforms.UriSanitizer;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoDatabase;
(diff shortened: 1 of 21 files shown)


v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
         options.getMaxWriteRatePerWorker(),
         options.getWriteRateRampUpMinutes(),
         options.getWriteRateRampUpSteps());
+    LOG.info("  Migration Mode:          {}", options.getMigrationMode());
+    LOG.info("  Change Stream Splits:    {}", options.getNumChangeStreamSplits());
```
michaeltle-goog: Only log the change stream config  if we're doing CDC or BACKFILL_AND_CDC mode.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
...
+          final String targetCollection =
+              (targetCollectionRaw == null || targetCollectionRaw.isEmpty())
+                  ? inputCollection
+                  : targetCollectionRaw;
+
+          if (includeBackfill) {
+            int targetChunkSize =
+                options.getTargetBackfillChunkSize() != null
+                    ? options.getTargetBackfillChunkSize()
+                    : 200000;
```
michaeltle-goog: Make this default a constant

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
...
+            int maxSplits =
+                options.getMaxBackfillSplits() != null ? options.getMaxBackfillSplits() : 256;
```
michaeltle-goog: Make the 256 default a constant

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
...
+                  inputCollection,
+                  e.getMessage());
+              cdcPartitions.addAll(
+                  MongoDbChangeStreamReader.generatePartitions(
+                      null,
+                      options.getSourceUri(),
+                      options.getSourceDatabase(),
+                      inputCollection,
+                      targetCollection,
+                      1,
```
michaeltle-goog: Add a comment here explaining the param

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
...
+            LOG.warn(
+                "Could not generate database-level change stream splits ({}). Falling back to single"
+                    + " split.",
+                e.getMessage());
+            cdcPartitions.addAll(
+                MongoDbChangeStreamReader.generateDatabasePartitions(
+                    null,
+                    options.getSourceUri(),
+                    options.getSourceDatabase(),
+                    1,
```
michaeltle-goog: Add a comment here explaining the param

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,73 +492,299 @@ public static void run(Options options) {
...
+
+      // Stateful Deduplication Stage: Only needed when reconciling historical backfill records
+      // with concurrent live CDC mutations. In pure STREAMING_CDC mode, change stream events
+      // for any given document are already strictly ordered and sequential from the oplog.
+      if (requiresDeduplication) {
+        documents = documents.apply("Deduplicate", StatefulDeduplication.of());
+      }
+
+      // Metrics Stage
+      documents = documents.apply("CountTotalProcessed", ParDo.of(new CountDocumentsFn()));
```
michaeltle-goog: Do we still need this stage, we're already capturing the counters else where IIUC.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -447,62 +823,60 @@ public void processElement(ProcessContext c) {
...
+                  .withNumWriteShards(options.getNumWriteShards())
+                  .withMaxBufferingDurationMs(options.getMaxBufferingDurationMs())
                   .withMaxConcurrentAsyncWrites(options.getMaxConcurrentAsyncWrites())
                   .withMaxWriteRetries(options.getMaxWriteRetries())
                   .withDlqMaxRetries(options.getDlqMaxRetries())
                   .withInitialWriteRatePerWorker(
                       options.getInitialWriteRatePerWorker() != null
                           ? options.getInitialWriteRatePerWorker()
-                          : 100)
+                          : 5000)
```
michaeltle-goog: Make this a constant

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -447,62 +823,60 @@ public void processElement(ProcessContext c) {
...
                   .withMaxWriteRatePerWorker(
                       options.getMaxWriteRatePerWorker() != null
                           ? options.getMaxWriteRatePerWorker()
-                          : 500)
+                          : 25000)
```
michaeltle-goog: Make this a constant

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1402 @@
...
+      int numSplits,
+      BsonTimestamp t0) {
+    return generatePartitions(
+        uri,
+        database,
+        sourceCollection,
+        targetCollection,
+        numSplits,
+        0,
+        Math.max(numSplits, 256),
```
michaeltle-goog: Make the max larger 10k, and make it a constant

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1402 @@
...
+                currentRestriction.getOffsetInCurrentPartition(),
+                currentRestriction.getLastSeenIdJson(),
+                currentRestriction.isDone()));
+      }
+      return null;
+    }
+
+    @Override
+    public void checkDone() throws IllegalStateException {
+      // Bounded stream completion check
```
michaeltle-goog: Do we also need to implement this checkDone?

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
```diff
@@ -0,0 +1,1166 @@
...
+
+          if (currentToken != null) {
+            stream.resumeAfter(currentToken);
+          } else if (partition.getStartAtOperationTimeSeconds() > 0) {
+            stream.startAtOperationTime(
+                new BsonTimestamp(
+                    (int) partition.getStartAtOperationTimeSeconds(),
+                    partition.getStartAtOperationTimeInc()));
+          }
+
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 77fa43249de194398e860a684c5c09998b9a7969. We now reassign `stream` upon every builder method invocation (`batchSize`, `maxAwaitTime`, `resumeAfter`, `startAtOperationTime`, `fullDocument`) and handle the `fullDocument` fallback to `UPDATE_LOOKUP` dynamically when calling `cursor()`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,759 @@
...
+                currentRestriction.isDone());
+        return SplitResult.of(primary, residual);
+      }
+      return null;
+    }
+
+    @Override
+    public void checkDone() throws IllegalStateException {
+      // Bounded stream completion check
+    }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 77fa43249de194398e860a684c5c09998b9a7969. Implemented `checkDone()` in `BackfillRestrictionTracker` to verify that either a split occurred or the restriction finished processing.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbTransforms.java
```diff
@@ -86,6 +107,13 @@ public static MongoClient createMongoClient(String uri) {
         || connectionString.getUuidRepresentation() == UuidRepresentation.UNSPECIFIED) {
       builder.uuidRepresentation(UuidRepresentation.STANDARD);
     }
+    if (connectionString.getReadPreference() == null) {
+      builder.readPreference(ReadPreference.secondaryPreferred());
+    }
+    builder.applyToConnectionPoolSettings(
+        pool -> pool.maxSize(256).minSize(64).maxWaitTime(30, TimeUnit.SECONDS));
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 77fa43249de194398e860a684c5c09998b9a7969. Reduced default connection pool `minSize` to 0 to eliminate connection churn during short-lived metadata probing queries.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/ReadSplitGenerator.java
```diff
@@ -70,126 +73,37 @@ public static List<BsonDocument> generateIndexSliceFilters(int numSplits) {
...
       } catch (Exception e) {
         LOG.warn(
-            "Data-driven splits failed for STRING type in '{}.{}' ({}). Falling back to uniform splits.",
+            "Failed generating type-isolated splits for '{}.{}' ({}). Falling back to algorithmic splits.",
             databaseName,
             collectionName,
             e.getMessage());
-        stringFilters = generateStringFilters(numSplits);
       }
     }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 77fa43249de194398e860a684c5c09998b9a7969. Parallelized `probeActiveTypeBounds` queries across type buckets using a dedicated `ExecutorService` with bounded daemon threads and cleaned up in a `finally` block.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Moved change stream configuration logging inside an `if (includeCdc)` check so it is only logged when running in `STREAMING_CDC` or `BACKFILL_AND_STREAMING` modes.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Defined `DEFAULT_TARGET_BACKFILL_CHUNK_SIZE = 200000` constant in `MongoDbToMongoDb`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Defined `DEFAULT_MAX_BACKFILL_SPLITS = 256` constant in `MongoDbToMongoDb`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Added `/* numSplits= */ 1` (and `/* client= */ null`) parameter documentation comments.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Added `/* numSplits= */ 1` (and `/* client= */ null`) parameter documentation comments.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Agreed. Removed the redundant `CountTotalProcessed` / `CountDocumentsFn` stage in commit b6fd0e3e7. Counters are already thoroughly captured by `MongoDbBackfillReader` (`backfillDocumentsRead`), `MongoDbChangeStreamReader` (`changeEvents*`), and `WriteWithDlq` (`successfulWrites` and operation breakdowns).

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Defined `DEFAULT_INITIAL_WRITE_RATE_PER_WORKER = 5000` constant in `MongoDbToMongoDb`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Defined `DEFAULT_MAX_WRITE_RATE_PER_WORKER = 25000` (as well as ramp-up minutes and steps) constants in `MongoDbToMongoDb`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
michaeltle-goog: Fixed in commit b6fd0e3e7. Defined `DEFAULT_MAX_SPLITS = 10000` in `MongoDbBackfillReader` and updated `Math.max(numSplits, DEFAULT_MAX_SPLITS)`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
michaeltle-goog: Yes, exactly. For bounded restriction trackers (`isBounded()` returning `IsBounded.BOUNDED`), Apache Beam requires `checkDone()` to verify that the restriction was either fully processed (`isDone() == true`) or stopped due to a split (`shouldStop == true`), preventing silent truncation. This was implemented in `BackfillSlotRestrictionTracker` in commit 77fa43249de194398e860a684c5c09998b9a7969.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/templates/MongoDbToMongoDb.java
```diff
@@ -342,74 +489,297 @@ public static void run(Options options) {
...
+        try {
+          long sec = Long.parseLong(options.getStartAtOperationTime());
+          t0 = new BsonTimestamp((int) sec, 0);
+        } catch (NumberFormatException nfe) {
+          LOG.warn(
+              "Could not parse startAtOperationTime '{}' as epoch seconds. Capturing current"
+                  + " cluster time.",
+              options.getStartAtOperationTime());
+          t0 = MongoDbChangeStreamReader.captureCurrentClusterTime(sourceUri, sourceDatabase);
+        }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 3ad10ac3c. Added ISO-8601 parsing fallback using `Instant.parse()` with explicit imports and unit test coverage (`options_iso8601Timestamp_parsedSuccessfully`).

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/DocumentWithMetadata.java
```diff
@@ -89,15 +152,58 @@ public Document getDocument() {
...
+        return col + "#" + new Document("_id", document.get("_id")).toJson(CANONICAL_JSON_SETTINGS);
+      } catch (Exception ignored) {
+        return col + "#" + document.get("_id");
+      }
+    }
+    Object id = getId();
+    if (id != null) {
+      return col + "#" + id;
+    }
+    return col + "#" + System.identityHashCode(this);
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 3ad10ac3c. Replaced `System.identityHashCode(this)` with a deterministic UUID name hash (`UUID.nameUUIDFromBytes(...)`) using explicit imports (`UUID`, `StandardCharsets`). Note that we intentionally exclude `operationType` and `timestampSortKey` from the hash so that multiple CDC events and backfill records for the same document collide on the same Beam worker state cell for correct monotonic deduplication.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1390 @@
...
+    public void teardown() {
+      if (cursorCache != null) {
+        for (PartitionCursorHolder holder : cursorCache.values()) {
+          holder.close();
+        }
+        cursorCache.clear();
+      }
+      if (clientCache != null) {
+        clientCache.clear();
+      }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 3ad10ac3c. Added an explicit loop to close each cached `MongoClient` instance prior to clearing `clientCache` in `@Teardown`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1390 @@
...
+    public void teardown() {
+      if (cursorCache != null) {
+        for (ProcessBackfillPartitionFn.PartitionCursorHolder holder : cursorCache.values()) {
+          holder.close();
+        }
+        cursorCache.clear();
+      }
+      if (clientCache != null) {
+        clientCache.clear();
+      }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 3ad10ac3c. Added an explicit loop to close each cached `MongoClient` instance prior to clearing `clientCache` in `@Teardown`.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
```diff
@@ -0,0 +1,1170 @@
...
+    public void teardown() {
+      if (cursorCache != null) {
+        for (PartitionCursorHolder holder : cursorCache.values()) {
+          holder.close();
+        }
+        cursorCache.clear();
+      }
+      if (clientCache != null) {
+        clientCache.clear();
+      }
```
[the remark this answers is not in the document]
michaeltle-goog: Fixed in commit 3ad10ac3c. Added an explicit loop to close each cached `MongoClient` instance prior to clearing `clientCache` in `@Teardown`.

michaeltle-goog:
Pushed `a33429afc`: fixes a dedup-key mismatch on **sharded** source collections.

**Problem:** the MongoDB change stream `documentKey` includes the shard key fields in addition to `_id` on sharded collections, whereas backfill events (`backfillEvent`) only ever carry `{_id}`. The same physical document therefore produced two different dedup keys depending on which stream it arrived from:

```
backfill: tgtCol#{"_id": {"$oid": "..."}}
cdc:      tgtCol#{"customerId": {"$numberInt": "12345"}, "_id": {"$oid": "..."}}
```

Two consequences in `BACKFILL_AND_STREAMING` mode:

1. `StatefulDeduplication` never compared the two events — they occupied separate Beam state cells, so the CDC-beats-backfill tie-break in `TimestampSortKey.compareTo` could never fire.
2. `KeyByWriteShard` hashes on `getDedupKey()`, so the two events routed to *different* write shards and could be written concurrently by different workers. A stale backfill snapshot row could land **after** a newer CDC update.

**Fix:** normalize `documentKey` down to `_id` inside `getDedupKey()` so both streams agree. Falls back to the raw `documentKey` if it is unparseable or has no `_id`, so malformed input behaves as before.

Unsharded collections are unaffected — MongoDB already emits `documentKey == {_id: ...}`, which matches the backfill form exactly. That is why this did not surface in earlier testing.

Added regression tests for both the sharded and unsharded cases. Full suite: 136/136 passing, Checkstyle 0 violations.

:warning: Note for deployment: this changes dedup keys for sharded collections, so it should not be rolled into an already-running job via in-place pipeline update (existing dedup state would be orphaned and write-shard assignment reshuffles). Fresh launches are fine.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1402 @@
...
+  /**
+   * RestrictionTracker for BackfillRestriction ensuring slice-per-offset claiming,
+   * claim-before-output, and checkpoint splitting.
+   */
+  public static class BackfillRestrictionTracker
+      extends RestrictionTracker<BackfillRestriction, BackfillRestriction> {
+
+    private BackfillRestriction currentRestriction;
+    private boolean shouldStop = false;
+
```
namita-l: synchronize access to these mutable members

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/StatefulDeduplication.java
```diff
@@ -0,0 +1,194 @@
...
+    private final Counter dedupCdcAccepted =
+        Metrics.counter(StatefulDeduplicationFn.class, "dedup_cdc_accepted");
+    private final Counter dedupBackfillDropped =
+        Metrics.counter(StatefulDeduplicationFn.class, "dedup_backfill_dropped");
+    private final Counter dedupCdcDropped =
+        Metrics.counter(StatefulDeduplicationFn.class, "dedup_cdc_dropped");
+    private final Counter dedupPassthrough =
+        Metrics.counter(StatefulDeduplicationFn.class, "dedup_passthrough");
+
+    @StateId(STATE_ID_LAST_SEEN_KEY)
```
namita-l, line 171: AI review: There are no @TimerId timers, no TTL expiration, and no state eviction:

Every document creates or updates state permanently
latestSeenState.write(currentKey);
Risk: In enterprise migrations with 100M+ to billions of documents, every distinct _id creates a permanent record in Cloud Dataflow's Windmill state.

For streaming and hybrid migrations (BACKFILL_AND_STREAM), Windmill persistent state and active cache will grow monotonically without bound.
This results in high state storage costs, worker memory bloat, cache thrashing, and eventually worker Out-Of-Memory (OOM) failures or pipeline stalls.
Recommended Fix: Introduce an event-time or processing-time cleanup timer to clear state for keys that have been idle past a configurable deduplication window:

java


@TimerId("stateExpiry")
private final TimerSpec expiryTimerSpec = TimerSpecs.timer(TimeDomain.PROCESSING_TIME);
@ProcessElement
public void processElement(
    ...,
    @StateId("latestSeen") ValueState<TimestampSortKey> latestSeenState,
    @TimerId("stateExpiry") Timer expiryTimer) {
  ...
  latestSeenState.write(currentKey);
  // Expire state entry after configurable dedup window (e.g., 24h or 48h)
  expiryTimer.offset(stateRetentionDuration).setRelative();
}
@OnTimer("stateExpiry")
public void onExpiry(
    @StateId("latestSeen") ValueState<TimestampSortKey> latestSeenState) {
  latestSeenState.clear();
}

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
```diff
@@ -0,0 +1,1176 @@
...
+  public static BsonDocument generateHashedMatchFilter(int numSplits, int splitIndex) {
+    if (numSplits <= 0) {
+      throw new IllegalArgumentException("numSplits must be greater than 0, got " + numSplits);
+    }
+    if (splitIndex < 0 || splitIndex >= numSplits) {
+      throw new IllegalArgumentException(
+          String.format("splitIndex must be in range [0, %d), got %d", numSplits, splitIndex));
+    }
+
+    BsonDocument modDoc =
```
namita-l, line 297: may want to adda flag to accept a configurable shard key (may not always be _id, though using that as a default should be fine

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/DocumentWithMetadataCoder.java
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    }
+    BOOLEAN_CODER.encode(true, outStream);
+
+    String docJson =
+        value.getDocument() != null ? value.getDocument().toJson(CANONICAL_JSON_SETTINGS) : null;
+    STRING_CODER.encode(docJson, outStream);
+
+    String originalDoc = value.getOriginalDocument();
+    boolean hasDistinctOriginal =
```
namita-l: AI review: If an element has originalDocument explicitly set, and originalDocument.equals(documentJson):

hasDistinctOriginal encodes as false.
On decode, originalDoc is decoded as null.
The resulting object has getOriginalDocument() == null instead of the original string.
original.equals(decoded) evaluates to false!
Risk: In Apache Beam, coders must satisfy CoderProperties.coderDecodeEncodeEqual. When decode(encode(x)).equals(x) is violated, Beam reshuffle, checkpointing, and deduplication transforms experience nondeterministic behavior and assertion failures.

Recommended Fix: Use a 2-bit flag or tri-state encoding:

0: originalDocument is null.
1: originalDocument is identical to documentJson (restore originalDoc = docJson on decode).
2: originalDocument is distinct (read string from stream).

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
```diff
@@ -0,0 +1,1176 @@
...
+
+/**
+ * Streaming reader that consumes MongoDB Change Streams across partitioned cursors.
+ *
+ * <p>Features:
+ *
+ * <ul>
+ *   <li>Multi-cursor partitioning per collection using orthogonal match filters on {@code
+ *       documentKey._id}.
+ *   <li>Capture-before-read starting timestamp support ({@code startAtOperationTime}).
```
namita-l, line 82: startAtTimeStamp can cause a silent data loss if the oplog has churned past the timestamp since it's best effort. Instead this should have a move to capture and resume from resume tokens instead

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbTransforms.java
```diff
@@ -897,87 +1082,771 @@ public void finishBundle(FinishBundleContext c) {
...
+      for (DocumentWithMetadata item : items) {
+        String targetCol = item.getTargetCollection();
+        if (targetCol == null) {
+          targetCol = item.getSourceCollection();
+        }
+
+        if (item.getOperationType() != null && item.getOperationType().isDelete()) {
+          Object id = item.getId();
+          if (id != null) {
+            LinkedHashMap<Object, CoalescedOp> colMap =
```
namita-l, line 1429: AI Review: In MongoDB, _id can be a BSON Binary (byte[]). Java arrays implement identity-based equals() and hashCode(), not value-based. If multiple updates to the same document with a binary _id occur in the same bundle, colMap.put(...) will treat them as different keys, defeating coalescing.

Recommended Fix: Ensure doc.getId() converts binary arrays to ByteBuffer or hexadecimal strings for hashing and equality comparison.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
```diff
@@ -0,0 +1,1402 @@
...
+    public IsBounded isBounded() {
+      return IsBounded.BOUNDED;
+    }
+  }
+
+  /**
+   * Splittable DoFn that executes partitioned MongoDB backfill reads on worker threads, streaming
+   * micro-batches of documents into Windmill without exceeding bundle commit limits.
+   */
+  public static class ProcessBackfillPartitionFn
```
namita-l: Is this dead code?

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
michaeltle-goog: Done.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
michaeltle-goog: We need to use _id here to guarantee uniqueness when deduping the CDC events in the BACKFILL_AND_STREAMING mode. _id is always required and must be unique for all documents in a collection so it's best to leave this as a hard coded behaviour of the template to avoid unintended corruption.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/DocumentWithMetadataCoder.java
michaeltle-goog: Done. Went with tri-state encoding.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
namita-l: We'd need to use id for sharding the change stream to the source, but we can keep sharding by id for the dataflow template itself?

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/StatefulDeduplication.java
michaeltle-goog: Added a 24hr timer which can be configured by a flag. This assumes that backfill will complete within 24hrs when running in BACKFILL_AND_STREAMING mode.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
michaeltle-goog: Let's  follow up on that in a separate PR. This one is prioritizing the backfill changes, CDC and BACKFILL_AND_STREAMING is still experimental.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbTransforms.java
michaeltle-goog: Added tests to assert that this won't be an issue.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbBackfillReader.java
michaeltle-goog: Yup removed.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/MongoDbChangeStreamReader.java
michaeltle-goog: We do capture the resume token in the ChangeStreamRestriction. This field is used for the initial query when no resume tokens are stored.

v2/mongodb-to-mongodb/src/main/java/com/google/cloud/teleport/v2/transforms/StatefulDeduplication.java
michaeltle-goog: Actually changed this to default to infinite, windmill stores the keys on disk so there's no risk of OOMs. With an infinite window there's also no chance of correctness issues.

namita-l (approved):

PR #4251, opened 2026-09-14
Add large MySQL to Spanner bulk migration load test

sm745052:
### Note

This load test is designed **not** to run as part of normal or scheduled CI: the class skips itself unless it is named explicitly in `-Dtest`.

### Summary

Adds `MySQLLargeLT`, a manually-invoked benchmark for the bulk `sourcedb-to-spanner` template against a large, pre-populated MySQL fleet. It complements the existing `*LT` classes, which spin up ephemeral sources and so cannot characterise behaviour at multi-terabyte scale.

Five scenarios are exposed as individual `@Test` methods:

| Test | Tables | Shards | Approx. size |
|---|---|---|---|
| `singleTableOneShard` | 1 | 1 | 64 GB |
| `allTablesOneShard` | 5,000 | 1 | 200 GB |
| `singleTableAllShards` | 1 | 1,000 | 64 TB |
| `topEightTablesAllShards` | 8 | 1,000 | 194 TB |
| `allTablesAllShards` | 5,000 | 1,000 | 200 TB |

Every other parameter is fixed in the class, so a run needs no arguments beyond `-Dtest`:

```
mvn test -pl v2/sourcedb-to-spanner -am \
  -Dtest=MySQLLargeLT#singleTableOneShard -Dsurefire.failIfNoSpecifiedTests=false
```

### Not picked up by the weekly schedule

The class carries `@Category(TemplateLoadTest.class)` so it can be selected by name from the Spanner load-test workflow's `workflow_dispatch` input. However, the weekly cron runs that whole category with **no** `-Dtest`, which would start a multi-hour job and resize a shared Spanner instance unattended.

`setUpClass()` therefore asserts the class was named explicitly and otherwise skips. The assumption is raised in `@BeforeClass`, so `ParentRunner` reports a failed assumption and never constructs an instance — no `@Before`/`@After` runs, and the shared instance is left alone.

### Notes for reviewers

- **The Spanner schema lives in GCS**, not in test resources. It is 5,000 `CREATE TABLE` statements (~585 KB), which is too large to check in, so the test reads it from the benchmark bucket at run time. The existing `MySQL5KTablesLT` generates its tables programmatically instead, but that is only possible because its source is ephemeral.
- Source credentials are **not** in the repo. The shard configs live in GCS and resolve the password via `secretManagerUri`.
- Test-only change: nothing under `src/main/` is touched.
- Teardown drops the generated database, deletes the run-scoped GCS output, and restores the instance node count in a `finally` block so a failed run cannot strand it.
- Necessary permissions to service accounts are granted

### Testing

`singleTableOneShard` was run end to end: 5,000-statement DDL applied in 67s, Dataflow job reached `JOB_STATE_DONE`, metrics collected, and teardown restored all resources. `BUILD SUCCESS`, `Tests run: 1, Failures: 0, Errors: 0`.


manitgupta:
How does this test work? Is the Spanner instance where the migration happens expected to be pre-provisioned?

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+      this.tables = tables;
+    }
+  }
+
+  @BeforeClass
+  public static void setUpClass() {
+    assumeTrue(
+        "Skipped: name this class in -Dtest to run it.",
+        System.getProperty("test", "").contains(MySQLLargeLT.class.getSimpleName()));
+    System.setProperty("project", DATAFLOW_PROJECT);
```
manitgupta: **Global system-property pollution.** `setUpClass()` mutates JVM-global system properties (`project` / `region` / `artifactBucket`) and never restores them. Surefire defaults to `reuseForks=true`, so running this alongside another LT class in the same JVM (e.g. `-Dtest=MySQLLargeLT#x,SomeOtherLT#y`) leaks `span-cloud-migrations-testing` / `asia-southeast1` / `smt-llt` into the other test — which then re-reads `TestProperties.project()` and silently launches its job into the wrong project/region/bucket. Consider capturing the prior values and restoring them in teardown, or scoping these overrides so they don't outlive the class.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+  @After
+  public void cleanUp() {
+    try {
+      if (spannerResourceManager != null) {
+        // On a static instance this drops only the generated database.
+        ResourceManagerUtils.cleanResources(spannerResourceManager, gcsResourceManager);
+      }
+    } finally {
+      try {
+        scaleInstance(IDLE_NODES);
```
manitgupta: **Teardown does not actually restore the shared instance.** Teardown hardcodes `scaleInstance(IDLE_NODES = 100)` rather than capturing the instance's pre-test node count. The permanent shared `nokill-smt-test-instance` normally runs at N != 100 nodes; any run scales it to 2000 then leaves it pinned at 100, silently changing capacity for every other user. The class javadoc claims it "restores the instance to IDLE_NODES nodes" but it never reads the original count. Suggest reading the current node count in `setUpClass()` and restoring that exact value in teardown.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+            .addEnvironment("ipConfiguration", "WORKER_IP_PRIVATE")
+            .addEnvironment("stagingLocation", "gs://" + BUCKET + "/staging")
+            .addEnvironment("tempLocation", "gs://" + BUCKET + "/temp");
+
+    PipelineLauncher.LaunchInfo jobInfo = launchJob(options);
+    LOG.info("Launched {} as job {}", scenario, jobInfo.jobId());
+
+    // Drains on timeout; otherwise teardown would drop the database under a running job.
+    PipelineOperator.Result result =
+        pipelineOperator.waitUntilDoneAndFinish(createConfig(jobInfo, JOB_TIMEOUT));
```
manitgupta: **No metrics captured on timeout.** On a 12h timeout, `waitUntilDoneAndFinish` returns `TIMEOUT` and the `assertThatResult(result).isLaunchFinished()` assertion throws *before* `getMetrics()` / `populateResourceManagerMetrics()` run. The largest scenarios (`topEightTablesAllShards` ~194 TB, `allTablesAllShards` ~200 TB) are exactly the ones most likely to hit `JOB_TIMEOUT=12h`, so the benchmark yields zero metrics plus a spurious failure for the very runs it exists to measure. Consider collecting metrics regardless of result state before asserting.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+    assertThatResult(result).isLaunchFinished();
+
+    Map<String, Double> metrics = getMetrics(jobInfo);
```
manitgupta: **Metrics collected but never exported to BigQuery.** `run()` re-implements two of the three steps of the existing base helper `collectAndExportMetrics(jobInfo)` (get + populate) but drops the export step. As a result the benchmark's RunTime / throughput / Spanner CPU / cost live only in the test log and are not persisted to the metrics table used for long-term performance tracking. Prefer reusing `collectAndExportMetrics(jobInfo)` so results are exported and the code doesn't diverge from the shared helper.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+            .addEnvironment("stagingLocation", "gs://" + BUCKET + "/staging")
+            .addEnvironment("tempLocation", "gs://" + BUCKET + "/temp");
+
+    PipelineLauncher.LaunchInfo jobInfo = launchJob(options);
+    LOG.info("Launched {} as job {}", scenario, jobInfo.jobId());
+
+    // Drains on timeout; otherwise teardown would drop the database under a running job.
+    PipelineOperator.Result result =
+        pipelineOperator.waitUntilDoneAndFinish(createConfig(jobInfo, JOB_TIMEOUT));
+    assertThatResult(result).isLaunchFinished();
```
manitgupta: **No correctness validation of the migration.** Unlike the base `runLoadTest`, this test performs no row-count / `SpannerRowsCheck`. A job that finishes having migrated nothing — a mistyped `SINGLE_TABLE` / `TOP_EIGHT_TABLES` value, or an empty/misconfigured source shard — still reaches `JOB_STATE_DONE`, passes `isLaunchFinished()`, and reports BUILD SUCCESS despite migrating 0 rows. Consider adding at least a minimal row-count assertion so a no-op run fails loudly.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+    LOG.info("Scaling {} to {} nodes", SPANNER_INSTANCE, nodeCount);
+    SpannerOptions options =
+        SpannerOptions.newBuilder().setProjectId(SPANNER_PROJECT).setHost(SPANNER_HOST).build();
+    try (Spanner spanner = options.getService()) {
+      InstanceInfo instance =
+          InstanceInfo.newBuilder(InstanceId.of(SPANNER_PROJECT, SPANNER_INSTANCE))
+              .setNodeCount(nodeCount)
+              .build();
+      spanner.getInstanceAdminClient().updateInstance(instance, InstanceField.NODE_COUNT).get();
+    } catch (Exception e) {
```
manitgupta: **Interrupt status swallowed.** `scaleInstance` catches `Exception` (which includes the `InterruptedException` thrown by the blocking `updateInstance(...).get()`) and rethrows as `IllegalStateException` without calling `Thread.currentThread().interrupt()`. If the call is interrupted, the interrupt signal is lost for downstream code. Suggest catching `InterruptedException` specifically and re-setting the interrupt flag.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+      this.shardConfig = shardConfig;
+      this.tables = tables;
+    }
+  }
+
+  @BeforeClass
+  public static void setUpClass() {
+    assumeTrue(
+        "Skipped: name this class in -Dtest to run it.",
+        System.getProperty("test", "").contains(MySQLLargeLT.class.getSimpleName()));
```
manitgupta: **Substring skip guard.** The guard uses `System.getProperty("test", "").contains("MySQLLargeLT")`. A run targeting a different class whose simple name merely *contains* `MySQLLargeLT` (e.g. a future `MySQLLargeLTv2` or a helper subclass) would satisfy the assumption and unintentionally launch this multi-hour, 2000-node-resizing benchmark. Consider a more exact match (word-boundary / exact class-name check).

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,266 @@
...
+ * </table>
+ *
+ * <p>The category makes this dispatchable by name from the Spanner load-test workflow, but the
+ * weekly schedule runs the whole category with no {@code -Dtest}, which would sweep in a run of
+ * this size. {@link #setUpClass()} therefore skips unless the class is named explicitly.
+ *
+ * <p>Each run scales the instance to {@value #TARGET_NODES} nodes, creates and drops its own
+ * database, and restores the instance to {@value #IDLE_NODES} nodes.
+ */
+@Category(TemplateLoadTest.class)
```
manitgupta: **Missing `SkipDirectRunnerTest` category.** Sibling `MySQL5KTablesLT` uses `@Category({TemplateLoadTest.class, SkipDirectRunnerTest.class})`, but this class omits `SkipDirectRunnerTest` and relies solely on the `assumeTrue` guard to avoid direct-runner sweeps. Adding the category gives a second, category-based safeguard against an accidental expensive run outside the `-Dtest` naming convention.

v2/sourcedb-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/loadtesting/MySQLLargeLT.java
```diff
@@ -0,0 +1,288 @@
...
+
+    Scenario(String shardConfig, String tables) {
+      this.shardConfig = shardConfig;
+      this.tables = tables;
+    }
+  }
+
+  @BeforeClass
+  public static void setUpClass() {
+    assumeTrue(
```
bharadwaj-aditya, line 151: This seems a bit hacky. Why not just mark this test as ignored, and when it needs to be run disable the check, deploy the template and run ?

sm745052: > when it needs to be run disable the check, deploy the template and run ?

We can ! it adds some manual steps though like
**1. remove ignore from code
2. push to a branch
3. run the workflow from that branch and** run with Dtest=...

PR #4260, opened 2026-09-15
test: de-flake AstraDb, reverse-migration, and datastream job-init ITs

manitgupta:
## What

De-flakes three integration tests across three modules. All changes are test-only (no template/product code touched).

### Fix 1 — `AstraDbToSpannerIT` DataStax `DriverTimeout: Query timed out after PT2S`
`v2/sourcedb-to-spanner/.../AstraDbToSpannerIT.java`

**Root cause:** the test builds `CqlSession`s against a real remote Astra DB with no `withConfigLoader(...)`, so it uses the DataStax driver default `basic.request.timeout = 2s`. Remote `CREATE TABLE` (`createAndPopulateTables`) and `DROP TABLE` (`tearDown`) round-trips exceed 2s and flake.

**Fix:** pass a programmatic `DriverConfigLoader` to both `CqlSession.builder()` calls, raising `REQUEST_TIMEOUT` and `CONTROL_CONNECTION_AGREEMENT_TIMEOUT` to 30s.

### Fix 2 — reverse-migration `ConditionCheck` waits too short for change-stream warm-up
`v2/spanner-to-sourcedb/.../SpannerToCassandraSourceDBCustomTransformationIT.java`, `SpannerToMySqlDataTypesPGDialectIT.java`, `SpannerToSourceDBShardedMySQLRetryAllDLQIT.java`

**Root cause:** jobs launch and stay RUNNING, but data arrives late due to change-stream warm-up, tripping the `waitForCondition` budget.

**Fix:** raise the wait windows — 10m→15m for the first two, 15m→20m for the DLQ check in the sharded retry test.

### Fix 4 — `MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT` first check ignores job-init latency
`v2/datastream-to-spanner/.../MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.java`

**Root cause:** the first (and only) condition check in each test method used a flat `Duration.ofMinutes(10)`, which does not account for the documented ~5-min Dataflow job-init latency, leaving only ~5 min of real data-wait.

**Fix:** use the documented `JOB_START_PROCESSING_WAIT_MINUTES` (=15) constant from `DataStreamToSpannerITBase` for those first checks.

## Verification status

**Compile:** `test-compile` passes for all three modules (`sourcedb-to-spanner`, `spanner-to-sourcedb`, `datastream-to-spanner`) on JDK 25 with `-Dspotless.check.skip=true`.

**End-to-end:**

- ✅ **Fix 1 (`AstraDbToSpannerIT`) — PASS.** Full real run against Astra DB + Dataflow + Spanner: `Tests run: 1, Failures: 0, Errors: 0`, `BUILD SUCCESS`. This directly exercises the fix — the remote Astra `CREATE TABLE`/inserts and `DROP TABLE` completed without the PT2S `DriverTimeout`, the job finished, and all 50 rows landed in Spanner.
- ⏳ **Fix 2 (reverse-migration ITs) — blocked in a local workstation environment, compile-verified only.** `SpannerToMySqlDataTypesPGDialectIT` staged and submitted the Dataflow job, but the job failed at launch with `Communications link failure … Connection refused`: the reverse-migration source DB is a local testcontainers (Docker) instance that GCP Dataflow workers cannot reach. This happens before the changed wait code runs. The other two reverse-migration ITs use the same local-Docker-source pattern. These need a CI/GCP environment with a reachable source.
- ⏳ **Fix 4 (`MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT`) — blocked on config, compile-verified only.** Fails in `setUp` because `CloudMySQLResourceManager` requires static Cloud SQL connection params (`-DhostIp`, `-DprivateConnectivity`, `-DcloudProxyHost`, `-DcloudProxyUsername`, `-DcloudProxyPassword`) that were not supplied.

Left as a draft: Fix 1 is green e2e; Fixes 2 and 4 still need a CI/GCP environment to validate end-to-end.

## Testing
Verified in Spanner PR CI (run https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/34948634705): all fixed ITs pass — AstraDbToSpannerIT ✅, SpannerToCassandraSourceDBCustomTransformationIT ✅, SpannerToMySqlDataTypesPGDialectIT ✅, SpannerToSourceDBShardedMySQLRetryAllDLQIT ✅, MySQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT ✅.


v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDBCustomTransformationIT.java
```diff
@@ -171,7 +171,7 @@ private void assertBasicRowInCassandraDB() throws InterruptedException {
     PipelineOperator.Result result =
         pipelineOperator()
             .waitForCondition(
-                createConfig(jobInfo, Duration.ofMinutes(10)),
+                createConfig(jobInfo, Duration.ofMinutes(15)),
```
bharadwaj-aditya, line 174: This should ideally be a constant in all the tests for a given template.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDBShardedMySQLRetryAllDLQIT.java
```diff
@@ -189,7 +189,7 @@ public void testSpannerToSrcDBRetryAllDLQ() throws Exception {
     PipelineOperator.Result dlqWaitResult =
         pipelineOperator()
             .waitForCondition(
-                createConfig(jobInfo, Duration.ofMinutes(15)),
+                createConfig(jobInfo, Duration.ofMinutes(20)),
```
bharadwaj-aditya, line 192: Same comment as above. Why is this custom ?

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDBCustomTransformationIT.java
manitgupta: It should. I only touched what failed. 

During the analysis I observed that the tests were timing out and the Dataflow job took a longer than typical time to boot up, which did not leave enough headroom for the job to actually work.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToSourceDBShardedMySQLRetryAllDLQIT.java
manitgupta: I only touched what failed. 

During the analysis I observed that the tests were timing out and the Dataflow job took a longer than typical time to boot up, which did not leave enough headroom for the job to actually work.

manitgupta: Resolving this one and replying on the other to have one thread.

v2/spanner-to-sourcedb/src/test/java/com/google/cloud/teleport/v2/templates/SpannerToCassandraSourceDBCustomTransformationIT.java
manitgupta: I took a 2nd pass at this and the values of timeouts for ITs in this location are all over the place. Some set 10, some 15 and some 20. Changing all will touch a lot of files. Should we keep this as is for now? I can change it too but it will change a number of ITs and I cannot predict how much time that adds to the overall suite execution time.

Let me know what you want me to do.

bharadwaj-aditya: ok. Keep as is for now.

bharadwaj-aditya (approved):
LGTM

PR #4261, opened 2026-09-15
Fix PostgreSQL inheritance IT to exercise CDC instead of backfill

manitgupta:
## Problem

`PostgreSQLDatastreamToSpannerInheritanceIT` was testing the wrong scenario.

The `datastream-to-spanner` template is used **only for CDC replication** in production; the snapshot/backfill is handled by the separate `sourcedb-to-spanner` template. But this test inserted the source rows into Postgres **before** the Datastream stream existed, so Datastream **backfilled** them.

Postgres backfill runs `SELECT * FROM parent_table`, and because the tables use table **INHERITANCE** (`child_table INHERITS parent_table`, `grandchild_table INHERITS child_table`), that query also returns the child and grandchild rows. As a result the Spanner `parent_table` wrongly received the descendant rows. On top of that, the assertion used `hasRecordsUnorderedCaseInsensitiveColumns`, which is only a *subset* check and silently ignores those extra inherited rows — so the bug was invisible.

## Fix

Make the test exercise CDC, matching how the template is used in production:

- **Source tables start empty at stream time.** Removed the 3 `INSERT` statements from `postgresql-schema.sql`, keeping only the `CREATE TABLE ... INHERITS` DDL. `createLogicalReplication()` still runs in `setUp()` after the DDL so the replication slot/publication exist and cover the tables before any writes.
- **Rows are inserted after the job is running.** In both `testPostgreSqlInheritance` and `testPostgreSqlInheritancePGDialect`, the 3 rows (parent id=1, child id=2, grandchild id=3) are now written to Postgres **after** `assertThatPipeline(jobInfo).isRunning()`, via a `writeCdcData()` `ConditionCheck` chained (with `ChainedConditionCheck`) before the row-count wait so the wait follows the inserts. Each insert targets its own physical table (`parent_table` / `child_table` / `grandchild_table`) so logical decoding tags the events per table. This mirrors the established sibling pattern in `DataStreamToSpannerWideRowForMaxTableNameIT` / `DataStreamToSpannerIT`.
- **Validation tightened to catch inherited-row duplication.** `validateResult` now asserts an exact per-table row count (`SpannerAsserts.assertThatStructs(rows).hasRows(expected.size())`) in addition to the existing unordered-contents check. So `parent_table` must contain exactly 1 row (id=1), not 3.

`getExpectedData()` is unchanged (parent→{1}, child→{2}, grandchild→{3}), and the 20-minute condition timeout is preserved.

## Verification

- `mvn -pl v2/datastream-to-spanner -am test-compile -Dspotless.check.skip=true`: **PASS** (exit 0).
- Full integration run (`mvn verify -PtemplatesIntegrationTests -Dtest=PostgreSQLDatastreamToSpannerInheritanceIT ...`): could **not** complete in this environment — it fails in `setUp()` while constructing `CloudPostgresResourceManager` because the static Cloud SQL Postgres connection requires the `-DcloudProxyHost` / `-DcloudProxyPostgresPort` (Cloud SQL Auth Proxy) properties and Datastream private connectivity, which are not provisioned here. The failure is upstream of the launch and of all the changed code, so it does not reflect on this change. Reviewers with the provisioned test environment can run the full IT.

## Testing

Verified in the Spanner PR integration-test CI:

- Run: https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/34959873746 (job: "Dataflow Templates Integration Tests").
- `PostgreSQLDatastreamToSpannerInheritanceIT` — `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`. Both `testPostgreSqlInheritance` (GoogleSQL dialect) and `testPostgreSqlInheritancePGDialect` pass under CDC.
- Under CDC the descendant rows no longer leak into the ancestor tables. For both dialects the run logged exactly one row per table and the `hasRows(expected.size())` exact-count assertion held:
  - `parent_table` -> `[1, Parent Row 1]` (exactly id=1)
  - `child_table` -> `[2, Child Row 1, 10]` (exactly id=2)
  - `grandchild_table` -> `[3, Grandchild Row 1, 5, New York]` (exactly id=3)

This confirms the fix: with the rows written after the stream is running (CDC), Datastream tags each change event to its own physical table, so Spanner `parent_table` ends with only id=1 instead of also receiving the inherited child/grandchild rows.

Note: the overall CI job is red only because of an unrelated infrastructure flake in a different test class — `PostgreSQLDatastreamToPGDialectSpannerTableAndIndexLimitsIT.testCellSize` errored in `setUp()` with a Datastream `ABORTED: unable to queue the operation` while starting its own stream. That failure is upstream of launch, in a separate test, and unrelated to this change.

### Update — review feedback (ON CONFLICT idempotency)

Addressed review feedback by making the three CDC inserts in `writeCdcData()` idempotent with `ON CONFLICT (id) DO NOTHING` (commit `1cbe212`), so a re-polled `ConditionCheck` (e.g. after a transient DB error) cannot fail on a primary-key conflict.

Re-verified in the Spanner PR CI: run https://github.com/GoogleCloudPlatform/DataflowTemplates/actions/runs/35202235100 finished green (whole workflow concluded `success`; the earlier unrelated infra flake did not recur). `PostgreSQLDatastreamToSpannerInheritanceIT` — `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`. Both `testPostgreSqlInheritance` and `testPostgreSqlInheritancePGDialect` passed under CDC, each with `parent_table` ending as exactly `[1, Parent Row 1]` (id=1), confirming the descendant rows still do not leak into the ancestor table.


v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerInheritanceIT.java
```diff
@@ -260,22 +281,47 @@ private void validateResult(
...
+      @Override
+      protected CheckResult check() {
+        try {
+          resourceManager.runSQLUpdate(
+              "INSERT INTO parent_table (id, name) VALUES (1, 'Parent Row 1')");
+          resourceManager.runSQLUpdate(
+              "INSERT INTO child_table (id, name, age) VALUES (2, 'Child Row 1', 10)");
+          resourceManager.runSQLUpdate(
+              "INSERT INTO grandchild_table (id, name, age, city) VALUES (3, 'Grandchild Row 1', 5,"
+                  + " 'New York')");
```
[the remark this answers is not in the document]
manitgupta: Applied in 1cbe21202bca1c7b637942464729f754cb3f636e. All three CDC inserts now use ON CONFLICT (id) DO NOTHING, so a re-polled ConditionCheck after a transient error cannot fail on a PK conflict.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerInheritanceIT.java
```diff
@@ -90,6 +92,12 @@ public void setUp() throws IOException {
         postgresResourceManager = CloudPostgresResourceManager.builder(testName).build();
         LOG.info(
             "PostgreSQL resource manager created with URI: {}", postgresResourceManager.getUri());
+        // Each method uses its own source database to keep the two tests independent.
+        LOG.info("Setting up PG dialect PostgreSQL resource manager...");
+        pgDialectPostgresResourceManager = CloudPostgresResourceManager.builder(testName).build();
```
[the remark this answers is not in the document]
manitgupta: No change needed here. Although both managers are built with the same testName, CloudSqlResourceManagerUtils.generateDatabaseName appends a yyyyMMdd_HHmmss timestamp plus a random 6-character suffix, so the two source databases get distinct names and cannot collide.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerInheritanceIT.java
```diff
@@ -71,6 +72,7 @@ public class PostgreSQLDatastreamToSpannerInheritanceIT extends DataStreamToSpan
 
   private static boolean initialized = false;
   private static CloudPostgresResourceManager postgresResourceManager;
+  private static CloudPostgresResourceManager pgDialectPostgresResourceManager;
```
bharadwaj-aditya, line 77: can you add a comment on what the purpose of both these managers is ? Is one for spanner ?

bharadwaj-aditya (commented):
Seems fine overall. But looking for a clarification on the resource managers.

v2/datastream-to-spanner/src/test/java/com/google/cloud/teleport/v2/templates/PostgreSQLDatastreamToSpannerInheritanceIT.java
manitgupta: Added a clarifying comment in 328489c13. Both are PostgreSQL *sources* (not Spanner): Datastream reads each via CDC, with a separate source database per @Test method so the GoogleSQL-dialect and PG-dialect runs stay independent (own source data + replication stream). The Spanner destinations are the separate spannerResourceManager / pgDialectSpannerResourceManager fields just below.

bharadwaj-aditya: ok makes sense. LGTM.

But maybe makes sense to rename the 1st one to googleSQLResource manager.

Can take that up in a follow up PR if it is easier.

bharadwaj-aditya (approved):
LGTM

PR #4276, opened 2026-09-21
fix: do not emit inline column OPTIONS for postgresql dialect

sm745052:
## Problem

PostgreSQL has no inline `OPTIONS` clause in a column definition, but `Column.prettyPrint()` appended one regardless of dialect. Any PG column carrying options renders DDL the server rejects:

```
INVALID_ARGUMENT: [ERROR] syntax error at or near "OPTIONS"; failed to parse the DDL statements.
```

This stayed latent while PG columns rarely carried options. It is now reachable: `INFORMATION_SCHEMA.COLUMN_OPTIONS` reports an implicit `locality_group = default` on every column.

`ShadowTableCreator` copies primary key columns verbatim from the information schema — clearing `isGenerated` / `generationExpression` but not `columnOptions` — so PG-dialect shadow tables are now unparseable:

```sql
CREATE TABLE "shadow_singers" (
	"singer_id"                             bigint NOT NULL OPTIONS (locality_group='default'),
	"timestamp"                             bigint,
	"lsn"                                   character varying(2621440),
	PRIMARY KEY ("singer_id")
)
```

Shadow tables are never created, the pipeline writes nothing, and the job sits in `JOB_STATE_RUNNING`. Affects **datastream-to-spanner** and **spanner-to-sourcedb** on PG-dialect databases; visible in CI as PG integration tests failing `expected: CONDITION_MET but was: TIMEOUT`.

## Fix

```java
if (dialect() == Dialect.POSTGRESQL || columnOptions() == null) {
  return;
}
```

Mirrors the existing dialect branch for generated columns just above. No input produced valid PG DDL previously, so no working behaviour changes. PG expresses column options with dedicated clauses rather than an options list — LOCALITY GROUP <name> inline, or ALTER TABLE ... ALTER COLUMN ... SET LOCALITY GROUP ...

## Tests

- **New** `ColumnTest.testPrettyPrintColumnOptions` — asserts PG suppresses the clause *and* GoogleSQL still emits it, so this cannot later be "fixed" by dropping options in both dialects.
- **Corrected** the PG golden in `InformationSchemaScannerTest.testScanPgSQLDdl`, which asserted `... NOT NULL OPTIONS (option1='SomeName')` — an expectation that pinned unparseable DDL.


v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Column.java
```diff
@@ -89,7 +89,10 @@ public void prettyPrint(Appendable appendable) throws IOException {
         appendable.append(" STORED");
       }
     }
-    if (columnOptions() == null) {
+    // PostgreSQL has no inline OPTIONS clause in a column definition; emitting one always
+    // produces a statement the PG dialect cannot parse. Column options are set separately
+    // via ALTER TABLE ... ALTER COLUMN ... SET OPTIONS (...) in that dialect.
+    if (dialect() == Dialect.POSTGRESQL || columnOptions() == null) {
       return;
     }
```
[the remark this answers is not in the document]
bharadwaj-aditya: please apply this check

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/ddl/ColumnTest.java
```diff
@@ -291,4 +292,31 @@ public void testPrettyPrint() {
...
+
+    // PostgreSQL has no inline OPTIONS clause. Emitting one yields DDL the PG dialect rejects
+    // with `syntax error at or near "OPTIONS"`, so the clause must be suppressed even when the
+    // information schema reports options for the column.
+    Column pg =
+        Column.builder(Dialect.POSTGRESQL)
+            .name("col1")
+            .type(Type.pgInt8())
+            .notNull(true)
+            .columnOptions(ImmutableList.of("locality_group='default'"))
```
bharadwaj-aditya, line 318: this should not be set in PG - right ?

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Column.java
```diff
@@ -89,7 +89,8 @@ public void prettyPrint(Appendable appendable) throws IOException {
         appendable.append(" STORED");
       }
     }
-    if (columnOptions() == null) {
+    // PostgreSQL has no inline OPTIONS clause in a column definition.
+    if (dialect() == Dialect.POSTGRESQL || columnOptions() == null) {
```
bharadwaj-aditya: does PG not support any column options ? Can you review the docs to be sure ?

v2/spanner-common/src/test/java/com/google/cloud/teleport/v2/spanner/ddl/ColumnTest.java
sm745052: in info schema it is there, so we are mocking that

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Column.java
sm745052: Spanner PG Dialect does have column options, but they are set in a different way that googlesql

for example in Gsql we do:
```
-- GoogleSQL
CREATE TABLE cars (
  car_id       INT64 NOT NULL,
  make         STRING(64),
  service_log  STRING(MAX) OPTIONS (locality_group = 'cold')
) PRIMARY KEY (car_id);
```

but in Pgsql we do
```
-- PostgreSQL
CREATE TABLE cars (
  car_id       bigint NOT NULL,
  make         varchar(64),
  service_log  text LOCALITY GROUP cold,
  PRIMARY KEY (car_id)
);
```


The way we add is different.

Today we use the Gsql format in PgSQL as well. This is a quick fix to enable the integration tests, we should later see how (or if?) to handle it.

bharadwaj-aditya: ok

bharadwaj-aditya (dismissed):
LGTM

v2/spanner-common/src/main/java/com/google/cloud/teleport/v2/spanner/ddl/Column.java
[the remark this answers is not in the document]
sm745052: ack

bharadwaj-aditya (approved):
LGTM

